PR #58, opened 2021-07-15
Activity stream

olivierdalang:
Initial implementation of activity stream using django-notifications.

I'm not 100% this is the best choice.
There's also django-activity-stream, which looks includes feature to follow/unfollow, but lacks seen/unseen.
There's also pinax-notifications, which looks like it it also takes care of the actual sending of the emails (the other ones don't do that)

But it is installed in a separate `activity` app, it should be relatively independent and not too hard to change afterwards. We also probably would need to sort out proper emailing solution on production as otherwise we'll very likely end up having all our sent emails flagged as spam.

Currently, it logs user creation, organization creation, membership addition/removal. But I didn't really think of what notification actually make sense. Can we list those ? (esp. for stuff linked to deltas/conflicts/etc).


olivierdalang:
Had some strange failures happening from tearDown methods of tests, when it deletes all created objects.

Not exactly sure why it failed though...

But anyways, the deletion in tearDown should not be needed, as Django's TestCase are ran in transactions that are rolled back, so I just removed that code.

docker-app/qfieldcloud/notifs/models.py
```diff
@@ -0,0 +1,260 @@
...
+        return team_members + organization_admins
+    elif user.user_type == User.TYPE_ORGANIZATION:
+        org_members = []
+        for m in user.organization.members.all():
+            org_members.extend(_subusers_for_user(m.member))
+        org_owners = _subusers_for_user(user.organization.organization_owner)
+        return org_members + org_owners
+
+
+def _users_for_project(project: Project):
```
suricactus, line 80: Should use the new `UsersQueryset.for_project`.

docker-app/qfieldcloud/notifs/models.py
```diff
@@ -0,0 +1,260 @@
...
+    notify.send(
+        authenticated_user,
+        verb=verb,
+        action_object=action_object,
+        target=target,
+        recipient=list(recipient),
+    )
+
+
+def _subusers_for_user(user: User):
```
suricactus, line 59: Can we move to the `User.queryset`?

docker-app/qfieldcloud/notifs/templates/notifs/notification_email_body.html
```diff
@@ -0,0 +1,13 @@
...
+<ul>
+    {% for notif in notifs %}
+    <li>{{notif}}</li>
+    {% endfor %}
+</ul>
+
+<hr/>
+
+<p>
+    <a href="https://{{hostname}}/settings/{{username}}/notifications_settings">Manage your notifications settings</a>
```
suricactus, line 12: Probably passing the URL from outside would easier to manage. Imagine `settings` renamed to `whatever` and you need to change at several places.

docker-app/qfieldcloud/notifs/templates/notifs/notification_email_body.txt
```diff
@@ -0,0 +1,9 @@
+Here is what happened recently on QFieldCloud:
+
+{% for notif in notifs %}
+- {{notif}}
+{% endfor %}
+
+-----
+
+Manage your notification settings here : https://{{hostname}}/settings/{{username}}/notifications_settings
```
suricactus, line 9: Same, pass the URL as parameter.

docker-app/qfieldcloud/core/models.py
```diff
@@ -126,6 +126,20 @@ class UserAccount(models.Model):
...
+    NOTIF_DAILY = "DAILY"
+    NOTIF_WEEKLY = "WEEKLY"
+    NOTIF_CHOICES = (
+        (NOTIF_IMMEDIATELY, _("Immediately")),
+        (NOTIF_HOURLY, _("Hourly")),
+        (NOTIF_DAILY, _("Daily")),
+        (NOTIF_WEEKLY, _("Weekly")),
+        (NOTIF_DISABLED, _("Disabled")),
+    )
+
```
suricactus, line 142: Can we use `models.TextChoices` instead? For me it's cleaner, rather that a bunch of capitalized properties. Also better for typing.

docker-app/qfieldcloud/notifs/models.py
```diff
@@ -0,0 +1,260 @@
...
+    """Returns a list of users concerned by updates to a project"""
+
+    proj_members = []
+    for m in project.collaborators.all():
+        proj_members.extend(_subusers_for_user(m.collaborator))
+    proj_owners = _subusers_for_user(project.owner)
+    return proj_members + proj_owners
+
+
+# Register signals
```
suricactus, line 90: Can we move all these signals into `signals.py`? So it's cleaner when we add more signals to listen for changes or modify the model.

docker-app/qfieldcloud/notifs/models.py
```diff
@@ -0,0 +1,260 @@
+"""
+This module logs activity streams using the `django-activity-stream` package.
+"""
+from django.db.models.signals import post_save, pre_delete
+from django.dispatch import receiver
+from django_currentuser.middleware import get_current_authenticated_user
```
suricactus, line 6: Not the best, but good enough workaround that we miss the request object.

docker-app/qfieldcloud/urls.py
```diff
@@ -61,4 +61,6 @@
     path("auth/", include("rest_framework.urls")),
     path("accounts/", include("allauth.urls")),
     re_path(r"^invitations/", include("invitations.urls", namespace="invitations")),
+    # We don't use those for now, but leaving that here until we're sure
+    # path("inbox/notifications/", include(notifications.urls, namespace="notifications")),
```
suricactus, line 65: What is this?

docker-app/qfieldcloud/core/models.py
```diff
@@ -144,6 +158,13 @@ class UserAccount(models.Model):
     is_email_public = models.BooleanField(default=False)
     avatar_uri = models.CharField(_("Profile Picture URI"), max_length=255, blank=True)
 
+    notifs_frequency = models.CharField(
+        max_length=25,
+        verbose_name=_("Email frequency for notifications"),
+        choices=NOTIF_CHOICES,
+        default=NOTIF_DAILY,
+    )
```
suricactus, line 166: Is it possible to have this a configuration like `notifs_frequency_m = models.UIntField(min=0, choices=IntChoices)`? This way it would be easier to change the configuration for the users.

docker-app/qfieldcloud/urls.py
olivierdalang: It's some preset views to manage notifications (provided by the notifications app), see https://github.com/django-notifications/django-notifications/blob/master/notifications/urls.py

I think it's more effort integrating those views in our view system than reimplementing... I'll just remove that comment

docker-app/qfieldcloud/core/models.py
olivierdalang: Currently there's just four different django_crons set up (SendNotificationsJobImmediately, SendNotificationsJobHourly, SendNotificationsJobDaily, SendNotificationsJobWeekly).
If we wanted this to be customizable, we'd run every minute for everybody, but would then need to store last execution time for each user.

Not sure it's worth going down that rabbit hole ?

docker-app/qfieldcloud/urls.py
suricactus: Ah, I see. Yeah, I guess removing the comments would be better.

docker-app/qfieldcloud/core/models.py
suricactus: hm, isn't it enough to have a `sent_at` field in the `Notification` model? Whenever the cron runs, you call:
```
notifs = Notification.objects.filter(user=user, sent_at__isnull=True)
send_notif_emails(notifs)
```
Or you mean that we don't have `sent_at` field at all?

olivierdalang: The issue is that if we wanted to make it easy to customize the frequency using an integer (minutes) rather than predefined choices, we'd also need some kind of scheduling strategy.

Currently it uses 4 different crons, on for each frequency, so we have predefined/hard-coded available schedules only.

It's not impossible to make it customizable, but then we'd need to store when the cron was last sending emails run for every user as they'd all run on the same minutely cron. IMO not really worth the complexity.

We could add more predefine frequencies if you think it's needed, but I think with immediate (minute), hourly, daily and weekly we've got most need covered ?

olivierdalang: Ok as per discussion implemented this (hehe not sure why this was overcomplex in my head ?). Used a DurationField instead of int field so it simplifies date arithmetic functions

docker-app/qfieldcloud/core/migrations/0046_useraccount_notifs_frequency.py
```diff
@@ -0,0 +1,32 @@
...
+            name="notifs_frequency",
+            field=models.DurationField(
+                blank=True,
+                choices=[
+                    (datetime.timedelta(0), "Immediately"),
+                    (datetime.timedelta(seconds=3600), "Hourly"),
+                    (datetime.timedelta(days=1), "Daily"),
+                    (datetime.timedelta(days=7), "Weekly"),
+                    (None, "Disabled"),
+                ],
```
suricactus, line 26:
```suggestion
                choices=[
                    (datetime.timedelta(0), _("Immediately")),
                    (datetime.timedelta(seconds=3600), _("Hourly")),
                    (datetime.timedelta(days=1), _("Daily")),
                    (datetime.timedelta(days=7), _("Weekly")),
                    (None, _("Disabled")),
                ],
```

PR #61, opened 2021-07-28
Change github workflow to push containers to public registry

robert197:
The goal of this PR is to build the public containers and push to public registry.

We will use the public application container as a base for our private containers.

In order to make it work completely we need to set following secrets inside the github secrets settings:

- DOCKER_HUB_USERNAME
- DOCKER_HUB_ACCESS_TOKEN


suricactus:
Awesome, thanks. @m-kuhn can we set the secrets to public?

PR #69, opened 2021-08-26
DRAFT workers labels

olivierdalang:
Initial implementation for workers labels

DONE
- [x] Added a new `worker_label` text field to models.Project
- [x] Added a new settings that makes the match between labels and docker images
- [x] Added a second worker in the docker-compose using QGIS 3.18 (based on the same dockerfile but different base image - made the dockerfile configurable by the way) (not yet tested, just so that we have two different workers to test)
- [x] Added `image` property to the workers definitions in the docker-compose, so the images get tagged

TODO
- [ ] Probably worth thinking how we want to deal with docker image builders (build in the same repo ? how do we version ? etc.)
- [ ] Test a little bit (did not really test if it works yet)
- [ ] Tests (automated)

Thoughts/remarks :

- This will still require many changes the day we want to support external workers, but we should be able to keep the `worker_image` field (hence will be backwards compatible for users)

- Not sure at all how this would work in a kubernetes setup. I guess in such a case, we'd need to host the images to a docker repo ?

- Initially I started with a more flexible approach where each project would just specify the docker image name instead of some hard-coded label, but I don't think it's going to be maintainable and poses more security issues. This means that now you need a migration if you are to add new workers, so not sure how it would work with external workers too... I'm not too sure how well this will all work in the end, maybe we need to discuss usage scenarios more in detail.

- Adding the worker definitions in docker-compose is quite convenient, but a bit messy (we for instance don't really want to start it on docker-compose up). We could keep them in a separate docker-compose, but these docker-compose files are really starting to proliferate too much :-)


mbernasocchi:
Task linked: [DEV-2872 Support multiple labeled qgis workers](https://app.clickup.com/t/2192114/DEV-2872)

suricactus:
Closing for now, as it is not expected to be merged soon and there are already a bunch of conflicts. Keeping the branch though!

PR #73, opened 2021-09-10
Get rid of host/qgis_tempdir confusion and use shared_tempdir only

suricactus:


mbernasocchi:
Task linked: [DEV-3147 Don't use multiple tmp dirpaths](https://app.clickup.com/t/2192114/DEV-3147)

PR #74, opened 2021-09-10
Create command to create superuser directly from command line

robert197:


docker-app/qfieldcloud/core/management/commands/create_superuser.py
```diff
@@ -0,0 +1,43 @@
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand
+
+class Command(BaseCommand):
+    """
+    The standard django command to create superuser does not allow to pass the password as argument.
+    With this command this is possible. Please use it only for testing purposes.
```
suricactus, line 7:
```suggestion
    Creates a normal or super user using the CLI. Unline the django's createsuperuser command, here we can pass the password as an argument. This is a utility function that is expected to be used only for testing purposes.
```

docker-app/qfieldcloud/core/management/commands/create_superuser.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    help = 'Create a super user with given username, email and password'
```
suricactus, line 10:
```suggestion
    help = 'Create a normal or super user with given username, email and password'
```

suricactus (approved):

PR #84, opened 2021-09-21
Fix adding a new user using the admin UI

suricactus:


olivierdalang (commented):
As far as I can tell, I think this is OK as the form only allows to create users of TYPE_USER, which support all these inlines.

However :
1. It may be safer for long term maintenance of the code to return false by default, meaning you'd have to first create the user, then only you can edit inlines, so that you always use the same logic (could make a difference if at some point we have a common base admin for users/organisations for instance). This is esp. true for UserAccount, which if I remember correctly is generated automatically anyway with a signal ?
2. This completely bypasses django's permissions system (but it's the case before the PR already).

suricactus:
1. I think there are still situations where manually adding a user is an option. E.g. adding a support user.
2. Good note. However, I think it is outside the scope of this PR.

marioba (approved):

mbernasocchi:
Task linked: [DEV-3165 Setup qfieldcloud support account](https://app.clickup.com/t/2192114/DEV-3165)

PR #85, opened 2021-09-22
Bump libqfieldsync

suricactus:
- Fix an error with basemap generation if the basemap layer is not present
- Add an option to disable basemap generation if needed


mbernasocchi:
Task linked: [DEV-3181 Disable basemap generation on QFieldCloud to save resources and time](https://app.clickup.com/t/2192114/DEV-3181)

PR #89, opened 2021-09-23
Support whitespace in the filename

suricactus:
if the uplaoded project file was names as `project   filename.qgs`, qfieldcloud was failing to perform any async task on it. this is no longer the case. Hide and seek championship goes to the entrypoint and quoting "$@"


marioba (approved):

mbernasocchi:
Task linked: [DEV-3186 Support whitespaces in project names](https://app.clickup.com/t/2192114/DEV-3186)

PR #90, opened 2021-09-24
Faster list files using list_objects and list_object_versions

suricactus:
There was an extra HEAD operation for each iteration, which proved to be quite slow. Now all happens in a single (well, maybe a few since it's 1000 objects per request) request.

### DIFF
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -326,18 +326,6 @@ def list_versions(
         yield S3ObjectVersion(name, v)
 
 
-def get_prev_string_for_sorting(s: str) -> str:
-    """Get a string that is one string before the given string, if sorted alphabetically."""
-    if len(s) == 0:
-        return s
-
-    s, c = s[:-1], ord(s[-1])
-    if c > 0:
-        s += chr(c - 1)
-    s += "".join(["\u7FFF" for _ in range(10)])
-    return s
-
-
 def list_files_with_versions(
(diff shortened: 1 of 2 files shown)


mbernasocchi:
Task linked: [DEV-3192 Faster S3 list files](https://app.clickup.com/t/2192114/DEV-3192)

docker-app/qfieldcloud/core/utils.py
```diff
@@ -296,3 +308,69 @@ def get_s3_object_url(
...
+def get_prev_string_for_sorting(s: str) -> str:
+    """Get a string that is one string before the given string, if sorted alphabetically."""
+    if len(s) == 0:
+        return s
+
+    s, c = s[:-1], ord(s[-1])
+    if c > 0:
+        s += chr(c - 1)
+    s += "".join(["\u7FFF" for _ in range(10)])
+    return s
```
olivierdalang, line 338: I don't see it being used, an not really clear what this does.

Or is that a test to see if I actually review the code ? :-D

olivierdalang (approved):
Ok looks good

docker-app/qfieldcloud/core/utils.py
suricactus: Haha, got me! 

Well, this was a artefact from the first implementation I used, that was not asuming the files are ordered. Since I realized that S3 always returns the object keys sorted, so there is no need for this magic anymore.

suricactus: Will remoe it, thanks.

PR #93, opened 2021-09-28
Support multiple tokens per user when the user logs in

suricactus:
Previously the login was handled by Django REST framework and Django REST auth. However, both of the modules does not support multiple tokens per user. As a result, once you logout from qfieldsync, your session on QField is destroyed too. 

With the following changes there is a new custom DB backend that stores the data about sessions. In addition there are columns that track:
- the token expiration
- the token last usage
- user-agent
- based on the user-agent, the API client is determined

### Features

- full controll over the login tokens
- allows some clients to have a single token (like QField)
- the logins are now tracked
- removed Django REST auth as it is no longer maintained

### API changes

- there is no way to register a new user using the command line anymore. This fixes a number of issues with non-filled fields via command line.

### Limitations

It is important to note the session in the browser does not use the same token backend.


nirvn:
@suricactus , nice.

docker-app/qfieldcloud/urls.py
```diff
@@ -54,9 +54,9 @@
     ),
     path("docs/", schema_view.with_ui("redoc", cache_timeout=0), name="schema-redoc"),
     path("admin/", admin.site.urls),
-    path("api/v1/auth/registration/", include("rest_auth.registration.urls")),
-    path("api/v1/auth/token/", auth_views.AuthTokenView.as_view()),
-    path("api/v1/auth/", include("rest_auth.urls")),
+    path("api/v1/auth/login/", auth_views.LoginView.as_view()),
+    path("api/v1/auth/token/", auth_views.LoginView.as_view()),
```
olivierdalang, line 58: Is it expected that the token view is the same as loginview ? Does that mean that to refresh the token, the app must send (and thus store) the login ?

Don't we need an endpoint to just get a new token, using the existing one ?

olivierdalang (commented):
I did a quick search and came across https://github.com/James1345/django-rest-knox, did you consider using that ? It looks maintained and largely used, and seems it aims to solve exactly your issue, and has some other benefits (encrypts tokens in the DB which is for sure a good idea, esp. with tokens that last 30 days).

Otherwise looked at the code, appart from my comment, it looks fine (though not super easy to review since it also reuses existing code). I would just suggest to put it in a separate app, so it's easier to swap if one day we need to and so it doesn't clutter core too much.

docker-app/qfieldcloud/urls.py
suricactus: We always had it like this, which is wrong in many ways. It can be addressed in another PR.

PR #95, opened 2021-09-29
Run the worker `apply_delta` without `--overwrite-conflicts`

suricactus:
We need to copy the command before appending to it, because it comes as static and is appended multiple times


mbernasocchi:
Task linked: [DEV-3264 Project's `overwrite_conflicts` flag is not taken into account](https://app.clickup.com/t/2192114/DEV-3264)

PR #96, opened 2021-09-29
Add permission checks whether the user can overwrite deltas

suricactus:


mbernasocchi:
Task linked: [DEV-3265 Forcefully overwrite the delta in case of conflict](https://app.clickup.com/t/2192114/DEV-3265)

PR #98, opened 2021-09-29
Get all the users involved into a project with their roles and role origins

suricactus:
This is the opposite of the already existing `Project.objects.for_user()`, now we add `User.objects.for_project`. It will return all the users that are somehow involved in the project: as project owner, organization owner, organiaztion admin, collaborator, member of a collaborating team. Note for a public project it will not return all the users.

Probably it is worth adding the users that have contributed in the past, but no longer are part of the project. Situation:
- `suricactus` owns a public project `OpenKebabMap`
- `suricactus` adds a few changes in the project via qfield
- `suricactus` becomes a vegan and transfers it to `kebab_destroyer_97`
- `suricactus` is no longer available in the people who contributed in the project

Easy to achive by looking in the `Delta.objects.filter(created_by=user)`

Anyway, I will add this in a follow-up PR on demand.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -25,8 +25,116 @@ from qfieldcloud.core.utils import get_s3_object_url
 # http://springmeblog.com/2018/how-to-implement-multiple-user-types-with-django/
 
 
+class UserQueryset(models.QuerySet):
+    """Adds for_project(user) method to the user's querysets, allowing to filter only users part of a project.
+
+    Users are annotated with the user's project role (`project_role`) and the origin of this role (`project_role_origin`).
+    If the project is public, it will return only the directly collaborated users.
+
+    Args:
+        project:               project to find users for
+
+    Usage:
+    ```
+    # List all users that are involved in OpenKebabMap.
+    Users.object.for_project(OpenKebabMap)
+    ```
+    """
+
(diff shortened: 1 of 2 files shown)


mbernasocchi:
Task linked: [DEV-3311 Add User.objects.for_project that returns all the users involved in a project](https://app.clickup.com/t/2192114/DEV-3311)

olivierdalang (approved):
Looks good, but so that there's less risk of inconsistencies when we change permission logic, maybe you could factor out the whole `permissions_config` object, and use the same for both `Project.objects.for_user()` and `User.objects.for_project()` ?

E.g. like this :
```
permissions_config = [
        # Direct ownership
        (
            lambda user: Q(owner=user),
            lambda project: Q(pk=project.owner.pk, user_type=User.TYPE_USER),
            V(ProjectCollaborator.Roles.ADMIN),
            V(ProjectQueryset.RoleOrigins.PROJECTOWNER),
        ),
        #...
] 
```

Btw it could be worth using namedtuples instead of just tuple so it's more clear what each item is.

But I find it also OK to keep like this, if so I'd just add a clear `Don't forget to also update XXX if you change this` comment on both sides.

suricactus:
I am afraid it might become harder to maintain if they are in the same tuple list. On the naming tuples - yes, I do agree would be better than list indeces. I will add the docs and merge.

PR #103, opened 2021-10-04
The response should contain all the keys if there was not packaging yet

suricactus:


mbernasocchi:
Task linked: [DEV-3351 Fix feedback of `GET /project/qfield-files/<id>` when there was no packaging yet](https://app.clickup.com/t/2192114/DEV-3351)

PR #104, opened 2021-10-05
Fix the new invitations command to use the correct website

suricactus:
- Fix the site name in `django_site`
- Use the same code as the rest of the invitations
- Remove the need for the request object, so we can send invitations programatically
- Make sure the user is not registered yet


mbernasocchi:
Task linked: [DEV-3360 Fix the new invitations command to use the correct website](https://app.clickup.com/t/2192114/DEV-3360)

PR #113, opened 2021-10-07
Query all members and owners of an organization

suricactus:


mbernasocchi:
Task linked: [DEV-3374 When displaying the organization members, include the owner too](https://app.clickup.com/t/2192114/DEV-3374)

PR #118, opened 2021-10-13
Enable organization owner checks, but only if we are updating an existing org

suricactus:


duke-nyuki:
Task linked: [CU-1mjp3vy Fix organization_owner not being set when validating the model](https://app.clickup.com/t/1mjp3vy)

PR #127, opened 2021-10-15
Guess QFieldSync useragent properly and allow only one token from QFieldSync

suricactus:


duke-nyuki:
Task linked: [CU-1mjqz7g Correctly guess the user-agent from QGIS](https://app.clickup.com/t/1mjqz7g)

PR #153, opened 2021-11-19
Only suggest teams that are part of the project-owning organization

suricactus:
Avoid situations like this when teams from other organizations are suggested:
image


duke-nyuki:
Task linked: [QF-1417 Disable listing all the teams in the collaborators suggestions](https://app.clickup.com/t/2192114/QF-1417)

PR #166, opened 2021-12-06
Use nginx instead of caddy as a web server

suricactus:
- it always has SSL certificate, thanks to the `mkcert` container. It allows to add it as trusted certificate locally.
- this local certificate is used to startup nginx
- if QFieldCloud is deployed, one can use `init_letsencrypt.sh` to create a real certificate.


.github/workflows/test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Export the env variables file
         run: |
           cp .env.example .env
-          export $(egrep -v '^#' .env | xargs)
+          set -a; source .env; set +a
```
suricactus, line 29: Quite a trickery to load an `.env` file. There are a few gotchas in case of whitespaces in the variable names. They should be quoted to make this work.

conf/nginx/templates/default.conf.template
```diff
@@ -0,0 +1,61 @@
...
+  location @proxy_to_app {
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Host $http_host;
+
+    proxy_redirect off;
+    proxy_pass http://django;
+  }
+
+  # location /minio {
```
suricactus, line 47: temporary disable minio, since it is not used for local development.

duke-nyuki:
Task linked: [QF-1425 Get rid of Caddy in favor of nginx docker-compose](https://app.clickup.com/t/2192114/QF-1425)

.github/workflows/test.yml
3nids: some are suggesting to rather use `export $(grep -v '^#' .env | xargs -d '\n')`
https://stackoverflow.com/a/20909045/1548052

suricactus: Didn't work with some of the variables, already tried that.

PR #168, opened 2021-12-06
Fix minio in the new nginx setup

suricactus:
However, downloading is still not working, but the access to the console is there. The proxy download will be fixed when we really start using minio for other than tests.


duke-nyuki:
Task linked: [QF-1463 Make minio work with the new nginx configuration](https://app.clickup.com/t/2192114/QF-1463)

PR #176, opened 2021-12-13
Dangerous race condition when invalidating rest of the tokens, it also invalidated the current token

suricactus:


duke-nyuki:
Task linked: [QF-1508 When uploading multiple files suddenly gets 401](https://app.clickup.com/t/2192114/QF-1508)

PR #177, opened 2021-12-14
Fix admin saving deltas

suricactus:
Well, it's funny I was so focused on the migrations I forgot to update the model itself.


duke-nyuki:
Task linked: [QF-1521 Cannot update deltas from the admin on master](https://app.clickup.com/t/2192114/QF-1521)

PR #208, opened 2022-01-03
API should return the latest package of the same project

suricactus:


m-kuhn:
Task linked: [QF-1577 When there is no need to trigger a new package job, the API should return the latest package of the same project, not the latest package in for all projects](https://app.clickup.com/t/2192114/QF-1577)

Issue #60, opened 2021-07-26
Refer to step outputs with "step0.returns.OUTPUT_NAME", instead of global per task "OUTPUT_NAME"


suricactus:
With the changes in #56 now we can define a list of steps per job. This way we can avoid overwriting the outputs, also it is more clear what comes from where.

PR #213, opened 2022-01-04
Improvements on the job execution inside the QGIS container

suricactus:
- a more refined workflow definition and validation
- versioned workflow output
- less settings to define a workflow step - less confusion
- a new "output" key in the root feedback object to avoid using arr indices
- better feedback when some tests are failing

Fixes https://github.com/opengisch/qfieldcloud/issues/60


duke-nyuki:
Task linked: [QF-1589 Refer to step outputs with "step0.returns.OUTPUT_NAME", instead of global per task "OUTPUT_NAME"](https://app.clickup.com/t/2192114/QF-1589)

Issue #211, opened 2022-01-03
Documentation doesn't succeed


stcz:
Thanks for the great project. I want to take a closer look at it and followed the instructions on the README.

I created a new Ubuntu 20.04 VM, installed `docker-compose` and checkout the master branch of this repository and followed the instructions.

The first error I receive is:
```
invalid sslmode value: "prefer # "prefer" OR "require" most of the times"
```
I solved this by removing the comment in `.env`.

Further I get:
```
ERROR: The Compose file './docker-compose.override.yml' is invalid because:
services.createbuckets.depends_on contains an invalid type, it should be an array
```
Referring to the [documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on), the depends_on condition is not supported anymore in docker-compose 3.x. I fixed this with:
```
--- docker-compose.override.local.yml   2022-01-03 18:36:06.136000000 +0000
+++ docker-compose.override.yml 2022-01-03 18:43:10.536000000 +0000
@@ -102,8 +102,7 @@
   createbuckets:
     image: minio/mc
     depends_on:
-      minio:
-        condition: service_healthy
+      - minio
     entrypoint: >
       /bin/sh -c "
       /usr/bin/mc config host add myminio ${STORAGE_ENDPOINT_URL} ${STORAGE_ACCESS_KEY_ID} ${STORAGE_SECRET_ACCESS_KEY};

```
After these changes the containers are built and running.

Then I'm able to access the WebUI and it seems working.

But if I'm running the tests (got this from [github workflows](https://github.com/opengisch/qfieldcloud/blob/master/.github/workflows/test.yml)):

```
docker-compose run app python manage.py test -v2 qfieldcloud
```
They will fail:
```
...
FAIL: test_change_and_delete_pushed_only_features (qfieldcloud.core.tests.test_delta.QfcTestCase)
FAIL: test_list_all_deltas_and_list_deltas_by_deltafile (qfieldcloud.core.tests.test_delta.QfcTestCase)
FAIL: test_non_spatial_delta (qfieldcloud.core.tests.test_delta.QfcTestCase)
FAIL: test_push_apply_delta_file (qfieldcloud.core.tests.test_delta.QfcTestCase)
FAIL: test_push_apply_delta_file_conflicts_overwrite_false (qfieldcloud.core.tests.test_delta.QfcTestCase)
[...]

PR #215, opened 2022-01-05
Fix docker-compose.override.local.yml Syntax Error

stcz:
This will fix a syntax error in `docker-compose.override.local.yml` mentioned in #211 

Referring to the documentation, the depends_on condition is not supported anymore in docker-compose 3.x.

I cannot estimate if there are any side effects by removing this, but i don't think so.

### DIFF
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -102,8 +102,7 @@ services:
   createbuckets:
     image: minio/mc
     depends_on:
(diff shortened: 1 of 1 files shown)


suricactus:
Nice catch!

suricactus:
I temporarily reverted this, since it broke the internal toolchain. Will have a look again in the near future.

PR #218, opened 2022-01-18
Download files using nginx x-accel-redirect, so the s3 redirect happens withing nginx

suricactus:
A huge improvement on consuming the API. However, the django API get's more tightly coupled with nginx now.


duke-nyuki:
Task linked: [QF-416 Download files directly to the S3 storage with internal redirect in nginx](https://app.clickup.com/t/2192114/QF-416)

PR #227, opened 2022-01-24
Increase timeouts when uploading huge files

### LINKED ISSUE #68: Network request timed out - Error / maximum Size of .GPKG file ?
Hi,
Until i could easily sync or upload our Project with the size of about 12MB to Qfieldcloud.

Today i added a giant layer with over 180.000 lines (we get this private-property-border-layer from our local government in Austria)
Adding this gib layer bloats the .gpkg to 145MB and the following Error occures while syncing.

Unbenannt2

With the result of freezing Qgis at40%  ...thats why i can actually not read the Protocol, cause i have to brutally-restart the crashed Qgis

Unbenannt

is there a known maximum size of .gpkg ? for cloud use right now ?

Cause if i export the same Project via cable to our tablet the project works surprisingly fine in Qfield ->even with the giant layer.
So the Problem must be a kind of uploadlimit to the Cloud, rather than ageneral filesize-regulation in Qfield.

greets and Thanks

Gernot

suricactus:
nginx has a default timeout of 60 seconds. No more. Fix #68

### DIFF
--- a/conf/nginx/templates/default.conf.template
+++ b/conf/nginx/templates/default.conf.template
@@ -45,6 +45,9 @@ server {
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_set_header Host $http_host;
 
+    proxy_read_timeout 300;
+    proxy_send_timeout 300;
+
     proxy_redirect off;
     proxy_pass http://django;
   }


duke-nyuki:
Task linked: [QF-1649 Timeout errors in QFieldSync when uploading large files](https://app.clickup.com/t/2192114/QF-1649)

PR #228, opened 2022-01-24
Ignore newcoming deltas with the same id

suricactus:
Proved to be very quirky to manage this from QField side. Therefore, just ignore the new comming delta with the same id, instead of throwing an error.


duke-nyuki:
Task linked: [QF-1657 Silently ignore duplicate deltas, instead of duplicate delta error](https://app.clickup.com/t/2192114/QF-1657)

nirvn:
@jmad1v07 , as discussed yesterday. That'll unlock synchronization for your devices in the field again :)

PR #231, opened 2022-01-27
Fix migrations on deltas for 3d geometry

suricactus:
@robert197 I think this was the issue you had. Found it on sentry.


duke-nyuki:
Task linked: [QF-1670 Fix migrations on deltas for 3d geometry](https://app.clickup.com/t/2192114/QF-1670)

PR #232, opened 2022-01-27
Fix migrations on deltas for nan coordinates in the geometry

suricactus:


duke-nyuki:
Task linked: [QF-1670 Fix migrations on deltas for 3d geometry](https://app.clickup.com/t/2192114/QF-1670)

PR #234, opened 2022-02-01
Add limit argument how many users should be invited per run

suricactus:


duke-nyuki:
Task linked: [QF-1579 Create a CRON job to invite 50 users at a time](https://app.clickup.com/t/2192114/QF-1579)

PR #241, opened 2022-02-02
Purge old versions of project files

olivierdalang:
This adds:
- a management command to purge old versions of files (`purge_old_file_versions`)
- a cronjob running this command once a day

It currently hardcodes the number of versions to keep (3 for COMMUNITY accounts, 10 for PRO accounts).

Some questions / things to consider:
- Using a number of versions to keep instead of a time span means that quickly pushing many changes is less safe than rarely pushing, which IMO is not nice for the user
- The command runs on all projects at once. Hard to predict how slow this will be in real life. We probably need to batch the queries somehow ? Do we have a way to test this without affecting actual data ?
- Unsure if a management command is the right place for this ? Other such tasks are directly implement in cron.py, maybe that's enough ?
- Do we want to audit these operations ?
- Do we want to add some kind of "model" layer to manage these files, to avoid duplicating code (e.g. how prefixes are built, or if/how these are audited)

@suricactus Let me know if it matches what is needed.

### DIFF
--- a/docker-app/qfieldcloud/core/cron.py
+++ b/docker-app/qfieldcloud/core/cron.py
@@ -1,6 +1,5 @@
 import logging
 
-from django.core.management import call_command
 from django_cron import CronJobBase, Schedule
 from invitations.utils import get_invitation_model
 
@@ -36,15 +35,3 @@ class ResendFailedInvitationsJob(CronJobBase):
         logger.info(
             f'Resend {len(invitation_emails)} previously failed invitation(s) to: {", ".join(invitation_emails)}'
         )
-
-
-class PurgeOldFileVersions(CronJobBase):
-    """Purges old version of files of all projects using the purge_old_file_versions
-    management command.
-    """
-
-    schedule = Schedule(run_every_mins=60 * 24 * 7)
-    code = "qfieldcloud.purge_old_file_versions"
-
-    def do(self):
-        call_command("purge_old_file_versions", "--force")
--- a/docker-app/qfieldcloud/core/cron.py
+++ b/docker-app/qfieldcloud/core/cron.py
@@ -43,7 +43,7 @@ class PurgeOldFileVersions(CronJobBase):
     management command.
     """
 
-    schedule = Schedule(runs_every_mins=60 * 24 * 7)
+    schedule = Schedule(run_every_mins=60 * 24 * 7)
     code = "qfieldcloud.purge_old_file_versions"
 
     def do(self):
--- a/docker-app/qfieldcloud/core/cron.py
+++ b/docker-app/qfieldcloud/core/cron.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.core.management import call_command
 from django_cron import CronJobBase, Schedule
 from invitations.utils import get_invitation_model
 
@@ -35,3 +36,15 @@ class ResendFailedInvitationsJob(CronJobBase):
         logger.info(
             f'Resend {len(invitation_emails)} previously failed invitation(s) to: {", ".join(invitation_emails)}'
         )
+
+
+class PurgeOldFileVersions(CronJobBase):
+    """Purges old version of files of all projects using the purge_old_file_versions
+    management command.
+    """
+
+    schedule = Schedule(runs_every_mins=60 * 24 * 7)
+    code = "qfieldcloud.purge_old_file_versions"
+
+    def do(self):
+        call_command("purge_old_file_versions", "--force")
--- a/docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
+++ b/docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
@@ -2,15 +2,17 @@ from collections import defaultdict
 
 from django.core.management.base import BaseCommand, CommandError
 from qfieldcloud.core import utils
-from qfieldcloud.core.models import Project
+from qfieldcloud.core.models import Project, UserAccount
 
 
 class Command(BaseCommand):
-    """
-    Deletes old versions of files
+
+    help = """
+    Deletes old versions of files. Will keep only the 3 most recent versions
+    for COMMUNITY accounts and the 10 most recent for PRO accounts.
     """
 
-    PROMPT_TXT = "This will purge old files for all projects. Type 'yes' to continue, or 'no' to cancel: "
+    PROMPT_TXT = "This will purge old files for all projects. Rerun with --force, or type 'yes' to continue, or 'no' to cancel: "
 
     def add_arguments(self, parser):
         parser.add_argument(
@@ -23,35 +25,41 @@ class Command(BaseCommand):
             action="store_true",
             help="Prevent confirmation prompt when purging all projects",
         )
-        parser.add_argument(
-            "--keep_count", type=int, default=10, help="How many versions to keep"
-        )
 
     def handle(self, *args, **options):
 
         # Determine project ids to work on
-        projects_ids = options.get("projects")
-        if not projects_ids:
-            if options.get("force") is not True and input(Command.PROMPT_TXT) != "yes":
-                raise CommandError("Collecting static files cancelled.")
-            projects_ids = Project.objects.values_list("id", flat=True)
-        else:
-            projects_ids = projects_ids.split(",")
+        proj_ids = options.get("projects")
 
         # Get the affected projects
-        projects_qs = Project.objects.all()
-        if projects_ids:
-            projects_qs = projects_qs.filter(pk__in=projects_ids)
+        if not proj_ids:
+            if options.get("force") is not True and input(Command.PROMPT_TXT) != "yes":
+                raise CommandError("Collecting static files cancelled.")
+            proj_instances = Project.objects.all()
+        else:
+            proj_instances = Project.objects.filter(pk__in=proj_ids.split(","))
+
+        # We'll need useraccount type
+        proj_instances = proj_instances.prefetch_related("owner__useraccount")
 
         bucket = utils.get_s3_bucket()
+        for proj_instance in proj_instances:
 
-        for project_id in projects_ids:
-            print(f"Processing project {project_id}")
+            print(f"Processing {proj_instance}")
 
-            prefix = f"projects/{project_id}/files/"
-            keep_count = options.get("keep_count")
+            # Determine account type
+            account_type = proj_instance.owner.useraccount.account_type
+            if account_type == UserAccount.TYPE_COMMUNITY:
+                keep_count = 3
+            elif account_type == UserAccount.TYPE_PRO:
+                keep_count = 10
+            else:
+                print(f"⚠️ Unknown account type - skipping purge ⚠️")
+                continue
+            print(f"Keeping {keep_count} versions")
 
-            # All version under prefix
+            # Get all files versions for that project
+            prefix = f"projects/{proj_instance.pk}/files/"
             all_versions = bucket.object_versions.filter(Prefix=prefix)
 
             # Organize the versions by file in a dict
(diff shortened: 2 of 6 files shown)


duke-nyuki:
Task linked: [QF-1550 Add functionality to automatically remove old versions per project](https://app.clickup.com/t/2192114/QF-1550)

suricactus:
Thanks, @olivierdalang 

> * a management command to purge old versions of files (`purge_old_file_versions`)
> * a cronjob running this command once a day

One question here, I do agree the command is nice, but I think this command will be better in `utils2.storage` so it can be easily called from Django if needed (I mean line 62 onwards). E.g. if you downgrade the number of versions to be kept from the UI, they can be deleted synchroniously, should not be considered super expensive operation with a bunch of API calls between Django and Storage API. Something like `delete_and_limit_file_versions(prefix: str, count: int = 10)`.

The django command can just call that utils function? What do you think, does it make sense?

> It currently hardcodes the number of versions to keep (3 for COMMUNITY accounts, 10 for PRO accounts).

It is ok for now.

> * Using a number of versions to keep instead of a time span means that quickly pushing many changes is less safe than rarely pushing, which IMO is not nice for the user

Yes, we considered this. If you think it would be easy to implement from now, you can add it as an alternative filter, e.g. Something like `delete_and_limit_file_versions(prefix: str, count: int = None, since: datetime = None)`. The problem with this approach is that one might have too many versions for short period of time. @marioba what do you think?

> * The command runs on all projects at once. Hard to predict how slow this will be in real life. We probably need to batch the queries somehow ? Do we have a way to test this without affecting actual data ?

Nothing that comes to my mind. I guess you need to write your own script for this.

> * Unsure if a management command is the right place for this ? Other such tasks are directly implement in cron.py, maybe that's enough ?

Already commented, I would say a utility function called by the management command.

> * Do we want to audit these operations ?

Not sure if audit or extensive logging. @marioba ideas? I would say no.

> * Do we want to add some kind of "model" layer to manage these files, to avoid duplicating code (e.g. how prefixes are built, or if/how these are audited)

YES! This is planned to happen after release. I hate we work with bare files, it's slow and cumbersome.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -35,3 +36,15 @@ def do(self):
...
             f'Resend {len(invitation_emails)} previously failed invitation(s) to: {", ".join(invitation_emails)}'
         )
+
+
+class PurgeOldFileVersions(CronJobBase):
+    """Purges old version of files of all projects using the purge_old_file_versions
+    management command.
+    """
+
+    schedule = Schedule(runs_every_mins=60 * 24 * 7)
```
suricactus, line 46: Too low, I would say at least once per hour.

docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
```diff
@@ -0,0 +1,95 @@
...
+                # The latest is not an old version
+                if version.is_latest:
+                    continue
+                old_versions_by_file[version.key].append(version)
+
+            # Process file by file
+            for filename, old_versions in old_versions_by_file.items():
+
+                # Sort by date (newest first)
+                old_versions.sort(key=lambda i: i.last_modified, reverse=True)
```
suricactus, line 77: Nice, so it will be relatively easy to develop cleanup by date.

docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
```diff
@@ -0,0 +1,95 @@
...
+
+                # Skip the newest N
+                old_versions_to_purge = old_versions[keep_count:]
+
+                # Debug print
+                all_count = len(old_versions)
+                topurge_count = len(old_versions_to_purge)
+                print(
+                    f"- {filename}: will purge {topurge_count} out of {all_count} old versions"
```
suricactus, line 86:
```suggestion
                    f'Purging {topurge_count} out of {all_count} old versions for "{filename}"...'
```

docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
```diff
@@ -0,0 +1,95 @@
...
+            print(f"Processing {proj_instance}")
+
+            # Determine account type
+            account_type = proj_instance.owner.useraccount.account_type
+            if account_type == UserAccount.TYPE_COMMUNITY:
+                keep_count = 3
+            elif account_type == UserAccount.TYPE_PRO:
+                keep_count = 10
+            else:
+                print("⚠️ Unknown account type - skipping purge ⚠️")
```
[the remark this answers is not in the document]
olivierdalang: Throwing an exception will stop the loop (thus one faulty user will stop the purge somewhere randomly, and only half of the projects will have been purged). Would one of `logger.error(...)` or `warnings.warn(...)` show in sentry ? Can you point me to another place in the code where such runtime checks are done ?

docker-app/qfieldcloud/core/cron.py
olivierdalang: I'm not exactly sure how cronjobs behave currently, but I guess this process can take some time in prod (ALL versions of ALL files of ALL projects)

Maybe it would be better to do that in a propre job queue ? It would allow to queue a project specific cleanup after file uploads, so that running it on all projects is only needed if we change the retention policy, and so that it's easier to scale if needed.

If we were to do that, would you suggest implementing this as a new Job class ?

docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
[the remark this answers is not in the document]
suricactus: Just search by `NotImplementedError` in the project. I am not sure how we can make sentry logging additional stuff, but I still believe and an obvious error is better than silent log message.

docker-app/qfieldcloud/core/cron.py
suricactus: > I'm not exactly sure how cronjobs behave currently, but I guess this process can take some time in prod (ALL versions of ALL files of ALL projects)

This is only once when we enforce this.

> Maybe it would be better to do that in a propre job queue ? It would allow to queue a project specific cleanup after file uploads, so that running it on all projects is only needed if we change the retention policy, and so that it's easier to scale if needed.

I do agree in the long term, but not for now. I would say we just call this function right after each upload. I know it's not super scalable, but it simple and will work for while, whithout all the complexity of queued jobs.

> If we were to do that, would you suggest implementing this as a new Job class ?

The current jobs implementation needs to be tweaked a bit if we go to that direction. Also, they are executed sequentially per project, while the clean-up job can be ran in parallel.

docker-app/qfieldcloud/core/management/commands/purge_old_file_versions.py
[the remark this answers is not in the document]
olivierdalang: Didn't find logging such issues in the code, only throwing exception. I changed to throw then, as long as there's a followup on these exceptions it's no big deal if for some time versions aren't purged.

docker-app/qfieldcloud/core/cron.py
olivierdalang: Ok did that:
- removed cron
- run purge_old_file_versions after uploads through API
- update tests accordingly

olivierdalang:
Ok this is ready for final review / merge (if tests pass)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -96,3 +100,54 @@ def remove_project_thumbail(project: "Project") -> None:  # noqa: F821
...
+        keep_count = 10
+    else:
+        raise NotImplementedError(f"Unknown account type {account_type}")
+
+    logger.debug(f"Keeping {keep_count} versions")
+
+    # Process file by file
+    for file in qfieldcloud.core.utils.get_project_files_with_versions(project.pk):
+
+        filename = file.latest.name
```
suricactus, line 128: I think this var is used only for logging. Can we remove it?

suricactus, line 128: I think this var is used only for logging. Can we remove it?

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -96,3 +100,54 @@ def remove_project_thumbail(project: "Project") -> None:  # noqa: F821
...
+        old_versions = file.versions
```
suricactus, line 129: Can you use `file.versions`  below directly, it's easier to keep track where it comes from, especially with typing enabled.

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -96,3 +100,54 @@ def remove_project_thumbail(project: "Project") -> None:  # noqa: F821
...
+
+        # Sort by date (newest first)
+        old_versions.sort(key=lambda i: i.last_modified, reverse=True)
+
+        # Skip the newest N
+        old_versions_to_purge = old_versions[keep_count:]
```
suricactus, line 135:
```suggestion
        old_versions_to_purge = sorted(file.versions, key=lambda i: i.last_modified)[-keep_count:]
```

Why I would prefer it:
- we make a copy of `file.versions`, instead of sorting in place,  which might affect the code in the future if the versions are cached.
- no need for the extra var `old_versions`

I would even move the sorting functionality directly to `S3ObjectWithVersions`. TBH the current QFC code already assumes that the versions are sorted chronologically by `boto3`, but I do agree and extra sorting would not hurt.

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -96,3 +100,54 @@ def remove_project_thumbail(project: "Project") -> None:  # noqa: F821
...
+
+        # Debug print
+        all_count = len(old_versions)
+        topurge_count = len(old_versions_to_purge)
+        logger.debug(
+            f'Purging {topurge_count} out of {all_count} old versions for "{filename}"...'
+        )
```
suricactus, line 142:
```suggestion
        logger.debug(
            f'Purging {len(old_versions_to_purge)} out of {len(old_versions)} '
            f'old versions for "{filename}"...'
        )
```

Sorry, I just hate extra variable mappings... I easily get confused by them.

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -191,6 +192,9 @@ def post(self, request, projectid, filename, format=None):
                 changes={filename: [None, new_object.latest.e_tag]},
             )
 
+        # Delete the old versions if the files
```
suricactus, line 195:
```suggestion
        # Delete the old file versions
```
I guess the `if` was a typo

suricactus (approved):
Nice, thanks! I would only ask to remove the extra renaming of already existing variables or object project properties, because I find it confusing and less maintainable.

docker-app/qfieldcloud/core/utils2/storage.py
olivierdalang: Ok changed so that it doesn't sort in place (didn't exactly take your rewrite as it changed the logic)

Yes indeed I saw that get_project_files_with_versions expects already sorted versions from boto3 (not sure boto3 guarantees that though ? may depend on the provider ?)

Anyway, since it's about deleting stuff, better be extra careful here.

docker-app/qfieldcloud/core/utils2/storage.py
olivierdalang: done

docker-app/qfieldcloud/core/utils2/storage.py
suricactus: yes, extra care makes sense.

PR #242, opened 2022-02-02
Fix storage size calculation to include old versions

olivierdalang:
I don't think this exactly matches what's used on disk (probably only stores changed chunks ?), because it just sums the size of all versions. I didn't find actual data usage exposed. 

Seems AWS has an API to calculate actual size per prefix (see https://stackoverflow.com/a/35095500/13690651), but didn't find anything like that with boto3...

Still, I think it makes sense to simply sum the versions, as it's easier to understand for users anyway.

Some questions:
- Currently, `Project.storage_size()` will result on API calls each time it is invoked, meaning that it's not really usable for anything else than displaying one at a time in the admin (which is the only usage of it currently as far as I can see). I guess we need to persist that in the database. But then when exactly do we update it ? In a cronjob ? Or rightaway after uploads ? Depends a bit how exactly we want to deal with quotas.

### DIFF
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -248,7 +248,7 @@ def get_deltafile_schema_validator() -> jsonschema.Draft7Validator:
 
 
 def get_s3_project_size(project_id: str) -> int:
-    """Return the size in MiB of the project on the storage, included the
+    """Return the size in MiB of the project on the storage, including the
     exported files and their versions"""
 
     bucket = get_s3_bucket()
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-1548 Complete storage size calculation function including versions](https://app.clickup.com/t/2192114/QF-1548)

docker-app/qfieldcloud/core/utils.py
```diff
@@ -249,15 +249,15 @@ def get_deltafile_schema_validator() -> jsonschema.Draft7Validator:
 
 def get_s3_project_size(project_id: str) -> int:
     """Return the size in MiB of the project on the storage, included the
```
suricactus, line 251:
```suggestion
    """Return the size in MiB of the project on the storage, including the
```

suricactus (approved):

PR #244, opened 2022-02-02
Allow running parallel jobs on the server

suricactus:


duke-nyuki:
Task linked: [QF-1687 Cannot download two projects in parallel](https://app.clickup.com/t/2192114/QF-1687)

PR #245, opened 2022-02-03
QF-1553: Restrict collaboration private projects

olivierdalang:
Implementation for QF-1553. collaborators can no longer be invited on private projects unless they are owned by an organisation they belong to.

I implemented this by adding another queryset class for ProjectCollaborator, which adds a `valid_only` method which filters out all invalid collaborator relationships. Invalid defined as `A membership to a private project not owned by an organization, or owned by a organization that the member is not part of is invalid`).

Such memberships are still allowed in the datamodel, so that they are not lost when ownership change.

It also contains tests (a bit redundent, testing the same behaviour both at the queryset level and through the API, but IMO worth keeping both as the queryset tests both project->user and user->projects relations).

Also I feel that this whole queryset approach is getting a bit out of hands :-D
Will maybe need revisiting when we battletest the application a bit more (I didn't inspect the generated SQL, not sure how it will perform on large datasets).


olivierdalang:
I need some more time to fix tests because of changes this introduced, but you can review implementation already if you want.

docker-app/qfieldcloud/core/models.py
```diff
@@ -987,6 +993,36 @@ def delete_project(sender: Type[Project], instance: Project, **kwargs: Any) -> N
         qfieldcloud.core.utils2.storage.remove_project_thumbail(instance)
 
 
+class ProjectCollaboratorQueryset(models.QuerySet):
+    def validate(self, keep=False):
```
suricactus, line 997:
```suggestion
    def validated(self, keep_invalid=False):
```

I would choose a different name, what if you go for "validated"? When you say "validate", I would expect to either return `true`/`false`, or to throw an Exception.

suricactus (approved):
Looks good to me. Nice tests too!

Yes, this queryset approach is exploding...

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -330,3 +336,53 @@ def test_public_projecs(self):
...
+            response = self.client.get(apiurl, follow=True)
+            self.assertEqual(response.status_code, 200)
+            json = response.json()
+            self.assertEqual(json["user_role"], user_role)
+            self.assertEqual(json["user_role_origin"], user_role_origin)
+
+        # Project is public, we have a public role
+        # TODO: this fails, is is expected the queryset does not contain public projects ?
+        # TODO: if so, it should 404 (or 403) instead of 500
+        # assert_role(user_role="reader", user_role_origin="public")
```
olivierdalang, line 387: @suricactus to review

suricactus: I would say we should not include public projects, yes. Or at least we should have a step to filter them. Returning 404 sounds good to me.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -101,18 +101,24 @@ class ProjectViewSet(viewsets.ModelViewSet):
...
-            include_public = True
 
         projects = Project.objects.for_user(self.request.user)
-        if not include_public:
-            projects = projects.exclude(
-                user_role_origin=ProjectQueryset.RoleOrigins.PUBLIC
+
+        # In the list endpoint, by default we filter out public projects. They can be
+        # included with the `include-public` query parameter.
+        if self.action == "list":
```
olivierdalang, line 109: I think this should fix the error 500 in test_project.py

(the queryset filtered out public projects, which makes sense in the list endpoint, but not in the get endpoint)

@suricactus Can you check this fix makes sense ? Did the get endpoint work for public projects so far ?

docker-app/qfieldcloud/core/tests/test_queryset.py
```diff
@@ -310,3 +310,63 @@ def p(proj, user):
...
+        # Initial user setup
+        u = User.objects.create(username="u")
+        o = Organization.objects.create(username="o", organization_owner=u)
+        p = Project.objects.create(name="p", owner=u, is_public=True)
+
+        u1 = User.objects.create(username="u1")
+
+        # A public project is readable
+        # TODO: this fails, is is expected the queryset does not contain public projects ?
+        # assert_role(p, u1, roles.READER, origins.PUBLIC)
```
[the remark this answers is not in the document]
olivierdalang: The docstring of `User.objects.for_projects` actually states that `If the project is public, it will return only the directly collaborated users.`, so this is fine

I expected `User.objects.for_projects` to by symmetric with `Project.objects.for_user`

docker-app/qfieldcloud/core/views/projects_views.py
suricactus: Ok, sounds good.

suricactus (approved):

PR #246, opened 2022-02-04
Fix needs_repackaging check, last_packaged < last_updated => need repackaging

suricactus:
Will also add tests so it never happens again.


duke-nyuki:
Task linked: [QF-1690 Deltas are applied when synced with QGIS, but not when synced with QField](https://app.clickup.com/t/2192114/QF-1690)

PR #247, opened 2022-02-04
QF-1555: Concept database schema for subscriptions

olivierdalang:
For discussion (so far only the datamodel part, no implementation at all).

@suricactus 
Additional changes compared to what we discussed: 
- renamed `UserAccountType` -> `AccountType` (we're namespaced under `subscription` app)
- renamed `ExtraPackageTypes` -> `ExtraPackage` (an user account has many ExtraPackages, not many ExtraPackagesTypes)
- Specific ExtraPackage as subclasses instead of types (more future-proof, also we don't mix megabytes and minutes in a "size" field)

Note that the draft in QF-1555 is not fully aligned with QF-234 (e.g. price field). Also it may be worth including already how it would work with billing, as it may have impact on the datamodel ?

Let me know what you think.


duke-nyuki:
Task linked: [QF-234 Implement the restrictions needed for the different subscriptions](https://app.clickup.com/t/2192114/QF-234)

duke-nyuki:
Task linked: [QF-1555 Design the database schema for subscriptions](https://app.clickup.com/t/2192114/QF-1555)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -112,13 +112,7 @@ def purge_old_file_versions(project: "Project") -> None:  # noqa: F821
...
 
     # Determine account type
-    account_type = project.owner.useraccount.account_type
-    if account_type == qfieldcloud.core.models.UserAccount.TYPE_COMMUNITY:
-        keep_count = 3
-    elif account_type == qfieldcloud.core.models.UserAccount.TYPE_PRO:
-        keep_count = 10
-    else:
-        raise NotImplementedError(f"Unknown account type {account_type}")
+    keep_count = project.owner.useraccount.account_type.storage_keep_versions
```
suricactus, line 115: The comment above is misleading, the account type is already determined in `useraccount`

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -0,0 +1,80 @@
...
+            AccountType.objects.update(is_default=False)
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return f"{self.display_name} ({self.code})"
+
+
+class ExtraPackageType(models.Model):
+    code = models.CharField(max_length=30, unique=True)
+    # TODO: match requirements in QF-234 (fields like automatic old versions)
```
suricactus, line 55: Not clear what you mean there?

docker-app/qfieldcloud/core/models.py
```diff
@@ -882,6 +881,11 @@ class Meta:
     data_last_updated_at = models.DateTimeField(blank=True, null=True)
     data_last_packaged_at = models.DateTimeField(blank=True, null=True)
 
+    repackaging_cache_expire = models.DurationField(
```
suricactus, line 884:
```suggestion
    repackaging_cache_expire_m = models.DurationField(
```

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -0,0 +1,80 @@
...
+    storage_keep_versions = models.PositiveIntegerField(default=10)
+    job_minutes = models.PositiveIntegerField(default=10)
+    can_add_storage = models.BooleanField(default=False)
+    can_add_job_minutes = models.BooleanField(default=False)
+    is_external_db_supported = models.BooleanField(default=False)
+    has_priority_support = models.BooleanField(default=False)
+    minimum_repackaging_time = models.DurationField(
+        default=timedelta(minutes=60),
+        validators=[MinValueValidator(timedelta(minutes=1))],
+    )
```
suricactus, line 38:
```suggestion
    can_configure_repackaging_cache_expire = models.BooleanField(default=False)
    min_repackaging_cache_expire_m = models.DurationField(
        default=timedelta(minutes=60),
        validators=[MinValueValidator(timedelta(minutes=1))],
    )
```
Did you ommit the unit of measure on purpose? On clickup it was specified as `minimum_repackaging_time_m`, so it is clear we store minutes?


Giving a second thought, now the name in `projects` and here match. I felt a bit confused by the `repackaging_time`, it sounded like a time of day (e.g. 23:30).

docker-app/qfieldcloud/core/models.py
olivierdalang: It's a duration field (stored as a postgres interval and converted to a python timedelta), so no units

docker-app/qfieldcloud/subscription/models.py
olivierdalang: same as above for the units, ok for the rename

docker-app/qfieldcloud/subscription/models.py
olivierdalang: (removed it, doesn't apply to extrapackagetype)

docker-app/qfieldcloud/core/utils2/storage.py
olivierdalang: fixed

docker-app/qfieldcloud/core/models.py
suricactus: you are right, sorry, I forgot to check it is no longer a simple `int` field

olivierdalang:
@suricactus One workflow passed and the other failed, but aren't they running the exact same commit ?

Looking into the error, I don't understand how it can happen, and not 100% sure it's related to this PR. Have these tests been flaky already ?

I think we could get rid of tearDown anyway, as TransactionTestCase already takes care of flushing the DB after each test case. The S3 storage would not get cleaned up, but since we use UUIDs that shouldn't cause trouble.

suricactus:
> Looking into the error, I don't understand how it can happen, and not 100% sure it's related to this PR. Have these tests been flaky already ?

They are. I think it is the worker wrapper doing so.

> I think we could get rid of tearDown anyway, as TransactionTestCase already takes care of flushing the DB after each test case. The S3 storage would not get cleaned up, but since we use UUIDs that shouldn't cause trouble.

Yes, agreed. It might be problem running locally though with files piling up, no?

olivierdalang:
> > I think we could get rid of tearDown anyway, as TransactionTestCase already takes care of flushing the DB after each test case. The S3 storage would not get cleaned up, but since we use UUIDs that shouldn't cause trouble.
> 
> Yes, agreed. It might be problem running locally though with files piling up, no?

IMO it's fine, one can reset the stack every now and then (`docker compose down --volumes`). In my case I do that sometimes already when resetting the DB when I switch between branches.

suricactus:
Then go for it.

docker-app/qfieldcloud/subscription/migrations/0001_initial.py
```diff
@@ -0,0 +1,147 @@
...
+                (
+                    "synchronizations_per_months",
+                    models.PositiveIntegerField(default=30),
+                ),
+                ("is_public", models.BooleanField(default=False)),
+                ("is_default", models.BooleanField(default=False)),
+            ],
+        ),
+        migrations.CreateModel(
+            name="ExtraPackageType",
```
suricactus, line 64: I am not 100% sure if the `Type` suffix is needed here?

PR #249, opened 2022-02-06
Restart nginx automatically and remove circular dependency

suricactus:


duke-nyuki:
Task linked: [QF-1701 Fix docker nginx does not restart](https://app.clickup.com/t/2192114/QF-1701)

PR #250, opened 2022-02-07
Add django-classy-tags requirement

suricactus:


duke-nyuki:
Task linked: [QF-1723 Allow permission checks in django templates to be stored as variable](https://app.clickup.com/t/2192114/QF-1723)

PR #251, opened 2022-02-08
Added delete_file_version function to deletes a specific file version

suricactus:
I have used the name "delete" as it is more straigh forward for me what it means. I guess the function you have recently added should be renamed too.

Ideally `purge_old_file_versions` should start using `delete_file_version`.


duke-nyuki:
Task linked: [QF-1549 Add ui functionality to remove old file versions](https://app.clickup.com/t/2192114/QF-1549)

PR #252, opened 2022-02-08
When a file is uploaded to a project, start a new process_projectfile job

suricactus:
This way the order in which files are uploaded is not that important anymore.
To save extra unnecessary process jobs, DCIM dir is ignored, since it
consists of staticfiles only.


duke-nyuki:
Task linked: [QF-1728 When a file is uploaded to a project, start a new process_projectfile job](https://app.clickup.com/t/2192114/QF-1728)

PR #253, opened 2022-02-09
QF-1551: Implement storage limit enforcement

olivierdalang:
(based on QF-1555 that must be merged/reviewed first)

Questions/comments:

- where exactly do we want to enforce this ? currently, it's done at the API level, but we may want to move this closer to the model if we want to reuse that same logic also from other places (admin, web frontend...). Same question for both file uploads and project reparenting
- currently, no caching is done about storage size, meaning that each time, each time the quota is checked (e.g. each upload), requests are made for **all versions of all files of all projects of the user**. I think we need to cache this at the project level. But maybe this already justifies investigating a bit more the idea of using django-storages and keep track of the files in the database ?


duke-nyuki:
Task linked: [QF-1555 Design the database schema for subscriptions](https://app.clickup.com/t/2192114/QF-1555)

duke-nyuki:
Task linked: [QF-1551 Implement the limitations for storage size limit](https://app.clickup.com/t/2192114/QF-1551)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -191,3 +192,13 @@ def delete_file_version(
         file_version._data.delete()
 
     return versions_to_delete
+
+
+def file_size_bytes(file: IO):
```
suricactus, line 197:
```suggestion
def file_size_bytes(file: IO) -> int:
```

suricactus: I think this is the best way to do it in python...

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -158,6 +158,15 @@ def post(self, request, projectid, filename, format=None):
             )
 
         request_file = request.FILES.get("file")
+
+        file_size_mb = file_size_bytes(request_file) / 1024 / 1024
```
suricactus, line 162:
```suggestion
        file_size_mb = request_file.size / 1024 / 1024
```
[Don't you have something like this?](https://docs.djangoproject.com/en/4.0/ref/files/uploads/#django.core.files.uploadedfile.UploadedFile.size)

suricactus:
> * where exactly do we want to enforce this ? currently, it's done at the API level, but we may want to move this closer to the model if we want to reuse that same logic also from other places (admin, web frontend...). Same question for both file uploads and project reparenting

Closer to the model, the better. I think the admin iface is the only exception to the enforcement of these checks.

> * currently, no caching is done about storage size, meaning that each time, each time the quota is checked (e.g. each upload), requests are made for **all versions of all files of all projects of the user**. I think we need to cache this at the project level. But maybe this already justifies investigating a bit more the idea of using django-storages and keep track of the files in the database ?

My two main concerns for "storing files info in the database" are:
- how are we sure we keep db and s3 in sync? 140% need for monitoring from day 0!
- how are we sure our transaction is finished: both s3 stored the file and the db saved a new record for the file?

If these two are solved, this would open a great area of optimizations. Give it a thought.

If not - project level cache is reasonable.

docker-app/qfieldcloud/core/views/files_views.py
olivierdalang: Indeed (got confused because that doesn't exist for io.BytesIO, but here we deal with a django UploadedFile)

docker-app/qfieldcloud/core/models.py
```diff
@@ -424,6 +424,28 @@ def avatar_url(self):
         else:
             return None
 
+    # TODO: definitely need to cache these, otherwise we must recompute from ALL projects each time !
+    def storage_quota_left(self) -> float:
```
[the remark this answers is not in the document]
olivierdalang: I think there's some kind of convention that says that property shouldn't be used for heavy lifting methods (as otherwise you can easily inadvertently make many queries), but will change once cache is in place

docker-app/qfieldcloud/core/migrations/0054_remove_project_files_count.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.2.11 on 2022-02-09 15:52
```
suricactus, line 1: Can we squash this migration with 0052?

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -158,6 +158,15 @@ def post(self, request, projectid, filename, format=None):
             )
 
         request_file = request.FILES.get("file")
+
+        file_size_mb = request_file.size / 1024 / 1024
+        quota_left = project.owner.useraccount.storage_quota_left
```
suricactus, line 163:
```suggestion
        quota_left_mb = project.owner.useraccount.storage_quota_left_mb
```

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -115,6 +116,33 @@ def get_queryset(self):
...
+        # But then it would also be enforced by admin, which we don't want I guess...
+
+        with transaction.atomic():
+            old_owner = serializer.instance.owner
+            super().perform_update(serializer)
+            new_owner = serializer.instance.owner
+
+            # If owner has not changed, no additional check is made
+            if old_owner == new_owner:
+                return
```
suricactus, line 134: I tried to thing if we can check this before opening the transaction, but failed to find a solution. Reminder just in case you have an idea to offload postgres bit.

docker-app/qfieldcloud/subscription/tests/test_storage_quota.py
```diff
@@ -0,0 +1,169 @@
...
+        u1 = User.objects.create(username="u1")
+        p1 = Project.objects.create(name="p1", owner=u1)
+
+        default = AccountType.get_or_create_default()
+
+        # Initially, we have the default account type storage
+        self.assertEqual(u1.useraccount.storage_quota_left, default.storage_mb)
+
+        # Changing account type changes the quota
+        micro = AccountType.objects.create(code="micro", storage_mb=1)
```
suricactus, line 40:
```suggestion
        account_type_1mb = AccountType.objects.create(code="account_type_1mb", storage_mb=1)
```

docker-app/qfieldcloud/subscription/tests/test_storage_quota.py
```diff
@@ -0,0 +1,169 @@
...
+        u1.useraccount.account_type = micro
+        u1.useraccount.save()
+        self.assertEqual(u1.useraccount.storage_quota_left, 1)
+
+        # Adding an extra package increses the quota
+        ep_med = ExtraPackageTypeStorage.objects.create(code="extra_med", megabytes=2)
```
suricactus, line 46:
```suggestion
        extra_2mb = ExtraPackageTypeStorage.objects.create(code="extra_2mb", megabytes=2)
```
Sorry, these cryptic names are already confusing for me.

docker-app/qfieldcloud/subscription/tests/test_storage_quota.py
```diff
@@ -0,0 +1,169 @@
...
+        # Deleting a version increases the quota
+        version = list(list_versions(bucket, storage_path))[0]
+        delete_file_version(p1, "test.data", version.id)
+        p1.save(recompute_storage=True)
+        self.assertEqual(u1.useraccount.storage_quota_left, 4)
+
+    def test_api_enforces_storage_limit(self):
+        u1 = User.objects.create(username="u1")
+        p1 = Project.objects.create(name="p1", owner=u1)
+        micro = AccountType.objects.create(code="micro", storage_mb=1)
```
suricactus, line 104:
```suggestion
        account_type_1mb = AccountType.objects.create(code="account_type_1mb", storage_mb=1)
```

docker-app/qfieldcloud/subscription/tests/test_storage_quota.py
```diff
@@ -0,0 +1,169 @@
...
+            apipath, {"file": self._make_file(mb=0.75)}, format="multipart"
+        )
+        self.assertEqual(response.status_code, 402)
+
+    def test_api_enforces_storage_limit_for_reparenting(self):
+        onemeg = AccountType.objects.create(code="onemeg", storage_mb=1)
+        twomeg = AccountType.objects.create(code="twomeg", storage_mb=2)
+        extra_onemeg = ExtraPackageTypeStorage.objects.create(
+            code="onemegextra", display_name="onemegextra", megabytes=1
+        )
```
suricactus, line 129:
```suggestion
        extra_1mb = ExtraPackageTypeStorage.objects.create(
            code="extra_1mb", display_name="onemegextra", megabytes=1
        )
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -912,25 +939,20 @@ def get_absolute_url(self):
...
     def private(self):
         # still used in the project serializer
         return not self.is_public
 
-    @property
+    @cached_property
     def files(self) -> Iterable[utils.S3ObjectWithVersions]:
-        return utils.get_project_files_with_versions(self.id)
+        """Gets all the files from S3 storage. This is potentially slow. Results are cached on the instance."""
+        return list(utils.get_project_files_with_versions(self.id))
```
suricactus, line 950: Why did you convert to a list?

suricactus:
Thanks! Except a few typing and naming details, I believe this matches the requirements.

I also added a few extra questions to explain the logic behind some decisions, but should not be considered blocking.

docker-app/qfieldcloud/core/models.py
```diff
@@ -877,6 +900,10 @@ class Meta:
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    # These cache stats of the S3 storage. These can be out of sync, and should be
+    # refreshed whenever retrieving/uploading files by passing `project.save(recompute_storage=True)`
+    storage_size = models.FloatField(default=0)
```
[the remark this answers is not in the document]
olivierdalang: ok

docker-app/qfieldcloud/core/migrations/0054_remove_project_files_count.py
olivierdalang: ok

docker-app/qfieldcloud/core/views/projects_views.py
olivierdalang: We could compare username in the json to the username of the instance, but not sure this is robust before validation (e.g. it could mix upper/lowercases, so that old_ownername != new_owndername, even if they represent the same object). And if we were to validate it before, we'd hit the DB anyway.

In any case, this doesn't happen too frequently, IMO not making a difference for the DB ? But I don't mind too much, I can change to the above idea if you prefer avoiding transaction (indeed a bit counterinutitive to do a change and rollback afterwards).

I switched to decorator syntax though.

docker-app/qfieldcloud/core/models.py
suricactus: Yes, clear! Thanks! Now I saw your commit message too.

docker-app/qfieldcloud/core/models.py
```diff
@@ -985,6 +1007,11 @@ def status(self) -> Status:
         else:
             return Project.Status.OK
 
+    def save(self, recompute_storage=False, *args, **kwargs):
```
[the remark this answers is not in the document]
suricactus: clear. thanks!

suricactus:
Second review from my side and seems this PR is ready. A few steps before merging is to fix the test and squash some extra commits, but seems we are ready to go.

suricactus (approved):

docker-app/qfieldcloud/core/views/projects_views.py
suricactus: It is good, thanks!

olivierdalang:
Ok I think we can squash this in just one commit ?

About 
> Closer to the model, the better. I think the admin iface is the only exception to the enforcement of these checks.

I didn't really change that, so it must still be done separately in each view. IMO it would be a good move, but only makes sense with a larger refactoring (moving all file related stuff to the Projects model, and letting that model take responsibility for adding/deleting files).

PR #254, opened 2022-02-09
Various suggestions to improve dev experience

olivierdalang:
This holds some suggestions to improve dev experience (to be reviewed commit by commit). Not a must as I can also do these changes just locally, but I guess it could be helpful for you too ?

### DIFF
--- a/docker-app/qfieldcloud/core/rest_utils.py
+++ b/docker-app/qfieldcloud/core/rest_utils.py
@@ -48,7 +48,7 @@ def exception_handler(exc, context):
             "detail": exc.detail,
         }
 
-    logging.exception(exc)
+    logging.info(exc)
 
     return Response(
         body,
--- a/docker-app/qfieldcloud/core/rest_utils.py
+++ b/docker-app/qfieldcloud/core/rest_utils.py
@@ -1,6 +1,7 @@
 import logging
 import sys
 
+from django.conf import settings
 from django.core import exceptions
 from qfieldcloud.core import exceptions as qfieldcloud_exceptions
 from rest_framework import exceptions as rest_exceptions
@@ -36,14 +37,16 @@ def exception_handler(exc, context):
     body = {
         "code": exc.code,
         "message": exc.message,
-        "debug": {
+    }
+
+    if settings.DEBUG:
+        body["debug"] = {
             "view": str(context["view"]),
             "args": context["args"],
             "kwargs": context["kwargs"],
             "request": str(context["request"]),
             "detail": exc.detail,
-        },
-    }
+        }
 
     logging.exception(exc)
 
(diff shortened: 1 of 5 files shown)


docker-app/qfieldcloud/core/rest_utils.py
```diff
@@ -1,14 +1,22 @@
...
 from rest_framework.response import Response
 
 logger = logging.getLogger(__name__)
 
 
 def exception_handler(exc, context):
+
+    # When running tests, we don't handle unexpected exceptions, as it swallows the
+    # stack trace, making debugging much more difficult
+    if sys.argv[1] == "test" and not isinstance(exc, QFieldCloudException):
```
suricactus, line 18: Is there a better way to get we are withing a test? E.g. `ENVIRONMENT=test` envvar or something? This might fail if the `test` command is renamed for whatever reason.

Not a blocker to merge, just looking for a cleaner solution.

olivierdalang: Yes I agree, but from what I found there's no out of the box way to detect that. I saw people suggesting to subclass django.test.runner.DiscoverRunner to add such a variable ?

Alternatively it we could use settings.DEBUG (which is more or less how django behaves with normal frontend view: if debug=False it shows the 500 page, if debug=True it shows the debug information).

But that would mean that the clients would not get a proper error json error message if debug=True. We would have to run with debug=False on deployment (including staging), but that's anyway a good idea IMO.

suricactus: yup, I got to the same conclusion a while ago.

The problem I have is that `DEBUG=True` is not directly linked to the tests. E.g. I can imagine one might want to run it on staging temporary to extract debug information, which might also alter the behaviour of some of the modules. 

Now I checked - we run CI tests with `DEBUG=1` on the CI, which I would propose to change, IMO tests with debug enabled should be run only locally for ... debugging.

Therefore I think subclass `django.test.runner.DiscoverRunner` is indeed probably the cleanest solution, which will improve our quality of life, unless I am misisng some extra complexity from it?

olivierdalang: I don't think there's any extra complexity, will make these changes

olivierdalang:
> 1. the request/response logging was implemented as such because there was a phase when I failed to develop it in caddy, then we had both caddy and nginx. Since we are now nginx only, I would propose to remove it completely from the source code.

Ok will do.

> 2. on `debugpy` -> nice addition! Is it possible to add a helper/util function `debug()`, that import the debug module, shows a message in the logs and starts the debug server? It would be much nicer dev experience.

Ok will do.

Closing this and will reopen separate PRs for each topic

PR #255, opened 2022-02-09
Add debugpy and document usage

olivierdalang:
Adds debugpy to the stack when debug=True and documents how to use it in readme.

(extracted from #254)


suricactus:
Love the last bit.

PR #261, opened 2022-02-10
Fix minio for local testing

olivierdalang:
We can simplify the minio situation by using a hostname that resolves both on the host and from within docker, so we can get rid of STORAGE_ENDPOINT_URL_EXTERNAL and have something more similar to what we have in production.

On Windows/Mac, that would be `http://host.docker.internal:9000` (see https://stackoverflow.com/a/67388365/)
On Linux, from what I read, `http://172.17.0.1:9000` should work (see https://stackoverflow.com/a/48547074/)

for reference, initial issue
> I'm trying out locally downloading files from qfieldsync (QGIS) but have issue with configuration (qfieldcloud redirects to http://minio:9000/ instead of http://localhost:8010/minio). I think it is due to utils.get_s3_client().generate_presigned_url(...) that doesn't make use of STORAGE_ENDPOINT_URL_EXTERNAL. The issue is specific to local testing (as on prod, STORAGE_ENDPOINT_URL and STORAGE_ENDPOINT_URL_EXTERNAL are the same), so not too much of an issue, but was wondering how you deal with that when manual testing ?


docker-compose.override.local.yml
```diff
@@ -98,6 +97,7 @@ services:
         retries: 5
     ports:
       - ${STORAGE_BROWSER_PORT}:9001
+      - 9000:9000
```
suricactus, line 100:
```suggestion
      - ${STORAGE_PORT}:9000
```

.env.example
```diff
@@ -15,10 +15,9 @@ STORAGE_ACCESS_KEY_ID=minioadmin
 STORAGE_SECRET_ACCESS_KEY=minioadmin
 STORAGE_BUCKET_NAME=qfieldcloud-local
 STORAGE_REGION_NAME=
-# Internal URL to the storage endpoint (from python code)
-STORAGE_ENDPOINT_URL=http://minio:9000
-# Public URL to the storage endpoint (external storage should be equivalent to STORAGE_ENDPOINT_URL, local development only, no trailing slash)
-STORAGE_ENDPOINT_URL_EXTERNAL=http://localhost:80/minio
+# URL to the storage endpoint (must be reachable both from withing docker and from the host)
+STORAGE_ENDPOINT_URL=http://172.17.0.1:9000
+# (on windows/mac, use http://host.docker.internal:9000 instead)
```
suricactus, line 20:
```suggestion
# URL to the storage endpoint either minio by default, or external (e.g. S3)
# the URL must be reachable both from within docker and from the host
# NOTE: to use minio on windows/mac, change the value to "http://host.docker.internal:8009"
STORAGE_ENDPOINT_URL=http://172.17.0.1:8009
STORAGE_ENDPOINT_PORT=8009
```

suricactus:
I use my machine extensively for other stuff too. Using ports within the 8000-9999 range causes a lot of conflicts.

PR #262, opened 2022-02-10
various suggestions cosmetics

olivierdalang:
Another round of suggestions (theoretically only cosmetic)

- [ ] use decorators to enable auditing on models (easier to see which model is audited and which one isn't)
- [x] moved pre/post-save/delete signals to save() delete() (unless required by loosely coupled apps, signals are not recommended as less readable/predictable (e.g. order of execution) - and they are functionally equivalent o)
- [x] included COMPOSE_FILES in .env so you don't need to type it or to provide a docker-compose wrapper script (and adapted readme and ci accordingly)
- [ ] use decorators in admin.py instead of registering the models at the end of the file

Let me know what you think (if there are some you like and some other you don't, I can open separate PRs)


suricactus:
> * [ ]  use decorators to enable auditing on models (easier to see which model is audited and which one isn't)
> * [ ]  use decorators in admin.py instead of registering the models at the end of the file

Funny enough I deliberately avoided decorators as less readable to get a quick overview what is and what isn't in the admin, let's say. TBH I don't have strong opinion on that, I am -0.001 to the idea, so let it be.

.env.example
```diff
@@ -1,3 +1,5 @@
+COMPOSE_FILE=docker-compose.yml:docker-compose.override.local.yml:docker-compose.override.dev.yml
```
suricactus, line 1:
```suggestion
COMPOSE_FILE=docker-compose.yml:docker-compose.override.local.yml:docker-compose.override.locall.yml
```

Note the recommended for local development is `local`, `dev` is for dev.qfield.cloud

olivierdalang:
> Funny enough I deliberately avoided decorators as less readable to get a quick overview what is and what isn't in the admin, let's say. TBH I don't have strong opinion on that, I am -0.001 to the idea, so let it be.

Ok moved those out of the branch.

PR #263, opened 2022-02-10
Debug improvements

suricactus:
Note the recommended local development docker-compose is the one with `local` suffix.

Add debugger to the worker wrapper too. It uses different port because:
- conflict with port exposure
- theoretically support parallel debuging, even though it would be clumsy with to VS code editors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ debugpy.wait_for_client()  # optional
 
 Or alternativley, prefix your commands with `python -m debugpy --listen 0.0.0.0:5678`.
 ```shell
-docker-compose run app python -m debugpy --listen 0.0.0.0:5678 manage.py test
+docker-compose run app -p 5678:5678 python -m debugpy --listen 0.0.0.0:5678 manage.py test
 ```
 
 Then, configure your IDE to connect (example given for VSCode's `.vscode/launch.json`, triggered with `F5`):
--- a/README.md
+++ b/README.md
@@ -103,10 +103,22 @@ Then, configure your IDE to connect (example given for VSCode's `.vscode/launch.
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "QFieldCloud - Remote attach",
+            "name": "QFC debug app",
             "type": "python",
             "request": "attach",
+            "justMyCode": false,
             "connect": {"host": "localhost", "port": 5678},
+            "pathMappings": [{
+                "localRoot": "${workspaceFolder}/docker-app/qfieldcloud",
+                "remoteRoot": "/usr/src/app/qfieldcloud"
+            }]
+        },
+        {
+            "name": "QFC debug worker_wrapper",
+            "type": "python",
+            "request": "attach",
+            "justMyCode": false,
+            "connect": {"host": "localhost", "port": 5679},
             "pathMappings": [{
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ debugpy.wait_for_client()  # optional
 
 Or alternativley, prefix your commands with `python -m debugpy --listen 0.0.0.0:5678`.
 ```shell
-docker-compose run app python -m debugpy --listen 0.0.0.0:5678 manage.py test
+docker-compose run app -p 5678:5678 python -m debugpy --listen 0.0.0.0:5678 manage.py test
```
olivierdalang, line 97: if using exec instead of run we wouldn't need to manually specify the port

docker-app/qfieldcloud/core/management/commands/dequeue.py
```diff
@@ -15,6 +16,13 @@
     ProcessProjectfileJobRun,
 )
 
+if os.environ.get("DEBUG_BUILD"):
+    import debugpy
+
+    debugpy.listen(("0.0.0.0", 5679))
+    print("debugpy ready for debugger on port 5679... 🐛")
```
olivierdalang, line 23: I'd suggest not starting debugpy by default (could happen that this lands on prod for some reason, in which case it's a huge security issue, and there's probably some overhead ?)

We could instead documenter debugging the workers as above: 
```
docker-compose run worker_wrapper python -m debugpy --listen 0.0.0.0:5678 manage.py dequeue
```

This way it's also more symetrical between app and worker_wrapper.

README.md
suricactus: Since I expose them from `docker-compose.local.yml` I guess this is extra that we don't need anymore.

suricactus: Wait, second though, why we need to run `exec`?

docker-app/qfieldcloud/core/management/commands/dequeue.py
suricactus: Ok, I see what you mean. I have removed it, now trying to provide docs how should it work....

README.md
olivierdalang: exec runs in the already running container with already open ports, while run will try to start another container (which I guess should fail due to already forwarded ports ?)

suricactus: > Since I expose them from `docker-compose.local.yml` I guess this is extra that we don't need anymore.

Hm, this does now work for whatever reason??? In docker compose I have `ports: ["5679:5679"]` in `docker-compose.override.local.yml` and I call:
```
docker-compose run worker_wrapper python -m debugpy --listen 0.0.0.0:5679 manage.py dequeue
``` 
and I cannot connect from VS code for whatever reason. If I remove `ports` from the compose definition and run:
```
docker-compose run -p 5679:5679 worker_wrapper python -m debugpy --listen 0.0.0.0:5679 manage.py dequeue
```
however it works? Ideas?

suricactus: Ah, as discussed on chat, when using `docker-compose run`, the ports are ignored so there is no conflict with already running containers, unless `--service-ports` is passed.

So finally what I would do:

- keep the ports in the docker compose config, so you can use `docker-compose exec manage.py tests`;
- recommend using `docker run -p 5678:5678 manage.py tests` in the docs;

PR #264, opened 2022-02-11
Reenable logs (e.g. when server error occurred)

suricactus:
Follow up of #256


suricactus:
In theory we can utilize JSON logging in k8s Loki. In practice we don't yet. I don't have strong opinion on either formatting, I got used to JSON output by now.

PR #265, opened 2022-02-11
Download file improvements, hiding the s3 urls; No more redirecting fallback

suricactus:
All files from s3 are now served either using the nginx X-Acccel redirect,
or directly served by django in DEBUG mode.

This way we got rid of the so problematic redirect that was breaking pretty
much every client.

Also, the `files/meta/<projectid>` endpoint looks a bit redundant, but
it all started with the `files/<projectid>` API being so unflexible. Too
late to change now.

### DIFF
--- a/docker-app/qfieldcloud/core/rest_utils.py
+++ b/docker-app/qfieldcloud/core/rest_utils.py
@@ -3,7 +3,6 @@ import logging
 from django.conf import settings
 from django.core import exceptions
 from qfieldcloud.core import exceptions as qfieldcloud_exceptions
-from qfieldcloud.testing import IN_TEST_SUITE
 from rest_framework import exceptions as rest_exceptions
 from rest_framework.response import Response
 
@@ -29,7 +28,7 @@ def exception_handler(exc, context):
     else:
         # When running tests, we rethrow the exception, so we get a full trace to
         # help with debugging
-        if IN_TEST_SUITE:
+        if settings.IN_TEST_SUITE:
             raise exc
         logging.exception(exc)
         exc = qfieldcloud_exceptions.QFieldCloudException(detail=str(exc))
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -1,10 +1,15 @@
 from __future__ import annotations
 
 import logging
+from pathlib import PurePath
 from typing import IO, List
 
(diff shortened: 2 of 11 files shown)


duke-nyuki:
Task linked: [QF-1741 Improvements when downloading project and user files](https://app.clickup.com/t/2192114/QF-1741)

olivierdalang:
Oops missed this PR, but commented on it on https://github.com/opengisch/qfieldcloud/pull/267 (that branch includes this too it seems)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -76,7 +76,7 @@ def file_response(
         response["redirect_uri"] = url
 
         return response
-    elif settings.DEBUG:
+    elif settings.DEBUG or settings.IN_TEST_SUITE:
```
olivierdalang, line 79: See discussion on #267

IMO this moves us away from testing the actual production stack (e.g. we wouldn't catch anymore bugs in the code just above and/or nginx config). Ideally we should find another way (either have download implementation that works without Nginx, or run the tests against the running docker compose stack)

suricactus: Yes, we have this difference between prod and test environment. However, downloading files is such a basic functionality, that 1) we would see it does not work immediately 2) we have semi integration tests, that are mostly focused on Django so far, so we don't test nginx at the moment anyway.

> either have download implementation that works without Nginx,

I am -1 for this, since I am the maintener of the client side of QFieldCloud, I would reduce networking on the clients at all costs. In this case it is proxying trough nginx which is already pretty good at proxying.

> run the tests against the running docker compose stack

I am +0.5 for this, because we would lose the debugger functionality, as all requests will be proxied with `nginx`.

The best of both worlds would be to have a dedicated integration tests for downloading files that uses `nginx` and we keep all the rest the same.

docker-app/qfieldcloud/core/rest_utils.py
```diff
@@ -29,7 +28,7 @@ def exception_handler(exc, context):
     else:
         # When running tests, we rethrow the exception, so we get a full trace to
         # help with debugging
-        if IN_TEST_SUITE:
+        if settings.IN_TEST_SUITE:
```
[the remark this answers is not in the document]
suricactus: I added a comment in `settings.py`

PR #266, opened 2022-02-11
If object storage (s3) returns non-ok status, show 404

suricactus:
I just don't feel ok we show the XML from the storage API.

### DIFF
--- a/conf/nginx/pages/404.html
+++ b/conf/nginx/pages/404.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-1742 If object storage (s3) returns non-ok status, show 404](https://app.clickup.com/t/2192114/QF-1742)

olivierdalang:
Maybe use 404 only for actual 404 errors, and 500 for all others, so at least it hints that some special error happened ? (otherwise the day we get a 404 error for an existing file it will be super tricky to debug)

suricactus:
Ok, I will add 500 for all the rest.

suricactus:
Ok, I will add 500 for all the rest. The only issue is that S3 return 403 not authorized by default.

PR #267, opened 2022-02-11
Make /package/latest/file responses use nginx X-Accel-Redirect instead of HTTP redirect

suricactus:
So no longer weird redirect cases for any type of files in QFC.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -23,7 +23,6 @@ from django.urls import reverse_lazy
 from django.utils.translation import gettext as _
 from model_utils.managers import InheritanceManager
 from qfieldcloud.core import geodb_utils, utils, validators
-from qfieldcloud.core.utils import get_s3_object_url
 from timezone_field import TimeZoneField
 
 # http://springmeblog.com/2018/how-to-implement-multiple-user-types-with-django/
@@ -420,7 +419,10 @@ class UserAccount(models.Model):
     @property
     def avatar_url(self):
         if self.avatar_uri:
-            return get_s3_object_url(self.avatar_uri)
+            return reverse_lazy(
+                "public_files",
+                kwargs={"filename": self.avatar_uri},
+            )
         else:
             return None
 
@@ -895,7 +897,10 @@ class Project(models.Model):
     @property
     def thumbnail_url(self):
         if self.thumbnail_uri:
-            return get_s3_object_url(self.thumbnail_uri)
+            return reverse_lazy(
+                "project_metafiles",
+                kwargs={"projectid": self.id, "filename": self.thumbnail_uri[51:]},
+            )
         else:
             return None
 
--- a/docker-app/qfieldcloud/core/tests/test_delta.py
+++ b/docker-app/qfieldcloud/core/tests/test_delta.py
@@ -4,9 +4,8 @@ import logging
 import time
 
 import fiona
-import requests
 import rest_framework
-from django.http.response import HttpResponse, HttpResponseRedirect
+from django.http.response import FileResponse, HttpResponse
 from qfieldcloud.authentication.models import AuthToken
 from qfieldcloud.core import utils
 from qfieldcloud.core.models import Job, Project, ProjectCollaborator, User
@@ -590,13 +589,13 @@ class QfcTestCase(APITransactionTestCase):
     def get_file_contents(self, project, filename):
         response = self.client.get(f"/api/v1/files/{project.id}/{filename}/")
 
-        if isinstance(response, HttpResponseRedirect):
-            response = requests.get(response.url)
-
         self.assertTrue(status.is_success(response.status_code))
         self.assertEqual(get_filename(response), filename)
 
-        return response.content
+        if isinstance(response, FileResponse):
+            return b"".join(response.streaming_content)
+        else:
+            return response.content
 
     def upload_deltas(self, project, delta_filename):
         delta_file = testdata_path(f"delta/deltas/{delta_filename}")
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -1,10 +1,15 @@
 from __future__ import annotations
 
 import logging
+from pathlib import PurePath
 from typing import IO, List
 
 import qfieldcloud.core.models
 import qfieldcloud.core.utils
+from django.conf import settings
+from django.core.files.base import ContentFile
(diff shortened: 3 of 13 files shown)


docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -590,13 +589,13 @@ def test_change_and_delete_pushed_only_features(self):
     def get_file_contents(self, project, filename):
         response = self.client.get(f"/api/v1/files/{project.id}/{filename}/")
 
-        if isinstance(response, HttpResponseRedirect):
```
olivierdalang, line 593: For next time, it's cleaner to set `self.client.get(f"/api/v1/files/{project.id}/{filename}/", follow=True)`

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -26,6 +31,69 @@ def staticfile_prefix(project: "Project", filename: str) -> str:  # noqa: F821
...
+        else:
+            url = qfieldcloud.core.utils.get_s3_object_url(key)
+
+        # Let's NGINX handle the redirect to the storage and streaming the file contents back to the client
+        response = HttpResponse()
+        response["X-Accel-Redirect"] = "/storage-download/"
+        response["redirect_uri"] = url
+
+        return response
+    elif os.environ.get("DEBUG", 1):
```
olivierdalang, line 79: better to use `settings.DEBUG`, or maybe even don't add this alternative behavior so we keep dev as close as possible to prod ?

(here I tend to always run the full stack, and think it's acceptable to have download not work if accessing to django directly)

docker-app/qfieldcloud/core/models.py
```diff
@@ -420,7 +419,10 @@ class UserAccount(models.Model):
     @property
     def avatar_url(self):
         if self.avatar_uri:
-            return get_s3_object_url(self.avatar_uri)
+            return reverse_lazy(
```
olivierdalang, line 422: I think there's no need for lazy resolution here (even if it doesn't hurt either)

docker-app/qfieldcloud/core/models.py
```diff
@@ -895,7 +897,10 @@ class Meta:
     @property
     def thumbnail_url(self):
         if self.thumbnail_uri:
-            return get_s3_object_url(self.thumbnail_uri)
+            return reverse_lazy(
```
olivierdalang, line 900: same

olivierdalang (commented):
Not 100% sure I completely understand the change (from what I see in the code, it seems we were already using X-Accel-Redirect in DownloadPushDeleteFileView, and that this PR moves this to utils2.storage.file_response ?)

More generally, not clear to me what do we gain from having our own endpoint that returns the files instead of just using the S3 URL directly ?

Even with X-Accel-Redirect, couldn't this be heavy on our server (file downloads stream through the server, without any real benefit such as security or authentication as we just pass the same url to the S3) ?

docker-app/qfieldcloud/core/models.py
suricactus: There is a strong bias towards `reverse_lazy` in the codebase so far, so I keep it.

docker-app/qfieldcloud/core/models.py
suricactus: There is a strong bias towards `reverse_lazy` in the codebase so far, so I keep it.

docker-app/qfieldcloud/core/utils2/storage.py
suricactus: Ah, now I saw I made default to 1, which was dangerous. Thanks for highlighting this!

suricactus:
> > We gain a lot -> no longer the need to handle redirects. This automatically reduces QField and QFieldSync by thousands of line of code.
> 
> But if you use directly S3's URL, there would be no redirect, or did I miss something ? (I'm talking about just serving files directly from the S3 endpoint, without any proxy/redirect/whatsoever). No idea how much changes this means (probably too much if URLs are somehow hardcoded in Qfield or qfieldsync, probably not much if URLs are returned by the API). IMO that must be even less lines of code as we let S3 handle files completely.

We need to check if you have permissions to access that object first, so django is involved. Then we either should return a response with the actual resource URL or redirect. The first one is basically manual redirect, so we use HTTP redirect.

> 
> Otherwise, if you want to completely hide the storage backend from the client:
> 
> * why do we still need presigned urls ? shouldn't we remove that, as we're supposed to only access S3 internally ?

Because history and it is easier to manage. I do agree would be better if we directly access them, but needs extra time to do so. Having presigned URLs does not hurt anyone when they are internal.

> * not sure where the buckets config is defined in the codebase, but we should make sure that the buckets are not publicly accessible and/or objects created with private ACLs.

This will now be possible, because some objects were defined public before (e.g. avatars).

> (sorry if my comments don't make sense, I'm not too familiar with that part of the stack yet :-) )

Perfectly fine, I like getting questioned.

However, if you don't have any strong objections, I would susggest to make the discussed improvements (e.g. drop the presigned URLs) in a follow up PR.

suricactus:
I think I understood you now, you mean:
```
GET qfield.cloud/api/v1/files/path/to/file -> returns s3.com/path/to/file
GET s3.com/path/to/file
````

I find this not very convinient API to consume, as the download process will be two step process. It adds additional client logic which I am trying to avoid as much as possible when using QT networking, which I find huge pain, especially in C++.

**Comment by olivierdalang** [comment]
You can include the URL in the step before (in your example, you assume the client has already retrieved a list of files)
```
GET qfield.cloud/api/v1/project/myproject -> returns {'name': 'project', 'files': [s3.com/path/to/file1, s3.com/path/to/file2, s3.com/path/to/file3], ...}
GET s3.com/path/to/file1
```

suricactus:
The main problem I see there is that you get a presigned url with X mins expiration. If you start downloading after X minutes, the client MUST refresh presigned URLs which is extra logic. This is real world situation, since there is a limit of 4-8 concurrent requests from a client (browser or Qt) at point of time. So you never know how much time it passed since you got your presigned URL and the actual time of request.

Anyway, if we don't merge this, we would need to keep 3 bug reports open and adjust 3 API clients to be able to consume it. I am open to discuss this as a fututure API change, but would propose to merge this for now,  so I can continue with the upcoming release these days.

olivierdalang:
> Anyway, if we don't merge this, we would need to keep 3 bug reports open and adjust 3 API clients to be able to consume it. I am open to discuss this as a fututure API change, but would propose to merge this for now, so I can continue with the upcoming release these days.

Yes sure !

PR #269, opened 2022-02-15
remove useless tearDown methods

olivierdalang:
(didn't run the tests yet, let's see if they really are useless)


suricactus:
Seems all good? In the worst case we revert this PR.

suricactus (approved):

suricactus:
@olivierdalang I guess the failure here is related https://github.com/opengisch/qfieldcloud/runs/5225583774?check_suite_focus=true

olivierdalang:
see followup PR : https://github.com/opengisch/qfieldcloud/pull/271

PR #271, opened 2022-02-17
remove useless tearDown methods followup

olivierdalang:
restored some parts of tearDown logic whose removal is probably linked to test failures such as https://github.com/opengisch/qfieldcloud/runs/5225583774?check_suite_focus=true


suricactus:
Lets see if it better now.

PR #272, opened 2022-02-23
Add container_id field to the job model and remove jobs without an active container

suricactus:
This PR add the option to clean up jobs that no longer have a running container for whatever reason. If we don't do so, the dangling jobs prevent QFieldCloud from accepting new jobs, which blocks the usage of the QFieldCloud project.


duke-nyuki:
Task linked: [QF-1760 Cleanup jobs that are no longer running](https://app.clickup.com/t/2192114/QF-1760)

docker-app/worker_wrapper/wrapper.py
```diff
@@ -226,6 +227,9 @@ def _run_docker(
             detach=True,
         )
 
+        self.job.container_id = container.id
+        self.job.save()
+
```
olivierdalang, line 232: Potentially more trouble some race condition here, as some time could elapse between the job creation above (without container id) and the starting of the container. In the meantime the cronjob could delete the job ?

Maybe better to not create the job until you have a container ID (e.g. move the lines 99:101 to here). You can create directly a failed job if it doesn't complete.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -35,3 +43,33 @@ def do(self):
...
+            status=Job.Status.FAILED,
+            finished_at=timezone.now(),
+            feedback={
+                "error_stack": "",
+                "error": "Worker unexpectedly terminated.",
+                "error_origin": "worker_wrapper",
+                "container_exit_code": -2,
+            },
+            output="Job unexpectedly terminated.",
+        )
```
olivierdalang, line 75: Does this output an error to Sentry ? If not, I'd recommend to do so, so we know something is wrong (or are tasks monitored in a different way)

olivierdalang (commented):
Added some comments.

Just a general question (don't remember exactly the whole setup): how come jobs with no active container prevent new jobs to be picked up ? Aren't the worker just ignoring these jobs, that should to them look just like any other job currently being processed ?

Or do you mean it prevents these jobs from being restarted ?

.env.example
```diff
@@ -74,6 +74,9 @@ COMPOSE_PROJECT_NAME=qfieldcloud
 QFIELDCLOUD_DEFAULT_NETWORK=qfieldcloud_default
 QFIELDCLOUD_ADMIN_URI=admin/
 
+# Timeout in seconds to wait for a job container to finish, otherwise terminate it.
+QFIELDCLOUD_WORKER_TIMEOUT_S=60
```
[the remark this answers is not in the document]
suricactus: Yes, the default is configured to 600, but this is going to be used for tests which should be pretty much instant. So I am avoiding github tests running forever.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -35,3 +43,31 @@ def do(self):
...
+        jobs.update(
+            status=Job.Status.FAILED,
+            feedback={
+                "error_stack": "",
+                "error": "Worker unexpectedly terminated.",
+                "error_origin": "worker_wrapper",
+                "container_exit_code": -2,
+            },
+            output="Job unexpectedly terminated.",
+        )
```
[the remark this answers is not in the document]
suricactus: Nope, good note. Will try to send info to sentry.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -35,3 +43,31 @@ def do(self):
...
+
+    def do(self):
+
+        client = docker.from_env()
+        qgis_containers = client.containers.list(
+            sparse=True, filters={"ancestor": QGIS_CONTAINER_NAME}
+        )
+        qgis_container_ids = [c.id for c in qgis_containers]
+
+        jobs = Job.objects.filter(
```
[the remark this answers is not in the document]
suricactus: There are two possibilities:

1) the container finishes properly
2) the container terminates

In this piece of code we only care for case 2). If it happens between 58 and 60, the worst that can be done is that we exclude a container that is already terminated and we will be able to clean up on the next cron run. So I would say we are safe and better not lock the DB.

docker-app/worker_wrapper/wrapper.py
suricactus: Good note, this is very valid situation.

docker-app/qfieldcloud/core/cron.py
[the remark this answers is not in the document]
olivierdalang: Yes the concern is more in the other way around, if a container is not yet created but the job is, but that's the comment below, so all good I think

docker-app/qfieldcloud/core/cron.py
```diff
@@ -35,3 +44,38 @@ def do(self):
...
+    schedule = Schedule(run_every_mins=3)
+    code = "qfieldcloud.set_terminated_workers_to_final_status"
+
+    def do(self):
+
+        client = docker.from_env()
+        qgis_containers = client.containers.list(
+            sparse=True, filters={"ancestor": QGIS_CONTAINER_NAME}
+        )
+        qgis_container_ids = [c.id for c in qgis_containers]
```
olivierdalang, line 60: Btw didn't think of it before, but is this guaranteed to return all workers in the stack ? and not only on the current host ? Otherwise it will terminate perfectly fine containers !

I'm wondering if it wouldn't be safer to also check the `job.started_at` timestamp and only keep those that outlasted the timeout ?

suricactus: While this is pefrectly fine on composer setup, you are very right this could be dangerous on k8s setup where the containers are isolated from one another and this will return false negatives. Good catch #2.

suricactus: Should be good now.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -52,16 +53,12 @@ class SetTerminatedWorkersToFinalStatusJob(CronJobBase):
...
-
-        client = docker.from_env()
-        qgis_containers = client.containers.list(
-            sparse=True, filters={"ancestor": QGIS_CONTAINER_NAME}
-        )
-        qgis_container_ids = [c.id for c in qgis_containers]
-
         jobs = Job.objects.filter(
             status__in=[Job.Status.QUEUED, Job.Status.STARTED],
-        ).exclude(container_id__in=qgis_container_ids)
```
olivierdalang: We don't use the container_id anywhere else, do we ? if so shouldn't we drop it ?

Also, could it not now happen that a container now runs, exceeds timeout (not sure if this can happen ?), then its job is set to "FAILED" by the cronjob, and then it finishes (maybe even successfully) ? I guess we can't really kill the container, but maybe in the worker, we could do a last check to ensure the status is still "STARTED" before setting it to "SUCCESSFUL" ?

Maybe not too problematic though, and probably will never happen...

PR #275, opened 2022-02-23
No need to have an invitation to register

suricactus:
image


duke-nyuki:
Task linked: [QF-1766 Remove the need to have an invitation to register to QFC](https://app.clickup.com/t/2192114/QF-1766)

PR #281, opened 2022-02-28
Block requests by IP, only allow requests with proper host header

suricactus:


duke-nyuki:
Task linked: [QF-1779 Block requests by IP, only allow requests with proper host header](https://app.clickup.com/t/2192114/QF-1779)

PR #282, opened 2022-02-28
Fix first_name, last_name etc fields in token serializer 

suricactus:


duke-nyuki:
Task linked: [QF-1780 Fix first_name, last_name etc fields in token serializer](https://app.clickup.com/t/2192114/QF-1780)

PR #286, opened 2022-03-02
Add a utility CLI that creates missing useraccounts if a user/org is missing one

suricactus:
Why they are missing I am not sure, but it happened twice.


duke-nyuki:
Task linked: [QF-1788 Fix users without useraccount](https://app.clickup.com/t/2192114/QF-1788)

PR #287, opened 2022-03-06
Fix amazon s3 download as extra headers were proxied and s3 returns 403

suricactus:


duke-nyuki:
Task linked: [QF-1801 QFieldCloud returns 403 when amazon s3 is used](https://app.clickup.com/t/2192114/QF-1801)

PR #289, opened 2022-03-08
Return the already running packaging job instead of error

suricactus:


duke-nyuki:
Task linked: [QF-1810 if you try to start a new packaing job, but there is a packaging job already running ,the server replies with error. Then QField shows that error, instead of just waiting for that packaging job to finish](https://app.clickup.com/t/2192114/QF-1810)

PR #293, opened 2022-03-11
Add message if data source file is not found

suricactus:


duke-nyuki:
Task linked: [QF-1852 Show more useful feedback if layer source file cannot be found](https://app.clickup.com/t/2192114/QF-1852)

PR #298, opened 2022-03-15
Set all ports from .env variables

suricactus:
Make it easier to test with different environment at the same time.


duke-nyuki:
Task linked: [QF-1872 Make all external ports available as env variables](https://app.clickup.com/t/2192114/QF-1872)

Issue #299, opened 2022-03-17
QFieldCloud Jobs : very long time to Process QGIS Project File (SVG files)


jpdupuy:
Hello,

The project I created refers to 385 SVG files in the project file folder. (SVG file size between 1K and 7K)

I am currently working with QGIS 3.22.4 LTR and QFieldSync 4.0.0. 

The SVG project files were uploaded successfully to the cloud, however the project is in busy status for a long time
![image](https://user-images.githubusercontent.com/72449015/158773390-5e413c18-5b0c-49b9-a465-5190c52b9f3d.png)

Checking the jobs with the WebUI, I could notice that it takes 25 seconds per SVG file (pending -> started > finished>
![image](https://user-images.githubusercontent.com/72449015/158774538-87575618-5b5e-49ea-bd54-4e39e9c89c2c.png)

So for 385 files, it should take 385 /2 = around 3 hours before the project status change from busy to ready.
Could you find out why it takes so long, and/or fix this issue.

suricactus:
The problem has been located and immeadiate action will follow today. I will keep you posted when it will not result in an avalanche of jobs again.

jpdupuy:
Thanks for the fix. I will test it tomorrow. 
However the SVG files are not part of the DCIM directory since they are used as road signs (and refer to the SCG directory in QGIS). 
Should I consider to copy them in the DCIM directory ? The symbology refers to the SVG file as a feature attribut.

PR #300, opened 2022-03-21
Prevent starting a new job on file uploads if there is already pending one

suricactus:
Fix https://github.com/opengisch/qfieldcloud/issues/299#issuecomment-1070769037


duke-nyuki:
Task linked: [QF-1880 Jobs in non finished status stays like this for way too long](https://app.clickup.com/t/2192114/QF-1880)

PR #302, opened 2022-03-21
Fix the Delta.Method enums to inherit str, so the string comparison works

suricactus:
This fix directly affects the permissions check for role reporter, which were failing before, as `delta.method == Delta.Method.Create` was returning false.


duke-nyuki:
Task linked: [QF-1705 Role reporter does not work properly](https://app.clickup.com/t/2192114/QF-1705)

PR #306, opened 2022-03-26
Update Port where django is running

stcz:
Related to 866c5032d1f2bfebec369db939c1f371d58f2217 the django port changed to 8111 from 8000. This needs to be changed in the README.md


suricactus:
Took some time to have a look. Thanks!

PR #316, opened 2022-04-21
Extract QFieldSync settings and wkb type for each layer in project details

suricactus:


duke-nyuki:
Task linked: [QF-1922 Add qfieldsync actions and geometry type in `project_details`](https://app.clickup.com/t/2192114/QF-1922)

PR #320, opened 2022-04-28
Prepare the list of deltas to be applied on the time of creating the job

suricactus:
This PR aims to simplify the execution of delta apply jobs within the worker, by always using a predefined list of deltas. This list is created on the time of creation of new delta apply job.

The code should follow the coditions explained here https://docs.qfield.org/reference/qfieldcloud/jobs/#delta-apply-delta_apply-job 

The highlight of this PR is that the M:N `ApplyJobDelta` is populated only within the Django app and the worker-wrapper is just reading from that list.


duke-nyuki:
Task linked: [QF-1921 Fix apply delta to work without trying to apply all previous changes for all projects](https://app.clickup.com/t/2192114/QF-1921)

PR #323, opened 2022-05-03
QF-1009: disable offline non-geopackages layers for community accounts

olivierdalang:
Based on #247 which should be reviewed/merged first.

This implements a limitation preventing to use offline layers that are not geopackages for community accounts.

~WIP : implementation should be there, but struggling with the tests as it seems no matter what I try, the workflows don't want to fail... let's see if this also happens in CI !~
See comment in test_packages.py about this.


duke-nyuki:
Task linked: [QF-1009 Disable non GeoPackage layers in COMMUNITY](https://app.clickup.com/t/2192114/QF-1009)

docker-app/qfieldcloud/core/tests/test_packages.py
```diff
@@ -534,3 +542,45 @@ def test_filename_with_whitespace(self):
...
+            code="without_nongpkg",
+            display_name="without_nongpkg",
+            is_nongpkg_supported=False,
+        )
+        self.user1.useraccount.save()
+        self.upload_files(self.token1.key, self.project1, files=files)
+        status = self.wait_for_project_status(self.project1)
+        # TODO: REVIEW: the job fails (see below), but the project status is still ok !
+        # Is this expected ? What does project.status actually represent ?
+        # self.assertEqual(status, Project.Status.FAILED)
```
olivierdalang, line 582: @suricactus Can you check that ? Is that expected ?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -31,6 +31,7 @@ def get_or_create_default(cls) -> "AccountType":
     can_add_storage = models.BooleanField(default=False)
     can_add_job_minutes = models.BooleanField(default=False)
     is_external_db_supported = models.BooleanField(default=False)
+    is_nongpkg_supported = models.BooleanField(default=False)
```
olivierdalang, line 34: @suricactus Am I correct understanding `is_external_db_supported` and the new `is_nongpkg_supported` are unrelated ?

docker-qgis/process_projectfile.py
```diff
@@ -119,13 +122,20 @@ def on_project_read(doc):
...
     logging.info(
         f'QGIS project layer checks\n{layers_data_to_string(details["layers_by_id"])}',
     )
 
+    # TODO: REVIEW: this only checks against "nongpkg_not_supported" error, not other types of errors,
+    # yet the InvalidLayersException was already in the code (but never used). Is it so that we
+    # ignore invalid layers for any other invalidity ?
+    for details in details["layers_by_id"].values():
+        if details["error_code"] == "nongpkg_not_supported_by_account":
+            raise InvalidLayersException()
```
olivierdalang, line 137: @suricactus Here also, careful review needed

olivierdalang: (don't look at this)

docker-qgis/process_projectfile.py
```diff
@@ -119,13 +122,20 @@ def on_project_read(doc):
...
     logging.info(
         f'QGIS project layer checks\n{layers_data_to_string(details["layers_by_id"])}',
     )
 
+    # TODO: REVIEW: this only checks against "nongpkg_not_supported" error, not other types of errors,
+    # yet the InvalidLayersException was already in the code (but never used). Is it so that we
+    # ignore invalid layers for any other invalidity ?
+    for detail in details["layers_by_id"].values():
+        if detail["error_code"] == "nongpkg_not_supported_by_account":
+            raise InvalidLayersException()
```
olivierdalang, line 137: @suricactus Here also careful review needed

docker-app/qfieldcloud/subscription/models.py
suricactus: Yes they are pretty much the same.

docker-app/qfieldcloud/core/tests/test_packages.py
suricactus: I close it because I don't know if this is still the issue.

docker-qgis/process_projectfile.py
suricactus: These changes are no longer needed, we agreed the check will be done live in the django side.

olivierdalang:
@suricactus Will stop for today, can you review this if CI passes ? (please don't look at the commit history which is pretty ridiculous in regard of the final changes :-D )

docker-app/qfieldcloud/subscription/tests/test_external_db.py
```diff
@@ -0,0 +1,145 @@
...
+            json_str = json.dumps(deltafile)
+            return io.StringIO(json_str)
+
+    def test_is_external_db_supported(self):
+        """This tests is_external_db_supported property of accounts types"""
+
+        u1 = User.objects.create(username="u1")
+        self._login(u1)
+
+        # Create a projet with a writable remote DB
```
suricactus, line 66:
```suggestion
        # Create a project with a writable remote DB
```

docker-compose.yml
```diff
@@ -168,7 +168,7 @@ services:
   mirror_transformation_grids:
     image: k3rnelpan1c/alpine-wget:latest
     command: |
-      wget --mirror https://cdn.proj.org/ -P /transformation_grids --no-host-directories
+      wget --no-verbose --mirror https://cdn.proj.org/ -P /transformation_grids --no-host-directories
```
suricactus, line 171: Does it at least show what we are downloading? If yes, then I am ok, if it is completely silent, it's problematic.

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -271,6 +271,12 @@ def can_create_delta(user: QfcUser, delta: Delta) -> bool:
     """Whether the user can store given delta."""
     project: Project = delta.project
 
+    if (
+        project._has_online_vector_data
+        and not project.owner.useraccount.account_type.is_external_db_supported
+    ):
+        return False
+
```
suricactus, line 279: The only potential problem I see here is that the UI might use this permission check.

docker-compose.yml
olivierdalang: It shows what is downloading, and errors too, but is a bit misleading when things work fine (looks like nothing happened). Will move that out of this PR as not critical anyway

docker-app/qfieldcloud/core/permissions_utils.py
olivierdalang: Yes indeed... In the case of the similar feature about checking quota when changing owner, we ended up doing the check separately both in the API view and in the GUI view (so not ideal in terms of not repeating oneself)...

I'll see if I can implement this in the Delta's `clean()` method ([docs](https://docs.djangoproject.com/en/4.0/ref/models/instances/#validating-objects-1)).

About the second part of your comment (whether we block all deltas, or only the once about these layers) it would be possible to implement this I guess. But not sure what's best ? IMO better to have no delta apply rather than just some, because that could break the data in case there are some relations. But not sure how it's done for other type of failures ?

Let me know what you prefer.

olivierdalang: Hmm actually `clean()` is not so well adapted, as it is supposed to be for hard validation (you can't save unless valid). Anyway, I'll just move that to a property on the model, so we can reuse it if needed (I think it's better on the model itself, but arguably we could also put in to permission_utils, or even the other way around on the account_type).

Then I guess it would be good for the user to know why the delta is failing (because that's not obvious). With current implementation he would just see that it's UNPERMITTED, but not why. In the frontend GUI, we could just call the same function to display ad hoc message. But how to deal with that through the API ? Is it needed ? Should we somehow store the reason it is UNPERMITTED ? Can we reuse `last_feedback`, or should we introduce a specific field ?

That makes me think that the initial approach (trying to apply the delta, and failing once we try) was not too bad, as this way we'd just tie into the existing delta failure implementation.

Let me know what you think (maybe we need another call to clarify exactly what we want, including all user facing aspects)

suricactus: > Then I guess it would be good for the user to know why the delta is failing (because that's not obvious). With current implementation he would just see that it's UNPERMITTED, but not why. In the frontend GUI, we could just call the same function to display ad hoc message. But how to deal with that through the API ? Is it needed ? Should we somehow store the reason it is UNPERMITTED ? Can we reuse last_feedback, or should we introduce a specific field ?

I think it would be more than enough to:
1) reply with an error when uploading deltas, that they do not match your current subscription plan
2) when opening the deltas page in the browser, show "You cannot apply deltas because your subscription plan"

 > That makes me think that the initial approach (trying to apply the delta, and failing once we try) was not too bad, as this way we'd just tie into the existing delta failure implementation.
This will bring extra jobs, which is the heaviest part in the whole QFC stack and would be nice to avoid.

> About the second part of your comment (whether we block all deltas, or only the once about these layers) it would be possible to implement this I guess. But not sure what's best ? IMO better to have no delta apply rather than just some, because that could break the data in case there are some relations. But not sure how it's done for other type of failures ?

Probably would be easier to describe that no deltas can be applied if you are not having a good subscription plan, agreed.

suricactus (approved):
Looks good.

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -99,13 +100,22 @@ def post(self, request, projectid):
...
-                            self.request.user, delta_obj
-                        )
-                        or not delta_obj.is_supported_regarding_owner_account
+                    if not permissions_utils.can_create_delta(
+                        self.request.user, delta_obj
                     ):
                         delta_obj.last_status = Delta.Status.UNPERMITTED
+                        delta_obj.last_feedback = {
+                            "msg": _(
+                                "User has no rights to create delta on this project. Try inviting him as a collaborator and try again."
```
suricactus, line 109:
```suggestion
                                "User has no rights to create delta on this project. Try inviting him as a collaborator with proper permissions and try again."
```

docker-qgis/apply_deltas.py
```diff
@@ -500,7 +500,8 @@ def committed_features_added_cb(layer_id, features):
             if feature.isValid():
                 _pk_attr_idx, pk_attr_name = find_layer_pk(layer)
 
-                assert pk_attr_name
+                if not pk_attr_name:
+                    raise DeltaException(f"Layer {layer.name()} has not primary key")
```
suricactus, line 504:
```suggestion
                    raise DeltaException(f'Layer "{layer.name()}" has no primary key.')
```
Even though not in place everywhere, I usually put placeholders in double quotes. Also, I am pretty sure it is "the layer has no primary key", but English is not my superpower.

PR #326, opened 2022-05-06
Clean up outdated project packages

suricactus:
#297 follow-up

- Clean up outdated project packages
- Add forgotten migrations: project.last_package_job
- Improved docs
- Disable the TestMiddleware if outside test environment


docker-app/qfieldcloud/core/cron.py
```diff
@@ -76,3 +77,27 @@ def do(self):
             },
             output="Job unexpectedly terminated.",
         )
+
+
+class DeleteDanglingProjectPackagesJob(CronJobBase):
```
olivierdalang, line 82: Just to be clear, by "dangling" you mean a package where a newer package exists ? (in my understanding, "dangling" would rather mean a package where the project doesn't even exist anymore)

If so, maybe call this "obsolete" instead ?

docker-app/qfieldcloud/core/cron.py
```diff
@@ -76,3 +77,27 @@ def do(self):
...
+    schedule = Schedule(run_every_mins=60)
+    code = "qfieldcloud.delete_dangling_project_packages"
+
+    def do(self):
+        # get only the projects updated in the last a little more than an hour,
+        # as we assume the rest of the projects have been cleaned from dangling
+        # packages in the previous CRON run
+        projects = Project.objects.filter(
+            updated_at__gt=timezone.now() - timedelta(minutes=65)
```
olivierdalang, line 91: What about waiting a little before deleting packages, just to avoid deleting brand new packages (which could not yet be completely uploded, and which potentially another user just started downloading because for him the brand new package was not yet there ?)
```diff
      projects = Project.objects.filter(
-            updated_at__gt=timezone.now() - timedelta(minutes=65)
+            updated_at__gte=timezone.now() - timedelta(minutes=125),
+            updated_at__lt=timezone.now() - timedelta(minutes=65)
      )
```

docker-app/qfieldcloud/core/middleware/test.py
```diff
@@ -6,8 +8,7 @@ class TestMiddleware:
     def __init__(self, get_response):
         self.get_response = get_response
 
-        # TODO check if in test mode, if not raise exception
-        if False:
+        if os.environ.get("ENVIRONMENT") != "test":
             raise MiddlewareNotUsed()
```
olivierdalang, line 12: Not sure if there's any particular logic about this in Django ? Is this catched by Django to disable the middleware ? Otherwise this will break production.

An alternative maybe more readable implementation would be to disable it in settings.py like this

```python
# settings.py
MIDDLEWARE = [
   # ...
]
if os.environ.get("ENVIRONMENT") == "test":
    MIDDLEWARE.append("qfieldcloud.core.middleware.test.TestMiddleware")
```

docker-app/worker_wrapper/wrapper.py
```diff
@@ -286,6 +287,22 @@ def after_docker_run(self) -> None:
...
+                    # keep the last package
+                    if package_id == str(self.job.project.last_package_job_id):
+                        continue
+
+                    storage.delete_stored_package(project_id, package_id)
+            except Exception as err:
+                logger.error(
+                    "Failed to delete dangling packages, will be deleted via CRON later.",
+                    exc_info=err,
+                )
```
olivierdalang, line 304: As said above, maybe better not to delete them too soon ?

And anyway, IMO better to use just one of these two implementations (either cron OR after successful package job), unless there is a particular reason we expect it to sometimes fail on the job ?

olivierdalang (commented):
Ok review done. Not sure how critical packages are, but IMO worth writing a test for that too ?

docker-app/qfieldcloud/core/middleware/test.py
suricactus: I was following the docs:

https://docs.djangoproject.com/en/3.2/topics/http/middleware/#marking-middleware-as-unused

docker-app/worker_wrapper/wrapper.py
suricactus: I always assume the job might crash, so it would be good if we have a self cleaning mechanism. Why do you prefer having only one of the cleanup mechanisms?

docker-app/qfieldcloud/core/middleware/test.py
olivierdalang: Ah ok fine then :-)

docker-app/qfieldcloud/core/cron.py
suricactus: Well spotted. I do agree this is better.

docker-app/qfieldcloud/core/cron.py
suricactus: Agree that dangling might be misinterpreted and "obsolete" is better.

docker-app/qfieldcloud/core/cron.py
olivierdalang: oops there should be a small overlap still:
```diff
      projects = Project.objects.filter(
-            updated_at__gt=timezone.now() - timedelta(minutes=65)
+            updated_at__gte=timezone.now() - timedelta(minutes=125),
+            updated_at__lt=timezone.now() - timedelta(minutes=60) # not 65
      )
```

docker-app/worker_wrapper/wrapper.py
suricactus: Addressed with the new check for the job id.

suricactus: re: two places same logic
I prefer to keep it like this, so the storage is available ASAP for the users.

PR #330, opened 2022-05-18
Added core_user.username UPPER index, as it is mostly compared with

suricactus:
Also:
- fix the order of the migrations
- added missing blank to `project.last_package_job`


suricactus:
While the DB distinguishes the upper and lower case, we do support case-insensitive login and URLs. It means we need to enforce only one case. 

Also, we have quite some `User.objects.filter(username__iexact='suricactus')`, which converts the string to uppercase and skips using the index.

This makes sure the values are unique case insensitive.

olivierdalang:
Ok, I got mistaken by the regex validator on username which accepts uppercase (despite converting to lowercase when creating an account).

In the longer term it could be safer to do exact comparisons only and convert user input to lowercase instead (I guess we get the username from the URL), and also enforce only lowercase usernames by changing the validator. This way if somehow user with uppercase make it to the DB (could happen with a batch import or using oauth), we don't risk collision which could be a security issue.

suricactus:
I think we should keep the case in the db. An example is `OPENGISch` which is the official name of one of the organizations. While it is safer to accept only one case, it is a bit limiting.

olivierdalang:
Ok to keep as is for now then (with your new UNIQUE index it should be safe anyway).

PR #332, opened 2022-05-22
Optimize getting project permissions query with a view

suricactus:
A little performance gain of 150 times faster query.

Also add `django-migrate-sql-deux` for easier raw SQL migration management.


suricactus:
The speed is no longer 150 times faster, as additional checks of the validity of project collaboration slowed it down.

docker-app/qfieldcloud/core/models.py
```diff
@@ -774,79 +775,17 @@ class RoleOrigins(models.TextChoices):
...
+        qs = (
+            Project.objects.select_related("user_role")
+            .defer("user_role__user_id", "user_role__project_id")
+            .annotate(user_role__user_id=Coalesce("user_role__user_id", None))
+            .filter(
+                Q(user_role__user=user)
+                | Q(user_role__origin=ProjectQueryset.RoleOrigins.PUBLIC),
+                user_role__is_valid=True,
+            )
+            .distinct("owner__username", "name")
```
suricactus, line 787: This distinct here is not ok. It prevents the `for_user` qs to be sorted, as it translates to `DISTINCT ON`. Ideas?

olivierdalang: Do the duplicates appear because an user can have several roles in the view ? If so, could it be sorted by having the DISTINCT clause in the view instead ?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1159,6 +1098,28 @@ def clean(self) -> None:
         return super().clean()
 
 
+class ProjectPermissionsView(models.Model):
+    user = models.ForeignKey(
+        User,
+        on_delete=models.DO_NOTHING,
+        related_name="related_projects",
```
olivierdalang, line 1105: I'd use `projects_permissions` as related name (otherwise, `user1.related_projects.all()` is confusing, as you'd expect it to be a list of projects)

docker-app/qfieldcloud/core/models.py
suricactus: The problem comes with the public projects, as other rows in the view already have `DISTINCT ON (project_id, user_id)`, but the public projects will return `NULL` as user_id in the query. The more I think, the more I believe the public projects should be handled separately.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1159,6 +1098,28 @@ def clean(self) -> None:
...
+    )
+    project = models.OneToOneField(
+        "Project",
+        on_delete=models.DO_NOTHING,
+        related_name="user_role",
```
olivierdalang, line 1110: Here too maybe we could come up with a more clear related name (either change the model name to `ProjectRoles`, or also use `projects_permissions` as related name and then maybe rename the `name` field to `role_name` ?

suricactus: I would go for `ProjectRolesView` in such case. We have more `roles` than permissions here.

docker-app/qfieldcloud/core/models.py
suricactus: What if we rename `related_name="project_roles"`?

docker-app/qfieldcloud/core/models.py
olivierdalang: yes sounds good

docker-app/qfieldcloud/core/models.py
olivierdalang: Yes, in the API list view for instance we had to handle public projects differently than other roles already (because you do not want to include them in your default project's list). And the day we want to materialize the view (either just materialized view, or a regular django model updated through signals), we won't be able to keep a list of all public relations anyway.

I think it's quite clear : `Project.objects.filter( Q(project_roles__user=user1) | Q(is_public=True) )`

suricactus: not really, because we need to explude the duplicates

olivierdalang: In my example above, if the permissions view has unique rows for project, user, I think there should be no duplicate? The | operator on Q objects is supposed to resolve to an OR on the where clause (not to be confused with | on queryset which would be an union).
Or am I missing something?

suricactus: Ok, we did it! This delayed union is doing mirracles.

suricactus: well, miracles only in tests here, failed in the UI part. Anyway, lessons learned that `row_number` is very slow.

PR #333, opened 2022-05-23
Run cron every 1min and restart ofelia automatically

suricactus:


duke-nyuki:
Task linked: [QF-1987 Investigate why some project jobs are never finished](https://app.clickup.com/t/2192114/QF-1987)

PR #334, opened 2022-05-25
Fix add members leak of teams from other organzations

suricactus:


duke-nyuki:
Task linked: [QF-1989 Fix add members leak of teams from other organzations](https://app.clickup.com/t/2192114/QF-1989)

PR #336, opened 2022-05-26
Project collaborators model tweaks: allows admins as collaborators and cleanup teams after member leaves the organization

suricactus:
Remove the limitation to add org admins or owner as project collaborators.

What if you remove them from org admins or owner, it should be easy to go back and forth with their permissions.


duke-nyuki:
Task linked: [QF-1996 When a user is removed from the organization, should be removed from all teams too](https://app.clickup.com/t/2192114/QF-1996)

docker-app/qfieldcloud/core/models.py
```diff
@@ -685,6 +705,15 @@ def clean(self) -> None:
...
         return super().clean()
 
+    @transaction.atomic
+    def delete(self, *args, **kwargs) -> None:
+        super().delete(*args, **kwargs)
+
+        TeamMember.objects.filter(
+            team__team_organization=self.organization,
+            member=self.member,
+        ).delete()
```
suricactus, line 715: I am not 100% sure if it is ok to mark the whole delete method as in transaction like this, but IMO should be in transaction in any case.

docker-app/qfieldcloud/core/models.py
```diff
@@ -648,7 +649,26 @@ def save(self, *args, **kwargs):
         return super().save(*args, **kwargs)
 
 
+class OrganizationMemberQueryset(models.QuerySet):
+    @transaction.atomic
```
suricactus, line 653: transaction

docker-app/qfieldcloud/core/models.py
```diff
@@ -648,7 +649,26 @@ def save(self, *args, **kwargs):
...
+            team__team_organization__in=self.values_list("organization"),
+            member__in=self.values_list("member"),
+        ).delete()
+
+        org_total, org_summary = super().delete(*args, **kwargs)
+
+        total = org_total + team_total
+        summary = {**org_summary, **team_summary}
+
+        return total, summary
```
suricactus, line 665: Transaction again?

PR #343, opened 2022-06-14
Make useraccount generation part of a transaction and do not create one for teams

suricactus:


duke-nyuki:
Task linked: [QF-2013 Fix users without accounts](https://app.clickup.com/t/2192114/QF-2013)

PR #344, opened 2022-06-15
When anonymous actor is passed to `audit`, assume no actor

suricactus:


duke-nyuki:
Task linked: [QF-2014 Users are unable to delete their own accounts when the account owns projects](https://app.clickup.com/t/2192114/QF-2014)

PR #347, opened 2022-06-16
Skip downloading attachments in workers

suricactus:
Attachments are unimportant files such as photos, video, documents etc
that do not affect the work of the geospatial project.

- Add `is_attachment` flag to the files/package response to indicate
whether the file is considered an attachment.
- Do not download files within the worker that are flagged as `is_attachment`
- List files from the package and add the files from the original project
files in the attachment directories.
- When a file is requested from the package files endpoint and has an
attachment prefix, first check if the file is available in the package
and if not, try to download it from the original project files.

### DIFF
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -314,6 +314,22 @@ def get_s3_project_size(project_id: str) -> int:
     return round(total_size / (1024 * 1024), 3)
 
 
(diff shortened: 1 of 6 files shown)


duke-nyuki:
Task linked: [QF-2025 Prevent downloading the whole DCIM directory in the worker](https://app.clickup.com/t/2192114/QF-2025)

PR #353, opened 2022-07-01
Replace null bytes in delta json in the delta to prevent pg error

suricactus:
```
unsupported Unicode escape sequence
LINE 1: ...id, '2c6f40b1-7d2a-4225-be8d-8190d7a691f9'::uuid, '{"clientI...
                                                             ^
DETAIL:  \u0000 cannot be converted to text.
```
no more


duke-nyuki:
Task linked: [QF-2035 Null string in deltas leads to server error 500](https://app.clickup.com/t/2192114/QF-2035)

PR #355, opened 2022-07-02
Remove `prefetch_related` for 1:n that caused memory leak and super slow responses

suricactus:
Note to myself:
Django does separate select for each of the N relationships, which makes everything incredibly slow.


duke-nyuki:
Task linked: [QF-2041 Remove prefetch_related for 1:n that caused memory leak and super slow responses](https://app.clickup.com/t/2192114/QF-2041)

PR #372, opened 2022-07-26
Add `deleteorphanedprojectfiles` command

suricactus:


duke-nyuki:
Task linked: [QF-2045 Delete old projects from an account if not used](https://app.clickup.com/t/2192114/QF-2045)

PR #379, opened 2022-08-07
Return nicer error when the project name already existing for given owner

suricactus:


duke-nyuki:
Task linked: [QF-2082 QFieldSync fix bad UX when creating a project with an existing name](https://app.clickup.com/t/2192114/QF-2082)

PR #387, opened 2022-08-11
Explicitly require a higher `requests` version, so it is satisfied by the system `urllib3`

suricactus:


duke-nyuki:
Task linked: [QF-2090 Fix warning shown in the docker container](https://app.clickup.com/t/2192114/QF-2090)

PR #390, opened 2022-08-11
Fix checking for permission when package is uploaded by a collaborator with lower perms

suricactus:
E.g. collaborator reader cannot package a project, since the last step of the package job: uploading the files, fails.


duke-nyuki:
Task linked: [QF-2094 Collaborators cannot package](https://app.clickup.com/t/2192114/QF-2094)

PR #395, opened 2022-08-15
Fix replacing null bytes with utf8 conversion but reading the whole file

suricactus:
Reading the whole file is not too much of a concern, as the delta files
should not grow indefinitely.

If we decode to utf8, we might get a chunk splitting the utf8 char in two,
therefore creating invalid utf8 character.

Tried to work with bytes directly, but failed. In theory we can safely
remove the null bytes without worrying too much, as null bytes appear
only for null characters in utf8.
Or at least according to this: https://stackoverflow.com/a/6907327/1226137


duke-nyuki:
Task linked: [QF-2136 UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd0 in position 65535: unexpected end of data](https://app.clickup.com/t/2192114/QF-2136)

PR #396, opened 2022-08-15
Prevent login with organizations or teams

suricactus:
Fix https://github.com/opengisch/QField/issues/3195

Quite interesting we didn't find it until now. A follow up would be to prevent resetting a password of an organization.


duke-nyuki:
Task linked: [QF-2133 QFieldSync list of projects empty & server error 500 #3195](https://app.clickup.com/t/2192114/QF-2133)

PR #397, opened 2022-08-15
Remove unnecessary `LoginSerializer`, use the `allauth` one as they are identical

suricactus:
No idea why I put it in the first place.


duke-nyuki:
Task linked: [QF-2138 Remove unnecessary `LoginSerializer`, use the `allauth` one as they are identical](https://app.clickup.com/t/2192114/QF-2138)

PR #399, opened 2022-08-16
Make user suggestions be filtered by username and ordered by best match

suricactus:


duke-nyuki:
Task linked: [QF-1955 Fix user suggestion list in "Add members" page](https://app.clickup.com/t/2192114/QF-1955)

PR #400, opened 2022-08-16
Qf 2134 fix sentry logging

olivierdalang:
Fixes for errors logged to sentry. To better review, look at the individual commits (referring to the sentry issue), since the last one refactors the error handler.

I'm not 100% sure doing this all in the exception handler is the right place, because stuff like `ObjectDoesNotExist` may be an user error (e.g. user entering a wrong URL) but also due to bugs in the code... IMO it would be cleaner to only take care about how exceptions are returned to the user in the handler, but otherwise catch/manage them in the code where we expect them to happen.


duke-nyuki:

suricactus (approved):
LGTM

PR #404, opened 2022-08-25
Qf 2134 fix sentry logging

olivierdalang:
Some more improvements to sentry:
- ~minor improvement to exception handler (log 4xx as info, 5xx as exceptions)~ Actually, they could still indicate some serious issues e.g. in QField (for example 3528247758 could indicate some but in QField). Let's continue to manually decide which exception we want to report or not (and in the longer term, do that in the code itself rather than in the uncaught exception handler)
- ~mute sentry issue 3528247758 (error -> info)~ (see above)
- include SENTRY_RELEASE env var, so that we can use features like "resolve in next release" in sentry

TODO BEFORE MERGE
- [ ] the release pipeline must be adapted to set a `SENTRY_RELEASE` env var (or maybe cleaner to write it to `.env` so it survives manual docker-compose up).


duke-nyuki:

docker-app/qfieldcloud/core/rest_utils.py
```diff
@@ -24,19 +23,17 @@ def exception_handler(exc, context):
...
     else:
         # Unexpected ! We rethrow original exception to make debugging tests easier
         if settings.IN_TEST_SUITE:
             raise exc
-        is_error = True
         qfc_exc = qfieldcloud_exceptions.QFieldCloudException(detail=str(exc))
 
-    if is_error:
+    # Log level is defined by the exception
+    if qfc_exc.log_as_error:
```
olivierdalang, line 36: From what I see, it seems only errors are logged to sentry. Would it make sense to also log info (which maybe we could promote to warning) for user caused issues ? I guess they could sometime indicate serious issues still (e.g. broken links, bugs in QField, etc).

Not sure how this would be done, found this: https://docs.sentry.io/platforms/python/guides/django/usage/set-level/

Also not sure where and how we use logging, just so that we don't fill sentry's servers with junk...

We also have settings.LOGLEVEL in our settings.py, but I think that is unused code ?

suricactus:
I think the only think we should do before merging this PR is to find an automatic way to set the `SENTRY_RELEASE` env variable. I believe we can get when building the containers with the `docker compose` setup we have.

docker-app/qfieldcloud/core/rest_utils.py
olivierdalang: Of course wasn't talking about logging debug messages to sentry... Did you check how many info messages there is ? At least just warnings would already be good, as it would include stuff like deprecation warning or unclosed resources, which are definitely worth knowing of. There definitely should not be too many of them (or if there are, we better know about them).

Or are we somehow checking the logs in a different way to find out about such issues ?

olivierdalang:
> I think the only think we should do before merging this PR is to find an automatic way to set the SENTRY_RELEASE env variable. I believe we can get when building the containers with the docker compose setup we have.

Yes, for that do you want to take over ? (I'm happy to do it, but then need some guidance as to how deployment is set up)

duke-nyuki:

PR #405, opened 2022-09-04
Make emails unique among regular users

suricactus:
No worries I modified an existing migration, I already patched the DB in
a compatible way


duke-nyuki:
Task linked: [QF-2062 Add unique constraint on core_user.emails when core_user.user_type = USER](https://app.clickup.com/t/2192114/QF-2062)

PR #430, opened 2022-10-28
Remove User.polymorph and use InheritanceManager

suricactus:


suricactus:
@olivierdalang might be interesting for you.

PR #438, opened 2022-11-11
Infrastructure for selenium tests

olivierdalang:
(not ready for review yet)

Groundwork for selenium testing:
- allow `nginx` as a hostname (so the selenium browser can access the app using the proxy instead of django, hence testing the full stack)
- include python-selenium dependency

TODO before merge
- [x] reenable preventiong of access by IP
- [x] consider using nginx as QFIELDCLOUD_HOST in the settings instead of hardcoding all that
- [x] see how to only install python-selenium on test setups (we don't need it on prod)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,7 +3,7 @@ ENVIRONMENT=test
 
 QFIELDCLOUD_HOST=localhost
 DJANGO_SETTINGS_MODULE=qfieldcloud.settings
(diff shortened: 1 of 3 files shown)


suricactus:
>  see how to only install python-selenium on test setups (we don't need it on prod)
If it makes sense?
https://github.com/opengisch/qfieldcloud/blob/master/docker-app/Dockerfile#L25

PR #474, opened 2023-01-16
Allow active users to work even if current_period since/until is not set

suricactus:


duke-nyuki:
Task linked: [QF-2364 dev.qfield.cloud -> view user list - not working](https://app.clickup.com/t/2192114/QF-2364)

PR #478, opened 2023-01-25
add checks before deletion on S3

olivierdalang:
(did not run tests locally, likely some typos/obvious errors left)

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -203,7 +203,7 @@ def remove_project_thumbail(project: "Project") -> None:  # noqa: F821
     """
     bucket = qfieldcloud.core.utils.get_s3_bucket()
     key = project.thumbnail_uri
-    if not key or not re.match(r"^projects/\w+/meta/\w+.(png|jpg)$", key):
+    if not key or not re.match(r"^projects/[\w-]+/meta/\w+.(png|jpg)$", key):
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -148,6 +149,8 @@ def remove_user_avatar(user: "User") -> None:  # noqa: F821
     """
     bucket = qfieldcloud.core.utils.get_s3_bucket()
     key = user.useraccount.avatar_uri
+    if not key or not re.match(r"^users/\w+/avatar.(png|jpg)$", key):
```
suricactus, line 152:
```suggestion
    if not key or not re.match(r"^users/\w+/avatar.(png|jpg|svg)$", key):
```

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -344,6 +356,12 @@ def delete_file_version(
             versions_to_delete.append(file_version)
 
     for file_version in versions_to_delete:
+
+        if (
+            not re.match(r"^projects/[\w-]+/.+$", file_version._data.key)
+            or not file_version.version_id
```
suricactus, line 362:
```suggestion
            or not file_version.id
```

olivierdalang: Ok committed, I saw that `version_id` in the stubs though (but nothing in the docs)
image

PR #479, opened 2023-01-25
Block project deletion

suricactus:


suricactus:
No longer makes sense. Closing.

PR #483, opened 2023-01-28
Recalculate project size when soft delete is performed in project files

suricactus:


duke-nyuki:
Task linked: [QF-2433 Cannot recalculate project size when soft delete is performed in project files](https://app.clickup.com/t/2192114/QF-2433)

PR #485, opened 2023-02-01
Setup sentry environment

suricactus:
Tried to set it up with:

```
ENVIRONMENT = os.environ.get("ENVIRONMENT")
sentry_sdk.init(environment=ENVIRONMENT)
```

but then:

```
sentry_sdk.client_get_options()
```

didn't return the correct environment. Dunno why, but this seems to work.


duke-nyuki:
Task linked: [QF-2437 Configure sentry environment](https://app.clickup.com/t/2192114/QF-2437)

PR #488, opened 2023-02-01
Log warning instead of asserting when trying to delete a file that does not exist

suricactus:


duke-nyuki:
Task linked: [QF-2411 Error when changing profile picture](https://app.clickup.com/t/2192114/QF-2411)

PR #497, opened 2023-02-09
Bump django-audit dependency to the latest 2.2.2

suricactus:
DB migrations:
- https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0008_action_index.py - adds an index, will lock temporarily
- https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0009_alter_logentry_additional_data.py - adds a nullable JSON field, should be fast
- https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0010_alter_logentry_timestamp.py - adds a non-nullable timestamp field with default value, will be slow but not terrible
- https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0011_logentry_serialized_data.py - adds a nullable JSON field, should be fast
- https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0012_add_logentry_action_access.py - no DB changes, changing the choices of an integer field

Changelog:
https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md#221-2022-11-28
NOTE: v2.2.2 is missing from the changelog, but fixes an issue with deleting an audit from the admin, see jazzband/django-auditlog#496

Changes:
jazzband/django-auditlog@v1.0a1...v2.2.2

Also enable audits when deleting a user.

With older versions of auditlog module, deleting a user that owns a project was causing an error, which seems to be fixed with 2.2.2 (see https://github.com/opengisch/qfieldcloud/pull/344).

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -8,7 +8,6 @@ from typing import List, Optional
 
 import django_cryptography.fields
 import qfieldcloud.core.utils2.storage
-from auditlog.context import disable_auditlog
 from deprecated import deprecated
 from django.contrib.auth.models import AbstractUser
 from django.contrib.auth.models import UserManager as DjangoUserManager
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-2464 Enable audits on most models on QFieldCloud](https://app.clickup.com/t/2192114/QF-2464)

PR #499, opened 2023-02-10
Store the actual total file bytes in the database, instead of estimateded MB

suricactus:
Storing the project file size in MB has precision errors and it's causing more trouble than good.

When uploading a file was actually depending only on "purge_old_file_versions" to get the project file storage recalculated via `Project.save(recompute_storage=True)`.
When we disabled purging, we no longer got storage size updated.

Since `Project.save(recompute_storage=True)` is not very well scalable, I would prefer to calculate the file size as we upload/delete files.

Aslo:
The file operations depend on HTTP calls to the S3 storage and they might fail,
We need to choose source of truth between DB and S3.
For now the source of truth is what is on the S3 storage,
because we do most of our file operations directly by calling the S3 API.
1. S3 storage modification. If it fails, it will cancel the transaction
and do not update anything in the database.
For simplicity we now assume S3 storage is transactional, but we should be aware it might not hold be true all the times (e.g. request sent, processed, but response lost).
2. DB modification. If it fails, the states betewen DB and S3 mismatch,
but can be easyly synced from the S3 to DB with a manual script.


duke-nyuki:
Task linked: [QF-2415 Make QFieldCloud project storage usage show correct numbers and updates as soon as a file is uploaded](https://app.clickup.com/t/2192114/QF-2415)

PR #504, opened 2023-02-12
Allow also teams as project collaborators when project is within organization

suricactus:
The only question is why we see this now...


duke-nyuki:
Task linked: [QF-2471 Cannot add team as project collaborator in an organization](https://app.clickup.com/t/2192114/QF-2471)

PR #505, opened 2023-02-12
Speed up QFieldCloud admin by referencing user with autocomplete view

suricactus:
By default django is rendering all possible values in a select field, which
is no longer scalable with 60k users. Therefore an autocomplete field is
rendered.

Also make UserAccount representation nicer from "Useraccount 1" to "suricactus (UserAccount)".


duke-nyuki:
Task linked: [QF-2472 Use autocomplete in admin when referencing to User/UserAccount, as the list is super long](https://app.clickup.com/t/2192114/QF-2472)

PR #510, opened 2023-02-14
Opening community projects causes an error because account.user is not an Organization instance

suricactus:


duke-nyuki:
Task linked: [QF-2478 Opening comminity projects in the web interface causes an error](https://app.clickup.com/t/2192114/QF-2478)

PR #511, opened 2023-02-14
Do not account staff users as active users

suricactus:
This is super important for billing.


duke-nyuki:
Task linked: [QF-2480 Do not count staff users as active users](https://app.clickup.com/t/2192114/QF-2480)

PR #513, opened 2023-02-15
Prevent worker wrapper exception when Project.project_details was already NULL

suricactus:
When using `update_fields` and there are no modified rows, django will throw an error.


duke-nyuki:
Task linked: [QF-2491 Prevent update_fields exception if the `Project.project_details` is already null.](https://app.clickup.com/t/2192114/QF-2491)

PR #518, opened 2023-02-16
Enforcing pos ints as storage_bytes

why-not-try-calmer:
Added setter to enforce pos ints as storage_bytes.


duke-nyuki:
Task linked: [QF-2488 Prevent QFieldCloud from storing negative values in storage_bytes field](https://app.clickup.com/t/2192114/QF-2488)

PR #519, opened 2023-02-16
Non-negative `file_storage_bytes` when deleting a file

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2488 Prevent QFieldCloud from storing negative values in storage_bytes field](https://app.clickup.com/t/2192114/QF-2488)

suricactus:
Hey, this will not fix the issue either, sorry. The `default` is evaluated [only when a new object is created](https://docs.djangoproject.com/en/4.1/ref/models/fields/#default).

What has to be changed is on this line here:
https://github.com/opengisch/qfieldcloud/blob/master/docker-app/qfieldcloud/core/utils2/storage.py#L510

I think it makes sense we do it like:
```
-         project.file_storage_bytes -= sum([v.size for v in file.versions])
+         file_storage_bytes = project.file_storage_bytes - sum([v.size for v in file.versions])
+         project.file_storage_bytes = max(file_storage_bytes, 0)

```

suricactus:
Ok, so now we wait for green. 

Only one small request on the commit message: "Committing suggestion." as a message will not tell us a lot in the future, simply using the PR's title or something more descriptive would be great.

Thanks! And huge +1 for forcing the branch history to only have the two commits we want.

suricactus:
@why-not-try-calmer ok, green. Just change the commit message and we merge this. :)

PR #523, opened 2023-02-20
Make sure QFieldCloud is compatible with `nan` coords sent by QField 2.6+

suricactus:
QGIS support of `nan` coordinates sucs. Actually pretty much any Geospatial library sucks at this.

A few changes in this PR:
- updated the gpkg file, so it contains a `points_xy`, `points_xyz` and `points_xyzm` layers
- replace all `nan` coordinates with `0`s in geometries
- check the created/updated geometries after each test in Django

Note the XYZM tests are disabled, as Fiona and checkly are not supporting it well. Enable them in better times.

Fix: https://github.com/opengisch/qfieldcloud/issues/496


duke-nyuki:
Task linked: [QF-2466 If geometry has `nan` values (Z or M, as from QField 2.6+), QFieldCloud does not report an error, but also does not save the geometry in the DB](https://app.clickup.com/t/2192114/QF-2466)

suricactus:
See also https://github.com/qgis/QGIS/pull/47034/

docker-qgis/apply_deltas.py
```diff
@@ -237,8 +237,9 @@ def get_geometry_from_delta(
             if geometry.isNull():
                 raise DeltaException(f"Null geometry from {wkt=}")
 
+            # E.g. Shapefile might report a `Polygon` geometry, even though it is a `MultiPolygon`
             if geometry.type() != layer.geometryType():
-                raise DeltaException(
+                logger.info(
```
nirvn, line 242: It's not so much that QGIS wrongly reports geometry, is that's some providers can tolerate feature geometry that doesn't match its overall stated geometry. In any case, good! :)

PR #528, opened 2023-02-23
Fix deleting versions with `include_older=True`

suricactus:
The function assumes the versions are sorted from latest/newest to oldest versions. However, this was no longer the case and was asserting.


duke-nyuki:
Task linked: [QF-2511 Problems deleting older file versions](https://app.clickup.com/t/2192114/QF-2511)

PR #533, opened 2023-02-27
Users getting error 500 when trying to register with username taken by an Organization

suricactus:
NOTE `allauth` depends on the `PRESERVE_USERNAME_CASING` to make `iexact` lookups. When set to `False` as it is now, all `Person` usernames are stored as lowercase. However, usernames for `Organization` or `Team` are not stored lowercase and then `allauth` does not check for them.

TODO check what are the consequences if we don't add all of these and change the `PRESERVE_USERNAME_CASING` to `True`.


duke-nyuki:
Task linked: [QF-2541 User instances with type=1 are missing corresponding Person instances](https://app.clickup.com/t/2192114/QF-2541)

PR #534, opened 2023-02-28
Fix `createuseraccount` script as plans are no longer stored there

suricactus:


duke-nyuki:
Task linked: [QF-2543 Fix `createuseraccount` script as plans are no longer stored there](https://app.clickup.com/t/2192114/QF-2543)

PR #539, opened 2023-03-05
Speed up delta admin by adding some indices

suricactus:


duke-nyuki:
Task linked: [QF-2559 Speed-up delta admin](https://app.clickup.com/t/2192114/QF-2559)

PR #541, opened 2023-03-05
Added logs when `file` is not found in `request.data`

suricactus:
Trying to debug why it seems the clients never sent the file object.


duke-nyuki:
Task linked: [QF-2540 Debug `EmptyContentError: Empty content` when uploading files](https://app.clickup.com/t/2192114/QF-2540)

PR #543, opened 2023-03-05
Always use boto3 session when creating different objects.

suricactus:
As suggested here https://github.com/boto/boto3/issues/801#issuecomment-875140699,

The occasional error we get `KeyError: 'endpoint_resolver'` should be fixed if we replace `boto3.client()` with `s3_session.client()` .


duke-nyuki:
Task linked: [QF-2560 Getting `KeyError: 'endpoint_resolver'` when calling the S3 storage](https://app.clickup.com/t/2192114/QF-2560)

PR #546, opened 2023-03-07
Add filter for finalized jobs to JobAdmin

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2557 Add a filter by finilized/non finilized jobs in Admin UI](https://app.clickup.com/t/2192114/QF-2557)

suricactus (approved):

suricactus:
Looking at the CI error, did you try filtering with your code change:
```
django.core.exceptions.FieldError: Cannot resolve keyword 'last_status' into field. Choices are: applyjob, created_at, created_by, created_by_id, feedback, finished_at, id, output, packagejob, processprojectfilejob, project, project_id, started_at, status, type, updated_at
```

Actually I decieded to simply open it on my PC and I got this Django error, which matches the CI one:
image

I assume you still do not manually test your changes. Please do, you should not rely 100% on CI to catch your errors, especially for UI things that can be easily verified.

suricactus (changes requested):
Please make the test pass.

PR #547, opened 2023-03-07
QF-2556 storage usage

why-not-try-calmer:
gn


duke-nyuki:
Task linked: [QF-2556 Show the storage usage of an account in the Person/Organization admin](https://app.clickup.com/t/2192114/QF-2556)

suricactus:
I believe we can add `docker-app/qfieldcloud/core/filters.py` to be an autoloaded filter in `TEMPLATES` setting.

why-not-try-calmer:
> I believe we can add `docker-app/qfieldcloud/core/filters.py` to be an autoloaded filter in `TEMPLATES` setting.

Okay, is it needed though? Couldn't find this notion on https://docs.djangoproject.com/en/4.1/howto/custom-template-tags/.

why-not-try-calmer:
In hindsight, since the function is used in both private and public, and since it is registered once in private, it could stay in `public/docker-app/qfieldcloud/core/filters.py` while not being registered there, correct?

PR #548, opened 2023-03-09
Erroneous mimetypes invalidated, not thrown

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2490 When QFieldCloud does not recognize the mimetype of the uploaded file, just invalidate the field, do not crash](https://app.clickup.com/t/2192114/QF-2490)

why-not-try-calmer:
The same idea could be used for `upload_project_thumbail`.

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -262,8 +263,14 @@ def file_response(
     )
 
 
+class Mimes(str, Enum):
```
suricactus, line 266: Some more descriptive name than `Mimes` would help a lot. Like `AvatarMimeTypes` or `ImageMimeTypes` or else?

why-not-try-calmer:
I am no able to reproduce the CI error locally. Tests pass A-OK here.

miili:
For what it's worth: Use `Literal` instead.

```py
ImageMimeType = Literal["image/svg+xml", "image/png", "image/jpeg"]
```

why-not-try-calmer:
> For what it's worth: Use `Literal` instead.
> 
> ```python
> ImageMimeType = Literal["image/svg+xml", "image/png", "image/jpeg"]
> ```

Enums are fine. The name and the value are distinct, and consumed differently.

suricactus (approved):

PR #553, opened 2023-03-13
Human readable file sizes

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2586 When a file is downloaded from the Admin and no particular version is selected, it returns a non HTTP 200 status](https://app.clickup.com/t/2192114/QF-2586)

PR #556, opened 2023-03-13
QF 2587 human readable sizes in downloadable files

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2587 Show the file size in a nicer MB/GB/etc format](https://app.clickup.com/t/2192114/QF-2587)

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
```diff
@@ -169,6 +169,17 @@
                     .then((files) => refreshTableContents({ files }))
                     .catch((error) => refreshTableContents({ error }));
             };
+            const toFixed1 = (n, f) => (n/f).toFixed(1);
```
suricactus, line 172: Avoid this kind of super functional type of writing, you saved 2-3 characters in the original place of calling.

```
toFixed1(n, 10 ** 3) + " KB";
```
vs
```
((n / 10) ** 3).toFixed(3)+ " KB";
```

I would propose we assume the relevant digits are 3, there are quite a lot of files in GIS that can fit within 100 bytes, which will show as 0 KB in your case.

Finally, any chance that we add this in a `<span title="{n} bytes">{N} KB</span>`, so the supporter has the full information about the file size?

suricactus (changes requested):
See my inline comment, please add a `title` showing the full number and show 3 digits.

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
why-not-try-calmer: Oh it's not really a matter of style, it's a matter of having 1 thing to change instead of 3 as soon as we need refactoring. And I am not counting call sites where it might be needed in the future.

Yes to the other two things.

suricactus: I think I get your intent.

But I believe it is more or less matter of style this time with this contents, because trying to abstract this will take more maintanance and congnitive effort to understand in the future.

I think you can achieve pretty much the same without the need of an extra function, just by refactoring your current  `humanReadableSizes`. Use single return in the end, something like `return `<span title="${bytes} bytes">${sizeValue} ${sizeUnit}</span>` and calculate `sizeValue`, and `sizeUnit` in the `if`s/`switch` statements before.

BTW `humanReadableSizes` is probably not the best name, especially the plural in the name, something like `bytesToHumanReadableSize` or `filesize10` as in Django will be better.

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
```diff
@@ -196,7 +207,8 @@
 
                         $trow.querySelector('td:nth-child(1)').innerHTML = file.name;
                         $trow.querySelector('td:nth-child(2)').innerHTML = file.last_modified;
-                        $trow.querySelector('td:nth-child(3)').innerHTML = file.size;
+                        const [sizeBytes, size10] = filesize10(file.size)
```
suricactus, line 210: Avoid returning "tuples" in JS. If you need to return multiple values, better return objects.

In this case, no need to return `sizeBytes` at all, since it is the same as `file.size`.

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
suricactus: Can you still please remove the now renamed `toFixed3` function?

suricactus (approved):

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
```diff
@@ -196,7 +206,7 @@
 
                         $trow.querySelector('td:nth-child(1)').innerHTML = file.name;
                         $trow.querySelector('td:nth-child(2)').innerHTML = file.last_modified;
-                        $trow.querySelector('td:nth-child(3)').innerHTML = file.size;
+                        $trow.querySelector('td:nth-child(3)').innerHTML = `<span title=${file.size} bytes">${filesize10(file.size)} KB</span>`;
```
suricactus, line 209:
```suggestion
                        $trow.querySelector('td:nth-child(3)').innerHTML = `<span title="${file.size} bytes">${filesize10(file.size)} KB</span>`;
```

PR #562, opened 2023-03-14
Tiny git workflow improvements

m-kuhn:
- Only run tests on `master`, `release` and `pull_request`
- Add `*.orig` to `.gitignore` (leftovers from conflicts)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,6 @@ name: Test
 
 on:
   push:
-    branches:
-      - master
-      - release
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -1,8 +1,11 @@
 name: Test
 
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - master
+      - release
+  pull_request:
```
suricactus, line 8: I depend on tests running on push and pull_request both for two reasons:
- there is a flacky test, which fails 20% of the time, with two tests in parallel it makes the probability both fail 20% * 20% -> 4%
- sometimes I test things without opening a PR at all, just to see if they pass

suricactus: While the latter is a matter of preference and I am open to change, the former has practical implications for now.

PR #563, opened 2023-03-14
Add release-drafter

m-kuhn:
This creates release drafts automatically, whenever something is committed to the `release` branch. If everything goes well :crossed_fingers: the new release will be automatically linked to this branch and this way avoid the risk of releasing from `master` (dev) branch.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,19 @@
+name: Release Drafter
+
+on:
+  push:
+    branches:
+      - release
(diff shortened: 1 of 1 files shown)


suricactus:
Should we keep two release drafters as you suggest, or you overlooked the existance of the current one? We already have a release drafter here: https://github.com/opengisch/qfieldcloud/blob/master/.github/workflows/release_drafter.yml .

I don't mind which file we keep, I have tiny preference over `_` in the names.

PR #568, opened 2023-03-16
Fix storage calculation when deleting multiple files at once

suricactus:
With parallel requests, the ORM bit me again by caching the old value. I am pretty sure there must be a smarter way of incrementing and decrementing data in the DB, but what I have though so far is not super Django-ish:

This is plain ugly:
```
Project.objects.filter(pk=project.pk).update(file_storage_bytes=F('file_storage_bytes')+file.size)
```


duke-nyuki:
Task linked: [QF-2545 Delete photos via sync - datastorage does not update properly](https://app.clickup.com/t/2192114/QF-2545)

Issue #570, opened 2023-03-16
Changes not applied - Unable to find feature


miili:
A bunch of changes failed with `Unable to find feature` and could not be applied.

The changes came in in small chunks (3 deltas). Applying the change again does fail with the same error.

This is related to https://github.com/opengisch/qfieldcloud/issues/324

suricactus:
Can you give me any details about username/projectname and ideally, delta ids that you want to know more about. Also what is the data format of the layers, GeoPackage, PostGIS?

miili:
Dataformat is GeoPackage. Change `eff4a084-aef5-4b9b-8efa-321fb23f0ab0` Project `Eifel-Large-N/Re-Deployment`

miili:
@suricactus when you see patch `dd6fff05-b9ad-47c8-a253-8ae54c405c10` the `sourcePk` and `localPk` is the same. However the `sourcePk` should be 383.

Is this due to an old QField client?

miili:
Now that QfieldCloud is offered as a paid service (which I am 💯 supportive of) bugs like this  should be ironed out!! I see hundreds of changes which appear applied, but are not. We were confident to use qfieldcloud in production, but I am getting a little nervous now.

miili:
All that was changed in-between the working versions was the styling in the project file.

suricactus:
Hi, @miili . This is definitely something that is of high importance.

> when you see patch dd6fff05-b9ad-47c8-a253-8ae54c405c10 the sourcePk and localPk is the same. However the sourcePk should be 383.

How do you know the source PK is wrong in this one? Adding a bit more context will help finding the root cause.

> Is this due to an outdated QField client app?

Should not be the case, can you please share the version you are using?

miili:
The matching `sourcePk` seems to be a common issue: https://github.com/opengisch/qfieldcloud/issues/324#issuecomment-1385168652

suricactus:
Seems there is some root problem might be from `QgsOfflineEditing` plugin. Will keep you updated what are our findings.

miili:
See for example these features:

* Only the diamond shaped features should have been modified: Move feature, update attributes
* (probably due to bad `sourcePk` another feature was moved and edited

This already is a big mess.

![image](https://user-images.githubusercontent.com/4992805/225707765-0f0bff11-e34f-4a66-905c-bb6973c49232.png)

GlaDal:
Hi,

Maybe I have the same problem as yours : 

[...]

Issue #324, opened 2022-05-05
Changes not applied


gabgarcies:
Hi everyone,

I do not understand why some changes of my field workers are not applied. They push changes once a day and I have seen on the website that some of those changes have the status "Not applied".
![image](https://user-images.githubusercontent.com/80522138/166916245-620d41f6-615a-4993-bed9-92a4a4413a17.png)
If I try to change the status to "Applied" manually, it does not work either. The action starts but, after a while, it says "Not applied" again (Server error 500).

Why this might happen?

I attach the raw json of one of those not applied changes. 
[raw_json.txt](https://github.com/opengisch/qfieldcloud/files/8630828/raw_json.txt)

Thanks,

Gabriel

gabgarcies:
It does not only happens with a single layer, but it happens occasionally with 3 different layers of the project. 

The field workers only edit attributes (not adding/deleting features).

gabgarcies:
Here I see the log of the **Delta Apply** that tries to apply the changes:
![image](https://user-images.githubusercontent.com/80522138/167103421-062a7c0a-a263-4506-af00-b950d70b1945.png)

ThKroschel:
Hey @gabgarcies, OPENGIS-Team and QField-Users!

We have the same problem here. Some deltas were made successfully, others not. 
![image](https://user-images.githubusercontent.com/106523421/212862403-fc5fe949-a24e-49bf-9ee8-de2c91a17e9d.png)

Trying to re-apply them, doesn't work. 
![image](https://user-images.githubusercontent.com/106523421/212865202-fb9a7688-ae63-4969-b502-4c813830d547.png)

Clicking the ID shows the OSM and some times the selected poly (like in this case). In the upper right I clicked "see jobs".
In this case it shows me a list of 4 jobs, because I tried 3 times to re-apply the changes. Clicking the last item at this list (the first try pushing field data from mobile device) it shows:
![image](https://user-images.githubusercontent.com/106523421/212868176-8520b0ee-60eb-44a3-be14-903982b613c3.png)
As you can see, the warning says, that "modifiedFeaturePK" and "sourcePK" doesn't match. In fact that's wrong - they match!

I wonder if there are any problems with data formats in the background? I also checked the content in HTML. For me it looks the same like the applied change underneath.

Any ideas, someone?

Greetings from Germany!

stenk

gabgarcies:
Hi @ThKroschel 

[...]

PR #571, opened 2023-03-16
apply_changes: better exceptions

miili:
Related to #570 #324.

I improved the `DeltaException` message to see what is going on and why the deltas fail erratically. I don't know QGIS Python well and wonder is there is an `error()` method analog to the `layer.dataProdiver().error()`. That could be helpful.

Took the liberty and rename `delta_log` to `DELTA_LOG`, as it is a global variable which is being reused. The global state here is a dangerous thing. Maybe pass along a lightweight context and fill it with information.

Other things I want to comment on:
* Don't use `TypedDict`. Go for `dataclass` where possible or even pydantic if you need serialisation/deserialistion.
* f-strings in log messages is discouraged, use `%s` syntax instead.
* Use `Literal` instead of `Enum` where feasible.

### DIFF
--- a/docker-qgis/apply_deltas.py
+++ b/docker-qgis/apply_deltas.py
@@ -166,7 +166,7 @@ class DeltaException(Exception):
 
 
 BACKUP_SUFFIX = ".qfieldcloudbackup"
-delta_log = []
+DELTA_LOG = []
 
 
 def project_decorator(f):
@@ -254,7 +254,7 @@ def delta_apply(
     inverse: bool,
     overwrite_conflicts: bool,
 ):
-    del delta_log[:]
+    DELTA_LOG.clear()
 
     project = QgsProject.instance()
     logging.info(f'Loading project file "{project_filename}"...')
@@ -275,8 +275,8 @@ def delta_apply(
     if not all_applied:
         logger.info("Some deltas have not been applied")
 
-    delta_log_copy = [*delta_log]
-    del delta_log[:]
+    delta_log_copy = [*DELTA_LOG]
+    DELTA_LOG.clear()
 
     return delta_log_copy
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-570 Form Submission - #2021-05-10T11:30:27Z](https://app.clickup.com/t/2192114/QF-570)

suricactus:
Hey, @miili , thanks for the PR.

The cleanup you did is something I appreciate and it is mergeable.

Note the typing comments are relevant by today, but not even started ;) . Let's leave them for a better day, actually I have some other ideas for this script in the future.

> Took the liberty and rename delta_log to DELTA_LOG, as it is a global variable which is being reused. The global state here is a dangerous thing. Maybe pass along a lightweight context and fill it with information.

It was considered to pass the context around in the past, it was a bit cumbersome to deal with at the time, but we might reconsider it.

---

Now to the point of this PR. The change of the message you did is a bit misleading. The reason why we check for invalid function is because `get_feature` always returns `QgsFeature`. If a new instance of `QgsFeature` is created, by default it is invalid. That being said, returning an invalid `QgsFeature` might be because:
- `getFeatures` failed to find the targeted feature and we return the default `QgsFeature` (line 1031)
- the provider fails to read the feature from the storage and it is invalid

At first glance it makes sense we return `None` if no feature is found and distinguish the two scenarios above, but this might have other implications and needs a bit more thinking and digging in the history. But note, changing the `get_feature` part will take longer to get reviewed, tested, merged and deployed.

To summarize, please remove the changes related to the delta exception and we can merge this PR.

miili:
I made some changes to `get_feature`, which now raises the exceptions `FeatureNotFoundError` and `DoublicateFeatureError`. Thus a validated `QgsFeature` is always returned.

Returning a `None` can be complicated as the caller has to to validation and a `None` can lead to bad and untraceable behavior. I strongly advocate against that pattern and rather use an explicit and meaningful try-except.

suricactus:
As I said, changing the `get_feature` will take longer to get reviewed, tested, merged and deployed. Will get back to you when clean up the immediate backlog.

PR #545, opened 2023-03-06
Added filter to Admin>Core>Delta filtering delta jobs on status pendi…

why-not-try-calmer:
…ng+started or not.

PR #574, opened 2023-03-17
Fixed missing handling of the 'ALL' case

why-not-try-calmer:


docker-app/qfieldcloud/core/admin.py
```diff
@@ -634,11 +634,16 @@ def lookups(self, request, model_admin):
...
+        not_finalized = Q(last_status="pending") | Q(last_status="started")
+        match self.value():
+            case "ALL":
+                return queryset
+            case "not finalized":
+                return queryset.filter(not_finalized)
+            case "finalized":
+                return queryset.filter(~not_finalized)
+            case _:
+                messages.warning(request, f"Status {self.value()} is not implemented!")
```
suricactus, line 646: Do not use `match/case` keywords in the qfieldcloud codebase. They are unintuitive for people coming from other languages and may lead to errors I would prefer we never open this door. Python survived for 30+ years with simple `if/else` and it is among most popular languages.

When the default "All" option is selected in the UI, the `self.value()` returns `None`, not the string `ALL`. In the initial comment in the original PR/commit I never explicitly said this is the returned value in such cases.

Finally, in the `case _:`, it should `assert` or raise an exception, because such situation means the developer forgot to add the option in the `lookups`.

suricactus:
IMO this PR was never tested locally before opening the PR, am I correct?

image

docker-app/qfieldcloud/core/admin.py
suricactus: Link to original discussion: https://github.com/opengisch/qfieldcloud/pull/545#discussion_r1139986268

suricactus: something like this should do it. I have not tested myself, some github python writing.

```
    def queryset(self, request, queryset):
        value = self.value()

        if value is None:
            return queryset

        q = Q(status="pending") | Q(status="started") | Q(status="queued")          
        if value == "not finalized":
            return queryset.filter(q)
        elif value == "finilized":
            return queryset.filter(~q)
        else:
            raise NotImplementedError(f"Uknown filter {value=}")
```

suricactus (changes requested):
Please make the fix also for both `Job` and `Delta` filter you added recently.

suricactus:
Please use the `bug` tag instead of `bugfix`, as the changelog generation is expecting a combination between any of these two:

- `bug` | `enhancement`
- `patch` | `minor` | `major`

I have removed the `OK despite flaky test`, because it is not ok yet.

why-not-try-calmer:
> IMO this PR was never tested locally before opening the PR, am I correct?
> 
> image

image

docker-app/qfieldcloud/core/admin.py
```diff
@@ -633,12 +633,21 @@ def lookups(self, request, model_admin):
...
-        q = Q(status="pending") | Q(status="started") | Q(status="queued")
-        if self.value() == "not finalized":
-            return queryset.filter(q)
+    def queryset(self, request, queryset) -> QuerySet:
+        value = self.value()
+
+        if value is None:
+            return queryset
+
+        not_finalized = Q(status="pending") | Q(status="started") | Q(status="queued")
```
suricactus, line 642: Last thing, which I spotted only now. Is there a reason to prefer to string representation that using `Job.Status` enum?

You can either keep it with multiple `Q(status=Job.Status.PENDING) | ...` or write it in a single line with `Q(status__in=(Job.Status.PENDING, Job.Status.STARTED, Job.Status.QUEUED))`. No strong opinion, whatever suits you more. The latter `__in` is more common in the codebase.

PR #575, opened 2023-03-17
Qf 2598 django admin speed

faebebin:
Speed up django admin interface for large lists

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -62,13 +62,17 @@ admin.site.unregister(LogEntry)
 Invitation = get_invitation_model()
 
 
-class QFieldCloudModelAdmin(admin.ModelAdmin):
+class ModelAdminEstimateCountMixin:
     # Avoid repetitive counting large list views.
     # Instead use pg metadata estimate.
     paginator = LargeTablePaginator
 
     # Display '(Show all)' instead of '(<count>)' in search bar
     show_full_result_count = False
+
+
+class QFieldCloudModelAdmin(ModelAdminEstimateCountMixin, admin.ModelAdmin):
+    pass
 
 
 def admin_urlname_by_obj(value, arg):
@@ -1118,7 +1122,7 @@ class UserAdmin(QFieldCloudModelAdmin):
         return False
 
 
-class LogEntryAdmin(QFieldCloudModelAdmin, BaseLogEntryAdmin):
+class LogEntryAdmin(ModelAdminEstimateCountMixin, BaseLogEntryAdmin):
     list_filter = ["action"]
     search_fields = [
         *BaseLogEntryAdmin.search_fields,
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -1005,10 +1005,7 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
         "storage_usage__field",
     )
 
-    list_select_related = (
-        "organization_owner",
-        "useraccount__current_subscription__plan",
-    )
+    list_select_related = ("organization_owner",)
 
     list_filter = ("date_joined",)
 
@@ -1020,9 +1017,12 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
             instance.organization_owner, instance.organization_owner.username
         )
 
-    @admin.display(description=_("Total Storage"))
+    @admin.display(description=_("Storage"))
     def storage_usage__field(self, instance) -> str:
-        return f"{instance.useraccount.current_subscription.plan.storage_mb} Mb"
+        used_storage = filesizeformat10(instance.useraccount.storage_used_bytes)
+        free_storage = filesizeformat10(instance.useraccount.storage_free_bytes)
+        used_storage_perc = instance.useraccount.storage_used_ratio * 100
+        return f"{used_storage} {free_storage} ({used_storage_perc:.2f}%)"
 
     def get_search_results(self, request, queryset, search_term):
         filters = search_parser(
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -634,6 +634,9 @@ class IsFinalizedJobFilter(admin.SimpleListFilter):
         )
 
     def queryset(self, request, queryset):
+        # Return All
+        if self.value() is None:
+            return queryset
         q = Q(status="pending") | Q(status="started") | Q(status="queued")
         if self.value() == "not finalized":
             return queryset.filter(q)
@@ -672,6 +675,7 @@ class JobAdmin(admin.ModelAdmin):
         "output__pre",
         "feedback__pre",
     )
+
 
     def get_object(self, request, object_id, from_field=None):
         obj = super().get_object(request, object_id, from_field)
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -6,6 +6,8 @@ from datetime import datetime
 from itertools import chain
 from typing import Any, Dict, Generator
 
+from django.contrib.admin.models import LogEntry
+
 from allauth.account.admin import EmailAddressAdmin
 from allauth.account.forms import EmailAwarePasswordResetTokenGenerator
 from allauth.account.models import EmailAddress
@@ -15,6 +17,7 @@ from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.admin.templatetags.admin_urls import admin_urlname
 from django.contrib.auth.models import Group
+from django.contrib.auth import get_user_model
 from django.core.exceptions import PermissionDenied
 from django.db.models import Q
 from django.db.models.fields.json import JSONField
@@ -52,6 +55,11 @@ from qfieldcloud.core.models import (
 from qfieldcloud.core.templatetags.filters import filesizeformat10
 from qfieldcloud.core.utils2 import jobs
 from rest_framework.authtoken.models import TokenProxy
+from auditlog.models import LogEntry
+from auditlog.admin import LogEntryAdmin as BaseLogEntryAdmin
+
+admin.site.unregister(LogEntry)
+
 
 Invitation = get_invitation_model()
 
@@ -637,6 +645,7 @@ class IsFinalizedJobFilter(admin.SimpleListFilter):
         # Return All
         if self.value() is None:
             return queryset
+
         q = Q(status="pending") | Q(status="started") | Q(status="queued")
         if self.value() == "not finalized":
             return queryset.filter(q)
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-2598 Enhance django admin interface](https://app.clickup.com/t/2192114/QF-2598)

docker-app/qfieldcloud/core/paginators.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        c = getattr(self.object_list, "count", None)
+        if callable(c) and not inspect.isbuiltin(c) and method_has_no_args(c):
+            estimate = 0
+            if not self.object_list.query.where:
+                try:
+                    cursor = connection.cursor()
+                    cursor.execute(
+                        "SELECT reltuples FROM pg_class WHERE relname = %s",
+                        [self.object_list.query.model._meta.db_table],
```
faebebin, line 32: FIXME: Rather check for db type instead of a try catch block.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -997,12 +1018,9 @@ def organization_owner__link(self, instance):
...
         )
 
-    @admin.display(description=_("Storage"))
+    @admin.display(description=_("Total Storage"))
     def storage_usage__field(self, instance) -> str:
-        used_storage = filesizeformat10(instance.useraccount.storage_used_bytes)
-        free_storage = filesizeformat10(instance.useraccount.storage_free_bytes)
-        used_storage_perc = instance.useraccount.storage_used_ratio * 100
-        return f"{used_storage} {free_storage} ({used_storage_perc:.2f}%)"
+        return f"{instance.useraccount.current_subscription.plan.storage_mb} Mb"
```
suricactus, line 1023: Is this causing a performance issue? The current calculation is quite handy for support.

Note that `instance.useraccount.current_subscription.plan.storage_mb` is not the same as `used_storage`.

I guess `storage_free_bytes` is having some performance issues, right?

Also, please use `filesizeformat10` for formatting, I guess it should be super cheap call.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1103,6 +1121,14 @@ def has_module_permission(self, request: HttpRequest) -> bool:
         return False
 
 
+class LogEntryAdmin(ModelAdminEstimateCountMixin, BaseLogEntryAdmin):
+    list_filter = ["action"]
+    search_fields = [
+        *BaseLogEntryAdmin.search_fields,
+        "content_type__model",
+    ]
+
```
faebebin, line 1130: Maybe should add some checks to avoid problems if upstream changes?

Performance gain:

The most expensive query comes from their filtering:

`    list_filter = ["action", ResourceTypeFilter, CIDFilter]` 
Where ResourceTypeFIlter ( https://github.com/jazzband/django-auditlog/blob/0cb3eb2b996cc14498dbdd94429169c64e23cfb9/auditlog/filters.py#LL5 ) gets slow on large dataset with many types.
image
(the red bar)

Instead could enable filtering the model type as search which then only takes time loading the specific filtered list. Or alternatively (TODO) provide a custom filter with options derived from available models list directly instead.

result:
image

Now next expensive queries are the two counts. With the `ModelAdminEstimateCountMixin` can complete avoid this:

Leading to only the actual list query taking up some time:
image


Note this is done on a large  number of entries (+ 5M) but not complete data. Thus on production the queries still take way longer.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -49,13 +51,25 @@
...
 from rest_framework.authtoken.models import TokenProxy
 
+admin.site.unregister(LogEntry)
+
+
 Invitation = get_invitation_model()
 
 
+class ModelAdminEstimateCountMixin:
+    # Avoid repetitive counting large list views.
```
faebebin, line 66: As on production all lists are quite long one could consider adding the mixin to all Tables (TODO)?

docker-app/qfieldcloud/core/paginators.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class LargeTablePaginator(Paginator):
+    """
+    Only for Postgres:
+    Overrides the count method to get an estimate instead of actual count when not filtered
+    inspired by: https://djangosnippets.org/snippets/2855/
+    """
+
+    EXACT_COUNT_LIMIT = 10000
```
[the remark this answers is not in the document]
suricactus: Yes, please. I am slowly trying to prefix everything that is QFC related with `QFIELDCLOUD_`, so a name such as `QFIELDCLOUD_ADMIN_EXACT_COUNT_LIMIT` looks fine?

docker-app/qfieldcloud/core/paginators.py
suricactus: hm, the db should always return float. Why don't you do the coerce in SQL already:

```
SELECT reltuples::int FROM pg_class WHERE relname = 'core_delta'
```

docker-app/qfieldcloud/core/admin.py
faebebin: Yes the `storage_free_bytes` is expensive. 

Ok, will have a close look again.

faebebin:
@suricactus thanks for your input - will implement.

What about indexing: `... index the fields you're filtering on for faster searches. You can add indexes to your models by adding the index=True attribute to your model fields ...`

I see alot of `CREATE INDEX ...` on the db. Did you already consider this specifically when adding any filter and search options? Do you think this has optimization potential should be investigated?

suricactus:
@faebebin thanks a lot for your PR!

Sure, indexing is an ongoing project. I am adding indices as soon as I detect their absense. A lot of columns were added without thinking about the Admin performance. So yes, feel free to add indices where it is needed. Adding more indices for sure will speed up stuff, e.g. #539 was exactly doing this.

docker-app/qfieldcloud/core/admin.py
suricactus: Thanks for the input. What is not clear is how `    list_filter = ["action", ResourceTypeFilter, CIDFilter]` is expensive when no filter is applied.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -634,14 +648,18 @@ def lookups(self, request, model_admin):
...
     def queryset(self, request, queryset):
+        # Return All
+        if self.value() is None:
+            return queryset
+
         q = Q(status="pending") | Q(status="started") | Q(status="queued")
         if self.value() == "not finalized":
             return queryset.filter(q)
         else:
             return queryset.filter(~q)
```
suricactus, line 659: This should be sorted in #574, you can revert the change to avoid conflicts.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1117,7 +1118,7 @@ def has_module_permission(self, request: HttpRequest) -> bool:
         return False
 
 
-class LogEntryAdmin(ModelAdminEstimateCountMixin, BaseLogEntryAdmin):
+class LogEntryAdmin(QFieldCloudModelAdmin, BaseLogEntryAdmin):
```
suricactus, line 1121: I am afraid this might cause troubles. We do have `QFieldCloudModelAdmin` that inherits `admin.ModelAdmin` and `BaseLogEntryAdmin` which most likely also inherits `admin.ModelAdmin`. I do not know what will be the consequences, but might cause some weird errorrs we better prevent from happening.

What if we combine the previous and the current approach in something like:

```
class ModelAdminEstimateCountMixin(): ...

class QFieldCloudModelAdmin(ModelAdminEstimateCountMixin, admin.ModelAdmin): ...

# pretty much all admin classes extend directly from QFieldCloudModelAdmin
class OrganizationAdmin(QFieldCloudModelAdmin): ...

# extending only the `ModelAdminEstimateCountMixin` to not extend `admin.ModelAdmin` twice.
class LogEntryAdmin(ModelAdminEstimateCountMixin, BaseLogEntryAdmin): ...
```

docker-app/qfieldcloud/core/admin.py
faebebin: For the `ResourceTypeFilter` they need to query and join `django_content_type` over the whole dataset to get the filter options

image

Because they do this expensive query in `def lookups(....)` method (s. here https://github.com/jazzband/django-auditlog/blob/0cb3eb2b996cc14498dbdd94429169c64e23cfb9/auditlog/filters.py#LL5 )

The best would be to just create an alternative filter by either hardcoding the lookup values or just all the model names from pg meta data (a bit like the estimate count) or retrieving this information somehow dynamically from the running python/django by getting all the admin registered model names?

suricactus: I see.

I did not follow the metadata part of your proposal, but I hope my suggestion below fits in the `somehow dynamically`.

Pseudo code suggestion:

```
from django_auditlog import ResourceTypeFilter

class QFieldCloudResourceTypeFilter(ResourceTypeFilter):
    def lookups(self, request, model_admin):
        qs = ContentType.objects.all().order_by("model")
        types = qs.values_list("id", "model")
        return types
```

suricactus: we can further make the function cacheable, as it is something that changes only if the app restarts. Actually there might be a django caching mechhanism for ContentTypes already, worth checking.

faebebin: > 

That looks very promising. Makes so much more sense than a distinct query over the whole dataset of logentries ... . We might also suggest this on their repo (upstream)?

faebebin: Yes makes sense

faebebin: Your pseudo code works exactly as is :D. Now the query for lookups only depends on model count not and quering the list entries (limited to page) is the main query

image

only the filter options are too many. IMO that is even acceptable for internal use?

docker-app/qfieldcloud/core/admin.py
```diff
@@ -641,7 +661,14 @@ def queryset(self, request, queryset):
...
 
-class JobAdmin(admin.ModelAdmin):
+class QFieldCloudResourceTypeFilter(ResourceTypeFilter):
+    def lookups(self, request, model_admin):
+        qs = ContentType.objects.all().order_by("model")
+        types = qs.values_list("id", "model")
+        return types
+
+
+class JobAdmin(QFieldCloudModelAdmin):
```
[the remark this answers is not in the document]
suricactus: :bow:

docker-app/qfieldcloud/core/admin.py
suricactus: ok, so we agreed in chat we revert this change for this time and have a follow-up PR implementing this.

suricactus (approved):

PR #583, opened 2023-03-19
Increase dh primes to 4096 bits and load `ssl-dhparams.pem` and `options-ssl-nginx.conf` configs

### LINKED ISSUE #527: DH prime number should be larger
The included Diffie-Hellman prime number in the nginx configuration is only 2048 bit long (see [here](https://github.com/opengisch/qfieldcloud/blob/master/conf/nginx/ssl-dhparams.pem))

Current recommendation (since 2023) from the [BSI](https://www.bsi.bund.de) is a number longer than 3000 bit (see [here](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile#subsection.8.2.1)

stcz:
Currently, the options-ssl-nginx.conf file is not included in the nginx configuration.

This will add it.

Further increase the size of the dh primes to 4096 bit (closes #527)


suricactus (approved):

stcz:
I just saw, that this might conflict with `scripts/init_letsencrypt.sh`

As I understand it correctly, this script should init the nginx configuration, if the configuration path is changed.

stcz:
`scripts/init_letsencrypt.sh` should now not be affected, but currently it would download only 2048 bit primes. This should not be a problem. 

Another solution might be to remove the files from the repository and create them with `scripts/init_letsencrypt.sh`

m-kuhn:
Thanks @stcz 
@suricactus I would suggest we add a freshly generated 4096 bit prime (generated from openssl) and merge

PR #585, opened 2023-03-19
Replace deprecated scale with deploy.replicas

stcz:
In the docker compose files `scale:` is used, which is deprecated:
```
WARN[0000] `scale` is deprecated. Use the `deploy.replicas` element 
```
This PR replaces it with deploy.replicas.

Since docker 1.18 Variables of the wrong type (what is the case here) [are casted](https://github.com/docker/compose/pull/5291).

### DIFF
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -20,7 +20,8 @@ services:
       - redis
 
   worker_wrapper:
-    scale: ${QFIELDCLOUD_WORKER_REPLICAS}
+    deploy:
+      replicas: ${QFIELDCLOUD_WORKER_REPLICAS}
     build:
       args:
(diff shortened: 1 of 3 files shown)


suricactus:
While I do agree it is a step in the right direction, we still have some environments that were experiencing troubles with this setting.

Cannot remember right now, but the deprecated setting is used on purpose for now.

Take this as a warning it might take a bit longer than anticipated to get this merged.

suricactus:
Closing since `scale` was un-deprecated, see https://github.com/compose-spec/compose-spec/commit/d53d354ccd74501289e4aa3f9ffa4bd2c3fa7963

In docker 24.x it is still causing the warning, but in 25.x it is no longer the case.

PR #586, opened 2023-03-19
Allow to change the auth token expiration time in .env

stcz:
It is useful to be able to change the `AUTH_TOKEN_EXPIRATION_HOURS` Variable in the environment. 

Please have a look at the variable placement in the `docker-compose.yml` because lines above it is mentioned, that Variables below are for worker_wrapper service. I cannot comprehend this, but pls just change when there is a better location.

Existing .env files should not be affected from this.


docker-app/qfieldcloud/settings.py
```diff
@@ -234,7 +234,7 @@
 
 # QFieldCloud variables
 AUTH_TOKEN_LENGTH = 100
-AUTH_TOKEN_EXPIRATION_HOURS = 24 * 30
+AUTH_TOKEN_EXPIRATION_HOURS = int(os.environ.get("QFIELDCLOUD_AUTH_TOKEN_EXPIRATION_HOURS", 24 * 30))
```
stcz, line 237:
```suggestion
AUTH_TOKEN_EXPIRATION_HOURS = int(
    os.environ.get("QFIELDCLOUD_AUTH_TOKEN_EXPIRATION_HOURS", 24 * 30)
)
```
Should I apply this patch to fix the test? I wouldn't because in online it looks nicer to me. `DEBUG` doesn't meet this too.

suricactus: It's not a matter of taste, it's a matter of tooling. Please apply the patch to get this merge-able.

suricactus:
@stcz do you need these changes or you propose them as nice to have?

stcz:
> @stcz do you need these changes or you propose them as nice to have?

I would say i need them.. ;)

We have tablets, that are used all two weeks by different persons, so it is useful when there is no need to tell them all the passwords to reauthenticate them.

suricactus:
Just wanted to make sure the proposed new configurations (this and the other similar PRs from you) are needed because you faced a limitation of the current configuration possibilites.

Thanks for the changes!

suricactus (approved):

.env.example
```diff
@@ -95,6 +95,10 @@ QFIELDCLOUD_WORKER_REPLICAS=1
 # DEFAULT: subscription.Subscription
 QFIELDCLOUD_SUBSCRIPTION_MODEL=subscription.Subscription
 
+# QFieldCloud auth token expiration hours
+# DEFAULT: 720 (30 days)
```
suricactus, line 99:
```suggestion
# QFieldCloud auth token expiration hours. E.g. 720 hours (30 days)
# DEFAULT: 720
```

suricactus: Please keep the lines with `DEFAULT:` with copy-pastable contents.

docker-app/qfieldcloud/settings.py
```diff
@@ -234,7 +234,9 @@
 
 # QFieldCloud variables
 AUTH_TOKEN_LENGTH = 100
-AUTH_TOKEN_EXPIRATION_HOURS = 24 * 30
+AUTH_TOKEN_EXPIRATION_HOURS = int(
+    os.environ.get("QFIELDCLOUD_AUTH_TOKEN_EXPIRATION_HOURS", 24 * 30)
+)
```
suricactus, line 240: Note this is not backwards compatible with the existing environments, because `QFIELDCLOUD_AUTH_TOKEN_EXPIRATION_HOURS` variable will be an empty string and the default value will not be taken. See #589 .

stcz: :see_no_evil: I will fix this in the evening

stcz: Done

PR #587, opened 2023-03-19
Allow to change the default timezone in .env

stcz:
It is useful to be able to change the default timezone variable in the environment.

Please have a look at the variable placement in the docker-compose.yml because lines above it is mentioned, that Variables below are for worker_wrapper service. I cannot comprehend this, but pls just change when there is a better location.

Existing .env files should not be affected from this.


suricactus:
Thanks for the PR. 

On the comment:
All variables are passed to the `app` service. `worker_wrapper` service extends the `app` service. But not all variables are used in `app`, e.g. `QFIELDCLOUD_WORKER_QFIELDCLOUD_URL` is passed in `app`, but actually used in `worker_wrapper`. AFAIR it is not so easy to extend the `services.<service>.environment` key in docker compose. Anyway, while suboptimal, it is working and not planned to change in the near future.

Even if you want to invest time and you find a more elegant way to do this, please propose it as an issue and not as PR for now, I will not merge it.

suricactus (approved):

PR #588, opened 2023-03-19
Fix comparing the deltas with current data for some some datatypes, mostly dates and times

### LINKED ISSUE #572: Conflict QDateTime and String
Saw this conflict more than a couple times:

```
The attribute "created" that has a conflict:
2023-03-02T10:21:21.000
PyQt5.QtCore.QDateTime(2023, 3, 2, 10, 21, 21)
```

suricactus:
`QgsFeature.attribute()` will return an instance of `QDate`, `QTime` or `QDateTime`,
depending on the attribute datatype. We need to explicitly handle such situations.

Note geopackages does not support `time` fields, so this case is untested, but should follow the same logic as date/datetime.

Fixes #572 .


duke-nyuki:
Task linked: [QF-2607 Conflict when applying deltas on attributes with QDateTime and QDate](https://app.clickup.com/t/2192114/QF-2607)

docker-qgis/apply_deltas.py
```diff
@@ -1268,9 +1277,23 @@ def compare_feature(
...
+                    )
+                elif isinstance(current_value, QDate):
+                    incoming_value = QDate.fromString(incoming_value, Qt.ISODate)
+                elif isinstance(current_value, QTime):
+                    incoming_value = QTime.fromString(incoming_value)
+
+            if current_value != incoming_value:
                 conflicts.append(
-                    f'The attribute "{attr}" that has a conflict:\n-{delta_feature_attrs[attr]}\n+{feature.attribute(attr)}'
+                    f'The attribute "{attr}" that has a conflict:\n-{current_value}\n+{incoming_value}'
```
suricactus, line 1296: @m-kuhn kind request for review. If I understand correctly, there are no more interesting return values from `QgsFeature.attribute(QString)`. Tried with binary and json, also checked `QVariant` docs, and seems everything else will be a datatype supported from JSON (as `incoming_value` is read from delta's JSON).

m-kuhn: I think that is correct

m-kuhn (approved):

PR #589, opened 2023-03-19
Fix backwards compatibility when `QFIELDCLOUD_DEFAULT_TIME_ZONE` is not set

suricactus:
In theory, we should mark all docker-compose envvars mandatory. But it's not the right moment for that right now, already in the backlog.

Fixes a bug introduced by #587 .


stcz:
:see_no_evil: I will fix this in the evening

PR #591, opened 2023-03-21
Added column showing number of recently active users to projets under `OrganizationAdmin`

why-not-try-calmer:
… OrganizationAdmin

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -2,7 +2,7 @@ import csv
 import json
 import time
 from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime
 from itertools import chain
 from typing import Any, Dict, Generator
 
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/admin.py
```diff
@@ -1025,17 +1026,20 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
...
     list_filter = ("date_joined",)
 
     autocomplete_fields = ("organization_owner",)
 
+    @admin.display(description=_("Active users (last 90 days)"))
+    def active_users(self, instance) -> int:
+        now = datetime.now()
+        three_months_ago = now - timedelta(days=90)
+        return len(list(instance.active_users(now, three_months_ago)))
+
```
suricactus, line 1042: Actually you can use `org.current_subscription.active_users_count`, which will do it for the last billing period, if the subscription was active and probably raise? otherwise. If the latter case we better return `None`, which makes the admin show `-`.

suricactus: And one remark, look at the page performance. If it get's 100%+ slower, probably it should not be a column in the List view, but readonly field in the edit view.

PR #593, opened 2023-03-23
Custom exception

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2539 When trying to add a new member in organization that reached the limit show nice error instead of HTTP 500](https://app.clickup.com/t/2192114/QF-2539)

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -167,3 +167,7 @@ class ProjectAlreadyExistsError(QFieldCloudException):
     code = "project_already_exists"
     message = "This user already owns a project with the same name."
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class ReachedMaxOrgaMembers(Exception):
```
suricactus, line 172:
```suggestion
class ReachedMaxOrgMembersError(ValidationError):
```

suricactus: So far I keep everything explicit without a lot of shortening. More typing, but clear explicitness.

why-not-try-calmer: The thing I don't like is that it's not a validation error, and by the same token, should not to have to declare a `status_code`.

suricactus: ok, second thought. Define it like the rest of exceptions in the file, but extend `QFieldCloudException`.

YOLO if someone relies on `ValidationError`.

why-not-try-calmer: Not sure if the just added status code is correct though.

PR #594, opened 2023-03-27
Configure locales and languages

suricactus:
Does not cause any harm if configured.


m-kuhn:
Task linked: [QF-371 Extract translations from QFieldCloud](https://app.clickup.com/t/2192114/QF-371)

PR #604, opened 2023-03-30
Don't throw if project deleted after job instantiation

faebebin:
Could reproduce and test locally and works. IMO we don't need to report when user does this.
`logger.info` is not reported to sentry, right?

Maybe we could even cancel the whole process somwhere earlier if project is deleted, but as it is a rare case not worth the effort?

### DIFF
--- a/docker-app/worker_wrapper/wrapper.py
+++ b/docker-app/worker_wrapper/wrapper.py
@@ -24,6 +24,7 @@ from qfieldcloud.core.models import (
     Job,
     PackageJob,
     ProcessProjectfileJob,
+    Project,
     Secret,
 )
 from qfieldcloud.core.utils import get_qgis_project_file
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-2634 Jobrun raises error if project was deleted in the meantime](https://app.clickup.com/t/2192114/QF-2634)

suricactus:
Yes, this is also my understandin of the problem:

1) one creates a new project
2) uploads a new .qgs file, which causes a new `process_projectfile` job is queued
3) the queue picks the job
4) the user being impatient, deletes the project, but the background process is still running, because it is not notified the project was deleted
5) the job finishes and tries to update the project status, but the project record no longer exists

I believe you can reproduce the error if you do step 4) in the right moment. Since the `process_projectfile` job is usually quite fast, you can temporary add a sleep in the job to make sure you can reproduce.

PR #605, opened 2023-03-30
Add provider name to get more detailed layer information

suricactus:


duke-nyuki:
Task linked: [QF-2633 Add provider name to get more detailed layer information](https://app.clickup.com/t/2192114/QF-2633)

Issue #592, opened 2023-03-22
Cannot add team as collaborator in Django backend


stcz:
Thank you again for the great project!

Currently, I'm not able to add a Team as a collaborator to a project on a local instance within the Django backend. There are only users listed.

Steps to reproduce:

1. Create an organization
2. Create a Team for the organization
3. Add Members to the team
4. Create a Project, that is owned by the organization
5. When now trying to add a team as collaborator, only users of the organization are listed, not the teams.

suricactus:
Have not tried myself but it well could be. It is not something in immediate plans, so if you need this functionality, feel free to open a PR or drop us an email at sales@qfield.cloud .

stcz:
Thx for the information. I will try to implement this.

stcz:
The SQL Query that is executed during the autocompletion is the following:
```
Executed SQL
    SELECT "core_user"."id",
           "core_user"."password",
           "core_user"."last_login",
           "core_user"."is_superuser",
           "core_user"."first_name",
           "core_user"."last_name",
           "core_user"."email",
           "core_user"."is_staff",
           "core_user"."is_active",
           "core_user"."date_joined",
           "core_user"."username",
           "core_user"."type",
           "core_person"."user_ptr_id",
           "core_person"."remaining_invitations",
           "core_person"."has_newsletter_subscription",
           "core_person"."has_accepted_tos",
           "core_organization"."user_ptr_id",
           "core_organization"."organization_owner_id",
           "core_organization"."is_initially_trial",
           "core_organization"."created_by_id",
           "core_organization"."created_at",
           "core_organization"."updated_at",
           "core_team"."user_ptr_id",
           "core_team"."team_organization_id"
      FROM "core_user"
      LEFT OUTER JOIN "core_person"
        ON ("core_user"."id" = "core_person"."user_ptr_id")
      LEFT OUTER JOIN "core_organization"
        ON ("core_user"."id" = "core_organization"."user_ptr_id")
      LEFT OUTER JOIN "core_team"
        ON ("core_user"."id" = "core_team"."user_ptr_id")
     WHERE ("core_user"."type" IN (1, 2) AND "core_user"."type" IN (1, 3))
     ORDER BY LOWER("core_user"."username") ASC
     LIMIT 2
```
The `WHERE` clause looks quite strange to me. 
[...]

PR #610, opened 2023-04-10
Remove filter from user autocompletion

stcz:
closes 592

This filter creates the SQL WHERE Clause, that causes the autocompletion filters to fail. (see [here](https://github.com/opengisch/qfieldcloud/issues/592#issuecomment-1500889942))

Usually, the users are filtered depending on the requirement of the field. 

I checked, the other autocompletion fields. They are still filtering as intended. 

I don't think that there are other side effects, because the class is only used by autocompletion.


stcz:
@suricactus can you review and merge this or tell me, what else is needed?

PR #611, opened 2023-04-11
Qf 2641 code style improvements

faebebin:
#QF-2641

### DIFF
--- a/docker-app/qfieldcloud/core/management/commands/dequeue.py
+++ b/docker-app/qfieldcloud/core/management/commands/dequeue.py
@@ -101,13 +101,6 @@ class Command(BaseCommand):
             if options["single_shot"]:
                 break
 
-    def run(self, job_id, *args, **options):
-        try:
-            job = Job.objects.get(id=job_id)
-            self._run(job)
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-2641 Code style](https://app.clickup.com/t/2192114/QF-2641)

docker-app/qfieldcloud/core/management/commands/dequeue.py
```diff
@@ -101,13 +101,6 @@ def handle(self, *args, **options):
             if options["single_shot"]:
                 break
 
-    def run(self, job_id, *args, **options):
-        try:
-            job = Job.objects.get(id=job_id)
-            self._run(job)
-        except Exception:
-            pass
-
```
faebebin, line 110: Seems dead code. And because there is also a JobRunner.run() it causes quite a bit of confusion.

suricactus: Agreed. Thanks for spotting.

suricactus (approved):

PR #613, opened 2023-04-12
Fix minio healthcheck local (and upgrade image)

faebebin:


suricactus:
@faebebin yes, `minio` is used only locally. 

Why did you hardcode the URL though?

m-kuhn:
Problem was, @faebebin ran `ufw enable`, which killed inter-docker-communication and therefor the healthcheck, but not local connections. With `ufw disable` things work again and this patch is obsolete.

PR #617, opened 2023-04-19
Qf 2634 Write project.status to db and cancel job workers

faebebin:
Consider: SoftDeletableModel (from model_utils)

Add project.status to db to facilitate cancelling of related job workers and reduce unnecessary exception. 
- Keep project.status updated via Job pre_save signal
- Further simplifies the code by quering project.status directly instead of related jobs. 
#QF-2634

TODO
- [ ]  Why Job pre_save signal not workng properly
- [ ]  Add more Status options to project
- [ ]  Better name for `static_status` - maybe `jobs_status` or simply `status` and rename property?
- [ ]  handle Exceptions
- [ ]  tests


duke-nyuki:
Task linked: [QF-2489 Ignore issues when the QFieldCloud project gets deleted before the jobs are finished](https://app.clickup.com/t/2192114/QF-2489)

PR #619, opened 2023-04-20
Add indexes on core job

m-kuhn:
Make things faster


docker-app/qfieldcloud/core/models.py
```diff
@@ -1501,12 +1501,12 @@ class Status(models.TextChoices):
...
     status = models.CharField(
-        max_length=32, choices=Status.choices, default=Status.PENDING
+        max_length=32, choices=Status.choices, default=Status.PENDING, db_index=True
     )
     output = models.TextField(null=True)
     feedback = JSONField(null=True)
     created_by = models.ForeignKey(User, on_delete=models.CASCADE)
-    created_at = models.DateTimeField(auto_now_add=True)
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
     updated_at = models.DateTimeField(auto_now=True)
```
faebebin, line 1510: @m-kuhn could you also index the `updated_at` as it is the order by field in the admin list display?

faebebin, line 1510: @m-kuhn could you also index the `updated_at` as it is the order by field in the admin list display?

#QF-2613

@suricactus I did not find anything remarkable or easy to improve on `JobAdmin` . The `DeltaAdmin` looks quite similar (same Joins) and has much faster list_display. Thinks it's just because it has an index on `updated_at`. Both get ordered after `-updated_at`.

#QF-2650

faebebin (changes requested):

faebebin (approved):

docker-app/qfieldcloud/core/models.py
faebebin: LGTM

suricactus:
LGTM too, thank you both!

PR #620, opened 2023-04-21
Rename subscription qs.active to current

faebebin:


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -320,11 +320,10 @@ class Package(models.Model):
...
 class SubscriptionQuerySet(models.QuerySet):
-    def active(self):
-        """Returns the subscriptions which are relevant to the current moment.
-
-        WARNING the name `active` is super misleading, should be `current`. Some of the subscriptions in the queryset might not be active, but cancelled or drafted.
-        TODO rename to `current`
+    def current(self):
+        """
+        Returns the subscriptions which are relevant to the current moment.
+        Note: Some of the subscriptions in the queryset might not be active, but cancelled or drafted.
```
suricactus, line 326:
```suggestion
        NOTE Some of the subscriptions in the queryset might not be active, but cancelled or drafted.
```

NOTE well noted, but I use "NOTE" to add notes when a note is noted :)

suricactus: Just to align with the rest of the codebase.

suricactus (approved):

PR #625, opened 2023-04-26
Rename active subscription to current, everywhere

faebebin:


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -595,14 +595,14 @@ def set_package_quantity(
...
-        """Returns the currently active subscription, if not exists returns a newly created subscription with the default plan.
+    def get_or_create_current_subscription(cls, account: UserAccount) -> "Subscription":
+        """Returns the current subscription, if not exists returns a newly created subscription with the default plan.
 
         Args:
             account (UserAccount): the account the subscription belongs to.
 
         Returns:
-            Self: the currently active subscription
+            Self: the current subscription
```
faebebin, line 606: IMO talking just of the `current subscription` works, thus we could also use `currently running subscription` ... ?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -823,5 +828,5 @@ class Meta:
     account = models.OneToOneField(
         UserAccount,
         on_delete=models.CASCADE,
-        related_name="current_subscription",
+        related_name="current_subscription_vw",
```
suricactus, line 831: I added the `_vw` suffix, or rename it to `_ro`, so we know it is read-only when using the ORM?

suricactus (approved):
Good job, thanks!

I am only not so convinced why we would rename `current_subscription_vw` to `current_subscription`.

docker-app/qfieldcloud/subscription/models.py
faebebin: Ah yes that would also work for me

faebebin:
You mean the other way round (current_subscription -> current_subscription_vw)? Because else it conflicts with the method. (I was also thinking about renaming the method 'active_subscription' -> 'running_subscription' , but finally dropped the idea)

suricactus (approved):
Looks good to me. Sorry for my messy comments above, I was not drunk when reviewing, just got my mind far away it seems :)

PR #627, opened 2023-04-26
Exclude the busy projects with a subquery

suricactus:
This prevents potential time gap between selecting the busy projects
and locking the job rows, because other worker_wrapper containers can
consume these jobs and cause troubles with multiple jobs running for
the same project in parallel.

Also simplify the job queue consumption with the second commit, which should not change any application logic.


duke-nyuki:

docker-app/qfieldcloud/core/management/commands/dequeue.py
```diff
@@ -66,28 +66,26 @@ def handle(self, *args, **options):
...
+                    .values("project_id")
                 )
 
-                busy_project_ids = [j["project_id"] for j in busy_projects_ids_qs]
-
                 # select all the pending jobs, that their project has no other active job
                 jobs_qs = (
                     Job.objects.select_for_update(skip_locked=True)
                     .filter(status=Job.Status.PENDING)
                     # TODO: use busy_project_ids_qs as subquery here instead of the result set busy_project_ids to prevent concurrency issues
```
m-kuhn, line 76: This TODO can be removed?

m-kuhn (approved):

docker-app/qfieldcloud/core/management/commands/dequeue.py
suricactus: Thanks for the reminder, yes!

PR #632, opened 2023-04-28
Build separate images for worker and app

m-kuhn:


.github/workflows/build_and_push.yml
```diff
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - master
-      - dispatch_deploy_after_build
```
suricactus, line 6: Isn't this the manual dispatch to build? Why do we remote it?

suricactus (approved):

PR #635, opened 2023-04-30
Increase stop grace period for worker wrapper

m-kuhn:
Let wrappers exit gracefully when scaling


suricactus:
- test with long running job if it is going to wait for these 15 mins (or more than the usual 10 secs)?
- test with no job running if it exists immediatly

suricactus (approved):
Did some tests last month, worked well. Let it be!

PR #639, opened 2023-05-04
Qf 2661 move certain exceptions to subscription

faebebin:


duke-nyuki:
Task linked: [QF-2661 move exceptions to subscriptions module](https://app.clickup.com/t/2192114/QF-2661)

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -169,6 +162,7 @@ class ProjectAlreadyExistsError(QFieldCloudException):
     status_code = status.HTTP_400_BAD_REQUEST
 
 
+@deprecated("moved to subscription")
```
suricactus, line 165: Do we really need to deprecate them, can't we just delete them?

docker-app/qfieldcloud/subscription/exceptions.py
```diff
@@ -1,6 +1,56 @@
...
+
+class SubscriptionException(QFieldCloudException):
     ...
 
 
 class NotPremiumPlanException(SubscriptionException):
     ...
+
+
+class ReachedMaxOrganizationMembersError(QFieldCloudException):
```
suricactus, line 13: Shouldn't all of these onwards extend `SubscriptionException`?

docker-app/qfieldcloud/core/exceptions.py
faebebin: oh sure, only keep  ReachedMaxOrganizationMembersError

suricactus (approved):
Thanks!

PR #640, opened 2023-05-04
Speed (and stabilize?) up test_packages

faebebin:
It did not fix. But think could still add this?


docker-app/qfieldcloud/core/tests/test_packages.py
```diff
@@ -45,10 +45,7 @@ def setUp(self):
             name="project1", is_public=False, owner=self.user1
         )
 
-        try:
-            delete_db_and_role("test", self.user1.username)
-        except Exception:
-            pass
+        delete_db_and_role("test", self.user1.username)
```
faebebin, line 49: IMO only hinders debugging. Locally I ran into error 'db already exsists'. An Error during db deletion is better here?

PR #641, opened 2023-05-05
Remove milliseconds from `active_since` and `active_until` respectively

suricactus:


duke-nyuki:
Task linked: [QF-2591 Subscription prevented because of active_since/active_until ms](https://app.clickup.com/t/2192114/QF-2591)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -770,6 +772,10 @@ def create_subscription(
 
         TODO Python 3.11 the actual return type is Self
         """
+        if active_since:
+            # remove milliseconds as there will be slight shift with the remote system data
+            active_since = active_since.replace(microsecond=0)
+
```
faebebin, line 778: Does this still work around the full second where one timestamp is `[date] hh:mm:31.01` and remote `[date] hh:mm:30.92` ?

(from clickup discussion / sentry issue)
2023-03-15 15:44:31.718879+00
2023-03-15 15:44:31.623984+00

suricactus: actually the remote timestamp is never with milliseconds.

suricactus: I don't think it would be a problem to have a gap between two subscriptions, the main worry here is to prevent overlaps. While not so nice for reports, it's a gap we can live with.

faebebin (approved):
LGTM

PR #643, opened 2023-05-05
Workaround vector layer type rename in QGIS 3.30

suricactus:
Calling `QgsVectorLayer().type().name` QGIS 3.30.x returns "Vector", while previous versions return "VectorLayer"

Related issue: https://github.com/qgis/QGIS/issues/53023

Fixes regression introduced in #637 .

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1119,9 +1119,11 @@ class Project(models.Model):
         has_online_vector_layers = False
 
         for layer_data in layers_by_id.values():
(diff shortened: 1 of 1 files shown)


m-kuhn (approved):

nirvn:
Hm, is that worth raising in QGIS itself? Looks like a broken python API promise that shouldn't happen until QGIS 4.0.

m-kuhn:
> Related issue: [qgis/QGIS#53023](https://github.com/qgis/QGIS/issues/53023)

See

> Related issue: https://github.com/qgis/QGIS/issues/53023

PR #644, opened 2023-05-08
Fix prevent job creation if unsufficient subscription

faebebin:
Does not change app logic. Only a fix. 

- [x] check project.owner.useraccount instead of created_by
- [x] move checks to permission_utils: NOT possible (circular import)
- [x] **ensure job status not stuck at QUEUED (or other) and always terminates (as FAILED) ! important**

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1576,12 +1576,12 @@ class Job(models.Model):
                 "The job ended in unknown state. Please verify the project is configured properly, try again and contact QFieldCloud support for more information."
             )
 
-    def raise_insufficient_subscription(self, check_online_layers=True) -> None:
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-2676 Do not start delta apply job if user is over quota](https://app.clickup.com/t/2192114/QF-2676)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1576,12 +1576,12 @@ def fallback_output(self) -> str:
                 "The job ended in unknown state. Please verify the project is configured properly, try again and contact QFieldCloud support for more information."
             )
 
-    def raise_insufficient_subscription(self, check_online_layers=True) -> None:
+    def check_can_create_job(self, check_online_layers=True) -> None:
```
faebebin, line 1579: Importing it from `.permissions_utils` led to circular import errors ...

PR #646, opened 2023-05-09
Speed up delta_apply jobs by using an indexed column for searching

suricactus:
Filtering by `content__clientId` is causing super slow query. While in pure SQL we can index on specific field in the JSON, like:

```CREATE INDEX ON core_delta(((content->>'clientId')::uuid))```

It is not so easy to run a query that uses the index in Django ORM.

Alternatively, we can create index directly on the JSON value:

```CREATE INDEX ON core_delta((content->'clientId'))```

But this will make the index 5x bigger.

In summary, it was chosen to add a new indexed column with the same data as in the JSON, because it's storage cheaper and faster solution.

Also:
Move `before_docker_run` to be before we set job status to `STARTED`


duke-nyuki:
Task linked: [QF-2663 Investigate a huge time gap between `Job.started_at` and first output timestamp](https://app.clickup.com/t/2192114/QF-2663)

duke-nyuki:
Task linked: [QF-2676 Do not start delta apply job if user is over quota](https://app.clickup.com/t/2192114/QF-2676)

faebebin (approved):
LTGM

Tried locally all works. 
BTW think there is probably many places still for those kind of performance issues in filtering (and sorting) mostly concerning the admin . Good to have this as an example.

PR #647, opened 2023-05-09
Make primary key extraction more strict, do not allow layers with missing primary keys

suricactus:


faebebin (approved):
LGTM

Did not test it though

PR #651, opened 2023-05-13
Avoid counting, use qs.exists() instead

faebebin:


suricactus (approved):
LGTM I guess it comes with a performace gain?

faebebin:
Yes, withiut actually analyzing but seems recommended way
```
The Django docs tells us to use querySet.count() if you only want the count, and use queryset.exists() if you only want to find out if at least one result exists.

The reason for this advice is because queryset.count() performs an SQL operation that scans every row in the database table to calculate the sum. On the other hand queryset.exists() simply reads a single record in the most optimized way:

remove ordering
remove grouping
clear any dev-defined select_related and distinct from the queryset
```

PR #652, opened 2023-05-14
[chore] Add and apply `pyupgrade` to `pre-commit`

m-kuhn:
Non critical python modernization, to be applied in a quiet moment.
I would propose to merge this in pieces (changes in tests and scripts first)


docker-app/qfieldcloud/core/models.py
```diff
@@ -523,7 +523,7 @@ def random_string():
         first_letter = secrets.choice(string.ascii_lowercase)
         letters_and_digits = string.ascii_lowercase + string.digits
         secure_str = first_letter + "".join(
-            (secrets.choice(letters_and_digits) for i in range(15))
+            secrets.choice(letters_and_digits) for i in range(15)
```
suricactus, line 526: Check `black` compatibility.

docker-app/qfieldcloud/core/tests/test_deleteorphanedfiles.py
```diff
@@ -25,7 +25,7 @@ def generate_projects(self, count: int):
         for i in range(1, count + 1):
             p = Project.objects.create(name=f"p{offset + i}", owner=self.u1)
             self.projects.append(p)
-            file = io.BytesIO("Hello world!".encode())
+            file = io.BytesIO(b"Hello world!")
```
suricactus, line 28: This is very ASCII centric. But I am ok with it.

docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -282,7 +282,7 @@ def test_push_apply_delta_file_not_json(self):
 
         response = self.client.post(
             f"/api/v1/deltas/{project.id}/",
-            {"file": open(delta_file, "r")},
+            {"file": open(delta_file)},
```
suricactus, line 285: Any chance we disable this rule. To me it is much more readable with explicit `open` flags, rather than memorizing the `open` defaults.

suricactus:
Good move. I would only miss the `open("path/to/file", "r")` as excplitily setting the file handler type.

docker-app/qfieldcloud/core/models.py
m-kuhn: Black is fine with it or we would see errors in pre-commit I guess?

docker-app/qfieldcloud/core/tests/test_delta.py
m-kuhn: I haven't found any options to do this (there are options for minimum python version `--py37-plus` etc, `--keep-percent-format`, `--keep-mock`, `--keep-runtime-typing`, none of the later ones looks particularly interesting to me) ...

I guess we either

- apply the changes and remove the pre-commit
- use all the rules and try to get used to this

suricactus: Let's go to with the latter then.

.pre-commit-config.yaml
```diff
@@ -34,6 +34,13 @@ repos:
     hooks:
       - id: black
 
+  #
```
suricactus, line 37:
```suggestion
  # tool to automatically upgrade syntax for newer versions of the language
```

docker-app/qfieldcloud/core/models.py
suricactus: Black is the previous step in the pre-commit.yml, so I though it might cause problems on the next run of black.

suricactus: But it would have failed with the CI, which is not the case, so ignore my comments.

PR #653, opened 2023-05-14
Some small improvements on type safety side

m-kuhn:


docker-qgis/utils.py
```diff
@@ -558,16 +558,14 @@ def run_workflow(
...
-            print("Feedback:")
-            print(
-                json.dump(
-                    feedback,
-                    feedback_filename,
-                    indent=2,
-                    sort_keys=True,
-                    default=json_default,
-                )
+        if type(feedback_filename) == IO:
```
suricactus, line 561: Hm, even if it works, I am not sure if it is a good idea to use `typing.IO` as a operand in runtime code.

What if we use something like `isinstance(feedback_filename, io.IOBase)`?

docker-qgis/utils.py
```diff
@@ -752,8 +750,8 @@ def format(self, record: logging.LogRecord) -> str:
         if isinstance(record.args, dict):
             for k in record.args.keys():
                 record.args[k] = self.redact(record.args[k])
-        else:
-            record.args = tuple(self.redact(arg) for arg in record.args)
+        elif isinstance(record.args, tuple):
+            record.args = tuple(self.redact(str(arg)) for arg in record.args)
```
suricactus, line 754:
```suggestion
        elif isinstance(record.args, tuple):
            record.args = tuple(self.redact(str(arg)) for arg in record.args)
        else:
            raise NotImplementedError()
```
I think this is a bit safer.

docker-qgis/utils.py
m-kuhn: Good hint !

suricactus (approved):
All good! Let's merge it after next release.

m-kuhn:
@suricactus is the do not merge label still applicable?

suricactus:
Yes, I did not forget it. Will merge after a small patch I want to do.

m-kuhn:
Good, I'll leave it to you

PR #650, opened 2023-05-11
Always accept file uploads from Qfield

faebebin:
This PR adds:
- in `files_views.py` that we do not throw QoutaError if client_type is `worker` or `qfield`
- in `qfield_files_views.py` throw early before packaging for qfield
- in `models.py` that we never throw on ProcessProjectFileJob

The rest is testing related because https://github.com/opengisch/qfieldcloud/pull/655 still needs some fixes

PR #655, opened 2023-05-15
Correctly calculate `Job.started_at`/`Job.finished_at`

suricactus:
Since `before_docker_run`/`after_docker_run` were called respenctively before/after setting the status and `started_at`/`finished_at`, it was causing timing issues in tests and not only.

This PR is aiming at fixing this.


faebebin (commented):
Tests fail - Should I take over fixing the tests - is what I am doing on https://github.com/opengisch/qfieldcloud/pull/650 too at the moment (in the waiting rooms ;)

suricactus:
If it's blocking - go ahead. The fix should be easy - the `project.packaged_at` .

docker-app/worker_wrapper/wrapper.py
```diff
@@ -323,46 +323,44 @@ def before_docker_run(self) -> None:
         self.data_last_packaged_at = timezone.now()
 
     def after_docker_run(self) -> None:
-        # only successfully finished packaging jobs should update the Project.data_last_packaged_at
-        if self.job.status == Job.Status.FINISHED:
```
faebebin, line 318: This check was never necessary, as before the `job.status = Job.status.FINISHED` was set right before calling `after_docker_run` .

docker-app/worker_wrapper/wrapper.py
```diff
@@ -323,46 +323,44 @@ def before_docker_run(self) -> None:
...
+        try:
+            project_id = str(self.job.project.id)
+            package_ids = storage.get_stored_package_ids(project_id)
+            job_ids = [
+                str(job["id"])
+                for job in Job.objects.filter(
+                    type=Job.Type.PACKAGE,
+                )
+                .exclude(
+                    status__in=(Job.Status.FAILED, Job.Status.FINISHED),
```
faebebin, line 344: But the problem is we use the status even more here ... WIP

docker-app/worker_wrapper/wrapper.py
```diff
@@ -314,46 +323,45 @@ def before_docker_run(self) -> None:
...
-                ]
+        try:
+            project_id = str(self.job.project.id)
+            package_ids = storage.get_stored_package_ids(project_id)
+            job_ids = [
+                str(job["id"])
+                for job in Job.objects.filter(
+                    type=Job.Type.PACKAGE,
+                )
+                .exclude(id=self.job_id)
```
faebebin, line 343: Before self.job was excluded with `FINISHED` jobs. Now need to exclude seperately.

PR #660, opened 2023-05-17
Cancel jobs on project delete

faebebin:


duke-nyuki:
Task linked: [QF-2634 Cancel the jobs running for a specific project, if the project gets deleted](https://app.clickup.com/t/2192114/QF-2634)

docker-app/worker_wrapper/wrapper.py
```diff
@@ -266,6 +262,10 @@ def _run_docker(
             cpu_shares=config.WORKER_QGIS_CPU_SHARES,
         )
 
+        # `docker_started_at`/`docker_finished_at` tracks the time spent on docker only
+        self.job.docker_started_at = timezone.now()
+        self.job.container_id = container.id
+        self.job.save()
```
suricactus, line 268: Hm, are you sure? In my understanding before it worked like:

1) `docker_started_at = now()`
2) initiate starting the container in detached mode. The container must have started when the line is finished.
3) block the python code until it finished via `container.wait()`
4) `docker_finished_at = now()`

While in your new case you swapped 1) and 2).

faebebin: I thought as the container.run command returns immediately it would not matter so much and just wanted to group the job.save() (could use also `update_fields`). But can revert this.

suricactus: It is not precisely immediately. It still call the docker daemon and might fail or not.

faebebin: Yep thought if it fails to start/run it is not wrong have NULL there?

suricactus: Absolutely.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -522,3 +547,31 @@ def after_docker_exception(self) -> None:
...
+        "container_id", flat=True
+    )
+
+    # Find all running worker containers who's Project and Job was deleted from the database
+    worker_without_job_ids = set(worker_ids) - set(worker_with_job_ids)
+
+    for worker_id in worker_without_job_ids:
+        container = client.containers.get(worker_id)
+        container.kill()
+        container.remove()
```
faebebin, line 573: Maybe we could add another service that takes additional care of cleaning the worker docker layer and does periodically some kind of:

```
def prune_workers():
    client: DockerClient = docker.from_env()
    # Delete stopped worker containers
    # client.containers.prune(filters={"label": "worker"})
    client.containers.prune()
    # TODO not working?
    #         filters={
    #        "label": "worker"
    #    }
```

docker-app/worker_wrapper/wrapper.py
```diff
@@ -264,15 +267,30 @@ def _run_docker(
             detach=True,
```
faebebin, line 267: @suricactus  Did you have bad experience with `auto_remove=True` , as you commented it?

suricactus: If I remember correctly, configuring the `auto_remove=True` will make it impossible to get the logs if the container is timed out.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -262,10 +262,10 @@ def _run_docker(
...
             mem_limit=config.WORKER_QGIS_MEMORY_LIMIT,
             cpu_shares=config.WORKER_QGIS_CPU_SHARES,
-            labels={
-                "worker": self.job_class.type,
-                "project_id": str(self.job.project_id),
-            }
+            labels=[
+                "worker",
+                str(self.job.project_id),
+            ]
```
suricactus: Why would you prefer lists instead of dictionaries? I think it's easier to consume the labels if they are named.

They are both supported https://docker-py.readthedocs.io/en/stable/containers.html .

docker-app/worker_wrapper/wrapper.py
suricactus: Probably I got the wrong impression, but I thought you were about to move `job.docker_started_at` back before calling `client.containers.run()`? I guess `job.container_id` should stay where it is.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -121,6 +124,10 @@ def run(self):
                 volumes=volumes,
             )
 
+            if exit_code == SIGKILL_EXIT_CODE:
+                # No further action required, Project and Jobs are deleted
```
suricactus, line 128:
```suggestion
                # No further action required, received by wrapper's autoclean mechanism when the `Project` is deleted
```

docker-app/worker_wrapper/wrapper.py
faebebin: Yes agree. I started with a dictionary: "worker: job.type". Then turned out that filtering only after the key did not work. I.e. `containers.list(filters={label: "worker}). Because I had a dynamic value I could not use the value in filtering. But I could just use {"type": "worker", "project_id": "..."} then filter after "type=worker" . 

BTW I removed the project_id in the recent revision. Should I let it there?

docker-app/worker_wrapper/wrapper.py
```diff
@@ -264,15 +267,30 @@ def _run_docker(
...
 
         try:
             # will throw an `requests.exceptions.ConnectionError`, but the container is still alive
             response = container.wait(timeout=self.container_timeout_secs)
+
+            if response["StatusCode"] == SIGKILL_EXIT_CODE:
+                logger.info(
+                    "Job canceled for deleted Project and Jobs.",
+                )
+                # No further action required, Project and Jobs are deleted
```
suricactus, line 291:
```suggestion
                # No further action required, received by wrapper's autoclean mechanism when the `Project` is deleted
```

docker-app/worker_wrapper/wrapper.py
suricactus:
```suggestion
        # `docker_started_at`/`docker_finished_at` tracks the time spent on docker only
        self.job.docker_started_at = timezone.now()
        self.job.container_id = container.id
        self.job.save(update_fields=["docker_started_at", "container_id"])
```

faebebin: ok, sure

docker-app/worker_wrapper/wrapper.py
suricactus: I think it is nice to have as much meta data as possible, so if you can resurrect `project_id`.

> Because I had a dynamic value I could not use the value in filtering. But I could just use {"type": "worker", "project_id": "..."} then filter after "type=worker" .

According to the docs it should work: https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.list

Diid you try with a more recent `docker-py` version if there is a bugfix?

faebebin: No, willl try

docker-app/worker_wrapper/wrapper.py
```diff
@@ -264,15 +270,31 @@ def _run_docker(
             detach=True,
             mem_limit=config.WORKER_QGIS_MEMORY_LIMIT,
             cpu_shares=config.WORKER_QGIS_CPU_SHARES,
+            labels={
+                "app": "worker",
+                "type": self.job.type,
+                "job_id": str(self.job.id),
+                "project_id": str(self.job.project_id),
```
faebebin, line 277: @suricactus added this on your request. Goood?

suricactus: Lovely!

suricactus (approved):
Nice, thanks!

PR #663, opened 2023-05-18
Some type updates in core app

m-kuhn:


docker-app/qfieldcloud/core/admin.py
```diff
@@ -1079,10 +1079,7 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
 
     @admin.display(description=_("Active users (last billing period)"))
     def active_users(self, instance) -> int:
-        try:
-            return instance.current_subscription_vw.active_users_count
-        except Exception:
-            return None
+        return instance.current_subscription_vw.active_users_count
```
m-kuhn, line 1082: @why-not-try-calmer do you remember why you had the `except Exception` here?
I changed it because `None` doesn't fit the return type `int` which might cause exceptions in other places, so I think in case of troubles it's better to have the root cause exception here (and handle specific exceptions if needed)

why-not-try-calmer: I remember understanding from https://github.com/opengisch/qfieldcloud/pull/591#discussion_r1146320252 that there was a potential exception-raiser upstream (I think it wasn't safe to assume that every organization had a current active subscription). But now looking at the code, `public/docker-app/qfieldcloud/core/models.py:Subscription.get_or_create_current_subscription(self)` suggests that the assumption is safe indeed, so your change seems justified.

m-kuhn: Thanks for the link, that comment suggests it's safe to return `None` which means we should adapt the return type to `Optional` instead.
Should we change the caught exception to `AttributeError` (most of the time I prefer specific exceptions, generic ones often risk masking other issues)? Or is there something else to do (since you are surprised)?

why-not-try-calmer: See commit below for suggestion.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -837,3 +837,8 @@ class Meta:
         on_delete=models.CASCADE,
         related_name="current_subscription_vw",
     )
+
+    @property
+    def current_subscription_vw(self) -> UserAccount | None:
```
m-kuhn, line 842: mypy says: `docker-app/qfieldcloud/subscription/models.py:842: error: Missing return statement  [return]`

docker-app/qfieldcloud/core/admin.py
```diff
@@ -147,7 +147,7 @@ def admin_urlname_by_obj(value, arg):
 )
 
 
-class EmailAddressAdmin(EmailAddressAdmin):
+class ExtendedEmailAddressAdmin(EmailAddressAdmin):
```
suricactus, line 150: A bit picky detail, but we do the other way around so far by renaming the import:

```suggestion
class EmailAddressAdmin(EmailAddressAdminBase):
```

suricactus (approved):
Looks good, thanks!

suricactus (approved):

PR #665, opened 2023-05-22
Check for busy project using workers list

faebebin:
This would add another mechanism to prevent jobs running on parallel on a project as an option if the actual reason just never gets found. (Discovered as sideeffect of running a job cancelling routine). One could also specify which job.type should never run in parallel.

- [ ] tested manually with 3 workers

But would need further testing if seen as an Option ...


duke-nyuki:

docker-app/worker_wrapper/wrapper.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, job_id: str) -> None:
             self.job = self.job_class.objects.select_related().get(id=job_id)
             self.shared_tempdir = Path(tempfile.mkdtemp(dir="/tmp"))
         except Exception as err:
-            feedback: Dict[str, Any] = {}
+            feedback = {}
             (_type, _value, tb) = sys.exc_info()
```
faebebin, line 65: formatter gone wild

docker-app/worker_wrapper/wrapper.py
```diff
@@ -392,9 +415,9 @@ def _prepare_deltas(self, deltas: Iterable[Delta]):
 
         client_pks_map = {}
 
-        for delta_with_modified_pk in local_to_remote_pk_deltas:
-            key = f"{delta_with_modified_pk['client_id']}__{delta_with_modified_pk['content__localPk']}"
-            client_pks_map[key] = delta_with_modified_pk["last_modified_pk"]
+        for delta in local_to_remote_pk_deltas:
+            key = f"{delta['client_id']}__{delta['content__localPk']}"
+            client_pks_map[key] = delta["last_modified_pk"]
```
faebebin, line 421: also only formatter gone wild

suricactus: There was a recent commit by @m-kuhn who changed that. Why the renaming was required?

suricactus: I guess you are branching from outdated master you might get these?

faebebin: yep correct

PR #673, opened 2023-06-01
Compare 'now' on db instead of timezonw.now()

faebebin:


duke-nyuki:
Task linked: [QF-2702 Get V("now") from db in querysets](https://app.clickup.com/t/2192114/QF-2702)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -355,16 +355,17 @@ class Package(models.Model):
 
 
 class SubscriptionQuerySet(models.QuerySet):
+    def is_period_active_condition(self):
+        return Q(active_since__lte=V("now")) & (
+            Q(active_until__isnull=True) | Q(active_until__gte=V("now"))
+        )
+
```
suricactus, line 362:
```suggestion
    @property
    def is_period_active_condition(self):
        return Q(active_since__lte=V("now")) & (
            Q(active_until__isnull=True) | Q(active_until__gte=V("now"))
        )
```

Isn't it easier to use? Or even better, since it is a static content, make it a static property?


```suggestion
    is_period_active_condition = (
        return Q(active_since__lte=V("now")) & (
            Q(active_until__isnull=True) | Q(active_until__gte=V("now"))
        )
    )
```

suricactus (approved):
I do agree making the property shared is a good move. Personally I would prefer to avoid the condition to be in a method, as it is clumsy to use and might make the future use to abuse it.

faebebin:
Yes, agree

PR #674, opened 2023-06-01
QF-2657 Improving tests on CI: parallel build + parallel testing

why-not-try-calmer:
- [x] adapt from https://app.clickup.com/t/2192114/QF-2657
- [x] resolve fun errors in setup
- [ ]  resolve fun errors in tests

`docker compose config`: https://drive.google.com/file/d/1XsoMgWGp3Mpxw6BtEZctuSZN34n0kIog/view?usp=drive_link


.github/workflows/test.yml
```diff
@@ -1,81 +1,293 @@
...
-          python-version: '3.10'
+          python-version: "3.10"
+
       - name: Check code formatting
-        uses: pre-commit/action@v2.0.3
+        uses: pre-commit/action@v3.0.0
+
+  # Build all images for services declaring a build
+  # Push them onto the repository's Github Container Registry
+  build_push__app:
```
m-kuhn, line 33: Can we put that all into a matrix?

```
jobs:
  build:
    runs-on: ubuntu-22.04
    strategy:
      matrix:
        include:
          - image_name: 'app'
            build_context: './docker-app'
            target: 'webserver_test'
          - image_name: ''worker_wrapper
            ...
```

why-not-try-calmer: What steps would be associated with `build`?  Something like:

```
uses: docker/build-push-action@v4
        with:
          context: ./docker-app
          target: webserver_test
          build-args: DEBUG_BUILD=1
          tags: ${{ env.REGISTRY }}/${{ github.repository }}/app:${{ github.head_ref }}.${{ github.sha }}
          cache-from: type=gha
          cache-to: type=gha,mode=max
          push: true
```
? If yes, every iteration has to fetch several values  -- a different numbers of values depending on the image -- so that  a combination of `context`, `build-args` and `tags` are well-defined. Isn't it more complicated than what is done now?

m-kuhn: Fair question, I think we can deal with context and tags, but build-args and target is complicated...

why-not-try-calmer:
@faebebin PR-questing you since you expressed interest in taking a look if you see something fishy -- thanks!

faebebin (commented):
Cannot spot / fix anything on first glance,  neither in the exported docker_config, sry. But seems the first test that triggers a worker and the Job hangs somewhere in [PENDING, QUEUED, STARTED]. 

I guess worker-wrappet never picked up the Job at all? Needs local debugging on e.g. `qfieldcloud.subscription.tests.test_external_db.QfcTestCase.test_is_external_db_supported` ...

PR #675, opened 2023-06-01
QF-2700 Run makemigrations --check right after building

why-not-try-calmer:
What looked like adding a single line of code turned into me arm-wrestling the Django migration writer. 

The issue is basically that the migration writer doesn't know how to serialize bound functions -- functions parameterized with an implicit reference -- and that static methods seem to be considered as bound functions.

I went for the "brutal" way of solving this: i.e. extracting the static method to a top-level function. There are other ways. Please make a commit suggestion if you are not happy with the way I went for.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,6 +54,7 @@ jobs:
 
       - name: Initial manage.py commands
         run: |
+          docker compose run app python manage.py makemigrations --no-input --check
           docker compose run app python manage.py migrate
           docker compose run app python manage.py collectstatic
       - name: Run unit and integration tests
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-2700 Run migrations check in CI](https://app.clickup.com/t/2192114/QF-2700)

m-kuhn:
Thank you, I am not able to comment on the arm-wrestling approach, but I believe the libqfieldsync changes are not needed in here?

why-not-try-calmer:
> Thank you, I am not able to comment on the arm-wrestling approach, but I believe the libqfieldsync changes are not needed in here?

By default my editor completely hide changes to submodules. I had I fixed this but a fresh reset brought the defaults back. Sorry

m-kuhn:
No worries, I think we should change that to a pypi package instead.

suricactus (approved):
LGTM

Interesting that the code had to be refactored like this, thanks for winning the wrestling contest!

I think it opens the discussion to retire the `Geodb` module once again.

PR #676, opened 2023-06-02
QF-2507 Configure the number of file versions per project

why-not-try-calmer:
This goes hand in hand with the Web UI implementation PR'ed in a distinct repository.


docker-app/qfieldcloud/core/models.py
```diff
@@ -1246,6 +1251,20 @@ def save(self, recompute_storage=False, *args, **kwargs):
...
+            self.storage_keep_versions,
+            self.owner.useraccount.current_subscription.plan.storage_keep_versions,
+        )
+
+    @keep_storage.setter
+    def keep_storage(self, value: Union[int, str]):
+        self.storage_keep_versions = min(
+            value if isinstance(value, int) else int(value),
+            self.owner.useraccount.current_subscription.plan.storage_keep_versions,
+        )
```
suricactus, line 1266: Personally I would avoid setters in the models as a pattern. However, bigger concern it introduces a new way of doing things in QFC codebase. Similar logic has been already implemented in the `save` model methods in other models.

Aslo, I guess the value can be string when directly input from the UI. IMO the type conversion should happen in the UI responsible for this, I guess a Django (admin?) view.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1246,6 +1251,20 @@ def save(self, recompute_storage=False, *args, **kwargs):
             self.file_storage_bytes = storage.get_project_file_storage_in_bytes(self.id)
         super().save(*args, **kwargs)
 
+    @property
+    def keep_storage(self) -> int:
```
suricactus, line 1255:
```suggestion
    def storage_versions(self) -> int:
```

I think the keyword here is "versions".

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: It's possible to avoid using setters. Putting everything inside `save()`.

UI elements from which the input is constructed are HTML <input> fields. I hope the Django template engine does everything automatically.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1036,10 +1032,25 @@ class Meta:
...
+        ),
+    )
+
+    use_storage_keep_versions = models.BooleanField(
+        _("Opt-in to project-based max. files versions"),
+        default=False,
+        null=False,
+        blank=False,
+    )
+
```
suricactus, line 1053: Can't we just avoid this field here? If a new project is created, then it automatically get's the owner current subscription plan `storage_keep_versions` value and it is set. Then the user can change it to whatever they want within the allowed values.

suricactus (changes requested):
Can you please:

- [x] add a max value for `storage_keep_versions`, let's say 100.
- [x] add a min value for `storage_keep_versions` to 1.
- [ ] drop `use_storage_keep_versions` - see my comment above.
- [x] merge the two migrations together, each migration makes the CI a few milliseconds slower.

Allowed myself to improve the clarity of the title about the changes according to the existing QFC terminology.

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: How does that work when the user wants to opt-out from this and just use the max. number of versions offered by their plan?

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -406,7 +406,7 @@ def purge_old_file_versions(
 
     # Number of versions to keep is determined by the account type
     # and by whether the user has opted-in to a project-based count
-    if project.use_storage_keep_versions:
+    if project.is_premium:
```
suricactus, line 409: As far as I remember our conversation with Marco and what I have posted on ClickUp, we should have used the plan's `is_premium` flag, not adding a `is_premium` on the project?

suricactus:
Please merge the migrations into one.

docker-app/qfieldcloud/core/utils2/storage.py
why-not-try-calmer: Absolutely, this is a leftover

why-not-try-calmer: Actually looking back at it, the agreement was:

```
project's owner's plan = community ?
- max versions = default 3 = default
else:
- max versions = `not defined`
- default = 10
```

So this is fine?

suricactus (approved):
THank you very much! Ready to merge once tests pass!

PR #680, opened 2023-06-05
Rename error_origin 'container' to 'worker

faebebin:


duke-nyuki:
Task linked: [QF-2659 Investigate deeper the failed jobs](https://app.clickup.com/t/2192114/QF-2659)

suricactus:
What is rationale behind this PR?

While the proposed naming is probably better, patching the JSONs for existing data is not realistic and having the two values just for the sake of clarity will make quering the database historically harder to write and memorize.

faebebin:
Agree. Was only a naming initiative. But agree with your concerns

faebebin:
-

PR #681, opened 2023-06-06
Filter organizations by active_users

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2604 Show a list of the active users in the Organization Admin](https://app.clickup.com/t/2192114/QF-2604)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1038,6 +1038,17 @@ def has_change_permission(self, request, obj):
...
 
+class ActiveUsersFilter(admin.SimpleListFilter):
+    title = _("active_users")
+    parameter_name = "active_users"
+
+    def lookups(self, request: Any, model_admin: Any) -> List[Tuple[Any, str]]:
+        return [("is_active", _("is active"))]
+
+    def queryset(self, request, qs: QuerySet) -> QuerySet:
+        return qs.filter(useraccount__current_subscription_vw__isnull=False)
```
faebebin, line 1049: I am not quite sure about what the ticket wants, but I guess it is more about active members? Or active owners? Maybe a list of active members in the detail view? I added a comment in the clickup task.

Regarding the filter: With this query I guess the naming should be similar to the People filter, either adjust one or the other:
image

faebebin (changes requested):

docker-app/qfieldcloud/core/admin.py
why-not-try-calmer: Thanks for the inputs, after discussing with Ivan the specs have become much clearer. I'll ping you when it's ready.

suricactus: Sorry for the vague task definition, my fault!

why-not-try-calmer:
I do no know how to pass data from the `organization_projects_active_users` custom field to the `OrganizationFormActiveUsersWidget` widget, and I don't know how to acquire the information.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1037,7 +1038,110 @@ def has_change_permission(self, request, obj):
...
+                    }
+
+                if scheduled["type"] == "job":
+                    users_scheduled[creator_id]["deltas_count"] += 1
+
+                elif scheduled["type"] == "delta":
+                    users_scheduled[creator_id]["jobs_count"] += 1
+
+            list_active_users = self.fields["list_active_users"]
+            list_active_users.initial = users_scheduled
```
why-not-try-calmer, line 1139: still no cigar

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1037,7 +1038,71 @@ def has_change_permission(self, request, obj):
...
+        return [
+            ("only_with_active_users", _("Yes")),
+            ("only_without_active_users", _("No")),
+        ]
+
+    def queryset(self, request: HttpRequest, qs: QuerySet) -> QuerySet:
+        if self.value() == "only_active_users":
+            return qs.filter(
+                useraccount__current_subscription_vw__isnull=False
+            ).order_by("+useraccount__current_subscription_vw")
```
faebebin, line 1055: does that work

why-not-try-calmer: Removed the "+". Good catch.

faebebin: Do we have any admin tests yet? can you try to somehow have this and other new admin features test covered? maybe we can just somewhere load admin pages list and navigate to one detail view ...

why-not-try-calmer: Hum, this PR is already 200 lines long and the business logic of the new feature (= counting jobs per organization and users) is already tested in a test that is part of this PR. There are two other changes which are basically pure Django-wiring:
- using the modelForm field as a source of data for the widget
- adding a new filter 

Wouldn't adding 2 extra tests amount to testing Django?

why-not-try-calmer: @faebebin is that OK or not yet? If not yet, maybe we can talk about this tomorrow, because I'd be interested if you have more comments / suggestions on other parts of this PR.

faebebin: will get back to it in afternoon

suricactus:
@why-not-try-calmer the django admin looks nice! I would only ask if in the list view we just should the number of active users. I guess not filter is needed for now, if sorting works should be enough.

docker-app/qfieldcloud/core/admin.py
suricactus: > Do we have any admin tests yet? can you try to somehow have this and other new admin features test covered? maybe we can just somewhere load admin pages list and navigate to one detail view ...

There is a test checking if admin pages are loading. See here https://github.com/opengisch/qfieldcloud/blob/master/docker-app/qfieldcloud/core/tests/test_admin.py#L62

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1072,18 +1137,23 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
...
     @admin.display(description=_("Active users (last billing period)"))
-    def active_users(self, instance) -> int | None:
-        # The relation 'current_subscription_vw' is not instantiated unless the organization
-        # does have a current subscription
-        if hasattr(instance, "current_subscription_vw"):
-            return instance.current_subscription_vw.active_users_count
-        else:
-            return None
+    def active_users(self, instance) -> int:
+        return instance.active_users_count
```
why-not-try-calmer, line 1156: Ensuring sorting on active users by default.

image

docker-app/qfieldcloud/core/models.py
```diff
@@ -691,6 +692,64 @@ def active_users(self, period_since: datetime, period_until: datetime):
...
         ).filter(Q(id__in=users_with_delta) | Q(id__in=users_with_jobs))
 
+    def list_active_users_jobs_deltas_count(self) -> List[Dict[str, Union[str, int]]]:
+        """Return a List of Dictionaries mapping user ids into the following values:
+        - user_id
+        - username
+        - number of jobs scheduled by the user
+        - number of deltas scheduled by the user
+        """
+        delta = timedelta(weeks=4)
```
faebebin, line 702: Is this referring to 'current billing period'. Are you sure this is 4 weeks and not time since last invoice?

why-not-try-calmer: The success of test added to this PR makes me want to say, 'Yes'.

docker-app/qfieldcloud/core/models.py
```diff
@@ -691,6 +692,64 @@ def active_users(self, period_since: datetime, period_until: datetime):
...
+        now = datetime.now()
+        four_weeks_ago = now - delta
+
+        # Get the organization's projects & active users
+        organization_projects_pk: Set[str] = set(
+            Project.objects.filter(owner=self.pk).values_list("pk", flat=True)
+        )
+        organization_active_users_dict = dict(
+            self.active_users(four_weeks_ago, now).values_list("pk", "username")
+        )
```
faebebin, line 712: It seems super indirect to me to first use the acitve_users method in a first place here. To get the active users we already query all jobs and deltas by `created_at=in timerange` .

I would refactor this somehow to:

```
            Delta.objects.filter(
                project__in=self.projects.all(),
                created_at__gte=period_since,
                created_at__lte=period_until,
            )
       ...
        )
        users_with_jobs = (
            Job.objects.filter(
                project__in=self.projects.all(),
                created_at__gte=period_since,
                created_at__lte=period_until,
            )
```
A method / qs or maybe even custom manager, that returns all the jobs and deltas created in the requested period. Then use this qs in `def active_users` and add a distinct user query to it. And this qs in `def list_active_users_jobs_deltas_count` and add some count to it?

docker-app/qfieldcloud/core/tests/test_organization.py
```diff
@@ -263,3 +263,48 @@ def _active_users_count(base_date=None):
...
+    def test_active_users_jobs_deltas_count(self):
+
+        # Let user2 be a member of organization 1
+        OrganizationMember.objects.create(
+            organization=self.organization1,
+            member=self.user2,
+            role=OrganizationMember.Roles.MEMBER,
+        )
+
+        # Idem for user3
```
faebebin, line 276: I think I know what `Idem` means, but IMO please use easy language ;)

docker-app/qfieldcloud/core/models.py
faebebin: I do not get what you mean with the `success of test added to this PR?` . The test does not assert anything that relates to a timerange or 'last billing period'?

Further, are you sure it is the 'last billing period' and not the current / recent one?

image

faebebin (changes requested):

docker-app/qfieldcloud/core/admin.py
faebebin: I would suggest focus on the core of the task first. You can branch off to continue later on filter or sorting and maybe even add a follwoup ticket? And then let filter / sort aside for now ?

docker-app/qfieldcloud/core/tests/test_organization.py
```diff
@@ -263,3 +263,48 @@ def _active_users_count(base_date=None):
...
+
+        results = self.organization1.list_active_users_jobs_deltas_count()
+        expected = [
+            {
+                "user_id": self.user2.pk,
+                "username": self.user2.username,
+                "jobs_count": 1,
+                "deltas_count": 1,
+            }
+        ]
```
faebebin, line 309: In the test also assert the time range that refers to billing period.

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: > Maybe we ask Ivan here how to best get the current / last billing period? But maybe look for some existing utility first.
> 
> I will continue in a bit ...
> 
> Looks good sofar :)

If you look at https://github.com/opengisch/qfieldcloud/blob/62c8419ab5359c3b8a7d2eb88a84d881cc8f2961/docker-app/qfieldcloud/core/models.py#L660-L661, does it answer your question?

faebebin: Don't think so. It does not mention billing period anywhere ... ?

docker-app/qfieldcloud/core/models.py
```diff
@@ -691,6 +692,64 @@ def active_users(self, period_since: datetime, period_until: datetime):
...
+        - username
+        - number of jobs scheduled by the user
+        - number of deltas scheduled by the user
+        """
+        delta = timedelta(weeks=4)
+        now = datetime.now()
+        four_weeks_ago = now - delta
+
+        # Get the organization's projects & active users
+        organization_projects_pk: Set[str] = set(
```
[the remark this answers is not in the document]
why-not-try-calmer: Yes indeed, self.projects.all() is equivalent. I have no preference.

I don't feel comfortable mixing refactoring other people's code with adding a feature, and I need the newly defined function to add the feature in a way that is easy to test.

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: I'd put it this way: the notion of active users used in the ClickUp task == the notion of active users used in OrganizationAdmin (already defined and used in master) == the notion of active users used in this PR. The only difference is that we need a specific time window. I suggest "the last 4 weeks ", but I have no specific criterion in mind. Is there something you disagree with? Or is there a specific criterion to determine the time window that I am unaware of?

docker-app/qfieldcloud/core/models.py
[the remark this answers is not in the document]
faebebin: You are very welcome to refactor when makes sense, IMO. Is important to stay DRY and readable.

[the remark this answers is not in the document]
faebebin: What you mean with you have no preference: You prefer a 3 liner to a readyble 1 liner?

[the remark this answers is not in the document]
faebebin: Just implement it please.

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: @suricactus Perhaps I am missing something about the requirements here. What notion of `active_users` should this PR be using?

docker-app/qfieldcloud/core/models.py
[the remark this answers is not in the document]
why-not-try-calmer: So the suggestion is that I merge this and the method above and rewrite all the tests that depend on the above method?

docker-app/qfieldcloud/core/models.py
faebebin: @why-not-try-calmer I asked in the support channel what they prefer. IMO you can also change the description to `1 month from now` or so. Just needs to be correct. Else it would cause more trouble then help...

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: You are onto something, yes, I can do without the `organization_active_users_dict`.

docker-app/qfieldcloud/core/models.py
faebebin: The added specifications are in the clickup task comment.

docker-app/qfieldcloud/core/models.py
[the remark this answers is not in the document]
faebebin: I suggested a little style improvement that also adds a common ground for the two 'active users' methods ... As long as the tests succeed (of the previous method) it should be fine. Maybe need to fix formatting though ..

docker-app/qfieldcloud/core/models.py
```diff
@@ -691,6 +692,61 @@ def active_users(self, period_since: datetime, period_until: datetime):
...
+                created_at__lte=now,
+            )
+            .values("created_by", "created_by__username")
+            .annotate(deltas_count=Count("created_by"), type=V("deltas"))
+            .order_by("-deltas_count")
+        )
+
+        # Join querysets' values
+        payload: Dict[str, Union[str, int]] = {}
+        for sched in chain(
```
faebebin, line 733: I would prefer `obj` . Eventhough it

I guess usually elsewhere it is the var name used:

```
for obj in some_queryset:
...
```

faebebin (changes requested):
I would prefer to always use naming style found elsewhere in the code, whether it is variable names or comments.

Only introduce new terms if there really is a semantic need for it (Dont't know whether this is correct senctence though ;) .

docker-app/qfieldcloud/core/tests/test_organization.py
why-not-try-calmer: I am not getting it. I usually use assertions to verify that a certain relation holds between inputs and outputs. What is the input-output relation that you would like me to make an assertion about?

docker-app/qfieldcloud/core/models.py
why-not-try-calmer: Okay, so all is fine or do you see something wrong in the current implementation?

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1072,18 +1159,26 @@ class OrganizationAdmin(QFieldCloudModelAdmin):
...
     autocomplete_fields = ("organization_owner",)
 
-    @admin.display(description=_("Active users (last billing period)"))
-    def active_users(self, instance) -> int | None:
-        # The relation 'current_subscription_vw' is not instantiated unless the organization
-        # does have a current subscription
-        if hasattr(instance, "current_subscription_vw"):
-            return instance.current_subscription_vw.active_users_count
-        else:
-            return None
```
faebebin, line 1086: That one seemed correct. Although I would use (not tested):

```
instance.useraccount.current_subscription().active_users_count()
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -691,6 +701,51 @@ def active_users(self, period_since: datetime, period_until: datetime):
...
+        recent_deltas: QuerySet = (
+            self.deltas(period_since, period_until)
+            .values("created_by", "created_by__username")
+            .annotate(deltas_count=Count("created_by"), type=V("deltas"))
+            .order_by("-deltas_count")
+        )
+
+        # Join querysets' values
+        builder: dict[str, str | int] = {}
+        for job_or_delta in chain(
```
faebebin, line 728: I would have just called it `obj`

docker-app/qfieldcloud/core/admin.py
why-not-try-calmer: Are you commenting code of this PR or of master? Is there an issue with the code of this PR?

faebebin: ... And then add a similar method in AbstractSubscription for the `list_active_users_jobs_deltas_count` , where  you can again maybe create some common base method (not sure though) to query the requested time range.

(Just in the end those two methods should be traceable  in the code as being closely related IMO.)

faebebin: Ah sorry for confusion. No this one above (master) seems right.

why-not-try-calmer: Which problem have you run into with the one offered by the PR?

faebebin: No, it is good that part. Sorry. But we should implement the `list_active_users_jobs_deltas_count` just the same way.

why-not-try-calmer: I think it would make it _much_ easier for me if you layered out your review into something like:
1) is the code correct (nothing broken?)
2) is the code complete (are all the target features there?)
3) is the code efficient and maintainable (DRY, SOLID, etc.)
4) does the code meet all the other conventions agreed upon by the maintainters of the codebase?
It's a bit difficult for me to go back and forth between all these aspects, and to keep asking questions about which aspect this or that comment is about, because in many cases it's just not clear. 

It should be possible to proceed with "rounds" (all comments about one layer before moving on to the next layer) or, if not appropriate, to just clarifiy what aspect a comment is about.

faebebin:
Split up into maintainable PRs

docker-app/qfieldcloud/core/admin.py
faebebin: @why-not-try-calmer yes agree with your suggestions - like this structure. Was a bit an unlucky history in this specific PR because I started with broken local stack but still found first corrections and wanted to post them asap. Then I came up with many more and you were already processing the first one. 

Further I was not very much into this part of codebase neither and found myself misunderstanding a couple of times. It is quite a complex task ...

I appreciate your input and will try to stay structured the next time.

PR #682, opened 2023-06-06
QF-2711 Adding a few failsafes around ListFilesView.get()

why-not-try-calmer:
_Source_:
https://app.clickup.com/t/2192114/QF-2711.

_Problem_:
The `download_project` step used in the `process_projectfile` workflow makes a call to `qfieldcloud_sdk` to the effect that the sdk will try and GET project files under a given project id. However, the attempt sometimes fails on "500: Internal Error", with the caller being offered little to no information about the error. 

_Proposal_:
This PR proposes to offer callers more information about such failures. It delineates all the I/O actions that might raise, collect the possible exceptions, and returns the first one raised as part of the response.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_qgis_file.py
+++ b/docker-app/qfieldcloud/core/tests/test_qgis_file.py
@@ -56,6 +56,16 @@ class QfcTestCase(APITransactionTestCase):
         else:
             return response.content
 
+    def test_try_to_get_nonexistent_project(self):
+        empty_string = ""
+        nonexistent_id = "007"
+        with self.subTest(
+            "Ensure that '/api/v1/files' handles missing resources correctly."
+        ):
+            one = self.client.get(f"/api/v1/files/{empty_string}")
+            two = self.client.get(f"/api/v1/files/{nonexistent_id}")
+            assert all(r.status_code == 404 for r in (one, two))
+
     def test_push_file(self):
         self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
 
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -14,7 +14,7 @@ from qfieldcloud.core.utils2.storage import (
     purge_old_file_versions,
 )
 from rest_framework import permissions, status, views
-from rest_framework.exceptions import bad_request, server_error
+from rest_framework.exceptions import NotFound, server_error
 from rest_framework.parsers import MultiPartParser
 from rest_framework.response import Response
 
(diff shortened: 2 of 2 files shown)


duke-nyuki:
Task linked: [QF-2711 Investigate and specify API_INTERNAL_SERVER_ERROR](https://app.clickup.com/t/2192114/QF-2711)

suricactus:
I think in this case we can already use `ProjectMixin`, which handles the situation with wrong project id?

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -38,9 +40,17 @@ class ListFilesView(views.APIView):
...
 
-        project = Project.objects.get(id=projectid)
-
-        bucket = utils.get_s3_bucket()
+        try:
+            project = Project.objects.get(id=projectid)
+            bucket = utils.get_s3_bucket()
+            assert hasattr(bucket, "object_versions")
+        except ObjectDoesNotExist:
+            return bad_request(
```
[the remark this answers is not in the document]
faebebin: 404 sounds reaonable, just an inexistent resource?

faebebin (changes requested):

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -37,10 +39,21 @@ class ListFilesView(views.APIView):
...
+            if not bucket.creation_date:
+                # Let DRF return 500 when bucket does not exist, since it needed but
+                # not what the client tried to get
+                return server_error(
+                    request=request,
+                    reason=f"Unable to fetch needed resource for {projectid}",
+                )
+        except ObjectDoesNotExist:
+            # map failure to get from db into failure to GET from API
+            raise NotFound(detail=projectid)
```
faebebin, line 54: Does this return a 404?

faebebin (changes requested):
Think looks very promising?

Maybe could you describe the error handling behaviour in a small test case in public/docker-app/qfieldcloud/core/tests/test_qgis_file.py ? 

e.g `returns 404 if project with requested id does not exist`

docker-app/qfieldcloud/core/views/files_views.py
why-not-try-calmer: yes

docker-app/qfieldcloud/core/tests/test_qgis_file.py
```diff
@@ -56,6 +56,16 @@ def get_file_contents(self, project, filename, version=None):
...
 
+    def test_try_to_get_nonexistent_project(self):
+        empty_string = ""
+        nonexistent_id = "007"
+        with self.subTest(
+            "Ensure that '/api/v1/files' handles missing resources correctly."
+        ):
+            one = self.client.get(f"/api/v1/files/{empty_string}")
+            two = self.client.get(f"/api/v1/files/{nonexistent_id}")
+            assert all(r.status_code == 404 for r in (one, two))
```
faebebin, line 67: Just subjective little taste comment from myside: I would keep it simple such a test and rather just write `assertEqual(one.status_code, 404) etc` but is ok

faebebin (approved):
LGTM, thanks :)

docker-app/qfieldcloud/core/tests/test_qgis_file.py
suricactus: Forget about it, I will handle it.

faebebin: Thanks for the correction

PR #684, opened 2023-06-08
QF-2711 Fine tune error handling in the `files` API endpoint

why-not-try-calmer:
Addressing #682's comments.

### DIFF
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -111,8 +111,6 @@ def get_s3_session() -> boto3.Session:
 def get_s3_bucket() -> mypy_boto3_s3.service_resource.Bucket:
     """
     Get a new S3 Bucket instance using Django settings.
-    If `bucket_name` is not provided it will be used from `settings.STORAGE_BUCKET_NAME`.
-    Assign 'allow_empty' False to raise when getting an empty s3 Bucket.
     """
 
     if not bucket_name:
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -16,6 +16,8 @@ from django.conf import settings
 from django.core.files.uploadedfile import InMemoryUploadedFile, TemporaryUploadedFile
 from redis import Redis, exceptions
 
+from .exceptions import EmptyS3Bucket
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


suricactus (approved):

suricactus:
I think the current state is ok-ish, still not 100% convinced we even need this check, but now it is mergeable. I only need:

- [ ] an answer what are the precise situations when the `.creation_date` is false-ish
- [ ] confirmation you have tested the exception handler is returning a proper error to the client

why-not-try-calmer:
> I think the current state is ok-ish, still not 100% convinced we even need this check, but now it is mergeable. I only need:
> 
> * [ ]  an answer what are the precise situations when the `.creation_date` is false-ish
> * [ ]  confirmation you have tested the exception handler is returning a proper error to the client

Thanks very much, will follow up with the unit test and the appropriate docstring

why-not-try-calmer:
@suricactus Used a different approach to check empty buckets; also added a test.

duke-nyuki:
Task linked: [QF-2711 Investigate and specify API_INTERNAL_SERVER_ERROR](https://app.clickup.com/t/2192114/QF-2711)

why-not-try-calmer:
@suricactus Can we merge?

docker-app/qfieldcloud/core/utils.py
```diff
@@ -109,13 +109,22 @@ def get_s3_session() -> boto3.Session:
...
-    # Get the bucket objects
+    session = get_s3_session()
     s3 = session.resource("s3", endpoint_url=settings.STORAGE_ENDPOINT_URL)
-    return s3.Bucket(settings.STORAGE_BUCKET_NAME)
+
+    # Ensure the bucket exists
+    s3.meta.client.head_bucket(Bucket=bucket_name)
+
+    # Get the bucket resource
+    return s3.Bucket(bucket_name)
```
suricactus, line 127: What I am still wondering is what will happen if we skip the line with `head_bucket` and the bucket does not exist on `s3.Bucket`. Aren't we going to get the same error?

why-not-try-calmer: What do you mean by "skip the line"?

suricactus: If we don't have it at all, sorry.

I think the error would be the same with or without that line (124) in case the bucket does not exist?

why-not-try-calmer: On an nonexistent Bucket, line 124 will raise `NoSuchBucket`. I cannot guarantee the same exception would be raised on line 127 if the Bucket doesn't exist.

suricactus: Not sure if you "cannot guarantee" because you observed different behaviour when you tested or because you didn't test?

why-not-try-calmer: In my tests, if you remove the line 124 and passes a non-empty string that references no bucket to `s3.Bucket(...)`, you get no error from `get_s3_bucket()`. Line 124 is there to prevent that from happening.

suricactus: Ok, let's merge it, but we need to monitor the average response time if it increases due to the extra HEAD call to the Object storage.

PR #686, opened 2023-06-09
Enable Groups model and groups field in People view in Django admin

suricactus:
This way we can now configure certain staff users to have access to  certain Django admin models only. What is more, we can make this access read-only. I will do this manually in the admin after releasing this  branch.

@m-kuhn @faebebin ideally I would like to have this in the upcoming release.


duke-nyuki:
Task linked: [QF-2728 Add staff permission in Django admin so users can read-only certain views](https://app.clickup.com/t/2192114/QF-2728)

faebebin (approved):
LGTM 

How about adding a group filter?

suricactus:
Group filter for People admin? Yeah, why not! But I think it is out of scope for this PR.

faebebin:
Yes agree. LGTM

PR #690, opened 2023-06-16
pip install libqfieldsync

m-kuhn:
By default installs libqfieldsync via pip
For local development, allows mounting a volume into the qgis container

Depends on https://github.com/opengisch/libqfieldsync/pull/49

### DIFF
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -128,12 +128,12 @@ services:
   qgis:
     build:
       target: dev
-    # Uncomment to use a local clone of the libqfieldsync repository
-    # that will be bind-mounted into this container for an editable installation
-      # environment:
-      #   - HOST_PATH_LIBQFIELDSYNC
-      # volumes:
-      #   - ${HOST_PATH_LIBQFIELDSYNC}:/libqfieldsync
+    environment:
+      - HOST_PATH_LIBQFIELDSYNC
+  # Uncomment to use a local clone of the libqfieldsync repository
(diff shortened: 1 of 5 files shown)


docker-compose.override.local.yml
```diff
@@ -125,6 +125,13 @@ services:
       exit 0;
       "
 
+  qgis:
+    build:
+      target: dev
+    volumes:
+# Uncomment to use a local libqfieldsync dev environment
+#      - /home/user/dev/libqfieldsync:/libqfieldsync
+
```
suricactus, line 134: Can we use a envvar for this? I would prefer we never keep local modifications in the `docker-compose` files. So far all of these situations are handled with an envvar.

.gitignore
```diff
@@ -1,14 +0,0 @@
-__pycache__/
```
suricactus, line 1: I guess it was a mistake that the global `.gitignore` was deleted. Can you please revert.

docker-compose.override.local.yml
m-kuhn: The first part before the `:` needs to point to a valid directory. We could by default mount some random directory from the source tree. Or a dummy volume. Both approaches should work but seem clumsy, that's why I didn't implement one of them in the first place. Do you see an option I forgot?

why-not-try-calmer: added

why-not-try-calmer:
I had some doubts about `pip install -e` but I must this is a good use-case!

suricactus:
Closing, #851 fixed this.

PR #692, opened 2023-06-17
New line handling

grzegorzb1990:
When the repository is cloned on Windows, all newline characters are converted to CLRF format. This causes a problem when running images that contain bash scripts. Linux scripts are not compatible with this newline character format and cause an error:
`standard_init_linux.go:228: exec user process caused: no such file or directory`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+# Handle line endings automatically for files detected as text
+# and leave all files detected as binary untouched.
+*           text=auto
+
+#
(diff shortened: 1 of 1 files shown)


m-kuhn:
Thank you for opening a pull request and welcome!

Most developers use linux, so the most important part for me is that this doesn't break there. And I don't see how it could break things there, so if this is a QOL improvement for Windows users, I'm happy to get this merged.

m-kuhn (approved):

PR #695, opened 2023-06-21
QF-2657 Improving tests on CI: parallel build/testing

why-not-try-calmer:
__TL;DR__
I am not able to make #674 work. I submit this as an ersatz. This PR borrows from design of the PR it supersedes, having each job in the matrix handle a separate batch of tests, grouped by Django app.

As for #674 it shaves off some precious minutes, and even more so if we are able to fix flakiness once and for all, because then we can run the tests on either `pull_request` or `push`.

__Comparison with #674__ 
- it runs to completion
- it is less efficient: while in #674 the Docker images where first built in parallel _once_ and then downloaded and used in each job in the matrix, now the images need to be built _for each job in the matrix_.
- it means much less maintainance
- it provides much faster feedback for PRs authors, especially for PRs not touching `core`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,15 @@ jobs:
   test:
     name: Run tests
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        django_apps:
+          - authentication
+          - notifs
+          - subscription
+          - core
+    continue-on-error: true
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-2657 Optimize testing locally and on CI](https://app.clickup.com/t/2192114/QF-2657)

suricactus:
I like this approach much more!

suricactus (approved):
IMO it is merge ready. Thanks a lot!

why-not-try-calmer:
The next low-hanging fruit for improving test running time is probably the (in)famous `--parallel` flag. I'll mention it on the ClickUp Task.

PR #697, opened 2023-06-22
Set dequeue transaction level to repeatable read so we have stable data

suricactus:


duke-nyuki:

m-kuhn:
I believe we should also add an assertion if `pg_is_in_recovery()` returns false

suricactus:
@m-kuhn I was just opening a separate PR https://github.com/opengisch/qfieldcloud/pull/698

m-kuhn (approved):

PR #699, opened 2023-06-22
QF-2704 Making sure strings are encoded when sending to Sentry

why-not-try-calmer:
In response to:

image

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_sentry.py
+++ b/docker-app/qfieldcloud/core/tests/test_sentry.py
@@ -38,5 +38,5 @@ class QfcTestCase(APITestCase):
             "buffer": StringIO("The traceback of the exception to raise"),
             "capture_message": True,  # so that sentry receives attachments even when there's no exception/event
         }
-        is_sent = report_serialization_diff_to_sentry(**mock_payload)
-        self.assertTrue(is_sent)
+        will_be_sent = report_serialization_diff_to_sentry(**mock_payload)
+        self.assertTrue(will_be_sent)
--- a/docker-app/qfieldcloud/core/tests/test_sentry.py
+++ b/docker-app/qfieldcloud/core/tests/test_sentry.py
@@ -1,11 +1,15 @@
 from io import StringIO
 
 from rest_framework.test import APITestCase
+from sentry_sdk import capture_message
 
 from ..utils2.sentry import report_serialization_diff_to_sentry
 
 
 class QfcTestCase(APITestCase):
+    def test_sending_message_to_sentry(self):
+        capture_message("Hello Sentry from test_sentry!")
+
     def test_logging_with_sentry(self):
         mock_payload = {
             "name": "request_id_file_name",
@@ -36,6 +40,7 @@ class QfcTestCase(APITestCase):
                 }
             ),
             "buffer": StringIO("The traceback of the exception to raise"),
+            "capture_message": True,
         }
-        result = report_serialization_diff_to_sentry(**mock_payload)
-        self.assertTrue(result)
+        is_sent = report_serialization_diff_to_sentry(**mock_payload)
+        self.assertTrue(is_sent)
--- a/docker-app/qfieldcloud/core/tests/test_sentry.py
+++ b/docker-app/qfieldcloud/core/tests/test_sentry.py
@@ -9,16 +9,16 @@ class QfcTestCase(APITestCase):
     def test_logging_with_sentry(self):
         mock_payload = {
             "name": "request_id_file_name",
-            "pre_serialization": {
-                "data": str(["some", "pre-serialization", "data", "keys"]),
-                "files": str(["some", "files", "to be", "listed"]),
-                "meta": str({"metadata": "of the request pre-serialization"}),
-            },
-            "post_serialization": {
-                "data": str(["some", "post-serialization", "data", "keys"]),
-                "files": str(["some", "files", "to be", "listed"]),
-                "meta": str({"metadata": "of the request post-serialization"}),
-            },
+            "pre_serialization": str({
+                "data": ["some", "pre-serialization", "data", "keys"],
+                "files": ["some", "files", "to be", "listed"],
+                "meta": ["metadata": "of the request pre-serialization"],
+            }),
+            "post_serialization": str({
+                "data": ["some", "post-serialization", "data", "keys"],
+                "files": ["some", "files", "to be", "listed"],
+                "meta": {"metadata": "of the request post-serialization"},
+            }),
             "buffer": StringIO("The traceback of the exception to raise"),
         }
         result = report_serialization_diff_to_sentry(**mock_payload)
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-2704 Sentry Analysis: Failed to upload photos. Debug `EmptyContentError: Empty content` when uploading files](https://app.clickup.com/t/2192114/QF-2704)

why-not-try-calmer:
This seems to work, according to the test this PR adds and the result from the previous workflow run: https://github.com/opengisch/qfieldcloud/actions/runs/5375876854/jobs/9752262489?pr=699#step:10:162

There's a way to properly unittest code interacting with Sentry but [it's only for Javascript](https://zivl.github.io/sentry-testkit/docs/api) as far as I know.

m-kuhn:
If it works on local tests, that's ok.
Unit testing external infrastructure is always risky and especially for this where it's a temporary measure it would be over-engineered.

m-kuhn:
Can you make sure you test with some non-basic ASCII content?
Do we have anywhere binary data involved (after all it's about images somewhere)?

why-not-try-calmer:
Sorry @suricactus:
- made sure SENTRY_DSN was present on `.env`
- run new test locally:
 
Capture d’écran du 2023-06-26 11-42-38

- not seeing anything on sentry:

Capture d’écran du 2023-06-26 11-41-37

why-not-try-calmer:
> Can you make sure you test with some non-basic ASCII content? Do we have anywhere binary data involved (after all it's about images somewhere)?

The approach used here doesn't use the contents directly but only key names and request meta data. Everything should be encodable to UTF8.

m-kuhn:
> The approach used here doesn't use the (binary) contents directly but only key names and request meta data (the idea is: if there is a difference in these items before and after serialization, then the serialization is the culprit).

:+1: 

> All the values used should be encodable to UTF8.

If that's the case, adding some non-basic ASCII chars to the tests shouldn't cause any hickups, correct?

why-not-try-calmer:
> > The approach used here doesn't use the (binary) contents directly but only key names and request meta data (the idea is: if there is a difference in these items before and after serialization, then the serialization is the culprit).
> 
> +1
> 
> > All the values used should be encodable to UTF8.
> 
> If that's the case, adding some non-basic ASCII chars to the tests shouldn't cause any hickups, correct?

done

m-kuhn:
> Sorry @suricactus:
> 
> * made sure SENTRY_DSN was present on `.env`
> * run new test locally.
> 
> * not seeing anything on sentry.
> 

@why-not-try-calmer that means tests are not yet successful? Can we help you somehow?

why-not-try-calmer:
> > Sorry @suricactus:
> > 
> > * made sure SENTRY_DSN was present on `.env`
> > * run new test locally.
> > 
> > * not seeing anything on sentry.
> > 
> 
> @why-not-try-calmer that means tests are not yet successful? Can we help you somehow?

The test is successful. But I am not seeing the attached files on Sentry, whether I run the test locally or in CI. There's no easy way to test that. Ivan suggested a way to do that using my private Sentry SSN but still nothing on Sentry.

m-kuhn:
That's what I was meaning with "testing external infrastructure is hard" -- can you force a local failure via some hacks and perform a manual test to see if it turns up?

why-not-try-calmer:
> That's what I was meaning with "testing external infrastructure is tricky" -- can you force a local failure via some hacks and perform a manual test to see if it turns up?

this guy [here](https://github.com/opengisch/qfieldcloud/pull/699/files#diff-b39d3e77223a080339a246fc1e89f4d890e9dcd250b773fe2ebe40e847243759) either reports to sentry or fails. It's not failing. So it's reporting to sentry. The only remaining issue, I understand the situation correctly, is just to configure the way it reports to sentry, so that we can see it on Sentry. I am waiting on @suricactus for going beyond what he and I discussed this morning, which prompted this new test. Do you suggest we proceed otherwise, @m-kuhn ?

m-kuhn:
> > That's what I was meaning with "testing external infrastructure is tricky" -- can you force a local failure via some hacks and perform a manual test to see if it turns up?
> 
> this guy [here](https://github.com/opengisch/qfieldcloud/pull/699/files#diff-b39d3e77223a080339a246fc1e89f4d890e9dcd250b773fe2ebe40e847243759) either reports to sentry or fails. It's not failing. So it's reporting to sentry. The only remaining issue, if I understand the situation correctly, is just to configure the way it reports to sentry, so that we can see it on Sentry. I am waiting on @suricactus for going beyond what he and I discussed this morning, which prompted this new test. Do you suggest we proceed otherwise, @m-kuhn ?

I'll believe that it reports to sentry when I see it there. You can set the `SENTRY_DEBUG=1` to get some insights into the communication channels to sentry.

why-not-try-calmer:
> > > That's what I was meaning with "testing external infrastructure is tricky" -- can you force a local failure via some hacks and perform a manual test to see if it turns up?
> > 
> > 
> > this guy [here](https://github.com/opengisch/qfieldcloud/pull/699/files#diff-b39d3e77223a080339a246fc1e89f4d890e9dcd250b773fe2ebe40e847243759) either reports to sentry or fails. It's not failing. So it's reporting to sentry. The only remaining issue, if I understand the situation correctly, is just to configure the way it reports to sentry, so that we can see it on Sentry. I am waiting on @suricactus for going beyond what he and I discussed this morning, which prompted this new test. Do you suggest we proceed otherwise, @m-kuhn ?
> 
> I'll believe that it reports to sentry when I see it there. You can set the `SENTRY_DEBUG=1` to get some insights into the communication channels to sentry.

https://opengisch.sentry.io/issues/4277631684/attachments/?project=5438993&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

Sentry needs events to attach files to. And there was no event.

docker-app/qfieldcloud/core/tests/test_sentry.py
```diff
@@ -0,0 +1,42 @@
+from io import StringIO
+
+from rest_framework.test import APITestCase
+
+from ..utils2.sentry import report_serialization_diff_to_sentry
+
+
+class QfcTestCase(APITestCase):
+    def test_logging_with_sentry(self):
```
m-kuhn, line 9: Decorate with skip test if SENTRY_DSN is not set

docker-app/qfieldcloud/core/utils2/sentry.py
```diff
@@ -7,8 +7,12 @@
 
 
 def report_serialization_diff_to_sentry(
-    name: str, pre_serialization: str, post_serialization: str, buffer: StringIO
-):
+    name: str,
+    pre_serialization: str,
+    post_serialization: str,
+    buffer: StringIO,
+    capture_message=False,
```
m-kuhn, line 14: Document with docstring

m-kuhn (approved):

m-kuhn:
Thanks

PR #705, opened 2023-06-27
Fix wrong test in test_organization

faebebin:
I am pretty sure this is meant?

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_organization.py
+++ b/docker-app/qfieldcloud/core/tests/test_organization.py
@@ -259,7 +259,7 @@ class QfcTestCase(APITestCase):
         # User 3 creates a job
         Job.objects.create(
             project=project1,
-            created_by=self.user3,
+            created_by=self.user4,
         )
-        # There are still 2 billable users, because self.user3 is staff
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/tests/test_organization.py
```diff
@@ -259,7 +259,7 @@ def _active_users_count(base_date=None):
         # User 3 creates a job
         Job.objects.create(
             project=project1,
-            created_by=self.user3,
+            created_by=self.user4,
         )
```
faebebin, line 263: Anyway I would suggest renaming all these tests so they are easier to read. The comments mention `User 1` and the refer to `self.user2` and `User 2` referring to `self.user3` ... . This is super hard to read and probably led to this bug?

I would suggest using `user2` directly in the comment or writing:
```
# 1. user2 creates ...
# 2. user3 does x
...
```
what you think?

suricactus: Absolutely in favor of everything that easily adds a readability improvement.

suricactus (approved):

PR #704, opened 2023-06-27
Show active members in the Organization Admin

faebebin:
Shows the active members, that have created a Job or a Delta for the Organization during the running stripe billing period in the admin.

The list of users in the detail view
image
(if no active members)
image

Dropped the active members count in the list_view - it would slow down too much:
It adds 4 Queries per organization. Without it is 6 Queries in total.

now:
image

before:
image

PR #706, opened 2023-06-28
[WIP] Qf 2604 org active users listview admin

faebebin:
A column with active users count 
based on https://github.com/opengisch/qfieldcloud/pull/704

Sorting or Filtering is out of reach because would need to `CurrentSubscription.objects.annotate(... define timerange ...)`.


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -602,6 +603,7 @@ def active_users(self):
 
     @property
     def active_users_count(self) -> int:
+        # TODO this should be a reliable count of active users, else needs to be renamed
```
faebebin, line 606: @suricactus eventhough there are tests describing the behaviour I do not really understand this. In the `active_users()` we set a default if `current_period_until` is not set and in the `active_users_count()` we return null. E.g. I get two users returned from `active_users` and 0 from `active_users_count()`. Not sure which one to use for the admin list_display and actually not even sure whether my calculation in https://github.com/opengisch/qfieldcloud/pull/704 was correct then ... ? At least we should rename the method to what it really does, e.g. `active_users_amount_for_billing` or so?

faebebin:
merging it int simplified detailview PR https://github.com/opengisch/qfieldcloud/pull/704 and therefore closing this one

PR #707, opened 2023-06-29
QF-2744 Quickfix Admin Projects download latest version button

why-not-try-calmer:
image

### DIFF
--- a/docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
+++ b/docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
@@ -238,7 +238,7 @@
                             const version = $versionsSelect.value || (
                                 file.versions
                                     .map(obj => obj.version_id)
-                                    .sort((a, b) => String(b).localeCompare(String(a).localeCompare()))[0]
+                                    .sort((a, b) => String(b).localeCompare(String(a)))[0]
                             );
                             window.open(buildApiUrl(pathToFile, { version }));
                         });
--- a/docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
+++ b/docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
@@ -44,13 +44,13 @@
                         <td class="qfc-admin-project-files-text-right"></td>
                         <td>
                             <select>
-                                <option value="">{% trans 'Download latest/Delete all version(s)' %}</option>
+                                <option value="">{% trans 'See all versions' %}</option>
                             </select>
                         </td>
                         <td>
-                            <button type="button" class="qfc-admin-projects-files-actions-info-btn">{% trans 'Info' %}</button>
-                            <button type="button" class="qfc-admin-projects-files-actions-download-btn">{% trans 'Download' %}</button>
-                            <button type="button" class="qfc-admin-projects-files-actions-delete-btn" data-csrf-token="">{% trans 'Delete' %}</button>
+                            <button type="button" class="qfc-admin-projects-files-actions-info-btn">{% trans 'File info' %}</button>
+                            <button type="button" class="qfc-admin-projects-files-actions-download-btn">{% trans 'Download latest version' %}</button>
+                            <button type="button" class="qfc-admin-projects-files-actions-delete-btn" data-csrf-token="">{% trans 'Delete file' %}</button>
                         </td>
                     </tr>
                 </template>
@@ -95,7 +95,8 @@
         width: 100%; /* Full width */
         height: 100%; /* Full height */
         overflow: auto; /* Enable scroll if needed */
-        background-color: rgb(0,0,0); /* Fallback color */
+        background-color: rgb(0,0,0
+        ); /* Fallback color */
         background-color: rgba(0,0,0,0.4); /* Black w/ opacity */
     }
 
@@ -215,9 +216,6 @@
                             $option.innerHTML = `${version.display} (${version.size} bytes)`;
 
                             $versionsSelect.appendChild($option);
-                            $infoBtn.disabled = true;
-                            $downloadBtn.disabled = true;
-                            $deleteBtn.disabled = true;
                         }
 
                         $tbody.appendChild($trow);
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-2744 Quickfix Admin Projects download latest version button](https://app.clickup.com/t/2192114/QF-2744)

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
```diff
@@ -224,6 +227,15 @@
...
 
+                        $versionsSelect.addEventListener('change', event => {
+                            versions_ids = file.versions.map(obj => obj.version_id)
+                            selected_version_id = event.target.value
+                            disable_if_not_version = !versions_ids.includes(selected_version_id)
+                            $infoBtn.disabled = disable_if_not_version;
+                            $downloadBtn.disabled = disable_if_not_version;
+                            $deleteBtn.disabled = disable_if_not_version;
+                        })
+
```
faebebin, line 238: ... so I would let the rest ...

faebebin: and only fix the `downloadBtn.addEventListener` ...

faebebin (changes requested):
see suggestion batch

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
```diff
@@ -215,6 +215,9 @@
                             $option.innerHTML = `${version.display} (${version.size} bytes)`;
 
                             $versionsSelect.appendChild($option);
+                            $infoBtn.disabled = true;
+                            $downloadBtn.disabled = true;
+                            $deleteBtn.disabled = true;
```
faebebin, line 220:
```suggestion
```

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
faebebin, line 238:
```suggestion
```

docker-app/qfieldcloud/core/templates/admin/project_files_widget.html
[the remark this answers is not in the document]
why-not-try-calmer: If nothing is selected, does it really make sense to let `info` or `delete` work?

why-not-try-calmer:
Actually, there is no need at all for a version selector, is there? After all, `Delete` and `Info` apply to the filename, and `Download` could simply get the value of the latest version from the JSON API, without the need to read any value from the form.

faebebin:
Ok, I do not know much about the file versioning actually. But then I would just to fix the current feature (which IMO my suggestion would do?) And then take it back to the support to ask what they really need before modifying the feature? Maybe comment your suggestions into clickup for discussion with support team

why-not-try-calmer:
> Ok, I do not know much about the file versioning actually. But then I would just to fix the current feature (which IMO my suggestion would do?) And then take it back to the support to ask what they really need before modifying the feature? Maybe comment your suggestions into clickup for discussion with support team

Yes, the selector only matter for downloads, and is therefore a bit misleading. It would make sense if we allowed also deletion to target specific files, so I'll forward this question.

I have reduced the scope of this PR to just fixing what needed to, relabeling here and there. Feel free to approve if you have no more objections!

faebebin:
@why-not-try-calmer did you ever try the following: Select a version > click 'delete file' . For me, it only deletes the chosen version? Should be master state I guess ...

faebebin:
Further the info button indeed shows info for all versions, so it could IMO be replaced on the right and maybe with a lilttle seperator to clarify that does not relate to the selected option.

why-not-try-calmer:
> @why-not-try-calmer did you ever try the following: Select a version > click 'delete file' . For me, it only deletes the chosen version? Should be master state I guess ...

No you are correct. I didn't expect the file version to be specified as part of the request's headers. I changed the html text of the Download button accordingly.

why-not-try-calmer:
^ updated screenshot in first commit message.

faebebin:
@why-not-try-calmer on master, if the default was selected and then clicked 'delete' then all versions (whole file) was deleted. So the title of 'delete version' button would be wrong in this case. And we do not want to change the logic of the default option. It should be a bugfix not new feature. IMO the feature was working with this one bug described in the task.

Please just implement my suggestion of just fixing downloadBtn.addEventListener https://github.com/opengisch/qfieldcloud/pull/708 . (In your version you do not take into account if there are no versions which IMO could be the case with some special files. At least I don't know and did not test it, so I just let this logic.) Further I don't know whether the sorting is necessary but maybe I am wrong there - then please correct me with specific code.

But anyway - let's not make a feature-PR out of this but stay with a bugfix. We should not lose more time on this. Else feel free to pass over to me. I won't approve it as a feature change.

why-not-try-calmer:
> @why-not-try-calmer on master, if the default was selected and then clicked 'delete' then all versions (whole file) was deleted. So the title of 'delete version' button would be wrong in this case. And we do not want to change the logic of the default option. It should be a bugfix not new feature. IMO the feature was working with this one bug described in the task.
> 
> Please just implement my suggestion of just fixing downloadBtn.addEventListener #708 . (In your version you do not take into account if there are no versions which IMO could be the case with some special files. At least I don't know and did not test it, so I just let this logic.) Further I don't know whether the sorting is necessary but maybe I am wrong there - then please correct me with specific code.
> 
> But anyway - let's not make a feature-PR out of this but stay with a bugfix. We should not lose more time on this. Else feel free to pass over to me. I won't approve it as a feature change.

Thanks for the suggestion, I think we are all good now. Can you checkout, test and confim?

suricactus:
Duplicate of https://github.com/opengisch/qfieldcloud/pull/708 .
Looking at the amount of changes here, it seems the PR went out of scope according to the original task on CU.
There are still some unresolved comments which probably cause this PR not to be mergeable yet.

Since we have an alternative implementation, I am closing this for now.

suricactus:
Ok, we did another round of reviewing with @why-not-try-calmer . We both agreed visual changes are best to be pushed in a separate PR and keep the bug fix PRs with less changes and therefore dicussion. For the future we will try to align on this.

@suricactus comment here:
The actual JS bugfix here is slightly more rubust and future proof because it does not depend on the default sorting that comes from the template (within the `<select><option>`) but really sorts the versions and makes sure we pick the most recent one.

@faebebin I hope you are ok with the current state of the PR. If yes, please close your PR here  https://github.com/opengisch/qfieldcloud/pull/708 and we keep the current state of this PR.

suricactus (approved):

PR #709, opened 2023-06-30
QF-2838 Public projects are not editable by default

why-not-try-calmer:


duke-nyuki:
Task linked: [QF-2838 Public projects are not editable by default](https://app.clickup.com/t/2192114/QF-2838)

suricactus (approved):

PR #710, opened 2023-07-04
Paginator for Projects as PoC

why-not-try-calmer:
Please no review yet.

### DIFF
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -1,7 +1,7 @@
 from rest_framework import pagination, response
 
 
-class PaginateResults(pagination.LimitOffsetPagination):
+class QfcLimitOffsetPagination(pagination.LimitOffsetPagination):
     """
     Based on LimitOffsetPagination. Custom implementation such that response.data = (DRF's blanket LimitOffsetPagination response).data.results
     For comparison, the DRF's blanket implementation defines:
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -3,12 +3,13 @@ from rest_framework import pagination, response
 
 class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
     """
-    Custom implemention such that response.data = response.data.results
-    For comparison, the DRF's blank implementation defines:
+    Custom implementation such that response.data = (DRF blanket implementation's response).data.results
+    For comparison, the DRF's blanket implementation defines:
+    - response.data.["results"]
     - response.data["count"]
     - response.data["next"]
     - response.data["previous"]
-    - response.data.["results"]"""
+    """
 
     def get_paginated_response(self, data) -> response.Response:
         return response.Response(data)
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -1,4 +1,5 @@
 from rest_framework import pagination
 
+
 class LimitOffsetPagination(pagination.LimitOffsetPagination):
     pass
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -5,7 +5,7 @@ class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
     """
     Custom implementation such that response.data = (DRF blanket implementation's response).data.results
     For comparison, the DRF's blanket implementation defines:
-    - response.data.["results"]
+    - response.data["results"]
     - response.data["count"]
     - response.data["next"]
     - response.data["previous"]
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -1,18 +1,14 @@
 from rest_framework import pagination, response
 
 
-class LimitOffsetPagination(pagination.LimitOffsetPagination):
-    """
-    Blank implementation defining:
-        - response.data["count"]
-        - response.data["next"]
-        - response.data["previous"]
-        - response.data.["results"]
-    """
-
-
 class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
-    """Custom implemention such that response.data = response.data.results"""
+    """
+    Custom implemention such that response.data = response.data.results
+    For comparison, the DRF's blank implementation defines:
+    - response.data["count"]
+    - response.data["next"]
+    - response.data["previous"]
+    - response.data.["results"]"""
 
     def get_paginated_response(self, data) -> response.Response:
         return response.Response(data)
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -1,5 +1,18 @@
-from rest_framework import pagination
+from rest_framework import pagination, response
 
 
 class LimitOffsetPagination(pagination.LimitOffsetPagination):
-    pass
+    """
+    Blank implementation defining:
+        - response.data["count"]
+        - response.data["next"]
+        - response.data["previous"]
+        - response.data.["results"]
+    """
+
+
+class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-2873 Add pagination on the QFC API endpoints](https://app.clickup.com/t/2192114/QF-2873)

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -26,3 +35,22 @@ def test_api_status_cache(self):
...
+        items_per_page = 1
+        view = ProjectViewSet.as_view({"get": "list"})
+        request = APIRequestFactory().get(
+            "/api/v1/projects/", {"items_per_page": items_per_page}
+        )
+        force_authenticate(request, user=user1, token=token)
+        response = view(request)
+        response_rendered = response.render()
+        results = response_rendered.data["results"]
+        self.assertEqual(len(results), items_per_page)
```
suricactus, line 56: Can you also verify that the current way of working (with no limit) is still returning the correct results. As far as I understand your PR, you are only testing when the new get parameter exists.

docker-app/qfieldcloud/core/paginators.py
```diff
@@ -35,3 +36,7 @@ def count(self):
             else:
                 return estimate
         return len(self.object_list)
+
+
+class DefaultPagination(pagination.LimitOffsetPagination):
+    limit_query_param = "items_per_page"
```
suricactus, line 42:
```suggestion
class LimitOffsetPagination(pagination.LimitOffsetPagination):
    ...
```
Just checked the base class has `limit_query_param="limit"`, which is ok name, I think there is no need to replace it. The one on ClickUp was just an example name.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -26,3 +41,42 @@ def test_api_status_cache(self):
         toc = time.perf_counter()
 
         self.assertGreater(toc - tic, 0)
+
+    def test_api_default_paginator_offset(self):
```
suricactus, line 45: Is it possible change your test to aligni with the already existing way of writing tests like here and the rest of the API testing code:

https://github.com/opengisch/qfieldcloud/blob/d7941b3614f161a08f8cbb6f03599496e2857c46/docker-app/qfieldcloud/core/tests/test_user.py#L101-L113

Or there is a technical limitation for that?

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -26,3 +41,42 @@ def test_api_status_cache(self):
...
+
+        # request-response without pagination (aka control test)
+        request_without_pagination = APIRequestFactory().get(
+            "/api/v1/projects/",
+        )
+        force_authenticate(request_without_pagination, user=self.user, token=self.token)
+        response = view(request_without_pagination)
+        response_rendered = response.render()
+        results_without_pagination = response_rendered.data
+        self.assertEqual(len(results_without_pagination), Project.objects.all().count())
```
suricactus, line 82: Are you sure there is no default pagination enforced, but with page size of 50 or so, so this kind of test is never going to catch it? IMO there should have been some default pagination enforced by the `Paginator`? Worth testing manually and if needed, incorporated in the test here.

docker-app/qfieldcloud/core/paginators.py
```diff
@@ -35,3 +36,7 @@ def count(self):
             else:
                 return estimate
         return len(self.object_list)
+
+
+class LimitOffsetPagination(pagination.LimitOffsetPagination):
+    pass
```
suricactus, line 42: I think the class should live in another file, called `pagination.py`, because they are a concept of DRF, and the current file is implementing concepts from django-admin.

docker-app/qfieldcloud/core/tests/test_api.py
why-not-try-calmer: I did check there was no enforced pagination independent of this PR at level of models or views. There is pagination using Django Table in `docker-app/qfieldcloud/web/views/projects_views.py`, but this is not something we can hope to re-use as is for any model or model-viewset.

docker-app/qfieldcloud/core/tests/test_api.py
why-not-try-calmer: I think this is the case now? Else I think the difference is necessary, because I don't need to play around with tokens.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -26,3 +56,56 @@ def test_api_status_cache(self):
         toc = time.perf_counter()
 
         self.assertGreater(toc - tic, 0)
+
+    def test_api_pagination_limitoffset(self):
+        expected_count = Project.objects.all().count()
+        assert expected_count == 500
```
suricactus, line 62:
```suggestion
        self.assertEqual(expected_count, 500)
```

docker-app/qfieldcloud/core/tests/test_api.py
why-not-try-calmer: Using now this style (client-based) style of authentication on the second test added in this PR. The first test still uses a factory because I need to be able to get a Request object to pass to the ProjectViewSet.as_view() closure. The reason for this is that it's the only way to ensure that the response with correspond to the actual rendered response (and not just the serialized response).

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import pagination, response
+
+
+class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
+    """
+    Custom implementation such that response.data = (DRF blanket implementation's response).data.results
+    For comparison, the DRF's blanket implementation defines:
+    - response.data.["results"]
```
why-not-try-calmer, line 8:
```suggestion
    - response.data["results"]
```

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import pagination, response
+
+
+class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
```
suricactus, line 4:
```suggestion
class LimitOffsetPagination(pagination.LimitOffsetPagination):
```

I would prefer a bit shorter name. We have to write it a few dozen times :)

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -26,3 +44,35 @@ def test_api_status_cache(self):
...
+        # Test page size
+        self.assertEqual(len(results_with_offset), page_size)
+
+        # Obtain without pagination (= control test)
+        results_without_pagination = self.client.get(
+            "/api/v1/projects/",
+        ).json()
+
+        # Test length (this is super slow -- 1 minute or so -- because of serialization)
+        self.assertEqual(len(results_without_pagination), unlimited_count)
```
suricactus, line 78: If there is no pagination by default, then just decrease the number of projects we have to save some CI time .

suricactus (approved):
Looks very good! Thanks! Few small comments above.

docker-app/qfieldcloud/core/pagination.py
suricactus: @why-not-try-calmer would you mind if you implement the proposed naming? I would have expected to either comment on the suggestion if you don't agree or accomodate it :)

why-not-try-calmer: In https://github.com/opengisch/qfieldcloud/pull/710/commits/f0eb815d8bc6fe646f9a99b0255ec6fe87ebf0e2 I accomodated the suggestion of renaming, but not with the particular name suggested. I have no preference about the name, but I think its unusual and a little bit misleading to give it the same name as its super-class. It's unusual in general to name classes after their super-classes, and it's misleading in this case because it really behaves differently.

suricactus: Thanks for the feedback. Yes, I overlooked the name matching.

`QfcLimitOffsetPagination` is a good name in this case, we have used similar patterns around the codebase.

suricactus (approved):
Good!

PR #711, opened 2023-07-04
Customize Model-level Pagination 

why-not-try-calmer:
Building upon #710.

### DIFF
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -31,7 +31,10 @@ class QfcLimitOffsetPagination(pagination.LimitOffsetPagination):
     default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
 
     def get_headers(self) -> dict[str, None]:
-        """Initializes a new header field to carry the number of paginated entries."""
+        """
+        Initializes a new header field to carry the number of paginated entries
+        if the class method 'count_entries' is True.
+        """
         if self.count_entries:
             return {"X-Total-Count": self.count}
         else:
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -20,16 +20,28 @@ def parameterize_pagination(_class: type) -> Callable:
 @parameterize_pagination
 class QfcLimitOffsetPagination(pagination.LimitOffsetPagination):
     """
-    Based on LimitOffsetPagination. Custom implementation such that response.data = (DRF's blanket LimitOffsetPagination response).data.results
-    For comparison, the DRF's blanket implementation defines:
-    - response.data["results"]
-    - response.data["count"]
-    - response.data["next"]
-    - response.data["previous"]
-    which are not available from instances of the present class.
+    Based on LimitOffsetPagination.
+    Custom implementation such that response.data = (DRF's blanket LimitOffsetPagination response).data.results
+    Optionally sets a new header ('X-Total-Count') to the number of entries in the paginated response.
+    Use it only if you can afford the performance cost.
+    Can be customized when assigning 'pagination_class'.
     """
 
     default_limit = QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
+    count_entries = True
+
+    def get_headers(self) -> dict[str, None]:
+        """Initializes a new header field to carry the number of paginated entries."""
+        return {"X-Total-Count": None}
 
     def get_paginated_response(self, data) -> response.Response:
-        return response.Response(data)
+        """
+        Sets the header field initialized in the previous method to the number of paginated entries.
+        Return just the entries in the response body.
+        """
+        if self.count_entries:
+            headers = self.get_headers()
+            headers["X-Total-Count"] = len(data)
+            return response.Response(data, headers=headers)
+        else:
+            return response.Response(data)
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -30,7 +30,7 @@ class QfcLimitOffsetPagination(pagination.LimitOffsetPagination):
     count_entries = True
     default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
 
-    def get_headers(self) -> dict[str, None]:
+    def get_headers(self) -> dict[str, Any]:
         """
         Initializes a new header field to carry the number of paginated entries
         if the class method 'count_entries' is True.
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -27,8 +27,8 @@ class QfcLimitOffsetPagination(pagination.LimitOffsetPagination):
     Can be customized when assigning 'pagination_class'.
     """
 
-    default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
     count_entries = False
+    default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
 
     def get_headers(self) -> dict[str, None]:
         """Initializes a new header field to carry the number of paginated entries."""
--- a/docker-app/qfieldcloud/core/pagination.py
+++ b/docker-app/qfieldcloud/core/pagination.py
@@ -1,7 +1,20 @@
 from rest_framework import pagination, response
 
 
-class LimitOffsetPaginationResults(pagination.LimitOffsetPagination):
+def parameterize_pagination(cls):
+    def capture_args(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


duke-nyuki:
Task linked: [QF-2873 Add pagination on the QFC API endpoints](https://app.clickup.com/t/2192114/QF-2873)

why-not-try-calmer:
Among several approaches (metaclass, mixin, manually slicing queryset in `get_queryset`..) I suggest a simple decorator to customize the class attributes of a target pagination class.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -102,7 +102,7 @@ class ProjectViewSet(viewsets.ModelViewSet):
     serializer_class = ProjectSerializer
     lookup_url_kwarg = "projectid"
     permission_classes = [permissions.IsAuthenticated, ProjectViewSetPermissions]
-    pagination_class = pagination.PaginateResults
+    pagination_class = pagination.QfcLimitOffsetPagination(default_limit=25)
```
suricactus, line 105: Nice we can control the `default_limit` via an argument. Is it possible to setup a value globally in `settings.py` too? Something like `QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT`?

docker-app/qfieldcloud/settings.py
```diff
@@ -431,6 +431,9 @@ def before_send(event, hint):
 # Use pg meta table estimate for pagination and display above n entries
 QFIELDCLOUD_ADMIN_EXACT_COUNT_LIMIT = 10000
 
+# Default limit for paginating data from views using QfcLimitOffsetPagination
+QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT = 15
```
suricactus, line 435:
```suggestion
QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT = 50
```

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -1,15 +1,46 @@
...
-    - response.data["next"]
-    - response.data["previous"]
+    Based on LimitOffsetPagination.
+    Custom implementation such that response.data = (DRF's blanket LimitOffsetPagination response).data.results
+    Optionally sets a new header ('X-Total-Count') to the number of entries in the paginated response.
+    Use it only if you can afford the performance cost.
+    Can be customized when assigning 'pagination_class'.
     """
 
+    default_limit = QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
```
why-not-try-calmer, line 30:
```suggestion
    default_limit = QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
    count_entries = False
```

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -102,7 +102,7 @@ class ProjectViewSet(viewsets.ModelViewSet):
     serializer_class = ProjectSerializer
     lookup_url_kwarg = "projectid"
     permission_classes = [permissions.IsAuthenticated, ProjectViewSetPermissions]
-    pagination_class = pagination.QfcLimitOffsetPagination
+    pagination_class = pagination.QfcLimitOffsetPagination(count_entries=True)
```
suricactus, line 105:
```suggestion
    pagination_class = pagination.QfcLimitOffsetPagination()
```

Make `count_entries` should be `True` for now by default. Also assume it is always not-None.

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -1,15 +1,46 @@
+from typing import Callable
+
+from qfieldcloud.settings import QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
```
suricactus, line 3:
```suggestion
from django.conf import settings
```

Imports of the settings should be like this, so Django can do a bit of magic and merge different app settings into one. Otherwise you might miss some overwrite of setting values from another app.

docker-app/qfieldcloud/core/pagination.py
suricactus, line 30:
```suggestion
    default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
```

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -1,15 +1,46 @@
...
+
     def get_paginated_response(self, data) -> response.Response:
-        return response.Response(data)
+        """
+        Sets the header field initialized in the previous method to the number of paginated entries.
+        Return just the entries in the response body.
+        """
+        if self.count_entries:
+            headers = self.get_headers()
+            headers["X-Total-Count"] = len(data)
```
suricactus, line 43: I think the `data` variable already only has the current slice of total data in the DB. However, what we actually need the total data, not the items of the current page.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -71,9 +72,24 @@ def test_api_pagination_limitoffset(self):
...
+
+    def test_api_headers_count(self):
+        """Test LimitOffset pagination custom 'X-Total-Count' headers implementation"""
+        # Authenticate client
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token.key)
+
+        response = self.client.get("/api/v1/projects/")
+        self.assertEqual(
+            int(response.headers["X-Total-Count"]),
+            self.total_projects,
```
why-not-try-calmer, line 94: Fixed: count from queryset, not from paginated results

suricactus: Thanks!

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -1,15 +1,45 @@
...
     def get_paginated_response(self, data) -> response.Response:
-        return response.Response(data)
+        """
+        Sets the header field initialized in the previous method to the number of paginated entries.
+        Return just the entries in the response body.
+        """
+        if self.count_entries:
+            return response.Response(data, headers=self.get_headers())
+        else:
+            return response.Response(data)
```
suricactus, line 45: I would move this `if` in the `get_headers` method. This way you don't branch you code regarding headers in other than `get_headers` callback. Makes life easier whenever extending is needed in the future.

why-not-try-calmer: Unfortunately that's not easy to do. Behind the curtains DRS manipulates the headers so that if `self.get_headers()` is not called, the extra header is not registered.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -71,9 +73,29 @@ def test_api_pagination_limitoffset(self):
...
+
+    def test_api_headers_count(self):
+        """Test LimitOffset pagination custom 'X-Total-Count' headers implementation"""
+        # Authenticate client
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token.key)
+
+        # Mutating project viewset for testing purposes
+        ProjectViewSet.pagination_class = pagination.QfcLimitOffsetPagination(
+            count_entries=True
+        )
```
suricactus, line 95: It would be interesting to add a test with `count_entries=False`.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -71,9 +73,35 @@ def test_api_pagination_limitoffset(self):
...
+            int(response.headers["X-Total-Count"]),
+            self.total_projects,
+        )
+
+        # Forcing 'count_entries' to be False
+        ProjectViewSet.pagination_class = pagination.QfcLimitOffsetPagination(
+            count_entries=False
+        )
+        response = self.client.get("/api/v1/projects/")
+        self.assertFalse("X-Total-Count" in response.headers)
```
suricactus, line 107:
```suggestion
        self.assertNotIn("X-Total-Count", response.headers)
```
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertNotIn

docker-app/qfieldcloud/core/pagination.py
```diff
@@ -1,15 +1,48 @@
...
+    Custom implementation such that response.data = (DRF's blanket LimitOffsetPagination response).data.results
+    Optionally sets a new header ('X-Total-Count') to the number of entries in the paginated response.
+    Use it only if you can afford the performance cost.
+    Can be customized when assigning 'pagination_class'.
     """
 
+    count_entries = True
+    default_limit = settings.QFIELDCLOUD_API_DEFAULT_PAGE_LIMIT
+
+    def get_headers(self) -> dict[str, None]:
```
why-not-try-calmer, line 33:
```suggestion
    def get_headers(self) -> dict[str, Any]:
```

suricactus (approved):
Final comments on the docs, looks good an will be merged.

PR #712, opened 2023-07-04
Same number in comment and test_organization

faebebin:
Some readability improvement as discussed here https://github.com/opengisch/qfieldcloud/pull/705#discussion_r1250774648


faebebin:
Yes good point :D  - I kinda wanted to stick to the numbers. But will just drop them. thanks

faebebin:
@why-not-try-calmer hope better maintainable now.

why-not-try-calmer (approved):

PR #713, opened 2023-07-05
Added more logging during job runs

suricactus:


suricactus:
The more the merrier!

PR #716, opened 2023-07-06
Set final status to jobs that have been killed by docker engine

suricactus:
Whenever the docker engine kills a container there is exit code 137. However, sometimes the "murder" happens when the job is still in the DB. In such cases, the job is stuck in "Started" status forever. This PR is setting the job into "failed" status.


faebebin (approved):
LGTM 

I know you wanted to also implement the name parameter to avoid interference with the test stack. I anyway approve this part because I will be off soon until Monday. IMO can add a followup PR.

faebebin (approved):
LGTM

Tested locally and still works + adds some useful info to the logs and feedback

PR #718, opened 2023-07-10
Fix email export and add extra columns to the export

suricactus:
Email export was not working due to order of URL patters to check.

Added columns `verified`, `last_login`, `owner_last_login`.

Check bottom right where the button is:

image


why-not-try-calmer:
Thanks, amazing time, I was literally scratching my head as to why the exportation failed.

why-not-try-calmer (approved):
Tested locally and it works.

PR #719, opened 2023-07-10
Create codeql.yml

m-kuhn:


suricactus:
@m-kuhn seems the system is super secure exept for the regex. Anyways, I think we should merge, seems it does not hurt?

m-kuhn:
I guess we need to close this one hole. And we can merge this whenever a review is greened.

suricactus (approved):
Thanks, look good to me.

PR #721, opened 2023-07-11
Skip downloading the S3 metadata if `skip_metadata` is passed to the API `/files/` list endpoint

suricactus:
The slowness is added by the line calling `version.head()`, which is additional HEAD request to S3 for each and every version. Far from optimal...

Some clients (e.g. QField, QFieldSync) are still requiring the `sha256` key to check whether the files needs to be reuploaded. Since we do not have control on these old client versions, we need to keep the API backward compatible for some time and assume `skip_metadata=0` by default.

TODO
Follow up PR that uses the `get_project_file_versions`, instead of custom implementation of it in `files_views.py`.


duke-nyuki:
Task linked: [QF-2910 Skip downloading the S3 metadata if `skip_metadata` is passed to the API `/files/` list endpoint](https://app.clickup.com/t/2192114/QF-2910)

faebebin (approved):
LGTM

Works locally :+1:

PR #723, opened 2023-07-11
Ignore validation errors for sentry

m-kuhn:
Will ignore the number one "error" there is --> wrong login credentials provided.
I was wondering if we should only ignore exceptions that are unhandled (e.g. this one is handled, so it will still be hidden). @faebebin any idea?

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -284,6 +284,7 @@ if SENTRY_DSN:
             PlanInsufficientError,
             QuotaError,
         )
+        from rest_framework.exceptions import ValidationError as RestValidationError
 
         ignored_exceptions = (
             ValidationError,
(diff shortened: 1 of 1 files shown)


suricactus:
> I was wondering if we should only ignore exceptions that are unhandled (e.g. this one is handled, so it will still be hidden).

I think we should ignore exception that are handled if we want to solve the problem once and forever. However, explicitly listing them brings less magic, which is also not bad.

m-kuhn:
The problem is, that currently we ignore the listed errors even if they are unhandled

faebebin:
Yes I would agree here that the proper way would be to send all unhandled errors to sentry, except some special cases of orphaned workers. IMO good for now but maybe we could work out a proper solution -> ticket?

m-kuhn:
There is a ticket we opened this morning in the current sprint ;-)

faebebin:
Cool, thanks - I was not in the best shape yesterday ;)

PR #726, opened 2023-07-12
Added `Project.problems` to list failing requirements for the project file

suricactus:
While not the best thing I have created in my life, I think for now it can serve the role of listing all problems that we know at QFieldCloud and already show them to the end user.


duke-nyuki:
Task linked: [QF-2892 Show information when the project will have some layers as readonly](https://app.clickup.com/t/2192114/QF-2892)

PR #730, opened 2023-07-12
Add `level` to `project.problems` and make `project.status` FAILED if `project.project_details` is missing

suricactus:


duke-nyuki:
Task linked: [QF-2892 Show information when the project will have some layers as readonly](https://app.clickup.com/t/2192114/QF-2892)

PR #733, opened 2023-07-13
QfcLimitOffsetPagination default limit

why-not-try-calmer:
- changed behaviour of QfcLimitOffset pagination class to not limit results unless a limit offset param is found in the request
- adjusted tests


suricactus (approved):
Looks good as code. Cannot test it myself, but testa look good.

PR #737, opened 2023-07-17
Opt-in controls for browsing paginated results

why-not-try-calmer:


suricactus (commented):
Code looks good but a bit of motivation why you do that would be good.

suricactus:
Thank you for the clarifications. I think this is out of scope of the task and supporting two serializations of the same API endpoint is undesired with the current manpower of the project. We might reconsider the final structure of the API results in the future with `api/v2`, but for now we should communicate all this metadata with headers only if explicitly required by any of the API consumers.

PR #739, opened 2023-07-18
Django command for extracting S3 version files metadata

why-not-try-calmer:
- [x] Download files concurrently
- [x] Extract metadata
- [x] Collect and save to disk as .csv
- [x] Add unit tests

__input__:
```
docker compose app python manage.py extracts3data
```

__output__:
```
id,key,e_tag,size,last_modified
20b479ac-f039-44f7-ab1e-9dcf0d7d07ea,projects/037ca201-747c-4b21-94eb-56bc5afe0212/files/aaa/file.txt,3bf4cfdddae3137d565094635a8ebcc9,13,2023-07-20 11:56:55.696000+00:00
a44433a6-cfce-48c6-aeb2-23b31ff1b126,projects/037ca201-747c-4b21-94eb-56bc5afe0212/files/aaa/file.txt,9af2f8218b150c351ad802c6f3d66abe,13,2023-07-20 11:56:52.387000+00:00
etc.
```


duke-nyuki:
Task linked: [QF-2879 Develop a script that produces a csv with s3 objects and their versions](https://app.clickup.com/t/2192114/QF-2879)

why-not-try-calmer:
Huh?
image

Yet the CI's check code step fails.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,141 @@
...
+
+                for field in interesting_fields:
+                    if hasattr(file, field):
+                        item = getattr(file, field)
+
+                        if not isinstance(item, str):
+                            item = str(item)
+
+                        item = item.strip('"')
+                        row.append(item)
```
suricactus, line 132: We do not need such level of abstraction with `interesting_fields`. You can directly get the keys the version object.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,141 @@
...
+            valid_config = cls.validate(maybe_valid_config)
+            cls._config = S3ConfigObject(**valid_config)
+
+        return cls._config
+
+    @classmethod
+    def inject_to_settings(cls):
+        """To avoid passing s3 config as parameters"""
+        for key, value in cls._config._asdict().items():
+            setattr(settings, key, value)
```
suricactus, line 80: This whole class is unnessesary. You could have just copy-pasted what we already had:

```
def get_s3_bucket(acces_key_id, secret_access_key, region_name, endpoint_url, bucket_name)
    session = boto3.Session(
        aws_access_key_id=acces_key_id,
        aws_secret_access_key=secret_access_key,
        region_name=region_name,
    )
    s3 = session.resource("s3", endpoint_url=endpoint_url)
    bucket = s3.Bucket(bucket_name)
    return bucket
```

Changing the settings like this is dangerous and quite far from what we needed...

suricactus (changes requested):

suricactus:
Fix is comming here hopefully: https://github.com/opengisch/qfieldcloud/pull/744

why-not-try-calmer:
Also it is possible that the CI has no access to any S3 files by default? The only failing test works A-OK on my machine.

suricactus (changes requested):
Please remove all changes related to the `S3ConfigObject`. Adding such a complexity is unneeded for the context of the task.

why-not-try-calmer:
> Please remove all changes related to the `S3ConfigObject`. Adding such a complexity is unneeded for the context of the task.

The task reads: 

> There should be a way to pass Object storage credentials if one wants to get the stats for a Object storage that is not already available.

My understanding is that the command should let users pass credentials independently from the application's settings. Is this correct?

Also, I think it's a good idea to validate credentials as soon as possible. Is this not correct? 

If you look at the code, all the class does is construct a datastructure from a validated user input, and store it for reuse. Since you asked that `get_s3_bucket` be refactored to make use of such a datastructure, why do you feel like this is adding unnecessary complexity? Wouldn't it much more complex to add 5 new parameters to all the functions that need this config independently of the application's settings?

suricactus:
> Since you asked that get_s3_bucket be refactored to make use of such a datastructure, why do you feel like this is adding unnecessary complexity?

I think this is where the misunderstanding is coming from. I guess I didn't highlight enough on "copy-pasted" part of my comment. I meant such function should be created in the script file, not refactoring the original function in `utils.py`.

The original task said: "Develop a script that produces a csv with s3 objects and their versions" . It also has "Implement in Python, ideally as a Django command." . This suggests that the script should be quite independent from the rest of the code. In the private conversation during the call I also tried to explain you the motivation behind the requested feature, which also one-time usage.

In the end I expected <50 lines of Python and I got 223 lines instead. Makes me feel a mistake that I said "ideally as a Django command", we don't want to solve all problems in the universe with such script, we just want a list of s3 objects and their versions.

---

Let's try to add another constraint to the task:

Do not change any runtime code outside `docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py`. If something is not yet available in the existing code, just copy-paste and modify it in the `extractstoragemetadata.py` file.

You can keep your fixes on the typings though.

why-not-try-calmer:
> > Since you asked that get_s3_bucket be refactored to make use of such a datastructure, why do you feel like this is adding unnecessary complexity?
> 
> I think this is where the misunderstanding is coming from. I guess I didn't highlight enough on "copy-pasted" part of my comment. I meant such function should be created in the script file, not refactoring the original function in `utils.py`.
> 
> The original task said: "Develop a script that produces a csv with s3 objects and their versions" . It also has "Implement in Python, ideally as a Django command." . This suggests that the script should be quite independent from the rest of the code. In the private conversation during the call I also tried to explain you the motivation behind the requested feature, which also one-time usage.
> 
> In the end I expected <50 lines of Python and I got 223 lines instead. Makes me feel a mistake that I said "ideally as a Django command", we don't want to solve all problems in the universe with such script, we just want a list of s3 objects and their versions.
> 
> Let's try to add another constraint to the task:
> 
> Do not change any runtime code outside `docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py`. If something is not yet available in the existing code, just copy-paste and modify it in the `extractstoragemetadata.py` file.
> 
> You can keep your fixes on the typings though.

okay, and what about the crucial

> There should be a way to pass Object storage credentials if one wants to get the stats for a Object storage that is not already available.

? Is that still on the menu?

why-not-try-calmer:
So for now the 2 blockers for me are:
1. YAML: Is that the right format to source credentials?
2. `create_time`: What should it be replaced with?

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @staticmethod
+    def from_file(path_to_config: Path) -> S3ConfigObject:
+        """Load a YAML file exposing credentials used by S3 storage"""
+        with open(path_to_config) as fh:
+            contents = yaml.safe_load(fh)
+
+        config = {
+            k: v for k, v in contents.items() if k in Command.S3ConfigObject._fields
+        }
```
suricactus, line 80: Can you rewrite this as proper `for` with `if` inside, in the qfc codebase inlining is acceptable for very simple cases?

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+        STORAGE_ENDPOINT_URL: str
+
+    @staticmethod
+    def from_file(path_to_config: Path) -> S3ConfigObject:
+        """Load a YAML file exposing credentials used by S3 storage"""
+        with open(path_to_config) as fh:
+            contents = yaml.safe_load(fh)
+
+        config = {
+            k: v for k, v in contents.items() if k in Command.S3ConfigObject._fields
```
suricactus, line 79: I am afraid we depend on property prefixed with `_` (underscore) here `Command.S3ConfigObject._fields`, which usually means hidden properties. I know this is the way for getting a list of properties in named tuples and it is highly unlikely they ever change it.

Nevertheless, I would prefer less magic and just explicitly get the keys from the config file.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+        }
+        return Command.S3ConfigObject(**config)
+
+    @staticmethod
+    def get_s3_bucket(config: S3ConfigObject) -> mypy_boto3_s3.service_resource.Bucket:
+        """Get a new S3 Bucket instance using S3ConfigObject"""
+
+        bucket_name = config.STORAGE_BUCKET_NAME
+        assert bucket_name, "Expected `bucket_name` to be non-empty string!"
```
suricactus, line 88: We either validate all the settings, or none of them. I don't have strong opinion.

I guess we want this validation to be when we read the config file?

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+        session = boto3.Session(
+            aws_access_key_id=config.STORAGE_ACCESS_KEY_ID,
+            aws_secret_access_key=config.STORAGE_SECRET_ACCESS_KEY,
+            region_name=config.STORAGE_REGION_NAME,
+        )
+        s3 = session.resource("s3", endpoint_url=config.STORAGE_ENDPOINT_URL)
+
+        # Ensure the bucket exists
+        s3.meta.client.head_bucket(Bucket=bucket_name)
+
```
suricactus, line 99:
```suggestion
```
No need, we were experimenting with this one.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+                except AttributeError:
+                    logger.error(
+                        f"Unable to find attribute '{field}', perhaps you were trying to get one of these instead {vars(file)}?"
+                    )
+                    sys.exit(1)
+
+                if not isinstance(item, str):
+                    item = str(item)
+
+                item = item.strip('"')
```
suricactus, line 124:
```suggestion
                # some fields (e.g. `e_tag`) have unnecessary `"` (double quotes) in their value.
                item = item.strip('"')
```

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,127 @@
...
+from qfieldcloud.core import utils
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Save metadata from S3 storage project files to disk."""
+
+    def add_arguments(self, parser: CommandParser):
+        parser.add_argument("-o", "--output", type=str, help="Name of output file")
```
suricactus, line 20: I think we can print to the `stdout` by default, so the `docker compose exec python manage.py` command can be redirected?

suricactus:
> @suricactus So for now the 2 blockers for me are:
> 
> 1. YAML: Is that the right format to source credentials?

If you can swith them to CLI arguments.

> 2. `create_time`: What should it be replaced with?

Each version has `last_modified`, which equals the `create_time`. Sorry, I didn't know the name of the field by heart.

3. Another request: Can you add an optional "prefix" argument, that shows only those object versions that have a key prefixed with that prefix.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+        s3 = session.resource("s3", endpoint_url=config.STORAGE_ENDPOINT_URL)
+        return s3.Bucket(bucket_name)
+
+    @staticmethod
+    def read_bucket_files(
+        bucket, fields: tuple[str], prefix: str = ""
+    ) -> Generator[list[str], None, None]:
+        """Yield file metadata 1 by 1. `...object_versions.filter(prefix="")` == `...object_versions.all()`"""
+
+        for file in bucket.object_versions.filter(Prefix=prefix):
```
suricactus, line 113:
```suggestion
        """Yield file metadata 1 by 1. Passing `prefix=""` will list all bucket objects."""

        for file in bucket.object_versions.filter(Prefix=prefix):
```

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+        parser.add_argument(
+            "-s3",
+            "--s3_credentials",
+            type=str,
+            help="""
+            Optional: S3 credentials as comma-separated key=value pairs.
+            Example: -s3 STORAGE_ACCESS_KEY_ID=...,STORAGE_SECRET_ACCESS_KEY=...,...
+        """,
+        )
```
suricactus, line 29: Please add all the credentials as separate arguments. It is more explcit and will reduce your code in terms of size and complexity, e.g. `from_shell` will become obsolete.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+        output_name = options.get("output")
+
+        if output_name:
+            name_suffix = output_name.rsplit(".", 1)
+
+            if len(name_suffix) != 2 or name_suffix[1] != "csv":
+                logger.error("The name of the output file should end with '.csv'")
+                sys.exit(1)
+        else:
+            output_name = "s3_storage_files.csv"
```
suricactus, line 50:
```suggestion
```

No need for that many checks. Your file extension should not be a reason for this command to fail.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+                    logger.warning(f"Blank or None value at key {key}")
+                builder[key] = val.strip()
+
+            config = Command.S3ConfigObject(**builder)
+        except Exception:
+            logger.error(
+                f"Unable to extract valid S3 storage credentials from your input: {user_input}"
+            )
+            sys.exit(1)
+        return config
```
suricactus, line 93: This is not needed code. We can use the native argparse functionality coming from django's `management.py`.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+            writer.writerows(rows)
+
+        logger.info(f"Successfully exported data to {output_name}")
+
+    class S3ConfigObject(NamedTuple):
+        STORAGE_ACCESS_KEY_ID: str
+        STORAGE_BUCKET_NAME: str
+        STORAGE_ENDPOINT_URL: str
+        STORAGE_REGION_NAME: str
+        STORAGE_SECRET_ACCESS_KEY: str
```
suricactus, line 73: If you insist of keeping this, please move it to the begining of the class. I would prefer to drop it and just use a simple dict. No strong opinion though.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+
+            for field in fields:
+
+                try:
+                    item = getattr(file, field)
+                except AttributeError:
+                    logger.error(
+                        f"Unable to find attribute '{field}', perhaps you were trying to get one of these instead {vars(file)}?"
+                    )
+                    sys.exit(1)
```
suricactus, line 124: Is there a real situation when this might fail? If not, I would prefer to not have this error handling in this case.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+
+                if not isinstance(item, str):
+                    item = str(item)
```
suricactus, line 127: Why is that? What we might have here that needs to be forcefully converted to str?

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
suricactus: Actually what will happend if the filename has `"` in it? We either call this only for etag or remove it entirely.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
```diff
@@ -0,0 +1,133 @@
...
+        fields = (
+            "id",
+            "key",
+            "e_tag",
+            "size",
+            "last_modified",
+        )
+        rows = self.read_bucket_files(bucket, fields)
+
+        with open(output_name, "w") as fh:
```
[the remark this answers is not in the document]
why-not-try-calmer: I don't think this bit of magic is required or helpful. Here all I am doing is adding the column name to the output, before writing the "value" rows.

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
why-not-try-calmer: size is an int

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
why-not-try-calmer: if the filename has " in non-wrapping position (first or last) it will work. if the filename has " in wrapping position it will be removed, but that should not be a problem, should it?

docker-app/qfieldcloud/core/management/commands/extractstoragemetadata.py
suricactus: Why do we prefer to be a str? This will make comparing the CSV harder oater, no?

why-not-try-calmer: csv is a text format it doesn't draw a distinction between Python int or str (in other words: reading from a csv always opens a text stream)

suricactus: The main question is why we do the explicit string conversion. Which I think is unneeded, there is no requirement to have string objects passed to "writeline".

What is more, this is also nice functionality in Python `csv` module:
```
 $ cat /tmp/csv2.py          
import csv
with open('example.csv', 'w') as csvfile:
    writer = csv.writer(csvfile, quoting=csv.QUOTE_NONNUMERIC)
    writer.writerow(['text', 42])

with open('example.csv', 'r') as csvfile:
    reader = csv.reader(csvfile, quoting=csv.QUOTE_NONNUMERIC)
    for row in reader:
        for c in row:
            print(c, type(c))
 $ python /tmp/csv2.py
text <class 'str'>
42.0 <class 'float'>
```

why-not-try-calmer: you cannot call .strip() on an int

docker-app/qfieldcloud/core/management/commands/extracts3data.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        parser.add_argument(
+            "--storage_region_name",
+            type=str,
+            help="S3 region name",
+        )
+
+    def handle(self, **options):
+        config = S3ConfigObject(
+            **{
```
suricactus, line 56: Avoid using dictionary comprehensions that take more than one line or even avoid them at all. THey are hard to read for people who are changing languages all the time.

docker-app/qfieldcloud/core/management/commands/extracts3data.py
```diff
@@ -0,0 +1,113 @@
...
+            "size",
+            "last_modified",
+        )
+        rows = self.read_bucket_files(bucket, fields)
+
+        handle = open(output_name, "w") if output_name else stdout
+        writer = csv.writer(handle, delimiter=",")
+        writer.writerow(fields)
+        writer.writerows(rows)
+        handle.close() if not stdout else ()
```
suricactus, line 77:
```suggestion
        if output_name:
            handle = open(output_name, "w")
        else:
            handle = stdout
        writer = csv.writer(handle, delimiter=",")
        writer.writerow(fields)
        writer.writerows(rows)

        if output_name:
            handle.close()
```

suricactus (commented):
Few small adjustments on the quirky syntax. Please use more straighforward language constructs in general, they are easier to read for everyone.

suricactus (commented):
Ok, looks good to me. Let's merge and see.

suricactus (approved):
Ok, looks good to me. Let's merge and see.

PR #743, opened 2023-07-19
Add remaining_trial_organizations to Person

faebebin:
Setting a counter `Person.remaining_trial_organizations` (default = 0) that is set to `Plan.max_trial_organizations` when checking out a new subscription, and decremented when starting a trial. It can be set also in the admin.

NOTE until now the check whether a trial can be started was based on the existing organizations (field: `is_initially_trial`). By deleting an organization one could start a new trial. The check will be based on the actual Person's history (counter).


duke-nyuki:
Task linked: [QF-2323 Add `UserAccount.remaining_trial_organizations`, so it's easy to change this from support's perspective](https://app.clickup.com/t/2192114/QF-2323)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -839,6 +839,12 @@ def create_subscription(
             )
             # NOTE to get annotations, mostly `is_active`
             trial_subscription_obj = cls.objects.get(pk=trial_subscription.pk)
+
+            if created_by.is_person and created_by.remaining_trial_organizations > 0:
+                # TODO Fix: for unknown reason created_by is sometimes an Organization
```
faebebin, line 844: TODO (see comment)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -852,6 +858,17 @@ def create_subscription(
             regular_plan = plan
             regular_active_since = active_since
 
+            if created_by.is_person:
+                # TODO Fix: for unknown reason created_by is sometimes an Organization
```
faebebin, line 862: TODO (see comment)

docker-app/qfieldcloud/subscription/models.py
faebebin: @suricactus do you know by channce, why created_by is an 'Organization' here? I am really a bit stuck here ...

faebebin:
Will force push after rebase. Added only one more commit else history unchanged

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -820,6 +814,12 @@ def create_subscription(
             # remove microseconds as there will be slight shift with the remote system data
             active_since = active_since.replace(microsecond=0)
 
+        if account.user.is_organization:
+            # NOTE sometimes `account.user` is not an organization instance for unknown reasons
+            created_by = Organization.objects.get(pk=account.pk).organization_owner
+        else:
+            created_by = account.user
+
```
faebebin, line 822: Could not find out neither what is wrong here. At least seams to work ...

docker-app/qfieldcloud/subscription/tests/test_subscription.py
```diff
@@ -724,6 +724,41 @@ def test_create_default_plan_subscription_raises_when_subscription_already_exist
...
+    def assert_create_org_decrements_count(
+        self, org_name, user, remaining_trial_organizations
+    ):
+        Organization.objects.create(
+            username=org_name, organization_owner=user, created_by=user
+        )
+        user.refresh_from_db()
+        self.assertEqual(
+            user.remaining_trial_organizations, remaining_trial_organizations
+        )
```
suricactus, line 760:
```suggestion
        def assert_create_org_decrements_count(
	        org_name, user, remaining_trial_organizations
	    ):
	        Organization.objects.create(
	            username=org_name, organization_owner=user, created_by=user
	        )
	        user.refresh_from_db()
	        self.assertEqual(
	            user.remaining_trial_organizations, remaining_trial_organizations
	        )

        # remaining_trial_organizations is decremented when creating a trial organization
        assert_create_org_decrements_count("org1", u1, 1)

        # remaining_trial_organizations is decremented down to 0
        assert_create_org_decrements_count("org2", u1, 0)

        # It doesn't directly prevent creating more trials, is just a counter that stays at 0
        assert_create_org_decrements_count("org3", u1, 0)
```

If you prefer to have the function, then define it withing the test. I think this is safer and easier to read. Ideally I usually avoid having such "helper" functions, as they may cause trouble in the future when reading. But in this case I think it is ok.

suricactus (approved):
LGTM. One suggestion about code reorganization.

docker-app/qfieldcloud/subscription/tests/test_subscription.py
faebebin: OK, thanks - will remember for future

PR #744, opened 2023-07-19
Fix regex error in tests

suricactus:


suricactus:
@m-kuhn now seems iti s fixed without causing more trouble.

m-kuhn:
Thx

PR #747, opened 2023-07-20
OpenAPI docs improvements

why-not-try-calmer:
See /swagger endpoint.

### DIFF
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -331,3 +331,13 @@ class PublicFilesView(views.APIView):
 
     def get(self, request, filename):
         return utils2.storage.file_response(request, filename)
+
+@extend_schema(exclude=True)
+class AdminDownloadPushDeleteFileView(DownloadPushDeleteFileView):
+    """Allowing `DownloadPushDeleteFileView` to be excluded from the OpenAPI schema documentation"""
+    pass
+
+@extend_schema(exclude=True)
+class AdminListFilesViews(ListFilesView):
+    """Allowing `ListFilesView` to be excluded from the OpenAPI schema documentation"""
+    pass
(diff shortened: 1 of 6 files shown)


duke-nyuki:
Task linked: [QF-2908 Migrate Swagger -> OpenAPI and revise docs if they make sense and make sure all "hidden" params are documented](https://app.clickup.com/t/2192114/QF-2908)

suricactus:
Thanks for removing the hook trick, I think this one has less maintanance weight for the future.

BTW, can you also double check if all the endpoints are properly documented. This can be easily tested using the swagger UI sending requests. If there is a missing parameter, then the requests will fail. Sometimes there are hidden parameters, also have a look on them too.

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -125,6 +141,27 @@ def has_permission(self, request, view):
...
+    post=extend_schema(
+        description="Upload a file to the project",
+        parameters=[
+            OpenApiParameter(
+                "file",
+                OpenApiTypes.STR,
+                OpenApiParameter.QUERY,
+                required=True,
+                description="Path to the file to upload",
+            )
```
suricactus, line 160:
```suggestion
            OpenApiParameter(
                "file",
                OpenApiTypes.BINARY,
                OpenApiParameter.QUERY,
                required=True,
                description="File to be uploaded",
            )
```

suricactus: This is file, did you test actually making requests using the Swagger UI?

why-not-try-calmer: It's on the menu -- once you confirm the API docs suits you :)

suricactus (commented):
Code looks ok.

Something I forgot to mention is that /qfield-files endpoint is deprecated, so please hide it from the OpenAPI docs too.

The rest should be verified if requests from the Swagger UI work and fixes should be accomodated.

PR #751, opened 2023-07-24
Fix all usage of walrus operators in the code, Currently the flake8 i…

why-not-try-calmer:
…s complaining about it


duke-nyuki:
Task linked: [QF-2958 Fix all usage of walrus operators in the code](https://app.clickup.com/t/2192114/QF-2958)

PR #752, opened 2023-07-27
Make sure keep-alive is passed to upstream proxy servers

suricactus:


m-kuhn (approved):
I would propose to give that a go on staging asap (patch release)

PR #759, opened 2023-08-03
Adding request body as str to attach_keys middleware

why-not-try-calmer:
A few clarifications and observations:
- `DownloadPushDeleteFileView.post()` does not trigger the `parse()` method of its declared parser_class (MultiPartParser), because I did subclass this parser class in a previous version of this PR, and my override of `parse()` never triggered in my tests
- room for improvement: ensure that the intended `parse()` function is called, so that we can debug more from that context; this might open the door to reading the request.body only for requests that fail, and not all of them as is the case here.

### DIFF
--- a/docker-app/qfieldcloud/core/management/commands/extracts3data.py
+++ b/docker-app/qfieldcloud/core/management/commands/extracts3data.py
@@ -1,0 +1,129 @@
+import csv
+import logging
+from sys import stdout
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-2704 Debug `EmptyContentError: Empty content` when uploading files (log raw request from Django)](https://app.clickup.com/t/2192114/QF-2704)

PR #760, opened 2023-08-03
Sending raw body to sentry

why-not-try-calmer:
A few clarifications and observations:
- Actually this is fine => ~~`DownloadPushDeleteFileView.post()` does not trigger the `parse()` method of its declared parser_class (MultiPartParser), because I did subclass this parser class in a previous version of this PR, and my override of `parse()` never triggered in my tests~~
- This is done => ~~room for improvement: ensure that the intended `parse()` function is called, so that we can debug more from that context; this might open the door to reading the request.body only for requests that fail, and not all of them as is the case here.~~

### DIFF
--- a/docker-app/qfieldcloud/core/middleware/requests.py
+++ b/docker-app/qfieldcloud/core/middleware/requests.py
@@ -25,7 +25,7 @@ def attach_keys(get_response):
         request_attributes = {
             "file_key": str(request.FILES.keys()),
             "meta": str(request.META),
-            "files": request.FILES.getlist("file")
+            "files": request.FILES.getlist("file"),
         }
         request.attached_keys = str(request_attributes)
         response = get_response(request)
--- a/docker-app/qfieldcloud/core/middleware/requests.py
+++ b/docker-app/qfieldcloud/core/middleware/requests.py
@@ -1,3 +1,7 @@
+import io
+import shutil
+
+
 def attach_keys(get_response):
     """
     QF-2540
@@ -6,12 +10,25 @@ def attach_keys(get_response):
     """
 
     def middleware(request):
+        input_stream = io.BufferedReader(io.BytesIO(request.body))
+
         request_attributes = {
             "file_key": str(request.FILES.keys()),
             "meta": str(request.META),
         }
+
+        # only report rawbody of less-than-10MBs multipart requests
+        if (
+            "multipart/form-data;boundary" in request.headers["Content-Type"]
+            and input_stream.tell() < 10000000
+        ):
+            output_stream = io.BytesIO()
+            shutil.copyfileobj(input_stream, output_stream)
+            request.body_stream = output_stream
+
         if "file" in request.FILES.keys():
             request_attributes["files"] = request.FILES.getlist("file")
+
         request.attached_keys = str(request_attributes)
         response = get_response(request)
         return response
--- a/docker-app/qfieldcloud/core/middleware/requests.py
+++ b/docker-app/qfieldcloud/core/middleware/requests.py
@@ -10,7 +10,7 @@ def attach_keys(get_response):
     """
 
     def middleware(request):
-        input_stream = io.BufferedReader(io.BytesIO(request.body))
+        input_stream = io.BytesIO(request.body)
 
         request_attributes = {
             "file_key": str(request.FILES.keys()),
@@ -27,6 +27,7 @@ def attach_keys(get_response):
             and "multipart/form-data;boundary" in content_type
             and input_stream.tell() < 10000000
         ):
+            request_attributes["content_type"] = content_type
             output_stream = io.BytesIO()
             shutil.copyfileobj(input_stream, output_stream)
             request.body_stream = output_stream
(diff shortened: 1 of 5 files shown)


duke-nyuki:
Task linked: [QF-2704 Debug `EmptyContentError: Empty content` when uploading files (log raw request from Django)](https://app.clickup.com/t/2192114/QF-2704)

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -222,7 +222,9 @@ def post(self, request, projectid, filename, format=None):
                 pre_serialization=request.attached_keys,
                 post_serialization=str(request_attributes),
                 buffer=buffer,
-                body_stream=request.body_stream,
+                body_stream=request.body_stream
+                if hasattr(request, "body_stream")
+                else None,
```
suricactus, line 227:
```suggestion
                body_stream=getatttr(request, "body_stream", None)
```
?

why-not-try-calmer:
Raw body:
[request_id_file_name_rawbody.txt](https://github.com/opengisch/qfieldcloud/files/12289040/request_id_file_name_rawbody.txt)

docker-app/qfieldcloud/core/tests/test_sentry.py
```diff
@@ -42,7 +60,9 @@ def test_logging_with_sentry(self):
                 }
             ),
             "buffer": StringIO("The traceback of the exception to raise"),
-            "capture_message": True,  # so that sentry receives attachments even when there's no exception/event
+            "body_stream": self.body_stream,
+            "capture_message": True,  # so that sentry receives attachments even when there's no exception/event,
         }
         will_be_sent = report_serialization_diff_to_sentry(**mock_payload)
         self.assertTrue(will_be_sent)
+        print(f"Body sent to Sentry: {self.body_stream.getvalue()}")
```
suricactus, line 68: Do we need this print?

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -220,8 +237,10 @@ def post(self, request, projectid, filename, format=None):
                 # using the 'X-Request-Id' added to the request by RequestIDMiddleware
                 name=f"{request.META.get('X-Request-Id')}_{projectid}",
                 pre_serialization=request.attached_keys,
-                post_serialization=str(request_attributes),
+                post_serialization=",".join(QfcMultiPartSerializer.errors)
+                + str(request_attributes),
```
suricactus, line 241:
```suggestion
                post_serialization=(",".join(QfcMultiPartSerializer.errors)
                + str(request_attributes)),
```
Does black accept this? It's super confusing like this. -1 for formatters these days :)

suricactus (approved):
Looks good, two small comments and one idea we communicate in the chat.

docker-app/qfieldcloud/core/tests/test_sentry.py
why-not-try-calmer: We don't!

docker-app/qfieldcloud/core/views/files_views.py
suricactus: Done

suricactus:
Not sure why the changes replaced in 989c7b66db65f09ff9ad17dc400aef8548922e9c was existing?

docker-app/qfieldcloud/core/middleware/requests.py
```diff
@@ -1,17 +1,35 @@
...
+            request.method == "POST"
+            and "Content-Length" in request.headers
+            and int(request.headers["Content-Length"])
+            < config.SENTRY_REQUEST_MAX_SIZE_TO_SEND
+        ):
+            input_stream = io.BytesIO(request.body)
+            output_stream = io.BytesIO()
+            shutil.copyfileobj(input_stream, output_stream)
+            request.body_stream = output_stream
+
```
faebebin, line 27: No idea whether this is possible:
But if sentry_sdk could be made available already here, couldn't we somehow check for some conditions and report raw_body already here:

```
if "file" not in request.data or not request.FILES.getlist("file"):
    sentry_sdk.capture_message(output_stream)
```
?

suricactus: We have not parsed the body yet, so we don't have `request.data`.

docker-app/qfieldcloud/core/middleware/requests.py
```diff
@@ -1,17 +1,35 @@
...
         request_attributes = {
             "file_key": str(request.FILES.keys()),
             "meta": str(request.META),
+            "files": request.FILES.getlist("file"),
         }
-        if "file" in request.FILES.keys():
-            request_attributes["files"] = request.FILES.getlist("file")
         request.attached_keys = str(request_attributes)
         response = get_response(request)
```
suricactus, line 34: What if we surround the `get_response` with try/catch here? Can we catch `EmptyContentError` here?

suricactus: NVM, discussed in chat already.

suricactus:
@why-not-try-calmer all works fine, seems it was a misconfiguration from my side. Allowed myself a final change before merging.

Thanks for sorting this out!

PR #761, opened 2023-08-03
Make `AuthenticationFailed` errors just that instead of using `ValidationError`

suricactus:
Makes QFieldSync nicer:

| before | after |
|--|--|
| image | image |

### DIFF
--- a/docker-app/qfieldcloud/authentication/serializers.py
+++ b/docker-app/qfieldcloud/authentication/serializers.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.contrib.auth import authenticate, get_user_model
 from django.utils.translation import gettext as _
 from rest_framework import serializers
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-3008 Make `AuthenticationFailed` errors just that instead of using `ValidationError`](https://app.clickup.com/t/2192114/QF-3008)

faebebin (approved):
LGTM

Tried it, cool :+1:

why-not-try-calmer:
This is a very minor point, but I think the UX would be better if the message read

> Authentication failed

instead of

> [authentication_failed] Authentication failed

suricactus:
@why-not-try-calmer It looks more scary indeed. However, showing the code helps a lot for internationalized UI. For me it's very hard to read the error on a screenshot if it is translated to chinese, but the code is something that is stable across translations.

PR #762, opened 2023-08-10
Add typing.py

faebebin:
Not sure about conventions of naming and where to put it


m-kuhn (approved):
I don't think we have bezt practices on this so far.
Let's put it here until someone sees a reason to move it.

faebebin:
No desired by the team. Preferably use `explicit stdlib types [int, str, List, ...] + sensible variable names`

PR #764, opened 2023-08-10
Inconsistent organization member lists, return `Person` instead of `OrganizationMember`

why-not-try-calmer:
Needed by https://github.com/opengisch/qfieldcloud-private/pull/460.

### DIFF
--- a/docker-app/qfieldcloud/core/querysets_utils.py
+++ b/docker-app/qfieldcloud/core/querysets_utils.py
@@ -25,20 +25,14 @@ def get_team_members(team):
     return TeamMember.objects.filter(team=team)
 
 
-def get_organization_members(organization):
-    return (
-        Person.objects.filter(organizationmember__organization=organization)
-        .annotate(
-            role=F("organizationmember__role"),
-            # organization=F("organizationmember__organization"),
-        )
-        .union(
-            Person.objects.filter(pk=organization.organization_owner.id).annotate(
-                role=V("owner"),
-                # organization=V(organization),
-            )
-        )
+def get_organization_members(organization) -> QuerySet[Person]:
+    non_owning_members = Person.objects.filter(
+        organizationmember__organization=organization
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-2365 Member list different depending where you are](https://app.clickup.com/t/2192114/QF-2365)

suricactus (approved):

docker-app/qfieldcloud/core/querysets_utils.py
```diff
@@ -24,8 +25,14 @@ def get_team_members(team):
...
-def get_organization_members(organization):
-    return OrganizationMember.objects.filter(organization=organization)
+def get_organization_members(organization) -> QuerySet[Person]:
+    non_owning_members = Person.objects.filter(
+        organizationmember__organization=organization
+    ).annotate(role=F("organizationmember__role"))
+    owner = Person.objects.filter(pk=organization.organization_owner.id).annotate(
+        role=V("owner")
+    )
+    return non_owning_members.union(owner)
```
suricactus, line 35:
```suggestion
    org_members = Person.objects.filter(
        organizationmember__organization=organization
    ).annotate(role=F("organizationmember__role"))
    owner = Person.objects.filter(pk=organization.organization_owner.id).annotate(
        role=V("owner")
    )
    return org_members.union(owner)
```

if they are members they are automatically non-owning?

why-not-try-calmer: I see there are two competing mental models here, and indeed acccording to the Django models it's not a tautology to say that owners are members, whereas on my mental model, it is a tautology. This is why for me the significant distinction is between non-owning members and owning members, since in principle it could be possible (if not right now, perhaps in a future version of Qfc) that the same organization has multiple owners.

Let me propose a middleground.

suricactus:
Thanks!

PR #767, opened 2023-08-15
Cleanup @deprecated code that affect only tests

faebebin:
Remove @deprecated objects that are referenced only in tests or nowhere.

Should remove some noise / unused code.


why-not-try-calmer (changes requested):
I'd suggest using a bytes_to_mb function instead of writing `* 1000 * 1000`  to make refactoring in the future a bit easier.

why-not-try-calmer (approved):

PR #768, opened 2023-08-16
Fix settings link in notifications email

faebebin:


duke-nyuki:
Task linked: [QF-3047 Notifications email settings link invalid url](https://app.clickup.com/t/2192114/QF-3047)

docker-app/qfieldcloud/notifs/cron.py
```diff
@@ -43,12 +44,14 @@ def do(self):
                     logging.warning(f"{user} has notifications, but no email set !")
                     continue
 
+                QFIELDCLOUD_HOST = os.environ.get("QFIELDCLOUD_HOST", None)
```
why-not-try-calmer, line 47:
```suggestion
                QFIELDCLOUD_HOST = os.environ.get("QFIELDCLOUD_HOST")
```
The form you used is already built-in.

why-not-try-calmer (changes requested):

docker-app/qfieldcloud/notifs/cron.py
faebebin: thanks

suricactus: Don't you have access to `settings.QFIELDCLOUD_HOST`? Ideally random reading from the env should be discouraged outside `settings.py`?

why-not-try-calmer (approved):

PR #771, opened 2023-08-17
Slight readability improvement on etag preparation for files

why-not-try-calmer:
As of current `master` the `e_tag` property of `S3ObjectVersion` will evaluate to `None` in certain edge cases, causing Exceptions (see for example https://github.com/opengisch/qfieldcloud-private/pull/474). This PR ensures `S3ObjectVersion.e_tag` never lies about its type.

### DIFF
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -83,7 +83,7 @@ class ListFilesView(views.APIView):
             path = PurePath(version.key)
             filename = str(path.relative_to(*path.parts[:3]))
             last_modified = version.last_modified.strftime("%d.%m.%Y %H:%M:%S %Z")
-            md5sum = version.e_tag.replace('"', "")
+            md5sum = version.e_tag.replace('"', "") if version.e_tag is not None else ""
 
             version_data = {
                 "size": version.size,
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -83,7 +83,11 @@ class ListFilesView(views.APIView):
             path = PurePath(version.key)
             filename = str(path.relative_to(*path.parts[:3]))
             last_modified = version.last_modified.strftime("%d.%m.%Y %H:%M:%S %Z")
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -83,10 +83,11 @@ def get(self, request: Request, projectid: str) -> Response:
             path = PurePath(version.key)
             filename = str(path.relative_to(*path.parts[:3]))
             last_modified = version.last_modified.strftime("%d.%m.%Y %H:%M:%S %Z")
+            md5sum = version.e_tag.replace('"', "") if version.e_tag is not None else ""
```
suricactus, line 86: Should be multiline `if` as in `core/utils.py`.

suricactus (changes requested):
Tried to find an explanation of a situation when the `e_tag` is `None`, but didn't find where it is. Would be interesting to share what is the situation when the `e_tag` is `None` and put it as a comment in the code.

IMO if such situation is real, we better return `None` than `""`, as we are lying whether the `e_tag` is present. So instead of adding the two `if`s, it seems changing the return type makes more sense to me. 

The above thoughts might be wrong when I see the reason  when this happens.

why-not-try-calmer:
> Tried to find an explanation of a situation when the `e_tag` is `None`, but didn't find where it is. Would be interesting to share what is the situation when the `e_tag` is `None` and put it as a comment in the code.
> 
> IMO if such situation is real, we better return `None` than `""`, as we are lying whether the `e_tag` is present. So instead of adding the two `if`s, it seems changing the return type makes more sense to me.
> 
> The above thoughts might be wrong when I see the reason when this happens.

My bad, I narrowed the issue down to its actual culprit.

suricactus:
@why-not-try-calmer I am still not sure when the `version.etag` is `None`? It only happens when we have a delete version, which should not be the case for any installation?

why-not-try-calmer:
Okay so thanks for showing how this in itself was not a bug. I've removed the `bug` label and this now boils down a simple convenience PR.

suricactus (approved):
Thanks!

It turned out very strange behaviour that confirmed my suspision - you had a version about deletion of the file.

PR #773, opened 2023-08-18
Cannot add organization owner as collaborator to project

why-not-try-calmer:
Evidence that this works:
Capture d’écran du 2023-08-18 08-37-34

Capture d’écran du 2023-08-18 08-37-03

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1478,7 +1478,12 @@ class ProjectCollaborator(models.Model):
             if self.collaborator.is_person:
                 members_qs = organization.members.filter(member=self.collaborator)
 
-                if not members_qs.exists():
+                # for organizations-owned projects, the candidate collaborator
+                # must be a member of the organization or the organization's owner
+                if not (
+                    members_qs.exists()
+                    or self.collaborator == organization.organization_owner
+                ):
                     raise ValidationError(
                         _(
                             "Cannot add a user who is not a member of the organization as a project collaborator."


duke-nyuki:
Task linked: [QF-2646 Adding the org owner as collaborator gives error 500 using the web UI](https://app.clickup.com/t/2192114/QF-2646)

suricactus (approved):
Thanks

docker-app/qfieldcloud/core/models.py
```diff
@@ -1478,7 +1478,12 @@ def clean(self) -> None:
             if self.collaborator.is_person:
                 members_qs = organization.members.filter(member=self.collaborator)
 
-                if not members_qs.exists():
+                # for organizations-owned projects, the candidate collaborator
+                # must be a member of the organization or the organization's owner
```
suricactus, line 1482:
```suggestion
                # for organizations-owned projects, the candidate collaborator
                # must be a member of the organization or and not organization's owner
```

suricactus: @why-not-try-calmer isn't it more correct now?

why-not-try-calmer: As of this PR, the `if` above will **not** lead to a ValidationError if the candidate collaborator is the organization's owner.

suricactus: Ok, I see, I misread the `not` in front of the condition.

suricactus (approved):

PR #778, opened 2023-08-24
Type hints in worker_wrapper

why-not-try-calmer:
While working on a recent task I cleaned up some type hints.


suricactus (approved):
Thanks!

Do you think this is ready to be `mypy`-ied? Would be nice to add more and more files to be automatically checked.

PR #780, opened 2023-08-25
Do not throw when creating a project on inactive subscription

suricactus:


faebebin (approved):
LGTM

Thanks for adapting the tests - IMO really helpful for future clarity.

PR #781, opened 2023-08-25
When project is deleted, it's files are not permanently deleted

suricactus:
See https://app.clickup.com/t/866av28j2


faebebin (approved):
LGTM

Just deleted a project with files locally without problem. I did not check for permanent deletion though but guess it works

PR #782, opened 2023-08-28
Improving project file validation user experience

why-not-try-calmer:
…fix through not re-enconding file's contents.

This will almost certainly _not_ be enough to fix the issue, but it can improve on the situation because it avoids one decoding ("fromstring"). To fix the issue I need to identity the source of the "corruption".

- [x] grabbing context around unparse-able contents
- [x] bubbling up feedback to user

### DIFF
--- a/docker-qgis/process_projectfile.py
+++ b/docker-qgis/process_projectfile.py
@@ -1,8 +1,5 @@
-import html
-import io
 import logging
 from pathlib import Path
-from typing import NamedTuple, Optional
 from xml.etree import ElementTree
 
 from qfieldcloud.qgis.utils import (
@@ -11,6 +8,7 @@ from qfieldcloud.qgis.utils import (
     InvalidXmlFileException,
     ProjectFileNotFoundException,
     get_layers_data,
+    get_qgis_xml_error_context,
     layers_data_to_string,
 )
 from qgis.core import QgsMapRendererParallelJob, QgsMapSettings, QgsProject
@@ -18,50 +16,6 @@ from qgis.PyQt.QtCore import QEventLoop, QSize
 from qgis.PyQt.QtGui import QColor
 
 logger = logging.getLogger("PROCPRJ")
-
-
-class XmlErrorLocation(NamedTuple):
-    line: int
-    column: int
-
-
-def get_qgis_xml_error_location(
-    invalid_token_error_msg: str,
-) -> Optional[XmlErrorLocation]:
-    """Get column and line numbers from the provided error message."""
-    if "invalid token" not in invalid_token_error_msg.casefold():
-        logger.error("Unable to find 'invalid token' details in the given message")
-        return None
-
-    _, details = invalid_token_error_msg.split(":")
-    line, column = details.split(",")
-    _, line_number = line.strip().split(" ")
-    _, column_number = column.strip().split(" ")
-
-    return XmlErrorLocation(int(line_number), int(column_number))
-
-
-def get_qgis_xml_error_context(
-    invalid_token_error_msg: str, fh: io.BufferedReader
-) -> Optional[tuple[str, str, str]]:
-    """Get a slice of the line where the exception occurred, with all faulty occurrences sanitized."""
-    location = get_qgis_xml_error_location(invalid_token_error_msg)
-    if location:
-        substitute = "?"
-        fh.seek(0)
-        for cursor_pos, line in enumerate(fh, start=1):
-            if location.line == cursor_pos:
-                faulty_char = line[location.column]
-                suffix_slice = line[: location.column - 1]
-                clean_safe_slice = suffix_slice.decode("utf-8").strip() + substitute
-
-                return (
-                    f"Unable to parse character: {repr(faulty_char)}",
-                    f"Replaced by '{substitute}' on line {location.line} that starts with:",
-                    html.escape(clean_safe_slice),
-                )
-
-    return None
 
 
 def check_valid_project_file(project_filename: Path) -> None:
--- a/docker-qgis/process_projectfile.py
+++ b/docker-qgis/process_projectfile.py
@@ -40,7 +40,7 @@ def get_location(invalid_token_error_msg: str) -> Optional[XmlLocationError]:
 
 def contextualize(
     invalid_token_error_msg: str, fh: io.BufferedReader
-) -> Optional[tuple[str]]:
+) -> Optional[tuple[str, str, str]]:
     """
     Get a sanitized slice of the line where the exception occurred, with all faulty occurrences sanitized.
     Returns the string as a 3-substring tuple to avoid tripping Docker Compose's stdout limitations.
--- a/docker-qgis/process_projectfile.py
+++ b/docker-qgis/process_projectfile.py
@@ -57,8 +57,8 @@ def contextualize(
                 clean_safe_slice = suffix_slice.decode("utf-8").strip() + substitute
 
                 return (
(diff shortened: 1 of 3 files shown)


why-not-try-calmer:
image

Still need to work on the line not rendering properly.

why-not-try-calmer:
__What is written to the database:__
image

__What is rendered to the user:__
image

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,61 @@
...
+    layers_data_to_string,
+)
 from qgis.core import QgsMapRendererParallelJob, QgsMapSettings, QgsProject
 from qgis.PyQt.QtCore import QEventLoop, QSize
 from qgis.PyQt.QtGui import QColor
 
 logger = logging.getLogger("PROCPRJ")
 
 
-class ProjectFileNotFoundException(BaseException):
```
why-not-try-calmer, line 14: Exceptions were moved to utils.py to be able to catch them there without incurring any circular dependency.

docker-qgis/process_projectfile.py
```diff
@@ -44,13 +65,16 @@ def check_valid_project_file(project_filename: Path) -> None:
...
-        try:
-            with open(project_filename) as f:
-                ElementTree.fromstring(f.read())
-        except ElementTree.ParseError as err:
-            raise InvalidXmlFileException(
-                project_filename=project_filename, xml_error=err
-            )
+        with open(project_filename, "rb") as fh:
+            try:
+                for event, elem in ElementTree.iterparse(fh):
```
why-not-try-calmer, line 70: `iterparse` parses in a piece-meal way, thus using a bit less memory.

why-not-try-calmer:
image

Thanks to @suricactus for finding why my implementation wasn't able to render properly!

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
 
-class InvalidFileExtensionException(BaseException):
-    message = (
-        'Project file "%(project_filename)s" has unknown file extension "%(extension)s"'
-    )
-
-
-class InvalidXmlFileException(BaseException):
-    message = "Project file is an invalid XML document:\n%(xml_error)s"
+def get_location(invalid_token_error_msg: str) -> Optional[XmlErrorLocation]:
```
suricactus, line 28: More of a `get_qgis_xml_error_location`?

suricactus, line 28: More of a `get_qgis_xml_error_location`?

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
 
+                return (
+                    f"Unable to parse this character: {repr(faulty_char)}",
+                    f"It was replaced by '{substitute}' on line {location.line} that starts with:",
+                    html.escape(clean_safe_slice),
+                )
 
-class FailedThumbnailGenerationException(BaseException):
-    message = "Failed to generate project thumbnail:\n%(reason)s"
+    return None
```
suricactus, line 65: Feels like all of these should go to `utils.py`?

suricactus, line 65: Feels like all of these should go to `utils.py`?

suricactus, line 65: Feels like all of these should go to `utils.py`?

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
+        fh.seek(0)
+        for cursor_pos, line in enumerate(fh, start=1):
+            if location.line == cursor_pos:
+                faulty_char = line[location.column]
+                suffix_slice = line[: location.column - 1]
+                clean_safe_slice = suffix_slice.decode("utf-8").strip() + substitute
 
+                return (
+                    f"Unable to parse this character: {repr(faulty_char)}",
+                    f"It was replaced by '{substitute}' on line {location.line} that starts with:",
```
suricactus, line 61:
```suggestion
                    f"Unable to parse character: {repr(faulty_char)}",
                    f"Replaced by '{substitute}' on line {location.line} that starts with:",
```

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
+                    html.escape(clean_safe_slice),
```
suricactus, line 62: The escape should not be here. The HTML injection is a visualization problem, not a storage problem.

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
+    _, column_number = column.strip().split(" ")
 
-class InvalidQgisFileException(BaseException):
-    message = 'Project file "%(project_filename)s" is invalid QGIS file:\n%(error)s'
+    return XmlErrorLocation(int(line_number), int(column_number))
 
 
-class InvalidLayersException(BaseException):
-    message = 'Project file "%(project_filename)s" contains invalid layers'
+def contextualize(
```
suricactus, line 42:
```
def get_qgis_xml_error_context(
```
?

docker-qgis/process_projectfile.py
```diff
@@ -1,40 +1,68 @@
...
+    invalid_token_error_msg: str, fh: io.BufferedReader
+) -> Optional[tuple[str, str, str]]:
+    """
+    Get an html-safe slice of the line where the exception occurred, with all faulty occurrences sanitized.
```
suricactus, line 46:
```
    Get a slice of the line where the exception occurred, with all faulty occurrences sanitized.
```

docker-qgis/utils.py
```diff
@@ -366,18 +404,6 @@ def eval(self, root: Path) -> Path:
         return path
 
 
-class BaseException(Exception):
```
suricactus, line 369: Just double check that `BaseException` is no longer needed.

suricactus (changes requested):
Thanks a lot, you went quite deep to fix this issue! 

Mostly comments on reordering and renaming stuff.

One note - the HTML escaping should not be here, as the semi-HTML data itself is ok to be stored. The problem is when it is visualized, the contents is not escaped.

suricactus (changes requested):
Thanks a lot, you went quite deep to fix this issue! 

Mostly comments on reordering and renaming stuff.

One note - the HTML escaping should not be here, as the semi-HTML data itself is ok to be stored. The problem is when it is visualized, the contents is not escaped.

suricactus (changes requested):
Thanks a lot, you went quite deep to fix this issue! 

Mostly comments on reordering and renaming stuff.

One note - the HTML escaping should not be here, as the semi-HTML data itself is ok to be stored. The problem is when it is visualized, the contents is not escaped.

docker-qgis/utils.py
why-not-try-calmer: Renamed to "QfcException" already.

docker-qgis/process_projectfile.py
suricactus: I do agree it is a quick solution, but I think it is quite dirty. Since we do direct queries in jobs to search for errors in the database, escaping the output in such a early stage will make our task much harder, as we have to remember the specific HTML escaping.

What is more, what if we want to put the outputs in CSV one day, should we already escape? Or JSON, yaml etc etc.

why-not-try-calmer:
Okay, I see the end of it :)
image

suricactus (approved):
Nice!

suricactus:
I like less and less code <3 . Thanks for your time!

PR #783, opened 2023-08-29
[WIP] #QF-3103 Use pg_advisory_locks against deadlocks and overlaps

faebebin:
I suspect locking race conditions to be the cause of deadlocks and overlapping jobs.

Using advisory_locks on blocks of code seems more solid than relying on `select_for_update`. I experimented locally with two workers doing update requests and timeouts. It seemed the locking did not work regardless the PG isolation level. Instead, using `SELECT pg_advisory_lock(lock_id)` worked as expected.

I guess it would not impact performance significantly as dequeueing jobs is not the bottleneck currently.

TODO take care of unlocking advisory locks in case of blackout.

#QF-3103

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/jobs.py
+++ b/docker-app/qfieldcloud/core/utils2/jobs.py
@@ -5,9 +5,13 @@ import qfieldcloud.core.models as models
 from django.conf import settings
 from django.db import transaction
 from django.db.models import Q
+from django_pglocks import advisory_lock
 from qfieldcloud.core import exceptions
 
 logger = logging.getLogger(__name__)
+
+# Any number
+DELTAS_APPLY_LOCK_ID = 1001
 
 
 @transaction.atomic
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-2664 Investigate Job error: deadlocks in it](https://app.clickup.com/t/2192114/QF-2664)

docker-app/qfieldcloud/core/utils2/jobs.py
```diff
@@ -28,71 +29,78 @@ def apply_deltas(
...
-    # 5. Check whether there are jobs found in the queue and exclude all deltas that are part of any pending job.
-    if apply_jobs.exists():
-        pending_deltas = pending_deltas.exclude(jobs_to_apply__in=apply_jobs)
-
-    # 6. If there are no pending deltas, do not create a new job and return.
-    deltas_count = pending_deltas.count()
-    if deltas_count == 0:
-        return []
+    with advisory_lock(project.id):
+        # Ensure there is no concurrent request trying to apply these deltas.
```
faebebin, line 33: NOTE no code modified in the block body except removing the `select_for_update` .

faebebin:
Close it because found better solutions. Further this would need more research and understanding to handle locks securely.

PR #785, opened 2023-08-31
Add a way to easily export deltafiles from Django Admin - Deltas list and Job view

suricactus:
image

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -1,6 +1,7 @@
 import csv
 import json
 import time
+import uuid
 from collections import namedtuple
 from datetime import datetime
 from itertools import chain
@@ -23,7 +24,7 @@ from django.db.models import Q, QuerySet
 from django.db.models.fields.json import JSONField
 from django.db.models.functions import Lower
 from django.forms import ModelForm, fields, widgets
-from django.http import HttpRequest
+from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.http.response import Http404, HttpResponseRedirect, StreamingHttpResponse
 from django.shortcuts import resolve_url
 from django.template.response import TemplateResponse
@@ -54,7 +55,7 @@ from qfieldcloud.core.models import (
 )
 from qfieldcloud.core.paginators import LargeTablePaginator
 from qfieldcloud.core.templatetags.filters import filesizeformat10
-from qfieldcloud.core.utils2 import jobs
+from qfieldcloud.core.utils2 import delta_utils, jobs
 from rest_framework.authtoken.models import TokenProxy
 
 admin.site.unregister(LogEntry)
@@ -787,6 +788,17 @@ class JobAdmin(QFieldCloudModelAdmin):
 
         return inlines
 
+    def get_urls(self):
+        urls = super().get_urls()
+        return [
+            path(
+                "<uuid:apply_job_id>/export-deltafile",
+                self.admin_site.admin_view(self.export_applyjob_deltafile),
+                name="export_applyjob_deltafile",
+            ),
+            *urls,
+        ]
+
     def error_type(self, instance):
         if instance.feedback and "error_type" in instance.feedback:
             return f"{instance.feedback['error_type']}".strip()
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-3110 Add a way to easily export deltafiles from ApplyJob or Deltas list views](https://app.clickup.com/t/2192114/QF-3110)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -982,29 +1016,54 @@ def set_status_unpermitted(self, request, queryset):
...
+        self.message_user(request, "Delta application started")
+
+        # we need to sleep 1 second, just to make sure the apply delta started
+        time.sleep(1)
+
+        return HttpResponseRedirect(".")
+
+    @admin.action(description="Download deltas as deltafile")
+    def download_deltafile(self, request, queryset: QuerySet) -> HttpResponse:
+        if queryset.values_list("project_id").distinct().count() != 1:
```
faebebin, line 1047: Throws for me: I tried it with some existing Delta's though
image

docker-app/qfieldcloud/core/admin.py
```diff
@@ -982,29 +1016,54 @@ def set_status_unpermitted(self, request, queryset):
...
+            messages.error(request, "Selected deltas must be part of the same project!")
```
faebebin, line 1048: It would be nice to have this information upfront, like in the action title `Download project deltas as file`. But I guess is hard to provide in a concise way. Enabling cross project downloads would be undesired or hard to implement?

docker-app/qfieldcloud/core/templates/admin/core/job/change_form.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'admin/change_form.html' %}
+{% load i18n %}
+
+{% block object-tools-items %}
+  {{ block.super }}
+  {% if original.type == 'delta_apply' %}
+  <li>
+    <a href="{% url 'admin:export_applyjob_deltafile' original.pk %}" class="historylink">
+      {% trans 'Download deltafile' %}
```
faebebin, line 9: That one works nicely :+1:

docker-app/qfieldcloud/core/admin.py
```diff
@@ -787,6 +788,17 @@ def get_inlines(self, request, obj=None):
...
+        urls = super().get_urls()
+        return [
+            path(
+                "/<uuid:apply_job_id>/export-deltafile",
+                self.admin_site.admin_view(self.export_applyjob_deltafile),
+                name="export_applyjob_deltafile",
+            ),
+            *urls,
+        ]
+
```
faebebin, line 801: Maybe we could somehow declare a specific `class ApplyJobAdmin(JobAdmin)` which adds this functionality and also the two previous methods where we have to check for `if obj and obj.type == Job.Type.DELTA_APPLY` and still show them in one job-list-view? Would be nice but not necessary if hard to  implement.

faebebin (changes requested):
Nice functionality :+1: 

For me downloading from Delta list view did not work.

Besides this I came up with some nice to have ideas ...

docker-app/qfieldcloud/core/admin.py
suricactus: Been there before, it's harder to maintain when we have only 3 job types. I would prefer to keep it as it is for now.

docker-app/qfieldcloud/core/admin.py
suricactus: Thanks! Changing the code before pushing is never a smart idea :)

docker-app/qfieldcloud/core/admin.py
suricactus: Cross project downloads are undesired. Keep in mind this is more for us rather than support or anyone else. So no need to be extra explanatory.

suricactus:
@faebebin allowed myself to forcepush to keep the history short. Anyways, I changed two lines to make it work now. Thanks for testing!

docker-app/qfieldcloud/core/admin.py
faebebin: sure - I actually thought you might have considered this before ;)

PR #786, opened 2023-08-31
Mypy app-wide scope

why-not-try-calmer:
This extends the scope of the static type-checker to also encompass docker-app and docker-redis.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/utils.py
+++ b/docker-app/qfieldcloud/core/tests/utils.py
@@ -51,16 +51,11 @@ def setup_subscription_plans():
 
 
 def set_subscription(
-    some_user: User | Iterable[User],
+    users: User | Iterable[User],
     code: str = None,
     **kwargs,
 ):
-    users: list[User] = []
-
-    if isinstance(some_user, User):
-        users.append(some_user)
-    else:
-        users += some_user
+    users: list[User] = [users] if isinstance(users, User) else users
     assert len(
         users
     ), "When iterable, the first argument must contain at least 1 element."
--- a/docker-app/qfieldcloud/core/tests/utils.py
+++ b/docker-app/qfieldcloud/core/tests/utils.py
@@ -2,7 +2,7 @@ import io
 import os
 from datetime import timedelta
 from time import sleep
-from typing import IO, Iterable, Union
+from typing import IO, Iterable
 
 from django.utils import timezone
 from qfieldcloud.core.models import Job, Project, User
@@ -51,13 +51,16 @@ def setup_subscription_plans():
 
 
 def set_subscription(
-    users: Union[User, Iterable[User]],
+    some_user: User | Iterable[User],
     code: str = None,
     **kwargs,
 ):
-    if isinstance(users, User):
-        users = [users]
+    users: list[User] = []
 
+    if isinstance(some_user, User):
+        users.append(some_user)
+    else:
+        users += some_user
     assert len(
         users
     ), "When iterable, the first argument must contain at least 1 element."
--- a/docker-app/qfieldcloud/core/utils2/projects.py
+++ b/docker-app/qfieldcloud/core/utils2/projects.py
@@ -53,7 +53,7 @@ def create_collaborator(
 
 def create_collaborator_by_username_or_email(
     project: Project, username: str, created_by: Person
-):
+) -> tuple[bool, str]:
     """Creates a new collaborator (qfieldcloud.core.ProjectCollaborator) if possible
 
     Args:
@@ -62,7 +62,7 @@ def create_collaborator_by_username_or_email(
         created_by (Person): the user that initiated the collaborator creation
 
     Returns:
-        Tuple[bool, str]: success, message - whether the collaborator creation was success and explanation message of the outcome
+        tuple[bool, str]: success, message - whether the collaborator creation was success and explanation message of the outcome
     """
     success, message = False, ""
     users = list(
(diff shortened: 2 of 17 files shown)


docker-app/qfieldcloud/core/tests/utils.py
```diff
@@ -51,13 +51,16 @@ def setup_subscription_plans():
...
 ):
-    if isinstance(users, User):
-        users = [users]
+    users: list[User] = []
 
+    if isinstance(some_user, User):
+        users.append(some_user)
+    else:
+        users += some_user
     assert len(
```
faebebin, line 64: could it be:

users: list[User] = [some_user] if isintance(some_user, User) else some_users

BTW i would rather keep the var name `users` because it is simpler?

faebebin (changes requested):
LGTM 

a couple of small proposals

docker-app/qfieldcloud/core/tests/utils.py
why-not-try-calmer: Yes that works too!

why-not-try-calmer:
> LGTM
> 
> a couple of small proposals

Thanks, ready for me

faebebin (approved):

docker-app/qfieldcloud/core/utils2/projects.py
```diff
@@ -81,10 +81,8 @@ def create_collaborator_by_username_or_email(
             username
         )
     elif users[0].is_organization:
-        message = (
-            _(
-                'Organization "{}" cannot be added. Only users and teams can be collaborators.'
-            ).format(username),
+        message = _(
+            f'Organization "{username}" cannot be added. Only users and teams can be collaborators.'
```
suricactus, line 85: @faebebin @why-not-try-calmer 

I have already commented this in another place, this replacing `"{}".format(1)` with `f"{1}"` is causing troubles for potential translations.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -794,7 +794,7 @@ def create_subscription(
         plan: Plan,
         created_by: Person,
         active_since: Optional[datetime] = None,
-    ) -> Tuple["Subscription", "Subscription"]:
+    ) -> tuple[Optional["AbstractSubscription"], "AbstractSubscription"]:
```
suricactus, line 797: Why don't you use the definition from the docs: `tuple[AbstractSubscription | None, AbstractSubscription]` ?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -794,7 +794,7 @@ def create_subscription(
         plan: Plan,
         created_by: Person,
         active_since: Optional[datetime] = None,
```
suricactus, line 796: This one should be `datetime | None` as all other places, no?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -804,14 +804,16 @@ def create_subscription(
...
+            tuple[AbstractSubscription | None, AbstractSubscription]: the created trial subscription if the given plan was a trial and the regular subscription.
 
         TODO Python 3.11 the actual return type is Self
         """
         if active_since:
             # remove microseconds as there will be slight shift with the remote system data
             active_since = active_since.replace(microsecond=0)
 
+        assert active_since
+
```
suricactus, line 816: @why-not-try-calmer @faebebin 

This assert is actually breaking the system, as we have situations where we DO allow no active since to be passed. You can clearly see that from the typing of the function argument.

suricactus (commented):
Hey, @why-not-try-calmer @faebebin . Thanks for the contribution and reviews.

There are a couple of problems that from this PR, the worst of them in [docker-app/qfieldcloud/subscription/models.py](https://github.com/opengisch/qfieldcloud/pull/786/files#diff-b70d2704585f8f83a71fbed4cde22e403c9bb41dba7516233f4f63f37245bca2) line 815 -> the assert there is changing the behaviour of the app, therefore breaking it.

I caught this error when trying to integrate this PR changes with web UI.

I would suggest for such quite fundamental changes we have a double review, cool down period and opening a PR on the web ui to verify the changes are working.

See https://github.com/opengisch/qfieldcloud/pull/791 for the fixes.

docker-app/qfieldcloud/core/utils2/projects.py
suricactus: Fixed.

docker-app/qfieldcloud/subscription/models.py
suricactus: Fixed.

docker-app/qfieldcloud/subscription/models.py
suricactus: Fixed.

docker-app/qfieldcloud/subscription/models.py
suricactus: Fixed.

why-not-try-calmer:
> Hey, @why-not-try-calmer @faebebin . Thanks for the contribution and reviews.
> 
> There are a couple of problems that from this PR, the worst of them in [docker-app/qfieldcloud/subscription/models.py](https://github.com/opengisch/qfieldcloud/pull/786/files#diff-b70d2704585f8f83a71fbed4cde22e403c9bb41dba7516233f4f63f37245bca2) line 815 -> the assert there is changing the behaviour of the app, therefore breaking it.
> 
> I caught this error when trying to integrate this PR changes with web UI.
> 
> I would suggest for such quite fundamental changes we have a double review, cool down period and opening a PR on the web ui to verify the changes are working.
> 
> See #791 for the fixes.

I merged this quite early after only 1 approval so it's on me. I think it makes sense indeed to let you / the project owner do the merges when anything non-trivial, especially when several pending PRs can impact each other.

docker-app/qfieldcloud/subscription/models.py
faebebin: thanks for fixing :+1:

PR #791, opened 2023-09-08
Fix incorrect assert for `active_since` when creating a `Subscription`

suricactus:
- Fixes an issue introduced by https://github.com/opengisch/qfieldcloud/pull/786/files.
- Fixes misleading docs in Authentication backend.
- Remove all `List, Dict, Tuple, Optional, Union` from docker-app

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -2,7 +2,7 @@ import logging
 import uuid
 from datetime import datetime, timedelta
 from functools import lru_cache
-from typing import Optional, TypedDict, cast
+from typing import Type, TypedDict, cast
 
 from constance import config
 from deprecated import deprecated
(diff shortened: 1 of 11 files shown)


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -794,15 +794,15 @@ def create_subscription(
         account: UserAccount,
         plan: Plan,
         created_by: Person,
-        active_since: Optional[datetime] = None,
-    ) -> tuple[Optional["AbstractSubscription"], "AbstractSubscription"]:
+        active_since: datetime | None = None,
+    ) -> tuple[Type["AbstractSubscription"] | None, "AbstractSubscription"]:
```
why-not-try-calmer, line 798:
```suggestion
    ) -> tuple[Type["AbstractSubscription"] | None, "AbstractSubscription"]:
```
I am not familiar with Type. Is it required? tuple["MyClass"] doesn't work?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -813,8 +813,7 @@ def create_subscription(
             # remove microseconds as there will be slight shift with the remote system data
             active_since = active_since.replace(microsecond=0)
 
-        assert active_since
-
+        regular_active_since: datetime | None = None
```
why-not-try-calmer, line 816: Ok!

docker-app/qfieldcloud/subscription/models.py
suricactus: Unfortunately, Python is a bit stupid in ignoring types defined as strings... It tried to run the `|` operators bettween `str` and `None` types instead of ignoring it as type definition, which lead to the following error:
```
  File "/usr/src/app/qfieldcloud/subscription/models.py", line 798, in AbstractSubscription
    ) -> tuple["AbstractSubscription" | None, "AbstractSubscription"]:
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

I tried so, had to force push to fix that.

suricactus: A good question is whether to use `Type` every time we use string types, or only on demand. I have no strong opionion.

Assuming we want to not use `Optional`, we do have two paths:

Without explicit `Type` it is easier for the developer to type, but changing a type to `Optional` will require to add the `Type` around the string and a bit cryptic error.

why-not-try-calmer (approved):

PR #794, opened 2023-09-14
Removed redis as a dependency

suricactus:
In the end of the day it was never actually used and only brings noise when QFC is deployed.


why-not-try-calmer (approved):
Nice cleanup, I found only 1 occurrence of `redis` in `public` but it's rather inocuous (docstring).

Looking forward to the next release providing the updated image, as I kinda need it to tweak an on-premise version.

PR #796, opened 2023-09-27
Remove the tmpdir with feedback and thumbnail after job is finished

suricactus:
Otherwise get out of /tmp storage with long uptime.

### DIFF
--- a/docker-app/worker_wrapper/wrapper.py
+++ b/docker-app/worker_wrapper/wrapper.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import shutil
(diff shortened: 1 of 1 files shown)


faebebin:
seems to fail for 
```
    self.fail("Worker failed with error")
  File "/usr/src/app/qfieldcloud/core/tests/test_qfield_file.py", line 78, in fail
    super().fail(msg)
AssertionError: Worker failed with error

FAIL: test_list_files_for_qfield (qfieldcloud.core.tests.test_qfield_file.QfcTestCase)
```

PR #797, opened 2023-09-27
Ensure subscription created_by Person and overwrite createsuperuser

faebebin:
#QF3203

https://app.clickup.com/t/2192114/QF-3203

Tested locally to create an organization and worked.


duke-nyuki:
Task linked: [QF-3203 Fix field remaining_trial_organizations does not exist](https://app.clickup.com/t/2192114/QF-3203)

docker-app/qfieldcloud/settings.py
```diff
@@ -96,6 +95,8 @@
     "auditlog",
     # Local
     "qfieldcloud.core",
+    # listed after core becuase we overwrite createsuperuser command
```
suricactus, line 98:
```suggestion
    # listed after core because we overwrite createsuperuser command
```

suricactus:
I think the `createuser` command is still used by the k8s setup, we should not remove it, but fix it.

faebebin:
Both `createuser.py` and `createsuperuser.py` create a Person instance now.

IMO we should nevertheless get rid of the createuser.py . Create followup task?

suricactus (approved):
Yes, I would prefer to have a single command too. I think it was mostly password related thing, it was robert197 who added this in the first place.

Thanks for the fixes!

PR #800, opened 2023-10-02
Catching exceptions earlier when trying to cancel orphans containers with empty references

why-not-try-calmer:
Please consider the associated ClickUp task for details.


duke-nyuki:
Task linked: [QF-3210 When listing orphaned containers, python-docker sometimes use empty references](https://app.clickup.com/t/2192114/QF-3210)

suricactus (approved):
Good catch!

why-not-try-calmer:
Hum does it mean that indeed `NotFound` was raised, but not caught by the `except` clause? In which case I'd assume the import path I used is incorrectly?

suricactus:
This is not deployed yet. IMO Sentry bot is just a bit stupid. Will try to deploy this week and we will see.

PR #722, opened 2023-07-11
Migration drf_yasg (Swagger) -> drf_spectacular (Open API 3)

why-not-try-calmer:
__Todo__:
- [x] add hidden params
- [x] fix errors

__Validation errors__:
```
/usr/src/app/qfieldcloud/core/views/files_views.py:40: Error [ListFilesView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
/usr/src/app/qfieldcloud/core/views/files_views.py:118: Error [DownloadPushDeleteFileView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
/usr/src/app/qfieldcloud/authentication/views.py:66: Error [LogoutView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
/usr/src/app/qfieldcloud/core/views/files_views.py:285: Error [ProjectMetafilesView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
/usr/src/app/qfieldcloud/core/views/files_views.py:297: Error [PublicFilesView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
/usr/src/app/qfieldcloud/core/views/jobs_views.py:21: Error [JobViewSet]: exception raised while getting serializer. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: tuple index out of range)
/usr/src/app/qfieldcloud/core/views/package_views.py:161: Error [PackageUploadFilesView]: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
[...]

PR #802, opened 2023-10-04
Fix OpenAPI and Swagger URLs and CORS

suricactus:
Since the schema endpoint has been changed, we broke the integration on https://docs.qfield.org/reference/qfieldcloud/api/ . I overlooked the endpoint renaming.

@why-not-try-calmer was there any reason you rename endpoints?

### DIFF
--- a/docker-app/qfieldcloud/urls.py
+++ b/docker-app/qfieldcloud/urls.py
@@ -42,14 +42,19 @@ urlpatterns = [
         name="index",
     ),
     path(
-        "schema/",
+        "swagger.yaml",
         SpectacularAPIView.as_view(),
-        name="schema",
+        name="openapi_schema",
     ),
     path(
         "swagger/",
-        SpectacularSwaggerView.as_view(url_name="schema"),
-        name="schema-swagger-ui",
+        SpectacularSwaggerView.as_view(url_name="openapi_schema"),
+        name="openapi_swaggerui",
+    ),
+    path(
+        "docs/",
+        SpectacularRedocView.as_view(url_name="openapi_schema"),
+        name="openapi_redoc",
     ),
     path(
         settings.QFIELDCLOUD_ADMIN_URI + "api/files/<uuid:projectid>/",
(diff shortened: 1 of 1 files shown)


why-not-try-calmer:
> Since the schema endpoint has been changed, we broke the integration on https://docs.qfield.org/reference/qfieldcloud/api/ . I overlooked the endpoint renaming.
> 
> @why-not-try-calmer was there any reason you rename endpoints?

Are you talking about https://github.com/opengisch/qfieldcloud/pull/722/files#diff-53c9c2dd0d12c371d3267292e547885a0189ce67b77ec67b3f6b590f140d959a? No, no reason I would have taken seriously in light of any evidence that this was breaking something.

docker-app/qfieldcloud/urls.py
```diff
@@ -42,14 +42,19 @@
         name="index",
     ),
     path(
-        "schema/",
+        "swagger.yaml",
```
why-not-try-calmer, line 45: I think it's not considered a good practice to use file extensions in urls, for the sake of not being hostage of an extension. But I am no expert in these conventions.

docker-app/qfieldcloud/urls.py
```diff
@@ -42,14 +42,19 @@
...
         SpectacularAPIView.as_view(),
-        name="schema",
+        name="openapi_schema",
     ),
     path(
         "swagger/",
-        SpectacularSwaggerView.as_view(url_name="schema"),
-        name="schema-swagger-ui",
```
why-not-try-calmer, line 52: Yes, using hyphens was not a good practice, so that's better!

why-not-try-calmer:
Also I should point out that the url linked from the documentation does point to a valid endpoint. Isn't it simply the embedded JS that is off?

docker-app/qfieldcloud/urls.py
suricactus: Well, it depends. For the sake of backwards compatibility, it should have the extension for now.

suricactus:
> Also I should point out that the [url](https://app.qfield.cloud/docs/) used in the documentation does point to a valid endpoint, any overlook notwithstanding. So isn't it simply the embedded JS being off?

I would not prefer not to change it now.

PR #807, opened 2023-10-19
Disable tmate sessions

suricactus:
Did not delete it in case it is really needed by someone


faebebin (approved):
LGTM

I never used this one

PR #817, opened 2023-11-03
Actually set the bad layer handler on project to support localized datapath layer packaging

suricactus:
See https://github.com/opengisch/libqfieldsync/pull/56


duke-nyuki:
Task linked: [QF-3337 Fix localized paths in libqfieldsync on the cloud](https://app.clickup.com/t/2192114/QF-3337)

PR #821, opened 2023-11-10
Fix the returned sha256 checksum on global files level

suricactus:


duke-nyuki:
Task linked: [QF-3350 Fix getting file sha256 checksum](https://app.clickup.com/t/2192114/QF-3350)

why-not-try-calmer:
After deploying this commit to on-prem I was able to make a change on QField and sync+load it in QFieldSync. I'd be relieved if @lucienicolier could confirm that this works for her as well. 
Thanks a lot for the patch!

PR #827, opened 2023-11-18
Bye release-drafter! So long and thanks for all the fish!

suricactus:


duke-nyuki:
Task linked: [QF-3652 Remove release drafter workflow from QField repos](https://app.clickup.com/t/2192114/QF-3652)

PR #829, opened 2023-11-20
Implement automatic Let's Encrypt certificate renewal

stcz:
### Problem
Currently, if you are using certbot and a Let's Encrypt Certificate, the certificate will be renewed by certbot, but it will not be copied to nginx. This needs to be done by manually running `scripts/init_letsencrypt.sh`.

Further, it is currently not possible to use a own certificate from another source.

### Solution
In this PR i propose a solution where you can manually set the path that NGINX is using for the Certificate. 
Further, I added the certbot certificates directly to the NGINX container.


suricactus:
Hey, @stcz . First, very highly appreciated you contribute your solution upstream! Thank you very much!

You properly identified `scripts/init_letsencrypt.sh` as something one has to run manually or from a cron from the host, which is also annoying on our side.

If I read your proposal correctly, you move the whole logic to handle certificates renewal cron on the host. In a perfect universe I would prefer to have all this logic within the docker-compose configuration. One project achieving something similar is https://hub.docker.com/r/gordonchan/auto-letsencrypt/ . Personally I never had time to investigate deeper what we can do here.

As for giving full path to your certificates to support external ones, I am fully ok with that.

stcz:
Hi @suricactus, 
Thanks for your response.

Just to be sure: I not sure what you mean with:
> If I read your proposal correctly, you move the whole logic to handle certificates renewal cron on the host.  In a perfect universe I would prefer to have all this logic within the docker-compose configuration.

To explain it in detail:
- The configuration is done in .env (what i think is a better place, than the docker-compose file).
- The "logic" is running inside the docker containers
- The configuration is stored in a folder mounted to the filesystem. (Could also be a volume, but would produce migration/backwards compatibility issues)

I would like it, when there is no change in the docker-compose-file is needed to change the certificate source. That's why I prefer the `.env`. 

I would prefer to not use images like the proposed auto-letsencrypt, as it doesn't look well maintained. 

If the proposed way is fine i would go for testing and improving.

stcz:
I worked again on it and changed a few things:

- The Diffie-Hellman parameters are moved to a folder
- Own 4096 bit DH parameters are only created and used for production
- The nginx tls config is not updated by `init_letsencrypt.sh` anymore. They should be maintained in the repository.

From my point of view, this PR is now working as intended. Can you review it and let me know if more changes are needed?

I set up a test Server and will check if the renewal is working. 

For merging, I would recommend squashing my commits to one.

suricactus:
Hey! Xmas time allowed me to have a look and it was looking quite solid, also understood why my question earlier was a bit weird. I also want to setup an internal server with this branch before merging. We might not merge it very soon, but it will become part of upstream in the next weeks.

stcz:
Thanks, that sounds good.

When set up my testserver I changed as few as possible, I figured out #843.  

I would suppose to quote all variables in  `.env.example`. But this should be done in another PR. What do you think about this?

Further, I thought about hardining `./scripts/init_letsencrypt.sh` against this issue. 

What do you think about this when loading .env?
```
# read and export the variables from the .env file for the duration of this script
set -o allexport
source <(cat .env | sed -e '/^#/d;/^\s*$/d' -e "s/'/'\\\''/g" -e "s/=\(.*\)/='\1'/g")
set +o allexport
```
It is from [here](https://stackoverflow.com/a/66118031). But i would like to avoid.

stcz:
I added an unrelated change in README by changing Infrastructure to h2. I or you can remove if this is unwanted.

suricactus:
> When set up my testserver I changed as few as possible, I figured out #843.
> 
> I would suppose to quote all variables in `.env.example`. But this should be done in another PR. What do you think about this?
> 
> Further, I thought about hardining `./scripts/init_letsencrypt.sh` against this issue.

Internally we double quote the envvars with special chars in them:

	ENVVAR="whatever@funky or even space"

> What do you think about this when loading .env?
> 
> ```
> # read and export the variables from the .env file for the duration of this script
> set -o allexport
> source <(cat .env | sed -e '/^#/d;/^\s*$/d' -e "s/'/'\\\''/g" -e "s/=\(.*\)/='\1'/g")
> set +o allexport
> ```
> 
> It is from [here](https://stackoverflow.com/a/66118031). But i would like to avoid.

Been there. Fails in certain cases unfortunately...

suricactus:
@stcz just to let you know there was another PR that already quoted the envvars with special chars, see https://github.com/opengisch/qfieldcloud/pull/846 .

README.md
```diff
@@ -221,18 +221,32 @@ Run the django database migrations
...
+    QFIELDCLOUD_TLS_CERT=/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/fullchain.pem
+    QFIELDCLOUD_TLS_KEY=/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/privkey.pem
+    QFIELDCLOUD_TLS_DHPARAMS=/etc/nginx/dhparams/dhparams4096.pem
 
-Note you may want to change the `LETSENCRYPT_EMAIL`, `LETSENCRYPT_RSA_KEY_SIZE` and `LETSENCRYPT_STAGING` variables.
+You can also use your own certificates by placing them in `conf/nginx/certs/` ànd changing `QFIELDCLOUD_TLS_CERT` and `QFIELDCLOUD_TLS_KEY` accordingly.
+Don't forget to create your Diffie-Hellman parameters.
 
-### Infrastructure
+## Infrastructure
```
m-kuhn, line 249: @suricactus happy with this change?

suricactus: If you meant the title change - yes.

scripts/init_letsencrypt.sh
```diff
@@ -33,10 +32,7 @@ docker compose run --rm --entrypoint "\
 
 echo
 
-echo "### Copy the certificate and key to their final destination ..."
-cp ${CONFIG_PATH}/certbot/conf/live/${QFIELDCLOUD_HOST}/fullchain.pem ${CONFIG_PATH}/nginx/certs/${QFIELDCLOUD_HOST}.pem
-cp ${CONFIG_PATH}/certbot/conf/live/${QFIELDCLOUD_HOST}/privkey.pem ${CONFIG_PATH}/nginx/certs/${QFIELDCLOUD_HOST}-key.pem
-echo
+chmod 755 "$CONFIG_PATH/nginx/99-autoreload.sh"
```
m-kuhn, line 35: I believe if we keep it executable in the git repo, that should be sufficient?

.env.example
```diff
@@ -14,6 +14,13 @@ DJANGO_ALLOWED_HOSTS="localhost 127.0.0.1 0.0.0.0 app nginx"
 
 SECRET_KEY=change_me
 
+# Certificates
+# For a local instance self signed certs are used.
+# For production use with certbot and letsencrypt uncomment the following lines:
+# QFIELDCLOUD_TLS_CERT=/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/fullchain.pem
+# QFIELDCLOUD_TLS_KEY=/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/privkey.pem
+# QFIELDCLOUD_TLS_DHPARAMS=/etc/nginx/dhparams/dhparams4096.pem
```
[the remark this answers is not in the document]
stcz: For me the main question here was how to make dev environments working without any changes in the .env and minimal changes to a production setup. When pointing default to the letsencrypt here, manually changes would be needed here for dev environments. 

As you have to/should change variables in .env in a production setup, i would prefer it this way with uncomments for production. But i can change it to whatever you prefer.

scripts/init_letsencrypt.sh
stcz: It is intentionally like this, so that the autoreload is not executed as long as it is not necessary. 

If it should always be executed, even if not necessary, it would be sufficient to make it executable in git.

m-kuhn: Ah I see.
In this case wouldn't it be more transparent if it was done with something more "visible" like renaming this file?

.env.example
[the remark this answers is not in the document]
suricactus: We are on the same page, I also prefer minimal changes. What I usually try to avoid is uncommeting stuff, to me it's never clear who, when and why commented/uncommented stuff. Isn't it going to work with minimal to no changes if we just do:

```
QFIELDCLOUD_TLS_CERT=""
QFIELDCLOUD_TLS_KEY=""
QFIELDCLOUD_TLS_DHPARAMS=""
```

And if one wants to have let's encrypt, add it in the comment documenting each value? Rough suggestion:

```

# TLS certificate filename from within the `nginx` container.
# When empty, the default self-signed certificate is used.
# For usage with Let's Encrypt certificate, use as:
# QFIELDCLOUD_TLS_CERT="/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/fullchain.pem"
# DEFAULT: ""
QFIELDCLOUD_TLS_CERT=""

```

scripts/init_letsencrypt.sh
stcz: This would be possible but would be way more complicated and i don't know if it is worth this. 

If we copy it with the script we should be aware of:
- We have to delete the folder that is created by docker-compose
- Should add the file to .gitignore
- We still need to apply the permissions

I can change it to your preference, but i would be fine with the simple chmod.

.env.example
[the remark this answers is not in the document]
stcz:
```suggestion
# TLS certificate filename from within the `nginx` container.
# When empty, the default self-signed certificate is used.
# For usage with Let's Encrypt certificate, use as:
# QFIELDCLOUD_TLS_CERT="/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/fullchain.pem"
# QFIELDCLOUD_TLS_KEY="/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/privkey.pem"
# QFIELDCLOUD_TLS_DHPARAMS="/etc/nginx/dhparams/dhparams4096.pem"
# DEFAULT for all: ""
QFIELDCLOUD_TLS_CERT=""
QFIELDCLOUD_TLS_KEY=""
QFIELDCLOUD_TLS_DHPARAMS=""
```
Is it fine like this?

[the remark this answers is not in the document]
suricactus:
```suggestion
# Certificates

# TLS certificate filename from within the `nginx` container.
# When empty, the default self-signed certificate is used.
# For usage with Let's Encrypt certificate, use as:
# QFIELDCLOUD_TLS_CERT="/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/fullchain.pem"
# DEFAULT: ""
QFIELDCLOUD_TLS_CERT=""

# TLS private key filename from within the `nginx` container.
# When empty, the default self-signed certificate is used.
# For usage with Let's Encrypt certificate, use as:
# QFIELDCLOUD_TLS_KEY="/etc/letsencrypt/live/${QFIELDCLOUD_HOST}/privkey.pem"
# DEFAULT: ""
QFIELDCLOUD_TLS_KEY=""

# TLS DF parameters filename from within the `nginx` container.
# When empty, no DF parameters are used.
# DEFAULT: "/etc/nginx/dhparams/dhparams4096.pem"
QFIELDCLOUD_TLS_DHPARAMS="/etc/nginx/dhparams/dhparams4096.pem"
```

The DH params can be set by default to use the one included in the repo, no matter if we use self-signed or let's encrypt params?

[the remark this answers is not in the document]
stcz: Thx.

About the DH Parameters:
- Using precomputed shared parameters increases the attack surface for precomputation attacks
- Using self generated increases the risk for weak primes. 
- Self generating consumes more Energy. 

Both has pros and cons and in the end it doesn't matter I think.

I will adapt the script later to precomputed.

scripts/init_letsencrypt.sh
m-kuhn: @suricactus what's your opinion on this?

suricactus: I am in favor of keeping things explicit. I do not see any significant risk by reloading nginx every X hours, so I would prefer to commit the file with the `+x` permission.

.env.example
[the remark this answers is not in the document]
stcz: I now use the RFC 3526 4096bit DH params by default, other DH-params are ignored by git.

I have not tested it jet again.

scripts/init_letsencrypt.sh
stcz: Done

stcz:
Should be fixed now.
- I added the LETSENCRYPT variables to the ignore list
- I removed the default values from the compose config

stcz:
One last question:
Has anyone done end-to-end tests? I haven't tested if it is really working and i also haven't tested if there might be any side effects on current installations when just updating.

In theory, there should be no issues. If it is merged, I would directly deploy it on my production server.

suricactus:
> One last question: Has anyone done end-to-end tests? I haven't tested if it is really working and i also haven't tested if there might be any side effects on current installations when just updating.
> 
> In theory, there should be no issues. If it is merged, I would directly deploy it on my production server.

Not yet, planning to do this on our internal staging server and if all good, straight to prod in few weeks.

> Do you have any ideas about the flanky test?

Forget about it, it is flaky for reasons beyond your immediate control.

stcz:
OK.

Is master your staging branch and release the production branch?

suricactus:
Both are using the same `release` branch. I will just temporarily deploy this branch on some of the internal dev servers (not necessarily staging) to see how it behaves before merging.

stcz:
Hi,
What is the status of the PR? Should I update the chnflicting changes?
Thanks for your response :)

suricactus:
Hey! This seems to work well for some internal tests. Let's update it and plan to merge it in the following weeks. Thanks for the proactiveness!

suricactus:
It took us some time to spend resources to make this merged, but the time has come. Thanks again to @stcz for initiating this and addressing our concerns. I will merge it in it's current state and continue in a follow-up here https://github.com/opengisch/qfieldcloud/pull/1005 .

PR #830, opened 2023-11-20
Make AbstractSubscription.__str__() harmless

why-not-try-calmer:
Fix for https://github.com/opengisch/QFieldCloud-on-prem/issues/15.

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -882,7 +882,11 @@ class AbstractSubscription(models.Model):
         return trial_subscription_obj, regular_subscription_obj
 
     def __str__(self):
-        return f"{self.__class__.__name__} #{self.id} user:{self.account.user.username} plan:{self.plan.code} total:{self.active_storage_total_mb}MB"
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -882,7 +882,11 @@ def create_subscription(
         return trial_subscription_obj, regular_subscription_obj
 
     def __str__(self):
-        return f"{self.__class__.__name__} #{self.id} user:{self.account.user.username} plan:{self.plan.code} total:{self.active_storage_total_mb}MB"
+        plan_code = self.plan.code if hasattr(self, "plan") else "None"
```
suricactus, line 885: I don't think there is a situation that a subscription can have no plan, as it is a non-nullable field. In short, this line is unneeded extra.

suricactus (approved):
Good catch! Small comment and ready to merge.

docker-app/qfieldcloud/subscription/models.py
why-not-try-calmer: Ok, I'll take your word for it!

suricactus (approved):

PR #844, opened 2023-12-28
Add draft for standalone config

stcz:
For small instance deployment, I added a standalone server configuration. It should be possible to simply run QFieldCloud with this only by launching it in docker. 

If this might be added and maintained to the repository, I would further work on this draft.

I think adding and get it working is not an issue here, but we should think about how to maintain and update the containers in the future. Any comments on this are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,9 @@ Code style done with precommit
 
 ### Launch a server instance
 
+For small self-hosted environments you can run QFieldCloud on a single server. Therefore use `docker-compose.override.standalone.yml`.
+For large instances use dedicated DB and Storage Servers.
+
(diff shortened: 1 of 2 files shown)


gounux:
Hi @stcz, this would be very much appreciated !

suricactus:
@stcz thank you for suggestion. While this would be nice, we as developers of QFieldCloud don't want to make false promises about the deployment of QFieldCloud - it is entirely administrator's responsibility how they do that. 

I have created #1052  that separates the additional services in a standalone configuration that can be reused for standalone services, but I have explicitly mentioned that this should be used at own risk.

PR #846, opened 2024-01-04
Escape `@` and `&` in .env example to fix init scripts

boardend:


suricactus (approved):

suricactus:
Very welcome first contribution! Since you are on this, it might be a good idea to add double quotes around all variables with `@` in their value too. It breaks on some shells. Feel free to do it here or in a follow-up PR.

I am also open to double quote all values.

faebebin (approved):
:partying_face:

boardend:
No more unescaped `@` in `&` in `.env.example`

suricactus (approved):
Thanks a lot!

PR #848, opened 2024-01-05
QF-3728 ensure qfieldcloud-qgis image name in docker-compose file

dddpt:
Problem is docker compose, depending on its version, uses dash (`qfieldcloud-qgis`) or underscore (`qfieldcloud_qgis`) as separator for image names.

Solution is to ensure the `qgis` always gets the `qfieldcloud-qgis` name using the `image` directive in `docker-compose.yml`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
           eval $(egrep "^[^#;]" .env | xargs -d'\n' -n1 | sed -E 's/(\w+)=(.*)/export \1='"'"'\2'"'"'/g')
 
       - name: Pull docker containers
-        run: docker compose pull
+        run: docker compose pull --ignore-buildable
 
       - name: Build and run docker containers
         run: |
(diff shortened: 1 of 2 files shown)


faebebin (changes requested):
Ok, thanks - didn't know compose versions use different seperators.

But this fix seems to break already on CI ... So we might need another way to fix that problem.

image

dddpt:
So, partial solution found.

What we want is that docker compose builds our `qgis` image and names it `qfieldcloud-qgis`. The problem is that now docker compose tries to pull a `qfieldcloud-qgis` image that does not exist.

The solution to that is to add the `--ignore-buildable` flag to the pull: `docker compose pull --ignore-buildable`.
This explicitly tells docker that the image is meant to be built and not pulled.

However it adds verbosity and I don't know how big a hassle it will be to add this flag to every pull in dev/prod/other.
Since the problem originates only with my setup, we can ignore this issue and I can only keep the change locally for myself.

What do you think?

dddpt:
Ok, it works with `pull_policy: build`. I think it is a better fix than `--ignore-buildable`.

Weird thing is that on my machine `docker compose pull` still tries to pull `qfieldcloud-qgis` and fails.
`docker compose up -d --build --force-recreate` works all fine.
imho, if we don't explicitly do a `docker compose pull` on deployment, this fix should work.

suricactus:
I am not too enthusiastic about changing things in the docker-compose files themself. I suggest an alternative approach here: #850 . This avoids any changes in docker-compose file, it just adds a new envvar if needed.

In the future we might support multiple QGIS containers anyways, so it's a good foundation for that future goal.

PR #851, opened 2024-01-09
Add multiple packaging offliners support and selection from the Admin

suricactus:
Allows support to multiple packaging offliners. This opens the road to multiple offliner implementations and slowly switching to python based code.

Admin:
image

Depends on https://github.com/opengisch/libqfieldsync/pull/54 .


docker-app/qfieldcloud/core/models.py
```diff
@@ -1035,6 +1039,17 @@ class Meta:
...
     )
 
+    # Packaging offliner to be used by the QGIS worker container.
+    packaging_offliner = models.CharField(
+        _("Packaging Offliner"),
+        help_text=_(
+            'The packaging offliner is used withing jobs to convert QGIS projects into QField projects. The new "Optimized Offliner" should be selected over the legacy "QGIS Core Offline Editing" for new projects.'
+        ),
+        max_length=100,
+        default=PackagingOffliner.PYTHONMINI,
```
suricactus, line 1049: NOTE All new projects will use the new packaging offliner by default.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1035,6 +1039,17 @@ class Meta:
         validators=[MinValueValidator(1), MaxValueValidator(100)],
     )
 
+    # Packaging offliner to be used by the QGIS worker container.
+    packaging_offliner = models.CharField(
+        _("Packaging Offliner"),
+        help_text=_(
+            'The packaging offliner is used withing jobs to convert QGIS projects into QField projects. The new "Optimized Offliner" should be selected over the legacy "QGIS Core Offline Editing" for new projects.'
+        ),
```
suricactus, line 1047: This help text should ideally match the needs for both support and endusers.

docker-app/qfieldcloud/core/models.py
```diff
@@ -943,6 +943,10 @@ class StatusCode(models.TextChoices):
         )
         TOO_MANY_COLLABORATORS = "too_many_collaborators", _("Too many collaborators")
 
+    class PackagingOffliner(models.TextChoices):
+        QGISCORE = "qgiscore", _("QGIS Core Offline Editing")
+        PYTHONMINI = "pythonmini", _("Optimized Offliner")
```
suricactus, line 948: Feel free to suggest better human readable names (the ones passed to `_()`) function.

docker-app/qfieldcloud/core/models.py
m-kuhn: That's a bold move, shouldn't we test it on a few chosen projects in the beginning?

docker-app/qfieldcloud/core/models.py
m-kuhn: What do you think about "[...] is used to package data for offline use with QField."?

docker-app/qfieldcloud/core/models.py
m-kuhn: QGIS Core Offline Editing sounds good, I wonder if we should already label it with "legacy" or (in case we wait for a few weeks before switching the default) we can add a "deprecated" to the name whenever we switch the default.

Personally I like "packager" more than "offliner", so I would probably go for "Optimized Packager".

m-kuhn:
Thanks for looking into this.
As you already indicated with your questions, I have mostly looked at the wording. If a technical review needs to be done, I think it would be better if @faebebin could have a look.

docker-app/qfieldcloud/core/models.py
suricactus:
```suggestion
        help_text=_(
            'The packaging offliner is is used to package data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for newer projects.'
        ),
```

docker-app/qfieldcloud/core/models.py
suricactus:
```suggestion
        QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
        PYTHONMINI = "pythonmini", _("Optimized Packager")
```

docker-app/qfieldcloud/core/models.py
suricactus: Ok, will do in phase 2.

suricactus:
Thanks for the review @m-kuhn . Will push a change that revert the default offliner to be qgis core. Still not there with knowledge transfer about packaging, but I believe we can do a share or interactive review with @dddpt and @faebebin with this one.

duke-nyuki:
Task linked: [QF-3333 Integrate the new minioffliner in qfieldcloud](https://app.clickup.com/t/2192114/QF-3333)

docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
```diff
@@ -14,13 +14,11 @@ class Migration(migrations.Migration):
...
-                    ("pythonmini", "Optimized Offliner"),
+                    ("qgiscore", "QGIS Core Offline Editing (deprecated)"),
+                    ("pythonmini", "Optimized Packager"),
                 ],
-                default="pythonmini",
-                help_text=(
-                    'The packaging offliner is used withing jobs to convert QGIS projects into QField projects. The new "Optimized Offliner" should be selected over the legacy "QGIS Core Offline Editing" for new projects.'
-                ),
+                default="qgiscore",
+                help_text='The packaging offliner is is used to package data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for newer projects.',
```
m-kuhn, line 21: Is is used

faebebin:
> Thanks for the review @m-kuhn . Will push a change that revert the default offliner to be qgis core. Still not there with knowledge transfer about packaging, but I believe we can do a share or interactive review with @dddpt and @faebebin with this one.

@suricactus feel free to ping me for a review session

docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
```diff
@@ -0,0 +1,26 @@
...
+        migrations.AddField(
+            model_name="project",
+            name="packaging_offliner",
+            field=models.CharField(
+                choices=[
+                    ("qgiscore", "QGIS Core Offline Editing (deprecated)"),
+                    ("pythonmini", "Optimized Packager"),
+                ],
+                default="qgiscore",
+                help_text='The packaging offliner is is used to package data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for newer projects.',
```
faebebin, line 21:
```suggestion
                help_text='The `packaging_offliner` packages data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for new projects.',
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1035,6 +1039,17 @@ class Meta:
         validators=[MinValueValidator(1), MaxValueValidator(100)],
     )
 
+    # Packaging offliner to be used by the QGIS worker container.
+    packaging_offliner = models.CharField(
+        _("Packaging Offliner"),
+        help_text=_(
+            'The packaging offliner is is used to package data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for newer projects.'
```
faebebin, line 1046:
```suggestion
            'The `packaging_offliner` packages data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for new projects.'
```

docker-qgis/entrypoint.py
```diff
@@ -6,37 +6,34 @@
...
-    QgsOfflineEditing,
-    QgsProject,
-    QgsRectangle,
-    QgsVectorLayer,
-)
+from qgis.core import QgsCoordinateTransform, QgsProject, QgsRectangle, QgsVectorLayer
 
 PGSERVICE_FILE_CONTENTS = os.environ.get("PGSERVICE_FILE_CONTENTS")
 
 logger = logging.getLogger("ENTRYPNT")
```
faebebin, line 28: Eventhough unrelated to the PR i am wondering:
Why not "__name__" here? Or like the full name "docker-qgis/entrypoint.py" if it would be "__main__"?
Same for other worker related scripts where we use such names.

docker-qgis/entrypoint.py
```diff
@@ -104,22 +99,29 @@ def _call_qfieldsync_packager(project_filename: Path, package_dir: Path) -> str:
...
     )
 
     attachment_dirs, _ = project.readListEntry("QFieldSync", "attachmentDirs", ["DCIM"])
-    offline_editing = QgsOfflineEditing()
+    if offliner_type == OfflinerType.QGISCORE:
+        offliner = QgisCoreOffliner()
+    elif offliner_type == OfflinerType.PYTHONMINI:
+        offliner = PythonMiniOffliner()
+    else:
+        raise NotImplementedError(f"Offliner type {offliner_type} is not supported.")
```
faebebin, line 112: Could we instead of raising an Error just log and set the default value?
```suggestion
    if offliner_type == OfflinerType.QGISCORE:
        offliner = QgisCoreOffliner()
    elif offliner_type == OfflinerType.PYTHONMINI:
        offliner = PythonMiniOffliner()
    else:
        logger.error(f"Offliner type {offliner_type} is not supported.")
        offliner = QgisCoreOffliner()
```

faebebin (commented):
I just looked through and added some suggestion. Did NOT yet test it.

docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
suricactus:
```suggestion
                help_text='The Packaging Offliner packages data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for new projects.',
```

We should not use developer names for the UI

docker-app/qfieldcloud/core/models.py
```diff
@@ -1035,6 +1039,17 @@ class Meta:
         validators=[MinValueValidator(1), MaxValueValidator(100)],
     )
 
+    # Packaging offliner to be used by the QGIS worker container.
+    packaging_offliner = models.CharField(
+        _("Packaging Offliner"),
```
dddpt, line 1044:
```suggestion
        _("Packaging ddd Offliner"),
```
my suggestion

docker-app/qfieldcloud/core/models.py
suricactus:
```suggestion
            'The Packaging Offliner packages data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for new projects.'
```

docker-qgis/entrypoint.py
suricactus: Honest asnwer: I don't remember. I am afraid to change without testing, because it can break some log parsing utilities.

There is one more place where this is used: `process_project_file.py`.  We can do a follow-up PR changing this behaviour with some tests if all the rest is good.

docker-qgis/entrypoint.py
suricactus: I would argue we should keep the `raise NotImplementedException` for enumerations, makes it easier for developer.

dddpt (approved):
Agree with Fabian's comment on description text.

PR #855, opened 2024-01-16
QF-3746bAdded README test instructions and .env.test.example file

dddpt:


suricactus (commented):
I am not 100% convinced on the two `.env` files approach. The `.env.sample` is already test ready, so now we will have two test-ready `.env` files. What if we just document the possibly colliding variables in the README instead?

Even if we keep the .env.test, I don't see logic why some variables have comments, some don't, some have double whitespace, some don't.

dddpt:
> I am not 100% convinced on the two `.env` files approach. The `.env.sample` is already test ready, so now we will have two test-ready `.env` files. What if we just document the possibly colliding variables in the README instead?

The idea is that if some variables unrelated to tests are changed in `.env`, you don't need to make sure `.env.test` stays consistent with `.env`. You only change the variables in `.env.test` if your test setup changes.

> Even if we keep the .env.test, I don't see logic why some variables have comments, some don't, some have double whitespace, some don't.

My bad. will improve.

suricactus:
I would prefer if we drop the `.env.text` file and if you want you can add it to the `README.md` as copy-pastable suggestion.

PR #857, opened 2024-01-16
When a layer is invalid due to localized datasource, show link to docs

suricactus:
| before | after |
|-|-|
| image | image |

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -22,6 +22,7 @@ from django.db.models.aggregates import Count, Sum
 from django.db.models.fields.json import JSONField
 from django.urls import reverse_lazy
 from django.utils.functional import cached_property
+from django.utils.safestring import SafeString, mark_safe
 from django.utils.translation import gettext as _
 from model_utils.managers import InheritanceManager, InheritanceManagerMixin
 from qfieldcloud.core import geodb_utils, utils, validators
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-3753 Show better message when localized datasource is used in QFieldCloud](https://app.clickup.com/t/2192114/QF-3753)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1186,21 +1186,32 @@ def problems(self) -> list[dict]:
                 layer_name = layer_data.get("name")
 
                 if layer_data.get("error_code") != "no_error":
+                    if layer_data.get("error_code") == "localized_dataprovider":
+                        description = _('Layer "{}" dataprovider is localized').format(
+                            layer_name
+                        )
+                        solution = _(
+                            'Make sure your <a href="https://docs.qfield.org/fr/how-to/outside-layers/">localized layer</a> is available on your QField device.'
```
faebebin, line 1194: not sure whether we should add the 

```
from django.utils.safestring import mark_safe

...

mark_safe(_(...))
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1186,21 +1186,32 @@ def problems(self) -> list[dict]:
...
+                    else:
+                        description = _(
+                            'Layer "{}" has an error with code "{}": {}'
+                        ).format(
+                            layer_name,
+                            layer_data.get("error_code"),
+                            layer_data.get("error_summary"),
+                        )
+                        solution = _(
+                            'Check the last "process_projectfile" logs for more info and reupload the project files with the required changes.'
```
faebebin, line 1205: And not related to PR changes:

we could here add the link with `project_jobs` url ... if we are already on it ?

faebebin (approved):
Looks good and simpel at first glance.

Put a comment because not sure whether we need the `mark_safe` and did not test. But if you are sure feel free to merge.

docker-app/qfieldcloud/core/models.py
suricactus: I had no idea what will happen if we pass the returning object to `json` serializer. All is good:

```
>>> a = mark_safe("<span>hello</span>")
>>> type(a)
<class 'django.utils.safestring.SafeString'>
>>> import json
>>> json.dumps(a)
'"<span>hello</span>"'
```

docker-app/qfieldcloud/core/models.py
suricactus: We cannot really, because we don't have the web UI here.

PR #860, opened 2024-01-25
QF-3715 In Subscriptions admin add a filter by plan

dddpt:
Screenshot from 2024-01-31 17-50-23

### DIFF
--- a/docker-app/qfieldcloud/subscription/admin.py
+++ b/docker-app/qfieldcloud/subscription/admin.py
@@ -108,6 +108,7 @@ class SubscriptionAdmin(QFieldCloudModelAdmin):
     list_filter = (
         SubscriptionPeriodFilter,
         "status",
+        "plan",
     )
 
     readonly_fields = (


faebebin (changes requested):
As you alter the model fields check `makemigrations` and add also the migration file to the PR.
`docker compose exec app python manage.py makemigrations`
Eventhough it is not really modifying the db schema django will still create a migration.

Besides this, Looks good - thanks alot!

dddpt:
done ;-)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -437,6 +437,7 @@ class UpdateSubscriptionKwargs(TypedDict):
 
     plan = models.ForeignKey(
         Plan,
+        verbose_name=_("Plan"),
```
suricactus, line 440: Why did you add the verbose name? What it was showing without it?

dddpt: I added the verbose name for UI consistency, so that all filters have a capital letter: "By Period", "By Status", "By Plan", etc...
Otherwise it shows "By plan".

suricactus: less code, more fun :)
And TBH I don't see a special reason why we need the model to be capital in this context.

docker-app/qfieldcloud/subscription/admin.py
```diff
@@ -105,10 +105,7 @@ class SubscriptionAdmin(QFieldCloudModelAdmin):
         "status",
     )
 
-    list_filter = (
-        SubscriptionPeriodFilter,
-        "status",
-    )
+    list_filter = (SubscriptionPeriodFilter, "status", "plan")
```
suricactus, line 108:
```suggestion
    list_filter = (
        SubscriptionPeriodFilter,
        "status",
        "plan",
     )
```

suricactus: Also if you have a trailing comma?

suricactus: Does not make sense a longer line to be kept in the same line, while the original tuple was in two lines. Also checked locally, black does respect the trailing commas.

Note we always prefer to use trailing commas on lists/tuples/dicts/whatever structs that have the potential to grow in time.

PR #862, opened 2024-01-26
Fix the usage of `GEODB_PORT`, it should be `HOST_GEODB_PORT`

suricactus:
The `GEODB_PORT` is used as part of the connection string within the compose stack. The `HOST_GEODB_PORT` is used to expose the PG port to the host. The code was trying to connect to the host port from within the compose stack.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,6 +58,7 @@ GEODB_PORT=5432
 GEODB_USER=postgres
 GEODB_PASSWORD="KUAa7h!G&wQEmkS3"
 GEODB_DB=postgres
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-3763 Fix failing test `test_connects_via_pgservice` (on CI)](https://app.clickup.com/t/2192114/QF-3763)

suricactus:
Ok, I fixed local run of the test, but the main issue that it fails on CI remains...

faebebin (approved):
wohoo works locally :partying_face: 


thanks alot

PR #865, opened 2024-01-26
Chore cleanup compose and env

faebebin:
Just some cleanup.

### DIFF
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -13,8 +13,6 @@ services:
     volumes:
       # mount the source for live reload
       - ./docker-app/qfieldcloud:/usr/src/app/qfieldcloud
-    environment:
-      DEBUG: 1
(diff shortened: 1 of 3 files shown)


docker-compose.override.local.yml
```diff
@@ -13,8 +13,6 @@ services:
     volumes:
       # mount the source for live reload
       - ./docker-app/qfieldcloud:/usr/src/app/qfieldcloud
-    environment:
-      DEBUG: 1
```
faebebin, line 17: As we already set it in .env we should not override it here?

docker-compose.override.test.yml
```diff
@@ -7,8 +7,6 @@ services:
       context: ./docker-app
       target: webserver_test
     environment:
-      # run flush, migrate and collectstatic in entrypoint
```
faebebin, line 10: Also remove unrelated comment

suricactus (approved):

suricactus:
Should be good like this.

PR #869, opened 2024-01-30
Remove redis related stuff

faebebin:
Think we have no more redis anywhere ?

### DIFF
--- a/docker-app/worker_wrapper/wrapper.py
+++ b/docker-app/worker_wrapper/wrapper.py
@@ -307,7 +307,7 @@ class JobRun:
                 "QT_QPA_PLATFORM": "offscreen",
             },
             volumes=volumes,
-            # TODO keep the logs somewhere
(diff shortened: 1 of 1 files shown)


docker-app/worker_wrapper/wrapper.py
```diff
@@ -307,7 +307,7 @@ def _run_docker(
                 "QT_QPA_PLATFORM": "offscreen",
             },
             volumes=volumes,
-            # TODO keep the logs somewhere or even better -> pipe them to redis and store them there
+            # TODO keep the logs somewhere
```
suricactus, line 310:
```suggestion
            # TODO stream the logs to something like redis, so they can be streamed back in project jobs page to the user live
```

I hope now with the comment it is clear the intent of the author. If you disagree, remove the whole comment and add it as a task in ClickUp instead of comment in the codebase.

PR #878, opened 2024-02-06
Fix geometry-less exported as empty layers

### LINKED ISSUE #874: Issue visibility child features after syncing over QFieldCloud, not cable
I am experiencing the following problems with qfieldcloud synchronization regarding child features.
1/ Existing child features are not shown in Qfield (the child fields are blanc)
2/ For some layers, which are linked to child features, the symbology and labels are not retrieved as they should
3/ Despite this, I can add new features in QField without a error message
I thought it was due to a change in my project and went back a few versions but the problem still persists.
I then synchronized over cable (with the same settings) and the problem does not occur.

Next I downloaded the sample (advanced) bee project which has some child tables (reviews and consumption) in the project.
The same problem seems to occur there. In QField, I don't see any child information under the corresponding tabs although I can see information in those tables in QGis. I also added a 'bee' review in QGis and synced this over the cloud but it doesn't show up in QField (see printscreens).

The problem started recently but occurs in both QField 3.1.9 Borneo and 3.0.7 Amazonia.
QGis version 3.34.3 Prizren on Ubuntu 22.04.3 LTS with QFieldSync plug-in 4.7.0

Could you have a look?

afbeelding
_Printscreen from QGis_

Screenshot_20240204_182435
_Printscreen from QField (no review visible in the app after cloud sync)_

Screenshot_20240204_185549
_Our project synced over cable_

Screenshot_20240204_185633
Our project synced over QFieldCloud

suricactus:
This was wrongly applied on non-spatial layers, which caused them to be exported as empty layers.

Fix #874

### DIFF
--- a/docker-qgis/requirements_libqfieldsync.txt
+++ b/docker-qgis/requirements_libqfieldsync.txt
@@ -1,1 +1,1 @@
-libqfieldsync @ git+https://github.com/opengisch/libqfieldsync@8ee725ca401b5d083edd8feed733803b8709d496
+libqfieldsync @ git+https://github.com/opengisch/libqfieldsync@e3647e9b0fcbaf74cbb4c3f72bc8f34d99cb44e0


duke-nyuki:
Task linked: [QF-3809 Geometry-less layers empty after synchronising via QFieldCloud](https://app.clickup.com/t/2192114/QF-3809)

PR #882, opened 2024-02-08
QF-3825 wait_for_services.py handles if pg never responds in the timeout window

dddpt:
Currently, if postgres never responds, `wait_for_services.py`  just hangs indefinitely, never returning.

### DIFF
--- a/docker-app/wait_for_services.py
+++ b/docker-app/wait_for_services.py
@@ -30,14 +30,16 @@ def wait_for_postgres():
             logger.info("Postgres is ready! ✨ 💅")
             conn.close()
             return True
-        except psycopg2.OperationalError as e:
+        except psycopg2.OperationalError as error:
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-3825 wait_for_services.py handles if pg never responds in the timeout window](https://app.clickup.com/t/2192114/QF-3825)

suricactus (approved):
We discussed this interactively. Just rename the name of the `error` variable and we are good to go.

suricactus (approved):

dddpt:
I tried many ways to get more details on errors as advertised in the [documentation](https://www.psycopg.org/docs/errors.html#sqlstate-exception-classes), [or here](https://www.psycopg.org/docs/module.html#dbapi-exceptions), with no success.

image

PR #885, opened 2024-02-14
Support `skip_metadata` `on packages/<project-id>/latest` request to boost performance

suricactus:
Serve only `md5` checksum instead of `sha256` checksum to gain extra performance when downloading packaged files in QField. The `sha256` causes N+1 requests to the S3 storage, therefore makes the QFieldCloud response on the `packages/<project-id>/latest` endpoint super slow.

See https://github.com/opengisch/QField/pull/5040

### DIFF
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -94,7 +94,7 @@ class ListFilesView(views.APIView):
                 "display": S3ObjectVersion(version.key, version).display,
             }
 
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-1724 Consider using etag instead of sha256 metadata in S3 to gain performance in files and package views](https://app.clickup.com/t/2192114/QF-1724)

boardend (approved):
I can confirm, that with the `skip_metadata` parameter a project with more than 2000 attachments can now be listed.

Code looks good to me :+1:

PR #886, opened 2024-02-14
QF-3716 Add possibility to rerun job in the Admin

dddpt:
The re-run button sets the job back to "pending" status.

However, on my local setup, jobs are systematically failing as the projects are missing their files.
Requires maybe some more tests?

Screenshot from 2024-02-14 17-32-26

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -26,7 +26,7 @@ from django.db.models.functions import Lower
 from django.forms import ModelForm, fields, widgets
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.http.response import Http404, HttpResponseRedirect, StreamingHttpResponse
-from django.shortcuts import redirect, resolve_url
+from django.shortcuts import resolve_url
 from django.template.response import TemplateResponse
 from django.urls import path, reverse
 from django.utils.decorators import method_decorator
@@ -878,7 +878,7 @@ class JobAdmin(QFieldCloudModelAdmin):
 
     def rerun_job(self, request, object_id):
         Job.objects.filter(pk=object_id).update(status="pending")
-        return redirect(request.META.get("HTTP_REFERER", "../.."))
+        return HttpResponseRedirect("..")
 
 
 class ApplyJobDeltaInline(admin.TabularInline):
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-3716 Add possibility to rerun job in the Admin](https://app.clickup.com/t/2192114/QF-3716)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -869,6 +876,10 @@ def export_applyjob_deltafile(
             },
         )
 
+    def rerun_job(self, request, object_id):
+        Job.objects.filter(pk=object_id).update(status="pending")
+        return redirect(request.META.get("HTTP_REFERER", "../.."))
```
suricactus, line 881: I guess the intention here is to return to the root of the admin? Does it really make sense? Why don't you stay on current job page, so you can see how it progress in statuses?

Also, I believe there is a better mechanism to move to another page than using the headers. Check the web UI views for examples.

docker-app/qfieldcloud/core/templates/admin/job_change_form.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'admin/change_form.html' %}
+
+{% block submit_buttons_bottom %}
+  {{ block.super }}
+
+  <div class="submit-row">
+    {% if original %}
+        <a href="{% url 'admin:rerun_job' original.pk %}" class="button">Re-run</a>
```
suricactus, line 8:
```suggestion
        <a href="{% url 'admin:rerun_job' original.pk %}" class="button">{% trans 'Re-run job' %}</a>
```

NOTE this needs the`i18n` to be imported in this template.

docker-app/qfieldcloud/core/admin.py
dddpt: Oops, that was a leftover from a different approach I tried, `return redirect("..")` is the simplest and sufficient.

suricactus: `HttpResponseRedirect(".")` is something that was already used as a pattern in the code

docker-app/qfieldcloud/core/templates/admin/job_change_form.html
dddpt: done

suricactus (approved):

PR #891, opened 2024-02-21
QF-3851 Remove S3 ACL parameter for avatar/projects thumbnails

dddpt:
Works fine for thumbnail, but I get `403 forbidden` for avatar at https://localhost/api/v1/files/public/users/<USER>/avatar.svg
Seems independant of `ACL` policy...

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -293,7 +293,6 @@ def upload_user_avatar(
         file,
         key,
         {
-            "ACL": "public-read",
             "ContentType": mimetype.value,
         },
     )
@@ -357,8 +356,6 @@ def upload_project_thumbail(
         file,
         key,
         {
-            # TODO most probably this is not public-read, since the project might be private
-            "ACL": "public-read",
             "ContentType": mimetype,
         },
     )


duke-nyuki:
Task linked: [QF-3851 Remove S3 ACL parameter for avatar/projects thumbnails](https://app.clickup.com/t/2192114/QF-3851)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -293,7 +293,6 @@ def upload_user_avatar(
         file,
         key,
         {
-            "ACL": "public-read",
```
suricactus, line 296: Ah, this one is more special. Is it the same django view-handling case as for the project thumbnail?

dddpt: Avatars are handled by [PublicFilesView](https://github.com/opengisch/qfieldcloud/blob/5fd1f5c338efd34c9b800760c6cda389bbdd1429/docker-app/qfieldcloud/core/views/files_views.py#L387)

Project thumbnails are handled by [ProjectMetafilesView](https://github.com/opengisch/qfieldcloud/blob/5fd1f5c338efd34c9b800760c6cda389bbdd1429/docker-app/qfieldcloud/core/views/files_views.py#L367C7-L367C27)

In any case, since the other s3 update where we use `s3.generate_presigned_url()`, "ACL": "public-read", is definitely no longer needed.

PR #893, opened 2024-02-22
QF-3857 Added dark theme

dddpt:
However, doesn't show as advertised.

For example, for the `superhero` dark theme, it is advertised in the [documentation](https://django-jazzmin.readthedocs.io/ui_customisation/#dark-mode-enabled) as looking like [this](https://bootswatch.com/superhero/).

If I use the integrated jazzmin UI tweaker, I see this, which is ok (note the grey buttons on the right which is not nice in any case):
Screenshot from 2024-02-22 15-15-11

However, If I set `"dark_mode_theme": "superhero"` and my browser to dark mode, I see this, which is rather unpalatable:
Screenshot from 2024-02-22 15-15-24

To be discussed, but the jazzmin dark themes don't seem so stable/ready yet.

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -629,4 +629,9 @@ JAZZMIN_SETTINGS = {
     "copyright": "OPENGIS.ch",
     # Additional custom CSS file for the Django Admin pages.
     "custom_css": "css/admin.css",
+    "show_ui_builder": True,
+}
+JAZZMIN_UI_TWEAKS = {
+    # CSS dark theme used, options here: https://django-jazzmin.readthedocs.io/ui_customisation/#light-themes
+    "dark_mode_theme": "superhero",
 }


duke-nyuki:
Task linked: [QF-3857 Jazzmin dark theme](https://app.clickup.com/t/2192114/QF-3857)

docker-app/qfieldcloud/settings.py
```diff
@@ -629,4 +629,9 @@ def before_send(event, hint):
     "copyright": "OPENGIS.ch",
     # Additional custom CSS file for the Django Admin pages.
     "custom_css": "css/admin.css",
+    "show_ui_builder": True,
```
suricactus, line 632: Shall we keep this? Or make it a setting for example?

dddpt: `show_ui_builder` is more about letting us devs test settings that we want to put in `settings.py`, definitely not to be kept.

Anyway, to be discussed in the sprint meeting ;-)

dddpt:
As stated in the PR opening, jazzmin admin dark mode is not yet production-ready, so giving up for now.
To be reopened if something changes on the jazzmin side ;-)

PR #894, opened 2024-02-24
Speed up packaging

nirvn:
This PR reworks the packaging job workflow to prevent _3_ project loads and instead rely on a single initial load. This can lead to massive speed gains on complex projects and/or projects containing remote sources that are slow.

Requires the following libqfieldsync improvements: https://github.com/opengisch/libqfieldsync/pull/68

### DIFF
--- a/docker-qgis/entrypoint.py
+++ b/docker-qgis/entrypoint.py
@@ -29,18 +29,13 @@ logger = logging.getLogger("ENTRYPNT")
 logger.setLevel(logging.INFO)
 
 
-def _call_libqfieldsync_packager(
-    project_filename: Path, package_dir: Path, offliner_type: OfflinerType
-) -> str:
+def _call_libqfieldsync_packager(package_dir: Path, offliner_type: OfflinerType) -> str:
     """Call `libqfieldsync` to package a project for QField"""
     logger.info("Preparing QGIS project for packaging…")
 
     project = QgsProject.instance()
-    if not project_filename.exists():
-        raise FileNotFoundError(project_filename)
-
-    if not project.read(str(project_filename)):
-        raise Exception(f"Unable to open file with QGIS: {project_filename}")
+    if not project.fileName():
+        raise Exception("Project is empty")
 
     layers = project.mapLayers()
     project_config = ProjectConfiguration(project)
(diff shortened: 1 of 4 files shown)


suricactus:
The decision to load the project 3 times was by design to make each step in a Job workflow independent from the rest. Now they will become tightly coupled and "load project" becomes mandatory. Do we have any specific proof how much do we gain for such an important change that is introduced here?

+1 in putting the logic in a single function, but I would prefer to still call it three times, unless clear evidence we really save 20+ seconds on this.

docker-qgis/entrypoint.py
```diff
@@ -138,15 +133,11 @@ def _call_libqfieldsync_packager(
...
+def _extract_layer_data() -> dict:
     logger.info("Extracting QGIS project layer data…")
 
-    project_filename = str(project_filename)
     project = QgsProject.instance()
-
-    with set_bad_layer_handler(project):
-        project.read(project_filename)
-        layers_by_id: dict = get_layers_data(project)
+    layers_by_id: dict = get_layers_data(project)
```
suricactus, line 141: This new approach is exploiting that the `BadLayerHandler` is not cleaning the bad layer handler info after the project load is done. This should not be the case and the `BadLayerHandler` might change the way it works as soon as a QGIS crash is fixed.

nirvn:
@suricactus , the speed gains are easy to replicate. Just take a QGIS project that loads in 10 seconds, then multiply that by 3 to get the time needlessly spent re-loading it 3 times when we can load it only once.

For one of our clients - the Groupements Forestiers Quebec - , a packaging job takes ~5 minutes, each project load taking 1 minute or so on the cloud server. I can provide job IDs, but I think we already looked at one together yesterday.

nirvn:
Avoiding the CPU cost of reloading a project three times is also not negligible, that surely reduces stress on the server infrastructure. Project loading is costly unless you have a single geopackage + an OSM basemap layers (e.g., every time you load the project, you re-create layouts, relationships, 2D and 3D map views, refetch metadata of layers for some data providers, etc.).

Here's a packaging that lasted >4 minutes: https://app.qfield.cloud/a/ggeneau/octobre_2023/jobs?id=91a62569-2a24-454a-b9eb-53ea66292374&job-preview=1#

suricactus:
Would be great to monitor back the actual speed improvement of these jobs. I have the gut feeling it will be negligible.

suricactus (approved):

suricactus:
Ready to merge.

PR #902, opened 2024-03-13
Fix setting status when status SIGKILL is sent by docker

suricactus:


duke-nyuki:
Task linked: [QF-3922 The job `status` field not saved on docker SIGKILL](https://app.clickup.com/t/2192114/QF-3922)

PR #903, opened 2024-03-13
Add note about changing subscription in the admin template

suricactus:


duke-nyuki:
Task linked: [QF-3901 Read the plan field in the Subscriptions, make it read-only for editing](https://app.clickup.com/t/2192114/QF-3901)

PR #905, opened 2024-03-14
If the srid is unsupported, just do not store delta geom instead of throw

suricactus:


duke-nyuki:
Task linked: [QF-3926 Changes fail to reach QFC if the delta uses SRID that is not supported by the PostGIS db storing old/new geom of the delta](https://app.clickup.com/t/2192114/QF-3926)

suricactus:
@boardend can you have a quick look if you find something fishy here? For me it's ok, but a post-merge review is better than nothing.

PR #909, opened 2024-03-18
Force garbage collection of all dangling QGIS objects around

suricactus:
Check the commit messages for more info.

### DIFF
--- a/docker-qgis/entrypoint.py
+++ b/docker-qgis/entrypoint.py
@@ -354,7 +354,7 @@ def cmd_process_projectfile(args: argparse.Namespace):
                 id="generate_thumbnail_image",
                 name="Generate Thumbnail Image",
                 arguments={
-                    "project": StepOutput("opening_check", "project"),
+                    "project_filename": WorkDirPathAsStr("files", args.project_file),
                     "thumbnail_filename": Path("/io/thumbnail.png"),
                 },
                 method=qfc_worker.process_projectfile.generate_thumbnail,
--- a/docker-qgis/entrypoint.py
+++ b/docker-qgis/entrypoint.py
@@ -16,12 +16,13 @@ from qfc_worker.utils import (
     Step,
     StepOutput,
     WorkDirPath,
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-3927 QGIS crashes on certain projects](https://app.clickup.com/t/2192114/QF-3927)

docker-qgis/entrypoint.py
```diff
@@ -146,9 +147,20 @@ def _extract_layer_data(project_filename: Union[str, Path]) -> dict:
     return layers_by_id
 
 
-def _open_project(project_filename: Union[str, Path]):
+def _open_project(project_filename: str) -> QgsProject:
```
nirvn, line 150: I think I'd rename this function to _open_read_only_project to avoid someone else using it in the future not appreciating the read-only nature of this method.

docker-qgis/qfc_worker/process_projectfile.py
```diff
@@ -99,47 +125,66 @@ def on_project_read(doc):
...
+    )
+    tmp_project.readProject.connect(
+        on_project_read_wrapper(tmp_project, tmp_layer_tree)
+    )
+    tmp_project.read(
+        project_filename,
+        tmp_project_read_flags,
+    )
 
     renderer = QgsMapRendererParallelJob(map_settings)
```
nirvn, line 189: Instead of the QEventLoop stuff, I'd just use renderer.waitForFinished()

docker-qgis/entrypoint.py
suricactus: Ideally I want to get the arguments already from the Workflow framework, but for now I will rename the function

docker-qgis/qfc_worker/process_projectfile.py
suricactus: We did it in another way, see the last commit.

nirvn (approved):
Very good stuff.

PR #912, opened 2024-03-19
Fix typo to fix timezones for users from the other side of the world

suricactus:
We did the wrong check to get the user timezone and we never entered this if body. Note we already assume that the user has a `useraccount` which is safe assumption.


duke-nyuki:
Task linked: [QF-3953 Incorrect timezome in the Project -> Changes page, shows UTC time instead of user's timezone](https://app.clickup.com/t/2192114/QF-3953)

nirvn (approved):
Ahhh, on behalf of the world, thanks. It'll make it so much easier to debug testing when I can rely on times being mine :wink:

PR #914, opened 2024-03-20
Fix missing active subscription creation by reverting the timezone fixes in #912

suricactus:
This reverts commit 8510d2369a5b53fda10c99f7cd683c861a6b0638. 

Reverts #912 . 

Prevents creating subscriptions active in the future. Proper fix will be issued in calmer times, sorry fellow friends from other parts of the world.


duke-nyuki:
Task linked: [QF-3964 When creating a new organization, the subscription `active_since` field takes value because of the user's timezone](https://app.clickup.com/t/2192114/QF-3964)

PR #917, opened 2024-03-23
Handle deltas with WKT geometry as empty string instead of null

suricactus:
Sometimes the deltas are sent with empty string instead of null for geometry. Since one cannot know what to expect from those users, better prepare for the future!

Note this should be already handled, as `delta_srid` would be either NULL or empty string, but double ensurance is better.

Related PR: https://github.com/opengisch/QField/pull/5139
Fix: https://github.com/opengisch/QField/issues/4796


duke-nyuki:
Task linked: [QF-3986 Fix delta not being saved when geometry WKT is an empty string](https://app.clickup.com/t/2192114/QF-3986)

PR #922, opened 2024-03-28
Do not remove broken WMS layers during packaging

suricactus:
See also https://github.com/opengisch/libqfieldsync/pull/76


duke-nyuki:
Task linked: [QF-4035 Do not remove the WMS layers if they are temporarily inaccessible from QFieldCloud packager](https://app.clickup.com/t/2192114/QF-4035)

PR #924, opened 2024-04-04
Fix localized path regression

nirvn:
@suricactus , am I missing something completely obvious here or are we missing a setBadLayerHandler call since v0.26.0?

### DIFF
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -20,7 +20,10 @@ from pathlib import Path
 from typing import IO, Any, Callable, NamedTuple, Optional
 
 from libqfieldsync.layer import LayerSource
-from libqfieldsync.utils.bad_layer_handler import set_bad_layer_handler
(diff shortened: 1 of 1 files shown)


suricactus:
@nirvn This should be handled by https://github.com/opengisch/qfieldcloud/commit/eaab36fcda0b04e13ee76d7c47eb3cd5fdc180f7#diff-09bee065a5c7767fb1960df829f4b0d812ede9b3f9299de26321eec6b42cb514R220 

I did not check why not, I will be able to do so after 15th April.

I guess with the other crash-free fixes, we can check if libqfieldsync `BadLayerHandler` can work as it was initially planned to:
https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/utils/bad_layer_handler.py#L55-L69

nirvn:
@suricactus , that line will not attach the bad layer handler to the project though, code requires us to do it manually. That's why it's not working. I've tested it when adding a test case in libqfieldsync.

PR #925, opened 2024-04-09
Fix init_letsencrypt.sh CONFIG_PATH

dddpt:
Problem for nginx cert update was `init_letsencrypt.sh` uses `CONFIG_PATH=./conf`, but qfieldcloud-nginx config has been moved to the `docker-nginx` folder.

Hence changing `CONFIG_PATH=./` by default and completing the subsequent paths in `init_letsencrypt.sh`

### DIFF
--- a/scripts/init_letsencrypt.sh
+++ b/scripts/init_letsencrypt.sh
@@ -7,12 +7,12 @@ set -o allexport
 source .env
 set +o allexport
 
-CONFIG_PATH="${CONFIG_PATH:-'./'}"
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-if [ ! -e "${CONFIG_PATH}/docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "${CONFIG_PATH}/docker-nginx/ssl-dhparams.pem" ]; then
+if [ ! -e "${SCRIPT_DIR}/../docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "${SCRIPT_DIR}/../docker-nginx/ssl-dhparams.pem" ]; then
   echo "### Downloading recommended TLS parameters ..."
-  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot-nginx/certbot_nginx/_internal/tls_configs/options-ssl-nginx.conf > "${CONFIG_PATH}/docker-nginx/options-ssl-nginx.conf"
-  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot/certbot/ssl-dhparams.pem > "${CONFIG_PATH}/docker-nginx/ssl-dhparams.pem"
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot-nginx/certbot_nginx/_internal/tls_configs/options-ssl-nginx.conf > "${SCRIPT_DIR}/../docker-nginx/options-ssl-nginx.conf"
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot/certbot/ssl-dhparams.pem > "${SCRIPT_DIR}/../docker-nginx/ssl-dhparams.pem"
   echo
 fi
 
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-4050 fix init_letsencrypt.sh CONFIG_PATH](https://app.clickup.com/t/2192114/QF-4050)

scripts/init_letsencrypt.sh
```diff
@@ -7,12 +7,12 @@ set -o allexport
 source .env
 set +o allexport
 
-CONFIG_PATH="${CONFIG_PATH:-'./conf'}"
+CONFIG_PATH="${CONFIG_PATH:-'./'}"
 
-if [ ! -e "docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "docker-nginx/ssl-dhparams.pem" ]; then
+if [ ! -e "${CONFIG_PATH}/docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "${CONFIG_PATH}/docker-nginx/ssl-dhparams.pem" ]; then
```
faebebin, line 12: in case of having .env in a parent repo it would either not find `.env` or the `./docker-nginx/...` depending on what location you run the script from...

dddpt: If I am not mistaken, the script is meant to be run from where you have your `.env` file.
So in the case of a parent repo with its `.env` file, you run it from there and the `CONFIG_PATH` variable allows to point to the correct path to find the `conf/` and `docker-nginx/` folders.

dddpt:
> Should not also the `docker-compose.yml` volume mount be included in the fix somehow?

Good question! I just tested: `docker-compose` handles relative paths gracefully.

It knows that paths given in `docker-compose.yml` are  relative to its own folder.

scripts/init_letsencrypt.sh
```diff
@@ -7,12 +7,12 @@ set -o allexport
 source .env
 set +o allexport
 
-CONFIG_PATH="${CONFIG_PATH:-'./'}"
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
faebebin, line 10:
```suggestion
SCRIPT_DIR=$(dirname "$(realpath "$0")")
```
would be a bit more elegant?

faebebin (changes requested):
thanks alot, looks good

Only one suggestion to get the script-dir bit more elegant

else good to merge for me

scripts/init_letsencrypt.sh
m-kuhn: I am not sure I am the correct person to ask. In any case, no objection from me.

faebebin (approved):

faebebin:
@dddpt can you test once again that this final version works. If it does you can merge it :+1: 

thanks alot

scripts/init_letsencrypt.sh
```diff
@@ -7,12 +7,12 @@ set -o allexport
 source .env
 set +o allexport
 
-CONFIG_PATH="${CONFIG_PATH:-'./conf'}"
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
```
faebebin, line 10: sorry forgot one thing:

Here we should use already something like

`ROOT_DIR="../${SCRIPT_DIR}"` but in one expression if possible

instead od having to `..` all the time after

faebebin (changes requested):

scripts/init_letsencrypt.sh
```diff
@@ -7,12 +7,12 @@ set -o allexport
...
 
-if [ ! -e "docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "docker-nginx/ssl-dhparams.pem" ]; then
+if [ ! -e "${SCRIPT_DIR}/../docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "${SCRIPT_DIR}/../docker-nginx/ssl-dhparams.pem" ]; then
```
dddpt, line 12:
```suggestion
if [ ! -e "${root_dir}/docker-nginx/options-ssl-nginx.conf" ] || [ ! -e "${root_dir}/docker-nginx/ssl-dhparams.pem" ]; then
```

dddpt:
tested on staging :ok_hand:

suricactus (commented):
Thanks all for the simplification and the fix. This was a leftover from pre-upstreaming some changes in February.

PR #938, opened 2024-04-23
Remove unnecessary QFIELDCLOUD_DEFAULT_NETWORK env

faebebin:


suricactus:
Can we still keep the variable with empty default value, but when the value is empty, get the network name as you do in this PR?

faebebin:
I was working towards reducing env vars.

But if you want to use it for running stacks in parallel or experimenting with workers? 
Though I do not directly see a usecase where the network is not derived from `COMPOSE_PROJECT_NAME` ... ?

suricactus (approved):
LGTM!

PR #945, opened 2024-05-15
Bump to latest Ubuntu 24.04 and QGIS 3.36.2

suricactus:
This will automagically bump `gdal` from 3.4.1 to 3.8.4. We actually care that the GPKG patch from 3.4.2 is present. See the QGIS patch here: https://github.com/qgis/QGIS/pull/47098/


duke-nyuki:
Task linked: [QF-4108 Bump QGIS container to latest QGIS and Ubuntu LTS](https://app.clickup.com/t/2192114/QF-4108)

faebebin (approved):
Tested locally the case with two gpkg layers. And it correctly only uploads the modified one :partying_face: 

Another probaly longliving bug solved!!
Thanks alot @suricactus

PR #951, opened 2024-05-27
Fix the failing worker_wrapper and tests by requests==2.32.0

suricactus:
The CI tests returned error messages like:
> Exception: Processing did no finish, did the worker hang ?

This was cause because the `worker_wrapper` gets on start:
> docker.errors.DockerException: Error while fetching server API version: Not supported URL scheme http+docker

NOTE since we are using multi-stage builds in docker, the `requests`
dependency was forced on the previous layer than the `worker_wrapper`
itself.

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -629,7 +629,7 @@ def get_project_file_storage_in_bytes(project_id: str) -> int:
     total_bytes = 0
     prefix = f"projects/{project_id}/files/"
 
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-4311 Fix the failing `worker_wrapper` and tests by `requests==2.32.0`](https://app.clickup.com/t/2192114/QF-4311)

boardend (approved):

boardend:
- Tested on my machine :+1: 
- When also the CI tests pass, this is good to go

PR #958, opened 2024-06-04
Limit QGIS project modifications to managers and admins

boardend:
Add a `has_restricted_projectfiles` flag to the project in order to restrict modifications of the QGIS project file to users that have the "Manager" or "Administrator" role

Screenshot from 2024-06-17 15-59-53


duke-nyuki:
Task linked: [QF-3822 Limit .qgs file edit to admins](https://app.clickup.com/t/2192114/QF-3822)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1016,6 +1016,14 @@ class Meta:
             "If enabled, QFieldCloud will automatically overwrite conflicts in this project. Disabling this will force the project manager to manually resolve all the conflicts."
         ),
     )
+
+    restrict_project_modification = models.BooleanField(
```
suricactus, line 1020:
```suggestion
    is_project_modification_restricted = models.BooleanField(
```

nirvn:
Open question: should we also limit modification of project plugins to admins? I think so

boardend:
> Open question: should we also limit modification of project plugins to admins? I think so

Sounds reasonable, but the question is up to @suricactus

If so, let's add a new task for that; This probably doesn't have to be configurable on the project like the `is_projectfile_restricted` flag. I would just add a new rule on the `POST` of `DownloadPushDeleteFileView`

nirvn:
@boardend , I would tie this to is_projectfile_restricted. I can't see need for further permission granularity to cover the project plugin .qml vs the project file .qgz

nirvn:
@boardend , you'll also need to take into consideration another project sidecar file. When saved as .qgs, QGIS saves an _attacments.zip file too. Logic should be updated here to catch all project sidecar files: 

https://github.com/opengisch/qfieldcloud/blob/330348bc54e18e272441b1a96cf8525aa624e4af/docker-app/qfieldcloud/core/utils.py#L238

suricactus:
Let's leave the plugins out of the scope of this PR.

As for the `_attachments.zip`, we had a dev discussion last week we can ignore it for now, as there are a few other places where this should be updated. We also need to verify what is the QGIS convention for the naming, I think it is `<qgs_file_basename>_attachhments.zip` but need to verify in QGIS source code. We should take this into consideration and not only look for the suffix, as I might have `family_photos_attachments.zip` as a valid file which is not a sidecar for `.qgs`.

IMO we should create a function `is_protected_project_file`, instead of raw `path.suffix.lower() in (".qgs", ".qgz")`. This way in the future we can add more sensitive files to that function.

nirvn:
@suricactus , you are correct about the convention. And yeah, same logic to identify the plugin would be used here. 

- Step 1: identify the one project file in the directory
- Step 2: identify <project file basename>_attachments.zip as protected
- Step 3: identify <project file basename>.qml as protected

The function your suggesting would allow for emancipation of that logic too, which is useful.

boardend:
@nirvn & @suricactus I just created the follow up task to refactor the QGIS/QField project file detection: https://app.clickup.com/t/2192114/QF-4375

boardend:
Did a final updated based on the discussion with @suricactus 

Let's accept this one and I will create a new PR to respect all QGIS/QField related projectfiles

suricactus (approved):
Thanks! In a follow-up would be great to add a test.

PR #963, opened 2024-06-18
Set initial created subscription plan status to ACTIVE_PAID

boardend:
When hosting a new QFieldCloud instance, a default subscription plan is created. This change ensures that the status of this plan is `ACTIVE_PAID`, so that created users can create/use projects on that instance as expected.

This change won't affect existing QFieldCloud instances in any way.

### DIFF
--- a/docker-app/qfieldcloud/subscription/migrations/0004_auto_20220923_1602.py
+++ b/docker-app/qfieldcloud/subscription/migrations/0004_auto_20220923_1602.py
@@ -77,6 +77,14 @@ def add_packages_to_accounts(apps, schema_editor):
         package.save()
 
 
+def set_existing_plans_to_active_status(apps, schema_editor):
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-3925 On public, when creating a user, set its default subscription status to "active paid"](https://app.clickup.com/t/2192114/QF-3925)

suricactus:
The `mypy` error here is because `QuerySet`  cannot take a typing arguments, feel free to fix it (I had to fix it for one of my branches I haven't pushed yet too). The other error in `core`... I don't know, let's retrigger the checks.

boardend:
@suricactus thanks for looking into this

Just did a test with a new `postgres_data` volume, executed the migrations and created an admin via the CLI. When then creating a new user via the admin, the user as a "Active Paid" billing subscription by default  :+1:

PR #965, opened 2024-06-19
Fix typing for QuerySet

suricactus:


suricactus:
Closing, there is a better fix coming :)

PR #969, opened 2024-06-21
Prevent error 500 when modifying a Person with an email that already exists

suricactus:


duke-nyuki:
Task linked: [QF-4431 Prevent error 500 when saving a Person with already existing email](https://app.clickup.com/t/2192114/QF-4431)

docker-qgis/Dockerfile
boardend, line 1: That feels a bit off topic here :) But I guess this change is intended?

boardend (approved):

docker-qgis/Dockerfile
suricactus: Yes, I put it in the commit message [We already have quite fresh pip, no need to force upgrade](https://github.com/opengisch/qfieldcloud/pull/969/commits/46d72da9cbaf1199cb940fda03582518eb23c538)

PR #974, opened 2024-06-26
Return sha256sum metadata if it does not exist.

wdcsimon:
Added code to return sha256sum metadata if it does not exist.
Error can occur if the data has been migrated from one s3 platform to another and metadata is lost.

### DIFF
--- a/docker-app/qfieldcloud/core/views/files_views.py
+++ b/docker-app/qfieldcloud/core/views/files_views.py
@@ -111,8 +111,14 @@ class ListFilesView(views.APIView):
                 metadata = head["Metadata"]
                 if "sha256sum" in metadata:
                     sha256sum = metadata["sha256sum"]
-                else:
+                elif "Sha256sum" in metadata:
                     sha256sum = metadata["Sha256sum"]
+                else:
+                    # Return "sha256sum" metadata if it does not exist.  Can occur if the data has been migrated from one s3 platform to another.
+                    s3 = utils.get_s3_client()
+                    obj = s3.get_object(Bucket=bucket.name, Key=version.key)
+                    data = io.BytesIO(obj['Body'].read())
+                    sha256sum = utils.get_sha256(data)
 
                 version_data["sha256"] = sha256sum
 


suricactus:
Hey! Thank you for your contribution!

There are few concerns I want to raise before proceeding with this PR.

1) The most obvious one is this is going to be painfully slow. Even if we agree this is ok for the first download after the said migration (which I am also not sure if it is really acceptable), the caclulated `sha256sum` is not stored permanently with the s3 objects. I would argue that the migration script you have used should take care the metadata is present, and QFieldCloud should always safely assume it is present.

2) Actually the `sha256sum` is deprecated and slowly phased out. We mostly use the `md5sum` instead, which is based on the `etag` calculated for the S3 object. The `etag` is integral part of the S3 design and is calculated anyway, so no need for additional metadata requests and calculations. Most of our official API clients already migrated from `sha256sum` to `md5sum`, I think only QFieldSync is still using `sha256sum` (need to double check this). That's why there is also `skip_metadata` flag available for the files endpoint. Consider using the `skip_metadata` parameter for your use-case. If you need to use QFieldSync (or other official API clients) and you are blocked by the deprecated `sha256sum`, please drop us a message at sales@qfield.cloud and we can find a way to speed up the development of using the new `md5sum` property instead. You gain speed and get future proof.

3) There is a very big change coming on files management in QFieldCloud, where most of the files metadata wll be stored within the database, instead of using S3 objects implementation details. This has multiple benefits:
- we will no longer be dependent on S3 versioning support (flexibility on S3 providers);
- we will no longer be dependent on S3 at all, one can use their host FS (flexibility on storage providers);
- we will no longer be dependent on S3 API calls to retrieve data about your files, it will be called only when one really does something with the file contents (speed, a lot of speed);
- ... and the list goes on
This change is expected to land in this repo in the end of the summer/early autumn. There will be a migrations scripts and so on.

I hope I have convinced while your contributions might fix your particular issue, probably is not of benefit for the upstream as of the upcoming months and the general direction this project is heading to. 

I also hope this long message will not discourage you to do future PRs on this repositories, they are more than welcome!

wdcsimon:
Thanks for the detailed reply, I'm happy to wait for the dm5sum development for QFieldSync - that will be a far better solution.  In the meantime, this fix solves my immediate problem even though getting file listing in QFieldSync is pretty slow.  Looking forward to the new file management development in QFieldCloud!

Issue #975, opened 2024-06-27
Unknow instruction Types in Dockerfile


vincentBenet:
Trying to install this project on my Synology NAS, I have this error while building project:

```
...

#9  [qfieldcloud-worker_wrapper internal] load metadata for docker.io/librairy/python:3.10.8-slim-buster
#9 CANCELED
failed to solve rpc error code = unknown desc = failed to solve with frontend dockerfile.v0: failed to create LBB definition: dockerfile parse error line 14: Unknow instruction: TYPES:
```

This is the YAML provided:

```
version: "3.9"

x-logging: &default-logging
  driver: "json-file"
  options:
    max-size: "100m"
    max-file: "10"
    tag: "{{.ImageName}}|{{.Name}}|{{.ImageFullID}}|{{.FullID}}"

services:
  app: &default-django
    build:
      context: ./docker-app
      target: webserver_runtime
      network: host
    restart: unless-stopped
    command: >
      gunicorn
        qfieldcloud.wsgi:application
        --bind 0.0.0.0:8000
        --timeout 300
        --max-requests 300
        --workers 3
        --threads 3
    volumes:
      - static_volume:/usr/src/app/staticfiles
      - media_volume:/usr/src/app/mediafiles/
    environment:
      DJANGO_ALLOWED_HOSTS: "localhost 127.0.0.1 0.0.0.0 app nginx"
      DJANGO_SETTINGS_MODULE: qfieldcloud.settings
      SECRET_KEY: 1234567890
      DEBUG: 1
      ENVIRONMENT: test
      SENTRY_DSN: 
      SENTRY_RELEASE: dev
      SENTRY_ENVIRONMENT: test
      SENTRY_SAMPLE_RATE: 1
      POSTGRES_DB: qfieldcloud_db
      POSTGRES_DB_TEST: test_qfieldcloud_db
      POSTGRES_USER: qfieldcloud_db_admin
      POSTGRES_PASSWORD: 3shJDd2r7Twwkehb
      POSTGRES_HOST: db
      POSTGRES_PORT: 5432
      POSTGRES_SSLMODE: prefer
      STORAGE_ACCESS_KEY_ID: minioadmin
      STORAGE_SECRET_ACCESS_KEY: minioadmin
      STORAGE_BUCKET_NAME: qfieldcloud-local
      STORAGE_REGION_NAME: 
      STORAGE_ENDPOINT_URL: http://172.17.0.1:8009
[...]

PR #976, opened 2024-06-27
Patch Dockerfile: EOF not valid under all build OS

vincentBenet:
Issue #975 Patch: Unknow instruction Types in Dockerfiles

The block COPY with EOF syntax return error under Synology container Builder


suricactus:
Thanks for the contribution. Looks ok for me, there should not be anyone hurt by this change.

PR #978, opened 2024-06-29
Set new user's default timezone to the settings one

gounux:
When creating a new user in the QFieldCloud admin interface, the user's timezone is always "Europe/Zürich", even if the `QFIELDCLOUD_DEFAULT_TIME_ZONE` / `TIME_ZONE` variables are different

This PR sets the timezone of a new user to the env timezone by default


suricactus (approved):

suricactus:
Thanks for spotting and fixing this eurocentric view we had :)

PR #979, opened 2024-07-02
Patch admin constance styles

faebebin:
Fixes styling in admin constance config page by adding bootstrap classes on two templates. Overrides 2 lines of code of the upstream templates.

makes this ...
image

look like this ...
image

NOTE an upstream pull was intended but would require rework the whole styling fix without boostrap probably and therefore stopped for now [(upstream PR closed):](https://github.com/jazzband/django-constance/pull/559)


suricactus (commented):
Happy with the new look, not so happy with the new HTML we have to maintain. But I guess you did your research, tried to push upstream, so let's agree this is the best approach for now.

suricactus (approved):

faebebin:
> Happy with the new look, not so happy with the new HTML we have to maintain. But I guess you did your research, tried to push upstream, so let's agree this is the best approach for now.

yes actually did an upstream PR but they said cannot not use the bootsrap classe (ofcourse). But I was too lazy to reconstruct with django native css only.
But I commented hopefully clearly enough that only added two classes to the CSS. So if things break at some point we just remove the patch or add it ontop of the new upstream html. Also is only one admin page affected therefore IMO not worth alot of effort.

PR #982, opened 2024-07-10
When adding a collaborator via the API, add created_by/updated_by

suricactus:
In addition, prevent attempts to add other `User` subclasses as project
collaborators

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -188,20 +188,34 @@ class OrganizationSerializer(serializers.ModelSerializer):
 
 class ProjectCollaboratorSerializer(serializers.ModelSerializer):
     collaborator = serializers.StringRelatedField()
-    role = serializers.CharField()
+    created_by = serializers.StringRelatedField()
+    updated_by = serializers.StringRelatedField()
 
     class Meta:
         model = ProjectCollaborator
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-4534 Ensure API endpoints for adding collaborators and members works](https://app.clickup.com/t/2192114/QF-4534)

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -188,20 +188,35 @@ class Meta:
...
+            "updated_by",
+            "created_at",
+            "updated_at",
+        )
 
 
 class OrganizationMemberSerializer(serializers.ModelSerializer):
+    organization = serializers.StringRelatedField()
     member = serializers.StringRelatedField()
     role = serializers.CharField()
```
faebebin, line 210:
```suggestion
```

faebebin (changes requested):
Looks good

One little cleanup suggestion

else great, thanks

docker-app/qfieldcloud/core/serializers.py
suricactus: Ah good spot.

faebebin (approved):
LGTM

just the one linting fix, but else good
thanks

suricactus:
The linting was fixed on `master`. Thanks!

PR #989, opened 2024-07-15
Fix admin constance styles with css only

faebebin:
This https://github.com/opengisch/qfieldcloud/pull/979 was a bit lazy way but rather just add some styles instead

The result should be the same
image


suricactus (approved):
Less code, more profit! Thanks for going back at this and fixing the things in the more maintainable way <3

PR #993, opened 2024-07-25
Enhance shell_plut with ipython if DEBUG_BUILD

faebebin:
I like ipython for block editing and much more. If installed, is used as default by `django manage.py shell_plus` . 
Suggest add it if DEBUG_BUILD is set.

### DIFF
--- a/docker-app/Dockerfile
+++ b/docker-app/Dockerfile
@@ -53,7 +53,7 @@ COPY --from=build /usr/local/lib/python3.10/site-packages/ /usr/local/lib/python
 
 # install debug dependencies
 ARG DEBUG_BUILD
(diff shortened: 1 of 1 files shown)


suricactus (approved):
I am good with it, would be great to share at some dev meeting how you use this.

faebebin:
@suricactus django shell will automatically chose ipython as its default. So nothing else to do.

Besides alot of handy stuff (see https://realpython.com/ipython-interactive-python-shell/ ) I mostly prefer the block editing. If you e.g. define a function or class in the shell and and press up arrow it navigates history in blocks, not lines. Further nicely supports vi keybindings and integrates alot of bash commands to use directly in the ipython shell etc

PR #997, opened 2024-08-01
Use optimized packager by default for new projects

gounux:
Since the `QGIS Core Offline Editing` packager is deprecated, why not use the "Optimized packager" for new project as default ?

image

### DIFF
--- a/docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
+++ b/docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
@@ -17,7 +17,7 @@ class Migration(migrations.Migration):
                     ("qgiscore", "QGIS Core Offline Editing (deprecated)"),
                     ("pythonmini", "Optimized Packager"),
                 ],
-                default="qgiscore",
+                default="pythonmini",
                 help_text='The Packaging Offliner packages data for offline use with QField. The new "Optimized Packager" should be preferred over the deprecated "QGIS Core Offline Editing" for new projects.',
                 max_length=100,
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/migrations/0073_project_packaging_offliner.py
```diff
@@ -17,7 +17,7 @@ class Migration(migrations.Migration):
                     ("qgiscore", "QGIS Core Offline Editing (deprecated)"),
                     ("pythonmini", "Optimized Packager"),
                 ],
-                default="qgiscore",
+                default="pythonmini",
```
suricactus, line 20: Instead of editing an already existing migration, please create a new migration with this change only.

suricactus (changes requested):
Thanks for your contribution! You are right that it is a bit strange why the deprecated one was used as default for quite some time. The reason was to allow test period with the old/new packagers. We have more that positive feedback with the new packager and this change should have happened already a few months ago, but it always slipped from the sprints.

Please use new migrations when changing models instead of changing already existing ones. While it will not change much in this case, it make it harder to follow the evolution of the code and the current behaviour of django might change in the future.

Also please rebase on the latest master and run pre-commit, there are some failures caused by the migration to `ruff` a few months ago: https://github.com/opengisch/qfieldcloud/actions/runs/10525606898/job/29164731554

Your change is more than welcome. Thanks again for contributing upstream!

suricactus:
Closing due to inactivity. Will open a PR myself.

gounux:
thanks @suricactus

PR #998, opened 2024-08-02
QF-4641 Deployment path decision for letsencrypt certificates

anajoy:
QFieldCloud build and run in containers extended by feature request to new support automated renewal of certificates by letsencrypt beside the already supported self-singed certificates by mkcert.

The code was handed over to me more or less working regarding the new functionality added. I further tested the certificate creation and automated renewal managed by certbot. I did now mainly test and adapt the development build for (local) dev and deployment build for test and production in the way, that both certificate creation methods can coexist with the same configuration file derived from .env.example as described in the README markdown.
The main change from my side is, that the deployer should decide about the certificate setup in the beginning, before continuing with any further build steps.
A review is appreciated


duke-nyuki:
Task linked: [QF-4641 Certificate autorenewal](https://app.clickup.com/t/2192114/QF-4641)

scripts/init_mkcert.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+if [[ ! -d ./docker-nginx/certs ]]; then
```
dddpt, line 3:
```suggestion
if [[ ! -d ./docker-nginx ]]; then
```
`./docker-nginx/certs` does not exist in a qfieldcloud installation from scratch, checking for `./docker-nginx` should be sufficient.

scripts/init_mkcert.sh
```diff
@@ -0,0 +1,44 @@
...
+docker stop qfieldcloud-mkcert-1
+
+echo "Unset COMPOSE_FILE"
+unset COMPOSE_FILE
+echo ${COMPOSE_FILE}
+
+echo "Deleting container qfieldcloud-mkcert-1"
+docker container rm qfieldcloud-mkcert-1
+
+echo "After deploying nginx connecting to localhost with curl should return no errors: curl https://localhost:8002/"
```
dddpt, line 44: When starting QFieldCloud using `sudo docker compose up -d --build`, there is nothing on https://localhost:8002 ("unable to connect"). Is it linked to some config in `.env`?

dddpt:
So, I was able to start up QFieldCloud following the README instructions, only a small change in `init_mkcert.sh` to do.

However, I still get a untrusted certificate error on the first connection as I should not, isn't it?
The errror:
Screenshot from 2024-08-14 16-04-32

The rootCA seems to be in my list of trusted certificates:
Screenshot from 2024-08-14 16-05-15

If I decide to trust, no problem to go to QFieldCloud
Screenshot from 2024-08-14 16-04-50

dddpt (changes requested):
So, I was able to start up QFieldCloud following the README instructions, only a small change in `init_mkcert.sh` to do.

However, I still get a untrusted certificate error on the first connection as I should not, isn't it?
Any idea where it could come from?

The errror:
Screenshot from 2024-08-14 16-04-32

The rootCA seems to be in my list of trusted certificates:
Screenshot from 2024-08-14 16-05-15

If I decide to trust, no problem to go to QFieldCloud
Screenshot from 2024-08-14 16-04-50

suricactus:
@anajoy any progress on this one? I would like to have it merged sooner than later.

scripts/init_mkcert.sh
anajoy: I couldn't neither confirm a connection to localhost:8002. So would be to clarify, what service is expected to listen on port 8002? I could not identify anyone? Probably a leftover of an older setup?

anajoy:
I try to answer your question, why we have to deal with a untrusted certificate?
I explored the net about this type or error. The most reasonable answer was, that any test certificate issued by letsencrypt has to be untrusted on productive browsers. This works therefore as designed. However they are bigger companies running their own PKI infrastructure beeing able to trust also test certificates. This type of certificates must be emitted based on trusted company onwed test root certificates.

anajoy:
I add a further evidence confirming the successful exchange of an expiring certificate (validity 90 days) after 83 days by applying a responding config change and awaiting the exptected day of exchange.

Reviews by Didier were checked, question answered and changes recommended applied.
I suggest to merge this change to master now.

[renewal_test_letsencrypt.txt](https://github.com/user-attachments/files/16729105/renewal_test_letsencrypt.txt)

docker-nginx/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM nginx:stable
 
 COPY pages /var/www/html/pages/
 COPY templates/ /etc/nginx/templates/
+COPY dhparams/ /etc/nginx/dhparams/
 COPY options-ssl-nginx.conf /etc/nginx/options-ssl-nginx.conf
 COPY 99-autoreload.sh /docker-entrypoint.d/99-autoreload.sh
-CMD chmod +x /docker-entrypoint.d/99-autoreload.sh
+RUN chmod +x /docker-entrypoint.d/99-autoreload.sh
```
suricactus, line 8: This was actually the missing bit that blocked the base PR. Thanks for spotting and fixing!

suricactus (commented):
Thanks for the work on this PR. Actually the main blocker was discovered here as I have put a comment on it: instead of `RUN` there was an `CMD` command in the nginx's `Dockerfile`.

However, the rest of the PR went into a direction which IMO is not needed and will cause more support. The `init_mkcert.sh` and yet another split of `docker-compose.localhost.yml` is not required, since the current `master` implementation already guarantees in a easy to understand and manage way that here is always a self-signed certificate.

I took a few other good ideas from this PR into #1005 . I am closing this for now.

anajoy:
Hi Ivan
Thank you for the review and feedback given.

I also thought about the options available to keep the  custom nginx directory as main location of certificate truth.
- by renaming and making permanent symlinks pointing to the default letsencrypt locations after an initial switch to letsencrypt.
- by running additional shell commands in directory /etc/letsencrypt/renewal-hooks/post, if symlink solution does not fit or work.

PR #1002, opened 2024-08-09
Add a Makefile to ease setup

mdouchin:
This is a proposal to add a Makefile to the project.

For example, the following commands can be used to launch a local instance:

```bash
# Copy environment file
make env

# Build (build stack)
make build

# Run (docker composer up)
make run

# Initialize (django migrations & static files)
make init

# Check status
make status

# Create super-user
make create-superuser

# Stop
make stop
```

This PR is a draft, as a first simple contribution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,45 @@
+SHELL:=bash
+ENV_FILE:=.env.example
+
+DJANGO_SUPERUSER_USERNAME="super_user"
+DJANGO_SUPERUSER_EMAIL="qfield@cloud.gis"
+DJANGO_SUPERUSER_PASSWORD="Ch@ngeM3Please;-)"
+
+env:
+	@echo "Configuring .env file from $(ENV_FILE)"
+	@grep -v -e '^#' -e '^$$' $(ENV_FILE) > .env
+
+build:
+	@echo "Build images"
+	docker compose build
+
+run:
+	@echo "Start QFieldCloud stack"
+	docker compose up -d
+
+#
+# Run initial configuration
+#
+# See https://github.com/opengisch/qfieldcloud#launch-a-local-instance
+# This is not really docker-friendly but django is to blame
+#
+init:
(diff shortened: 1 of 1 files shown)


mdouchin:
> I am closing this PR now, but I wish this will not discourage you at all to upstream other changes. PRs are more than welcome!

To be honest, I mostly agree with you. This adds a layer of abstraction, which can lead to more complexity.
It was a simple way to propose a first simple PR as a thank-you. No offence taken 👌

I have another subject in mind. Shouldn't we use pure HTTP in nginx service and let the system administrator decide how encryption should be added with which tools ? Let's encrypt is a way, but some would prefer to use a reverse proxy to encrypt the data, or have a dedicated certificate, etc. 

If I understand correctly, even the dev and local stack use encryption ? As a developer or tester, I would prefer to use unencrypt http://localhost:7080/admin/ URL to test the stack locally. Or am I missing something ? If not, I would try to propose a way/documentation to choose if encryption/let's encrypt should be added or not.

I can open an issue to talk about this subject and let this PR closed ;)

suricactus:
> I have another subject in mind. Shouldn't we use pure HTTP in nginx service and let the system administrator decide how encryption should be added with which tools ? Let's encrypt is a way, but some would prefer to use a reverse proxy to encrypt the data, or have a dedicated certificate, etc.
> 
> If I understand correctly, even the dev and local stack use encryption ? As a developer or tester, I would prefer to use unencrypt [http://localhost:7080/admin](http://localhost:7080/admin/) URL to test the stack locally. Or am I missing something ? If not, I would try to propose a way/documentation to choose if encryption/let's encrypt should be added or not.
> 
> I can open an issue to talk about this subject and let this PR closed ;)

The whole idea is that the stack is plug'n play with close to zero dependencies from the host. If you need to make the HTTPS optional, feel free to open an issue where we can discuss approaches. TBH in the last 4 years I use self-signed SSL for local development and testing and I have no issues with that.

PR #1010, opened 2024-08-30
Split requirements and freeze

m-kuhn:
Add a file requirements.in which specifies what we *want* and use pip-compile to build requirements.txt which freezes what we know works


suricactus:
Appreciated PR! I had a similiar branch locally but left it for better days to deduct direct and indirect dependencies. It is a good start. Would you mind if you already put a `README.md` instructions for pip compile?

m-kuhn:
It is "documented" in the `requirements.txt`

```
pip-compile requirements.in
```

https://github.com/opengisch/QFieldCloud/pull/1010/files#diff-6d1d18ca0d67cd16278085ea3947f2f19f91acf1b388e3534f46f561f03b1361R5

So I think eventually anyone in the desire to work on dependencies will eventually run into that. Do you think it's worth adding more docs related to that?

FTR: I also considered a pre-commit integration but we don't want to slow commits to 30 seconds. I think it needs to be on-demand on the dev machine. It could be integrated into a CI check.

m-kuhn:
AI think we should get rid of (many of the) fixed version numbers in the .in file, as they are mostly a relict of the old frozen requirements.
Apart from django itself, are there other consciously pinned deps?

suricactus:
I contininuosly push them to newer versions. I am pretty sure AI would suggest we bump to Django 5, would love to see it upgrading the codebase :) . Let's keep the numbers and reduce the direct dependencies.

suricactus:
Ok, I think this is the final push to this PR for now. I think the end result is quite good, however there is a follow up task to create a CI test that the developer has pushed their `requirements.in` and `requirements.txt` changes. I would let this to be a future onboarding task for someone :)

docker-compose.override.local.yml
```diff
@@ -17,6 +17,18 @@ services:
     depends_on:
       - db
 
+  pipcompile: &default-django
```
m-kuhn, line 20: Nice 👍

PR #1015, opened 2024-09-05
Make the information whether an error should be logged to sentry  part of the error class

suricactus:
(we reuse the http code to determine if it's sever error or client error)


duke-nyuki:

PR #1028, opened 2024-09-17
Do not check for email uniqueness if the email is not populated

### LINKED ISSUE #1022: Cannot leave the "Email address" field blank when adding a new user from the Web Admin interface
When adding a new user in the Web Admin interface (Home > Core > People > Add Person), the "Email address" field is not marked as mandatory but it cannot be left blank because there is a unique validation on this field.

It can be blank for the first user created but it can't be blank for any subsequent new user.

Error: "This email is already taken by another user!"

Expected behavior: Since the field is not marked as mandatory, I would expect no unique validation on the field.

I would suggest marking it as mandatory in the Web Admin interface (or maybe disable the unique validation).

Tested with 0.28.0

image

suricactus:
Fix #1022

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -362,15 +362,16 @@ class Person(User):
         verbose_name_plural = "people"
 
     def clean(self):
-        person_qs = self.__class__.objects.filter(email__iexact=self.email)
+        if self.email:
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-4849 Do not validate empty email address from Admin](https://app.clickup.com/t/2192114/QF-4849)

nirvn (approved):
Easy approval! ;-P

PR #1031, opened 2024-09-17
Add Teams Endpoints to manage teams

tdrivas:
New endpoints have been created for managing both teams and member under organization for CRUD actions

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -185,6 +185,17 @@ class UserManager(InheritanceManagerMixin, DjangoUserManager):
     def get_queryset(self):
         return super().get_queryset().select_subclasses()
 
+    def fast_search(self, username_or_email: str) -> "User":
+        """Searches a user by `username` or `email` field
+
+        Args:
+            username_or_email (str): username or email to search for
+
+        Returns:
+            User: The user with that username or email.
+        """
+        return self.get(Q(username=username_or_email) | Q(email=username_or_email))
+
 
 class PersonManager(UserManager):
     def get_queryset(self):
@@ -822,6 +833,14 @@ class Team(User):
     @property
     def teamname(self):
         return self.username.replace(f"@{self.team_organization.username}/", "")
+
+    @staticmethod
+    def format_team_name(organization_name: str, team_name: str) -> str:
+        """Returns the actual team username formatted as `@<organization_name>/<team_name>`."""
+        if not team_name:
+            raise ValueError("Team name is required.")
+
+        return f"@{organization_name}/{team_name}"
 
 
 class TeamMember(models.Model):
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.sites.models import Site
+from django.utils.translation import gettext as _
 from qfieldcloud.authentication.models import AuthToken
 from qfieldcloud.core import exceptions
 from qfieldcloud.core.models import (
@@ -13,10 +14,13 @@ from qfieldcloud.core.models import (
     Project,
     ProjectCollaborator,
     Team,
+    TeamMember,
     User,
 )
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
+
+from typing import Any
 
 
 def get_avatar_url(user: User) -> str | None:
@@ -171,6 +175,7 @@ class OrganizationSerializer(serializers.ModelSerializer):
     membership_is_public = serializers.BooleanField(
         read_only=True, source="membership_role_is_public"
     )
+    teams = serializers.SerializerMethodField()
 
     def get_members(self, obj):
         return [
@@ -178,6 +183,13 @@ class OrganizationSerializer(serializers.ModelSerializer):
             for m in OrganizationMember.objects.filter(
                 organization=obj, is_public=True
             ).values("member__username")
+        ]
+
+    def get_teams(self, obj: Organization) -> list[str]:
+        """Implementation of `SerializerMethodField` for `teams`. Returns list of team names."""
+        return [
+            t.teamname
+            for t in Team.objects.filter(team_organization=obj).values("username")
         ]
 
     def get_avatar_url(self, obj):
@@ -195,6 +207,7 @@ class OrganizationSerializer(serializers.ModelSerializer):
             "membership_role",
             "membership_role_origin",
             "membership_is_public",
+            "teams",
         )
 
 
@@ -535,3 +548,51 @@ class LatestPackageSerializer(serializers.Serializer):
     package_id = serializers.UUIDField()
     packaged_at = serializers.DateTimeField()
     data_last_updated_at = serializers.DateTimeField()
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    organization = serializers.StringRelatedField(source="team_organization")
+    team = serializers.CharField(required=True, allow_blank=False, source="username")
+
+    class Meta:
+        model = Team
+        fields = ("team", "organization")
+
+    def to_representation(self, instance: Team) -> dict[str, Any]:
+        representation = super().to_representation(instance)
+        representation["team"] = instance.teamname
+
+        return representation
+
+
+class TeamMemberSerializer(serializers.ModelSerializer):
+    member = serializers.CharField()
+
+    def to_internal_value(self, data: dict[str, Any]) -> dict[str, Any]:
+        validated_data = super().to_internal_value(data)
+        email_or_username: str = data["member"]
+
+        try:
+            existing_user = User.objects.fast_search(email_or_username)
+
+            if TeamMember.objects.filter(member=existing_user):
+                raise serializers.ValidationError(
+                    {
+                        "member": _('Team member "{}" already exists.').format(
+                            email_or_username
+                        )
+                    }
+                )
+
+            validated_data["member"] = existing_user
+
+        except User.DoesNotExist:
+            raise serializers.ValidationError(
+                {"member": _('User "{}" does not exists.').format(email_or_username)}
+            )
+
+        return validated_data
+
+    class Meta:
+        model = TeamMember
+        fields = ("member",)
--- a/docker-app/qfieldcloud/core/tests/test_team.py
+++ b/docker-app/qfieldcloud/core/tests/test_team.py
@@ -1,0 +1,435 @@
+import logging
+from qfieldcloud.authentication.models import AuthToken
+from qfieldcloud.core.models import (
+    TeamMember,
+    Organization,
+    OrganizationMember,
+    Person,
+    Team,
+)
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from .utils import setup_subscription_plans
+
+logging.disable(logging.CRITICAL)
+
+
+class TeamsTestCase(APITestCase):
+    def setUp(self):
+        setup_subscription_plans()
+
+        self.u1 = Person.objects.create_user(username="user1", password="abc123")
+        self.token1 = AuthToken.objects.get_or_create(user=self.u1)[0]
+
+        self.u2 = Person.objects.create_user(username="user2", password="abc123")
+        self.token2 = AuthToken.objects.get_or_create(user=self.u2)[0]
+
+        self.u3 = Person.objects.create_user(
+            username="user3", email="user3@email.com", password="abc123"
+        )
+        self.token3 = AuthToken.objects.get_or_create(user=self.u3)[0]
+
+        self.o1 = Organization.objects.create(
+            username="organization1",
+            password="abc123",
+            type=Organization.Type.ORGANIZATION,
+            organization_owner=self.u1,
+        )
+
+        OrganizationMember.objects.create(
+            organization=self.o1,
+            member=self.u2,
+            role=OrganizationMember.Roles.ADMIN,
+        )
+
+        OrganizationMember.objects.create(
+            organization=self.o1,
+            member=self.u3,
+            role=OrganizationMember.Roles.MEMBER,
+        )
+
+        team1_username = Team.format_team_name(self.o1.username, "team1")
+        team2_username = Team.format_team_name(self.o1.username, "team2")
+
+        self.t1 = Team.objects.create(
+            username=team1_username,
+            team_organization=self.o1,
+        )
+        self.t2 = Team.objects.create(
+            username=team2_username,
+            team_organization=self.o1,
+        )
+
+    def test_non_authenticated_cannot_list_teams(self):
+        """Ensure non-authenticated users cannot list teams."""
+        response = self.client.get(f"/api/v1/organizations/{self.o1.username}/teams/")
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_non_authenticated_cannot_create_team(self):
+        """Ensure non-authenticated users cannot create a team."""
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/",
+            {"username": "newteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_non_authenticated_cannot_update_team(self):
+        """Ensure non-authenticated users cannot update a team."""
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {"username": "newteamname"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+    def test_non_authenticated_cannot_delete_team(self):
+        """Ensure non-authenticated users cannot delete a team."""
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+    def test_non_member_cannot_create_team(self):
+        """Ensure non-members cannot create a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/",
+            {"username": "newteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(Team.objects.filter(username="newteam").count(), 0)
+
+    def test_non_member_cannot_update_team(self):
+        """Ensure non-members cannot update a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {"username": "newteamname"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_non_member_cannot_delete_team(self):
+        """Ensure non-members cannot delete a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_org_owner_can_create_team(self):
+        """Test that the organization owner can create a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/",
+            {"team": "newteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_org_owner_can_update_team(self):
+        """Test that the organization owner can update a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {"team": "updatedteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_org_owner_can_delete_team(self):
+        """Test that the organization owner can delete a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    def test_org_member_role_admin_can_create_team(self):
+        """Test that an admin can create a team."""
+        team_username = Team.format_team_name(self.o1.username, "newteam")
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/",
+            {"team": "newteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(Team.objects.filter(username=team_username).count(), 1)
+
+    def test_org_member_role_admin_can_update_team(self):
+        """Test that an admin can update a team."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {"team": "newteamname"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        updated_team_username = Team.format_team_name(self.o1.username, "newteamname")
+
+        self.t1.refresh_from_db()
+
+        self.assertEqual(self.t1.username, updated_team_username)
+        self.assertEqual(Team.objects.filter(username=updated_team_username).count(), 1)
+
+    def test_org_member_role_admin_can_delete_team(self):
+        """Test that an admin can delete a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(Team.objects.filter(username=self.t1.teamname).count(), 0)
+
+    def org_member_role_admin_can_add_member_to_team_by_username(self):
+        """Test that an admin can add a member to a team."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/",
+            {
+                "member": self.u3.username,
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(
+            TeamMember.objects.filter(team=self.t1, member=self.u3).count(), 1
+        )
+
+    def org_member_role_admin_can_add_member_to_team_by_email(self):
+        """Test that an admin can add a member to a team."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/",
+            {
+                "member": self.u3.email,
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(
+            TeamMember.objects.filter(team=self.t1, member=self.u3).count(), 1
+        )
+
+    def test_org_member_role_admin_can_delete_member_from_team(self):
+        """Test that an admin can remove a member from a team."""
+        TeamMember.objects.create(team=self.t1, member=self.u3)
+
+        self.assertEqual(TeamMember.objects.filter(member=self.u3).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/{self.u3.username}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(
+            TeamMember.objects.filter(team=self.t1, member=self.u3).count(), 0
+        )
+
+    def test_org_member_role_admin_cannot_create_team_member_with_empty_payload(self):
+        """Test that an admin cannot add a member to a team without a payload."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/",
+            {},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_org_member_role_admin_cannot_update_team_with_empty_payload(self):
+        """Test that an admin cannot update team without a payload."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token2.key)
+
+        old_team_name = self.t1.teamname
+
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        self.t1.refresh_from_db()
+
+        self.assertEqual(self.t1.teamname, old_team_name)
+
+    def test_org_member_role_member_can_list_teams(self):
+        """Test that a non-admin member can list teams."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.get(f"/api/v1/organizations/{self.o1.username}/teams/")
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        teams = sorted(response.json(), key=lambda t: t["team"])
+
+        self.assertListEqual(
+            teams,
+            [
+                {
+                    "team": self.t1.teamname,
+                    "organization": self.o1.username,
+                },
+                {
+                    "team": self.t2.teamname,
+                    "organization": self.o1.username,
+                },
+            ],
+        )
+
+    def test_org_member_role_member_can_retrieve_team_details(self):
+        """Test that an authorized user can get team's details."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.get(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        self.assertJSONEqual(
+            response.content.decode(),
+            {
+                "team": self.t1.teamname,
+                "organization": self.o1.username,
+            },
+        )
+
+    def test_org_member_role_member_can_list_members_of_team(self):
+        """Test that a non-admin member can list member of a team."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        TeamMember.objects.create(team=self.t1, member=self.u2)
+        TeamMember.objects.create(team=self.t1, member=self.u3)
+
+        self.assertEqual(TeamMember.objects.filter(team=self.t1).count(), 2)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.get(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        members = sorted(response.json(), key=lambda m: m["member"])
+
+        self.assertListEqual(
+            members,
+            [
+                {
+                    "member": "user2",
+                },
+                {
+                    "member": "user3",
+                },
+            ],
+        )
+
+    def test_org_member_role_member_cannot_create_team(self):
+        """Ensure non-admin members cannot create a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/",
+            {"team": "newteam"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_org_member_role_member_cannot_update_team(self):
+        """Ensure non-admin members cannot update a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.put(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/",
+            {"username": "newteamname"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+    def test_org_member_role_member_cannot_delete_team(self):
+        """Ensure non-admin members cannot delete a team."""
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+    def test_org_member_role_cannot_delete_member_from_team(self):
+        """Test that a member cannot remove a member from a team."""
+        TeamMember.objects.create(team=self.t1, member=self.u3)
+
+        self.assertEqual(TeamMember.objects.filter(member=self.u3).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/{self.u3.username}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(
+            TeamMember.objects.filter(team=self.t1, member=self.u3).count(), 1
+        )
+
+    def test_org_member_role_cannot_add_member_to_team(self):
+        """Test that a member cannot add a member to a team."""
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token3.key)
+
+        response = self.client.post(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/",
+            {
+                "member": self.u3.username,
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(
+            TeamMember.objects.filter(team=self.t1, member=self.u3).count(), 0
+        )
--- a/docker-app/qfieldcloud/core/urls.py
+++ b/docker-app/qfieldcloud/core/urls.py
@@ -10,6 +10,7 @@ from qfieldcloud.core.views import (
     qfield_files_views,
     status_views,
     users_views,
+    teams_views,
 )
 from rest_framework.routers import DefaultRouter
 
@@ -105,4 +106,24 @@ urlpatterns = [
         deltas_views.ListDeltasByDeltafileView.as_view(),
     ),
     path("deltas/apply/<uuid:projectid>/", deltas_views.ApplyView.as_view()),
+    path(
+        "organizations/<str:organization_name>/teams/",
+        teams_views.TeamListCreateView.as_view(),
+        name="team_list_create",
+    ),
+    path(
+        "organizations/<str:organization_name>/teams/<str:team_name>/",
+        teams_views.GetUpdateDestroyTeamDetailView.as_view(),
+        name="team_detail",
+    ),
+    path(
+        "organizations/<str:organization_name>/teams/<str:team_name>/members/",
+        teams_views.ListCreateTeamMembersView.as_view(),
+        name="team_member_list_create",
+    ),
+    path(
+        "organizations/<str:organization_name>/teams/<str:team_name>/members/<str:member_username>/",
+        teams_views.GetUpdateDestroyTeamMemberView.as_view(),
+        name="team_member_retrieve_update_destroy",
+    ),
 ]
--- a/docker-app/qfieldcloud/core/views/teams_views.py
+++ b/docker-app/qfieldcloud/core/views/teams_views.py
@@ -1,0 +1,264 @@
+from django.core.exceptions import ObjectDoesNotExist
+from drf_spectacular.utils import extend_schema, extend_schema_view
+from django.db.models import QuerySet
+
+
+from qfieldcloud.core import permissions_utils
+from qfieldcloud.core.models import (
+    Organization,
+    TeamMember,
+    Team,
+)
+from qfieldcloud.core.serializers import (
+    TeamMemberSerializer,
+    TeamSerializer,
+)
+
+from rest_framework import generics, permissions
+from rest_framework.request import Request
+from rest_framework.views import View
+
+from django.shortcuts import get_object_or_404
+
+
+class TeamMemberDeleteViewPermissions(permissions.BasePermission):
+    def has_permission(self, request: Request, view: View) -> bool:
+        user = request.user
+        organization_name = permissions_utils.get_param_from_request(
+            request, "organization"
+        )
+
+        try:
+            organization = Organization.objects.get(username=organization_name)
+        except ObjectDoesNotExist:
+            return False
+
+        if request.method == "DELETE":
+            return permissions_utils.can_delete_members(user, organization)
+
+        return False
+
+
+class TeamMemberPermission(permissions.BasePermission):
+    """
+    Permission class to handle CRUD operations for team members.
+    """
+
+    def has_permission(self, request: Request, view: View) -> bool:
+        organization_name = view.kwargs.get("organization_name")
+        team_name = view.kwargs.get("team_name")
+        team_username = Team.format_team_name(organization_name, team_name)
+
+        try:
+            organization = Organization.objects.get(username=organization_name)
+            _team = Team.objects.get(
+                username=team_username, team_organization=organization
+            )
+
+        except (Team.DoesNotExist, Organization.DoesNotExist):
+            return False
+
+        if request.method == "GET":
+            return permissions_utils.can_read_members(request.user, organization)
+
+        elif request.method == "POST":
+            return permissions_utils.can_create_members(request.user, organization)
+
+        elif request.method in ["PUT", "PATCH"]:
+            return permissions_utils.can_update_members(request.user, organization)
+
+        elif request.method == "DELETE":
+            return permissions_utils.can_delete_members(request.user, organization)
+
+        return False
+
+
+class TeamPermission(permissions.BasePermission):
+    """
+    Permission class to handle CRUD operations for teams.
+    """
+
+    def has_permission(self, request: Request, view: View) -> bool:
+        organization_name = view.kwargs.get("organization_name")
+
+        try:
+            organization = Organization.objects.get(username=organization_name)
+        except Organization.DoesNotExist:
+            return False
+
+        if request.method == "GET":
+            return permissions_utils.can_read_members(request.user, organization)
+
+        elif request.method in ["PUT", "POST"]:
+            return permissions_utils.can_update_members(request.user, organization)
+
+        elif request.method == "DELETE":
+            return permissions_utils.can_delete_members(request.user, organization)
+
+        return False
+
+
+@extend_schema_view(
+    get=extend_schema(description="Retrieve details of a team"),
+    put=extend_schema(description="Update a team's name"),
+    delete=extend_schema(description="Delete a team"),
+)
+class GetUpdateDestroyTeamDetailView(generics.RetrieveUpdateDestroyAPIView):
+    """
+    Handles GET, PUT, and DELETE for teams under a specific organization.
+    """
+
+    permission_classes = [permissions.IsAuthenticated, TeamPermission]
+    serializer_class = TeamSerializer
+
+    def get_object(self) -> Team:
+        organization_name = self.kwargs.get("organization_name")
+        team_name = self.kwargs.get("team_name")
+
+        organization = get_object_or_404(Organization, username=organization_name)
+        full_team_name = Team.format_team_name(organization_name, team_name)
+
+        return get_object_or_404(
+            Team, team_organization=organization, username=full_team_name
+        )
+
+    def perform_update(self, serializer: TeamSerializer) -> None:
+        """
+        Handle team name updates with validation.
+        """
+        organization_name = self.kwargs.get("organization_name")
+
+        serializer.is_valid(raise_exception=True)
+
+        new_team_name = serializer.validated_data["username"]
+
+        full_team_name = Team.format_team_name(organization_name, new_team_name)
+
+        if Team.objects.filter(
+            team_organization=serializer.instance.team_organization,
+            username=full_team_name,
+        ).exists():
+            raise permissions.ValidationError(
+                {"error": "A team with this name already exists."}
+            )
+
+        serializer.save(username=full_team_name)
+
+
+@extend_schema_view(
+    get=extend_schema(description="List all teams under an organization"),
+    post=extend_schema(description="Create a new team under an organization"),
+)
+class TeamListCreateView(generics.ListCreateAPIView):
+    """
+    Handles GET (list all teams) and POST (create a new team) under an organization.
+    """
+
+    permission_classes = [permissions.IsAuthenticated, TeamPermission]
+    serializer_class = TeamSerializer
+
+    def get_queryset(self) -> QuerySet[Team]:
+        organization_name = self.kwargs.get("organization_name")
+        organization = get_object_or_404(Organization, username=organization_name)
+        return Team.objects.filter(team_organization=organization)
+
+    def perform_create(self, serializer: TeamSerializer) -> None:
+        organization_name = self.kwargs.get("organization_name")
+        organization = get_object_or_404(Organization, username=organization_name)
+
+        serializer.is_valid(raise_exception=True)
+        team_name = serializer.validated_data["username"]
+
+        full_team_name = Team.format_team_name(organization_name, team_name)
+
+        serializer.save(team_organization=organization, username=full_team_name)
+
+
+@extend_schema_view(
+    get=extend_schema(description="Retrieve, update, or delete a team member"),
+    delete=extend_schema(description="Delete a team member"),
+)
+class GetUpdateDestroyTeamMemberView(generics.RetrieveDestroyAPIView):
+    """
+    View to handle adding and listing team members. --> organizations/<str:organization_name>/team/<str:team_name>/members/"
+    """
+
+    permission_classes = [permissions.IsAuthenticated, TeamMemberPermission]
+    serializer_class = TeamMemberSerializer
+
+    def get_queryset(self) -> QuerySet[TeamMember]:
+        organization_name = self.kwargs.get("organization_name")
+        team_name = self.kwargs.get("team_name")
+
+        full_team_name = Team.format_team_name(organization_name, team_name)
+
+        return TeamMember.objects.filter(
+            team__team_organization__username=organization_name,
+            team__username=full_team_name,
+        ).select_related("member")
+
+    def get_object(self) -> TeamMember:
+        organization_name = self.kwargs.get("organization_name")
+        team_name = self.kwargs.get("team_name")
+        member_username = self.kwargs.get("member_username")
+
+        full_team_name = Team.format_team_name(organization_name, team_name)
+
+        return get_object_or_404(
+            TeamMember,
+            team__team_organization__username=organization_name,
+            team__username=full_team_name,
+            member__username=member_username,
+        )
+
+    def get_full_team_name(self) -> str:
+        organization_name = self.kwargs.get("organization_name")
+        team_name = self.kwargs.get("team_name")
+
+        return Team.format_team_name(organization_name, team_name)
+
+    def perform_update(self, serializer: TeamMemberSerializer) -> None:
+        """
+        Update logic for the team member.
+        """
+        team_member = self.get_object()
+        new_username = self.request.data.get("member")
+
+        user_member = team_member.member
+        user_member.username = new_username
+        user_member.save()
+
+        serializer.save()
+
+
+@extend_schema_view(
+    get=extend_schema(description="List all members of a team"),
+    post=extend_schema(description="Add a new member to a team"),
+)
+class ListCreateTeamMembersView(generics.ListCreateAPIView):
+    permission_classes = [permissions.IsAuthenticated, TeamMemberPermission]
+    serializer_class = TeamMemberSerializer
+
+    def get_queryset(self) -> QuerySet[TeamMember]:
+        organization_name = self.kwargs.get("organization_name")
+
+        return TeamMember.objects.filter(
+            team__team_organization__username=organization_name,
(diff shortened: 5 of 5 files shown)


docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +523,43 @@ class LatestPackageSerializer(serializers.Serializer):
...
+# Add the necessary imports for Team and TeamMember models
+from qfieldcloud.core.models import Team, TeamMember
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    """
+    Serializer for creating a Team, accepts only 'username', 'team_organization', and optional 'members'.
+    """
+    
+    # members = serializers.PrimaryKeyRelatedField(queryset=OrganizationMember.objects.all(), many=True, required=False)
```
suricactus, line 538: Do we need the `members` variable? If not, please remove commented code, nobody is going too look for it in the future.
 If you really need it, please add a `# TODO we commented this because X and Y` on the line before.

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+        member = get_object_or_404(User, username=member_username)
+
+        # Check if the member belongs to the team
+        team_member = get_object_or_404(TeamMember, team=team, member=member)
+
+        # Remove the member
+        team_member.delete()
+
+        serializer = DeleteMemberSerializer(data={'message': f"Member {member.username} removed from team {team.username}"})
+        return Response(serializer.data, status=status.HTTP_204_NO_CONTENT)
```
suricactus, line 192: How come we both say: 204 no content, but we pass the `message` in the response? I would suggest to drop the `message` part, it should be just empty content if all other endpoints do that (double check if this is correct memory).

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+
+        # return Response({"message": f"Member {member.username} removed from team {team.username}"}, status=status.HTTP_204_NO_CONTENT)
```
suricactus, line 194: Commented?

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+        # team_details = [
+        # {
+        #     "username": team.username,
+        #     "organization": organization.username,
+        # }
+        # ]
+
+        # return Response(team_details, status=200)
+
+        # return Response(team_details, status=status.HTTP_200_OK)
```
suricactus, line 221: Commented?

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+        Delete a team within the organization.
+        """
+        organization = get_object_or_404(Organization, username=organization_name)
+        team_name = f'@{organization_name}/{team_name}'
+        team = get_object_or_404(Team, team_organization=organization, username=team_name)
+
+        # Delete the team
+        team.delete()
+        serializer = TeamDeleteSerializer(data={'message': "Team deleted successfully"})
+        return Response(serializer.data, status=status.HTTP_204_NO_CONTENT)
```
suricactus, line 262: Same as the other delete above.

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+
+        # team_list = [
+        #     {
+        #         "Team Name": team.username,
+        #         "Actual name:": team.teamname
+        #     }
+        #     for team in teams
+        # ]
+
+        # return Response(team_list, status=status.HTTP_200_OK)
```
suricactus, line 290: Same.

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+
+    def post(self, request, organization_name):
+        """
+        Handle POST request: Create a new team under a specific organization.
+        """
+        # Get the organization by its name
+        organization = get_object_or_404(Organization, username=organization_name)
+
+        # Extract the 'username' field from the request data
+        team_name = request.data.get('username')
```
suricactus, line 300: I think these are quite obvious, so you can drop the comments.
```suggestion
        organization = get_object_or_404(Organization, username=organization_name)
        team_name = request.data.get('username')
```

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +168,235 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+
+        if not team_name:
+            return Response({"error": "Team name is required."}, status=status.HTTP_400_BAD_REQUEST)
+
+        team_name = f'@{organization_name}/{team_name}'
```
suricactus, line 305: Ideally this should be upstreamed as a static method on the `Team` model.

docker-compose.override.local.yml
```diff
@@ -101,7 +101,7 @@ services:
       - ${HOST_GEODB_PORT}:5432
 
   minio:
-    image: minio/minio:RELEASE.2023-04-07T05-28-58Z
+    image: minio/minio:latest #RELEASE.2023-04-07T05-28-58Z
```
suricactus, line 104: Why is that? This should be fixed.

suricactus (changes requested):
Hey! There are a few things cleanup to be done before you continue working on these. Would you mind to address the feedback first, so it will be much easier for review.

In short:
- don't leave random commented code here and there
- no need to comment obvious things
- please rebase on master
- remove `docker-compose.override.local.yml` changes

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +523,91 @@ class LatestPackageSerializer(serializers.Serializer):
...
+
+
+# Add the necessary imports for Team and TeamMember models
+from qfieldcloud.core.models import Team, TeamMember, Organization
+
+
+# class TeamSerializer(serializers.ModelSerializer):
+#     """
+#     Serializer for creating a Team, accepts only 'username', 'team_organization', and optional 'members'.
+#     """
```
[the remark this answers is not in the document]
tdrivas: Accepted and Solved!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Accepted and Solved!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Accepted and Solved!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Accepted and Solved!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Accepted and Solved!

docker-app/qfieldcloud/core/serializers.py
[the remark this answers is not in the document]
suricactus: I still see the commented code here?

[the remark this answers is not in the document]
tdrivas: Its the serliazers.py, which i have not pushed yet as i still change things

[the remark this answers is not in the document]
tdrivas: serliazers has been pushed! You can see the changes now!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: New version has been pushed. Solved.

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Model has been updated. Solved

.gitignore
```diff
@@ -16,3 +16,4 @@ Pipfile*
 **/site-packages
 docker-qgis/libqfieldsync
 docker-qgis/qfieldcloud-sdk-python
+.env
```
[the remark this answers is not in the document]
tdrivas: Indeed. Line 19 was deleted

docker-app/qfieldcloud/core/serializers.py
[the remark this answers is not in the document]
tdrivas: Updated Serializers are now up.

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Updated users_views.py have been pushed.

.gitignore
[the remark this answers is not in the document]
suricactus: I guess you did not push it?

docker-app/qfieldcloud/core/models.py
```diff
@@ -810,6 +810,11 @@ def save(self, *args, **kwargs):
     def teamname(self):
         return self.username.replace(f"@{self.team_organization.username}/", "")
 
+    @staticmethod
+    def format_team_name(organization_name, team_name):
+        if not team_name:
+            raise ValueError("Team name is required.")
+        return f'@{organization_name}/{team_name}'
```
suricactus, line 817:
```suggestion
            raise ValueError("Team name is required.")

        return f'@{organization_name}/{team_name}'
```

Make sure that before/after `if` blocks and `raise` and `return` keywords there is an empty line.

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -844,3 +844,29 @@ def check_can_upload_file(project, client_type, file_size_bytes: int) -> bool:
...
+
+
+def can_create_teams(user: QfcUser, organization: Organization) -> bool:
+    """Check if the user can create a team within the organization."""
+    return user_has_organization_roles(
+        user, organization, [OrganizationMember.Roles.ADMIN]
+    )
+
+def can_manage_team(user: QfcUser, team: Team) -> bool:
+    """Check if the user can manage a team."""
```
suricactus, line 856:
```suggestion
    """Check if the user can add or remove members from a team."""
```

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -844,3 +844,29 @@ def check_can_upload_file(project, client_type, file_size_bytes: int) -> bool:
...
+    Ensures that the new member belongs to the same organization as the team."""
+    if user_has_organization_roles(user, team.team_organization, [OrganizationMember.Roles.ADMIN]):
+        # Ensure the new member is part of the same organization
+        if user_has_organization_roles(new_member, team.team_organization, [OrganizationMember.Roles.MEMBER]):
+            return True
+        raise UserOrganizationRoleError(
+            _("User '{}' is not part of the organization '{}' that owns the team.")
+            .format(new_member.username, team.team_organization.name)
+        )
+    return False
```
suricactus, line 872: Instead of adding new permission checks for team members, we can reuse the organization member checks (e.g. `can_create_members`). If needed, we can update their documentation so it is clear they are used not only for org member, but also for team members. This is how we check team permissions on app.qfield.cloud.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -14,7 +14,10 @@
     ProjectCollaborator,
     Team,
     User,
+    Person,
+    TeamMember
```
suricactus, line 18:
```suggestion
    TeamMember,
```

Do you have pre-commit, this should have a trailing comma.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -167,6 +170,12 @@ def get_members(self, obj):
                 organization=obj, is_public=True
             ).values("member__username")
         ]
+    
+
+    # added
```
suricactus, line 175:
```suggestion

```

Not needed comment.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+    def _validate_user_membership(self, user, organization):
+        is_org_member = OrganizationMember.objects.filter(
+            organization=organization,
+            member=user
+        ).exists()
+        
+        is_org_owner = organization.organization_owner == user
+        
+        if not (is_org_member or is_org_owner):
+            raise serializers.ValidationError("User is not a member of the organization.")
```
suricactus, line 587: I think you can do it like this?

```
org = Organization.objects.get(username='admin')
org.members.filter(member=user).exists()
```

suricactus, line 587: You can remove these, they should not be validated in the serializer, but the view that calls the serializer. Check `core/views/members_views.py` for example.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+        model = TeamMember
+        fields = ('team', 'member')
+
+class AddMemberSerializer(serializers.Serializer):
+    username = serializers.CharField(required=True)
+    email = serializers.EmailField(required=False, allow_blank=True)
+
+    def validate(self, data):
+        username = data.get('username')
+        user = self._validate_user_exists(username)
```
suricactus, line 557: This should also work with email too.


Basically, check if the user is a member of the organization already. That's all.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+            organization=organization,
+            member=user
+        ).exists()
+        
+        is_org_owner = organization.organization_owner == user
+        
+        if not (is_org_member or is_org_owner):
+            raise serializers.ValidationError("User is not a member of the organization.")
+
+class TeamDetailSerializer(serializers.ModelSerializer):
```
suricactus, line 589: I guess we don't need this as we already do the same here:

```
class TeamSerializer(serializers.ModelSerializer):
```

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+    organization = serializers.StringRelatedField(source='team_organization')
+
+    class Meta:
+        model = Team
+        fields = ('username', 'organization')
+
+
+class TeamListSerializer(serializers.ModelSerializer):
```
suricactus, line 597: I guess the same as above:

```
class TeamSerializer(serializers.ModelSerializer):
```

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+        model = Team
+        fields = ('username', 'teamname')
+
+
+class DeleteMemberSerializer(serializers.Serializer):
+    message = serializers.CharField(read_only=True)
+
+
+class TeamDeleteSerializer(serializers.Serializer):
+    message = serializers.CharField(default="Team deleted successfully", read_only=True)
```
suricactus, line 608: Remove these as when we delete, there should be no message returned. Or at least this is how all the other endpoints work.

docker-app/qfieldcloud/urls.py
```diff
@@ -79,5 +91,14 @@
...
+
+   
+    
+    # # ADDED
+    # path("organizations/<str:organization_name>/team/', TeamListCreateView.as_view(), name='team-list-create'),
+    
     path("__debug__/", include("debug_toolbar.urls")),
+   
+   
+
```
suricactus, line 103:
```suggestion
    re_path(r"^invitations/", include("invitations.urls", namespace="invitations")),
    path("__debug__/", include("debug_toolbar.urls")),
```

We can remove all this noise.

docker-app/qfieldcloud/urls.py
```diff
@@ -70,6 +73,15 @@
...
     ),
+
+      # New team-related endpoints
+    path("organizations/<str:organization_name>/team/", TeamListCreateView.as_view(), name='team-list-create'),
+    path("organizations/<str:organization_name>/team/<str:team_name>/members/", TeamMemberView.as_view(), name='team_members'),
+    path("organizations/<str:organization_name>/team/<str:team_name>/", TeamDetailView.as_view(), name='team-detail'),
+    path("organizations/<str:organization_name>/team/<str:team_name>/members/<str:member_username>/", TeamMemberDeleteView.as_view(), name='team-member-delete'),
+   
+
+
```
suricactus, line 84: This should be moved to `qfieldcloud.core.url` not here in `qfieldcloud.url`. If you keep it here, these endpoints are not going to be versioned with `api/v1/`. This file is for global endpoints, not really DRF endpoints.

docker-app/qfieldcloud/urls.py
```diff
@@ -29,8 +29,11 @@
 )
 from qfieldcloud.authentication import views as auth_views
 from qfieldcloud.core.views import files_views
+from qfieldcloud.core.views.users_views import *
```
suricactus, line 32: Never do this, `*` in imports are forbidden for reasons that google can help with ;)

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+            member=self.user2,
+            role=OrganizationMember.Roles.ADMIN,
+        )
+        OrganizationMember.objects.create(
+            organization=self.organization1,
+            member=self.user3,
+            role=OrganizationMember.Roles.MEMBER,
+        )
+
+    def test_list_teams(self):
```
suricactus, line 52: 1. make sure you can list teams in an organization without any teams.
2. continue with adding a few teams as you do

suricactus:
```suggestion
    def test_org_owner_can_list_teams(self):
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+            f"/organizations/{self.organization1.username}/team/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        json_response = response.json()
+        team_usernames = [team["username"] for team in json_response]
+
+        self.assertIn(team1.username, team_usernames)
+        self.assertIn(team2.username, team_usernames)
```
suricactus, line 73:
```suggestion
        payload = response.json()
        team_usernames = [team["username"] for team in payload]

        self.assertIn(team1.username, team_usernames)
        self.assertIn(team2.username, team_usernames)
```

What if the json_response is not sorted in the way you expected?

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def test_create_team(self):
```
suricactus, line 75:
```suggestion
    def test_organization_owner_can_create_team(self):
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+        response = self.client.post(
+            f"/organizations/{self.organization1.username}/team/",
+            {"username": "newteam"},
+        )
+
+        self.assertTrue(status.is_success(response.status_code))
```
suricactus, line 82:
```suggestion
        self.assertEqual(response.status_code, status.HTTP_201_CRETED)
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+
+        team_username = f"@{self.organization1.username}/newteam"
```
suricactus, line 84: Use `Team.format_team_name` here instead of home baked solution, this might bite you in the future if we change the username format.

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+        team = Team.objects.create(
+            username=f"@{self.organization1.username}/teamdetail",
+            team_organization=self.organization1,
+        )
+
+        # Authenticate with the correct user token
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        # Make a GET request to retrieve the team details
+        url = f"/organizations/{self.organization1.username}/team/teamdetail/"
```
suricactus, line 101:
```suggestion
        url = f"/organizations/{self.organization1.username}/team/{team.teamname}/"
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,156 @@
...
+        team_username = f"@{self.organization1.username}/newteam"
+        team_exists = Team.objects.filter(
+            username=team_username, team_organization=self.organization1
+        ).exists()
+        self.assertTrue(team_exists)
+
+    def test_retrieve_team_detail(self):
+        """Test retrieving the details of a specific team."""
+        team = Team.objects.create(
+            username=f"@{self.organization1.username}/teamdetail",
```
suricactus, line 93:
```suggestion
            username=f"@{self.organization1.username}/newteam",
```

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +172,185 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+
+    def get(self, request, organization_name):
+        """
+        Handle GET request: Retrieve all teams under a specific organization --> organizations/<str:organization_name>/team/
+        """
+        organization = get_object_or_404(Organization, username=organization_name)
+
+        teams = querysets_utils.get_organization_teams(organization)
+
+        serializer = TeamListSerializer(teams, many=True)
```
suricactus, line 274:
```suggestion
        organization = get_object_or_404(Organization, username=organization_name)
        teams = querysets_utils.get_organization_teams(organization)
        serializer = TeamListSerializer(teams, many=True)
```

It's all variable assignment, so no need fow the newlines. Anyways, this most probably goes away.

suricactus (changes requested):
General comment:

- there are still archeological treasures from the initial development, please make sure you remove these and push them
- merge all the tests in one `test_teams.py` file
- the views can be a bit more `django-rest-framework`-like (DRF) , e.g. you can use `RetrieveUpdateDestroyAPIView` or similar for the views baseclasses for "teams" and "team members". This will greatly reduce your codebase and simplify maintainence.


1) cleanup the initial comments in this PR and push all the changes.
2) rebase to the latest master and push
3) run `pre-commit` and push
4) address most of the requests not related the the views.py or tests
5) try to automate views behavior with DRF
6) continue with tests

docker-app/qfieldcloud/core/models.py
tdrivas: Solved

docker-app/qfieldcloud/core/serializers.py
tdrivas: Has been addressed!

.gitignore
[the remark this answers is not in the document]
tdrivas: pushed!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Message Deleted!

docker-compose.override.local.yml
tdrivas: Done!

docker-app/qfieldcloud/core/serializers.py
tdrivas: Deleted!

docker-app/qfieldcloud/urls.py
tdrivas: Removed!

docker-app/qfieldcloud/urls.py
tdrivas: Direct urls moved to core urls. Only global left here

docker-app/qfieldcloud/urls.py
tdrivas: Google helped me ;)

docker-app/qfieldcloud/core/urls.py
```diff
@@ -40,6 +40,7 @@
 """
 
 urlpatterns = [
+
```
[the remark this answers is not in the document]
tdrivas: Done

docker-app/qfieldcloud/core/serializers.py
tdrivas: pre-commit applied!

docker-app/qfieldcloud/core/serializers.py
tdrivas: Done!

docker-app/qfieldcloud/urls.py
tdrivas: Changed!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -523,3 +532,106 @@ class LatestPackageSerializer(serializers.Serializer):
...
+
+
+class TeamAccessSerializer(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
tdrivas: Deleted

docker-app/qfieldcloud/core/serializers.py
tdrivas: Done

docker-app/qfieldcloud/core/permissions_utils.py
tdrivas: Done!

suricactus (changes requested):
Forgot to post the comments from last time.

docker-app/qfieldcloud/core/urls.py
```diff
@@ -105,4 +105,24 @@
         deltas_views.ListDeltasByDeltafileView.as_view(),
     ),
     path("deltas/apply/<uuid:projectid>/", deltas_views.ApplyView.as_view()),
+    path(
+        "organizations/<str:organization_name>/team/",
```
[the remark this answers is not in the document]
tdrivas: Done!

docker-app/qfieldcloud/core/urls.py
```diff
@@ -105,4 +105,24 @@
...
+        users_views.TeamListCreateView.as_view(),
+        name="team-list-create",
```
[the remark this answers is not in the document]
tdrivas: Done!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,80 @@ class LatestPackageSerializer(serializers.Serializer):
     package_id = serializers.UUIDField()
     packaged_at = serializers.DateTimeField()
     data_last_updated_at = serializers.DateTimeField()
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    organization = serializers.CharField(
+        source="team_organization.username", read_only=True
+    )
```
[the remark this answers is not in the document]
tdrivas: Done

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -844,3 +844,36 @@ def check_can_upload_file(project, client_type, file_size_bytes: int) -> bool:
...
+    """Check if the user can manage a team."""
+    return user_has_organization_roles(
+        user, team.team_organization, [OrganizationMember.Roles.ADMIN]
+    )
+
+
+def can_add_members_to_team(user: QfcUser, team: Team, new_member: QfcUser) -> bool:
+    """Check if the user can add a new member to the team.
+    Ensures that the new member belongs to the same organization as the team."""
+    if user_has_organization_roles(
```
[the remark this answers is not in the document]
tdrivas: Done

docker-app/qfieldcloud/core/serializers.py
tdrivas: It has been deleted

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,80 @@ class LatestPackageSerializer(serializers.Serializer):
...
+            try:
+                return User.objects.get(email=email)
+            except User.DoesNotExist:
+                raise serializers.ValidationError(
+                    "User with the given email does not exist."
+                )
+
+        raise serializers.ValidationError(
+            "Invalid user lookup."
+        )  # Should not reach here
```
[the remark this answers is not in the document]
tdrivas: Done!

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +158,192 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+        """
+        organization_name = self.kwargs.get("organization_name")
+        team_name = self.kwargs.get("team_name")
+        member_username = self.kwargs.get("member_username")
+
+        organization = Organization.objects.get(username=organization_name)
+        team_full_name = Team.format_team_name(organization_name, team_name)
+        team = Team.objects.get(team_organization=organization, username=team_full_name)
+
+        return TeamMember.objects.filter(team=team, member__username=member_username)
```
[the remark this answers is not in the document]
tdrivas: Applied!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Done

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -180,6 +180,10 @@ def get_members(self, obj):
             ).values("member__username")
         ]
 
+    def get_teams(self, obj):
+        teams = Team.objects.filter(team_organization=obj)
+        return TeamSerializer(teams, many=True).data
+
```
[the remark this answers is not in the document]
tdrivas: Correct. 

We have to add in the declaration of the attributes of the class the command:

`teams = serializers.SerializerMethodField() 
`
The SerializerMethodField searches for the corresponding` get_<AttributeName>` method

added

docker-app/qfieldcloud/core/serializers.py
tdrivas: `def _validate_user_membership(self, user, organization):
    if not (organization.organization_owner == user or organization.members.filter(member=user).exists()):
        raise serializers.ValidationError("User is not a member of the organization.")
`

However, this method has been deleted as the related functionality is handled via the views.

suricactus: resolved

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Done!

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Changed!

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: We kept the "assertEqual"

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Done

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -146,3 +172,185 @@ class ListUserOrganizationsView(generics.ListAPIView):
...
+            role=OrganizationMember.Roles.ADMIN,
+        ).exists()
+        if not (is_org_owner or is_org_admin):
+            raise PermissionDenied("You do not have permission to remove members from this team.")
+
+        team_member.delete()
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+class TeamDetailView(APIView):
```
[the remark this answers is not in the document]
tdrivas: Changed but still POST operation is pending!

docker-app/qfieldcloud/core/urls.py
```diff
@@ -105,4 +105,24 @@
...
+    path(
+        "organizations/<str:organization_name>/teams/<str:team_name>/members/<str:member_username>/",
+        users_views.GetUpdateDestroyTeamMemberView.as_view(),
+        name="team-member-retrieve-update-destroy",
+    ),
+    path(
+        "organizations/<str:organization_name>/teams/<str:team_name>/",
+        users_views.TeamDetailView.as_view(),
+        name="team_detail",
+    ),
```
suricactus, line 127:
```suggestion
    path(
        "organizations/<str:organization_name>/teams/",
        team_views.ListCreateTeamView.as_view(),
        name="team_list_create",
    ),
    path(
        "organizations/<str:organization_name>/teams/<str:team_name>/",
        team_views.GetUpdateDestroyTeamView.as_view(),
        name="team_detail",
    ),
    path(
        "organizations/<str:organization_name>/teams/<str:team_name>/members/",
        team_views.ListCreateTeamMembersView.as_view(),
        name="team_member_list_create",
    ),
    path(
        "organizations/<str:organization_name>/teams/<str:team_name>/members/<str:member_username>/",
        team_views.GetUpdateDestroyTeamMemberView.as_view(),
        name="team_member_retrieve_update_destroy",
    ),
```

A bit of renaming here :)

tdrivas: Done!

docker-app/qfieldcloud/core/views/users_views.py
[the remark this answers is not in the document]
tdrivas: Done

docker-app/qfieldcloud/core/models.py
tdrivas, line 1: When update takes place, we must check whether a user with the username already exists so to avoid confict. Therefore, we must check it and fast_search yields an error if user does not exist. To this direction, another function has been implemenented (fast_count) so to return a value, even if the user does not exist.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+                raise serializers.ValidationError(
+                    {"member": "Username must be provided"}
+                )
+
+            if User.objects.fast_count(email_or_username) > 0:
+                raise serializers.ValidationError(
+                    {
+                        "member": f"A user with username '{email_or_username}' already exists."
+                    }
+                )
```
suricactus, line 572:
```suggestion
        try:
            member = User.objects.fast_search(email_or_username) > 0:
        except User.DoesNotExistError:
            raise serializers.ValidationError(
                    {
                        "member": f"A user with username '{email_or_username}' already exists."
                    }
                )
```

Actually there is no need for a new method, that's why this is throwing.

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -1,17 +1,22 @@
...
     CompleteUserSerializer,
     OrganizationSerializer,
     PublicInfoUserSerializer,
 )
+
 from rest_framework import generics, permissions
 from rest_framework.response import Response
 
-User = get_user_model()
+from django.contrib.auth.models import User
```
suricactus, line 19: Why you change this? It is not by coinsidence like this. Check the documentation of `get_user_model`  https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#django.contrib.auth.get_user_model

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -1,17 +1,22 @@
...
 from django.core.exceptions import ObjectDoesNotExist
 from drf_spectacular.utils import extend_schema, extend_schema_view
+
+
 from qfieldcloud.core import pagination, permissions_utils, querysets_utils
-from qfieldcloud.core.models import Organization, Project
+from qfieldcloud.core.models import (
+    Organization,
+    Project,
+)
```
suricactus, line 9:
```suggestion
    Project
)
```

Can you remove all changes that are not relevant to this PR. This file should not be modified in this PR since you moved the code to the new file.

docker-app/qfieldcloud/core/serializers.py
tdrivas: I will check it!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Back to the initial state!

docker-app/qfieldcloud/core/views/users_views.py
tdrivas: Back to the initial state!

docker-app/qfieldcloud/urls.py
```diff
@@ -29,6 +29,7 @@
 )
 from qfieldcloud.authentication import views as auth_views
 from qfieldcloud.core.views import files_views
+
```
suricactus, line 32: Can you please remove this line from the diff.

docker-app/qfieldcloud/core/views/teams_views.py
```diff
@@ -0,0 +1,285 @@
...
+#         """
+#         organization_name = self.kwargs.get("organization_name")
+#         team_name = self.kwargs.get("team_name")
+#         member_username = self.kwargs.get("member_username")
+
+#         return TeamMember.objects.filter(
+#             team__organization__username=organization_name,
+#             team__username=Team.format_team_name(organization_name, team_name),
+#             member__username=member_username,
+#         )
```
suricactus, line 62: Can you remove the commented code or add a comment why this is commented

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -277,9 +283,7 @@ def to_internal_value(self, data):
             if self._choices[i] == data:
                 return i
         raise serializers.ValidationError(
-            "Invalid status. Acceptable values are {}.".format(
-                list(self._choices.values())
-            )
+            "Invalid status. Acceptable values are {}.".format((self._choices.values()))
```
suricactus, line 286: Why did you change that?
```suggestion
            "Invalid status. Acceptable values are {}.".format(
            list(self._choices.values())
        )
```
Or if it is needed, can you please remove the inner braces.
```suggestion
            "Invalid status. Acceptable values are {}.".format(self._choices.values())
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -823,6 +848,13 @@ def save(self, *args, **kwargs):
     def teamname(self):
         return self.username.replace(f"@{self.team_organization.username}/", "")
 
+    @staticmethod
+    def format_team_name(organization_name, team_name):
```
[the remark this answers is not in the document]
tdrivas: Sure!

docker-app/qfieldcloud/core/serializers.py
tdrivas: It does not work. 

This 'if' checks via fast_search if a user with the new username exists already and therefore prevent another user to get this new username.

So, if we pass a non-existing username (which means that our user can change his/her username), the fast_search will raise an exception as it makes a get request to the db and does not find this username.

docker-app/qfieldcloud/urls.py
tdrivas: Done!

docker-app/qfieldcloud/core/models.py
```diff
@@ -185,6 +185,31 @@ class UserManager(InheritanceManagerMixin, DjangoUserManager):
...
+
+        Args:
+            username_or_email (str): The username or email to search for.
+
+        Returns:
+            int: The count of users with that username or email.
+        """
+        return self.filter(
+            Q(username=username_or_email) | Q(email=username_or_email)
+        ).count()
```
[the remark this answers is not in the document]
tdrivas: Copy my answer also here:

It does not work.

This 'if' checks via fast_search if a user with the new username exists already and therefore prevent another user to get this new username.

So, if we pass a non-existing username (which means that our user can change his/her username), the fast_search will raise an exception as it makes a get request to the db and does not find this username.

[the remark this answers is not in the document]
suricactus: Okk, let's keep all in the other comment thread.

docker-app/qfieldcloud/core/serializers.py
suricactus: @tdrivas not sure if I follow.

> So, if we pass a non-existing username (which means that our user can change his/her username), the fast_search will raise an exception as it makes a get request to the db and does not find this username.

Indeed, this is the reason why I suggest wrapping the `fast_search` in a `try/except` block?

I see that the suggestion I gave has a typo in there, check the updated version.

tdrivas: You are correct, it does work now. Thanks.

docker-app/qfieldcloud/core/views/teams_views.py
tdrivas: Done and also decorators have been added

docker-app/qfieldcloud/core/serializers.py
tdrivas: Done

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+
+class TeamMemberSerializer(serializers.ModelSerializer):
+    member = serializers.CharField()
+
+    def to_internal_value(self, data):
+        validated_data = super().to_internal_value(data)
+        email_or_username = data.get("member")
+        existing_username = self.context["view"].kwargs.get("member_username")
+
+        if existing_username is not None:
```
tdrivas, line 560: Managed to check if the request method is post or put/patch by checking the kwargs and not by checking the type of request

docker-app/qfieldcloud/core/models.py
```diff
@@ -823,6 +834,13 @@ def save(self, *args, **kwargs):
     def teamname(self):
         return self.username.replace(f"@{self.team_organization.username}/", "")
 
+    @staticmethod
+    def format_team_name(organization_name: str, team_name: str) -> str:
```
suricactus, line 838:
```suggestion
    def format_team_name(organization_name: str, team_name: str) -> str:
    """Returns the actual team username formatted as `@<organization_name>/<team_name>`."""
```

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -277,9 +283,7 @@ def to_internal_value(self, data):
             if self._choices[i] == data:
                 return i
         raise serializers.ValidationError(
-            "Invalid status. Acceptable values are {}.".format(
-                list(self._choices.values())
-            )
+            "Invalid status. Acceptable values are {}.".format(self._choices.values())
```
suricactus, line 286: Why is this change here? Please revert to the original state.

I am afraid I mention this problem here for the third time...

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -180,6 +182,10 @@ def get_members(self, obj):
             ).values("member__username")
         ]
 
+    def get_teams(self, obj):
+        teams = Team.objects.filter(team_organization=obj)
+        return TeamSerializer(teams, many=True).data
```
suricactus, line 187:
```suggestion
    def get_teams(self, obj: Organization) -> list[str]:
        """Implementation of `SerializerMethodField` for `teams`. Returns list of team names."""
        return [
            t.team_name
            for t in Team.objects.filter(
                team_organization=obj
            ).values("username")
        ]
```

Just print the team names, not all the info. Similar to what we do with the organization members.

suricactus (changes requested):
Can you please rebase on the latest master and squash all the changes in a single commit.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+        existing_username = self.context["view"].kwargs.get("member_username")
+
+        if existing_username is not None:
+            if not email_or_username:
+                raise serializers.ValidationError(
+                    {"member": "Username must be provided"}
+                )
+
+            try:
+                if User.objects.fast_search(email_or_username) > 0:
```
suricactus, line 567: What is the point of this line? Fast search is going to throw if the email username/email does not exist. You are checking a function that returns type `User` with `int`.

I am pretty sure we already have discussed this.

docker-app/qfieldcloud/core/serializers.py
suricactus: I guess this comment related to the line below. In any case, you should not go to such extent to check for the kwargs.

The PATCH and PUT on already existing team members is not needed. One should just delete the old one and add a new member.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+                    raise serializers.ValidationError(
+                        {
+                            "member": f"A user with username '{email_or_username}' already exists."
+                        }
+                    )
+            except User.DoesNotExist:
+                email_or_username = self.context["view"].kwargs.get("member_username")
```
suricactus, line 574:
```suggestion
```

We might only need the check for uniqueness.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +539,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+                    {"member": "Username must be provided"}
+                )
+
+            try:
+                if User.objects.fast_search(email_or_username) > 0:
+                    raise serializers.ValidationError(
+                        {
+                            "member": f"A user with username '{email_or_username}' already exists."
+                        }
+                    )
```
suricactus, line 572: You are not adding a user, I guess the message should be different.

```suggestion
                    raise serializers.ValidationError(
                        {
                            "member": f"A member with username '{email_or_username}' already exists."
                        }
                    )
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,501 @@
...
+        response = self.client.patch(
+            f"/api/v1/organizations/{self.org1.username}/teams/{self.t1.teamname}/members/{self.u3.username}/",
+            {},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        team_member.refresh_from_db()
+        self.assertEqual(team_member.member.username, self.u3.username)
+
```
suricactus, line 488:
```suggestion

```

Does not really make sense to update the username. One can simply delete and add a new user.

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,501 @@
...
+        self.u2 = Person.objects.create_user(username="user2", password="abc123")
+        self.token2 = AuthToken.objects.get_or_create(user=self.u2)[0]
+
+        self.u3 = Person.objects.create_user(
+            username="user3", email="user3@email.com", password="abc123"
+        )
+        self.token3 = AuthToken.objects.get_or_create(user=self.u3)[0]
+
+        self.org1 = Organization.objects.create(
```
suricactus, line 33:
```suggestion
        self.o1 = Organization.objects.create(
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,501 @@
...
+        response = self.client.get(f"/api/v1/organizations/{self.org1.username}/teams/")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        response_content = response.json()
+
+        team_usernames = sorted([team["username"] for team in response_content])
+
+        self.assertEqual(len(team_usernames), 2)
+        self.assertEqual(self.t1.username, team_usernames[0])
+        self.assertEqual(self.t2.username, team_usernames[1])
```
suricactus, line 77:
```suggestion
    def test_authorized_user_can_list_teams(self):
        """Test that an authorized user can list teams."""
        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)

        response = self.client.get(f"/api/v1/organizations/{self.org1.username}/teams/")

        self.assertEqual(response.status_code, status.HTTP_200_OK)

        payload = response.json()
        team_usernames = sorted([team["username"] for team in response_content])

        self.assertEqual(len(team_usernames), 2)
        self.assertEqual(self.t1.username, team_usernames[0])
        self.assertEqual(self.t2.username, team_usernames[1])
```

- No need for the top empty line after the docstring.
- `payload` is a standard name for the request/response, well, payload
- Keep the asserts lines separated with newlines before/after.
- Group the variable definition in the same block without newlines between them.

docker-app/qfieldcloud/core/views/teams_views.py
```diff
@@ -0,0 +1,265 @@
...
+        serializer.save(team_organization=organization, username=full_team_name)
+
+
+@extend_schema_view(
+    get=extend_schema(description="Retrieve, update, or delete a team member"),
+    patch=extend_schema(description="Partially update a team member"),
+    put=extend_schema(description="Update a team member"),
+    delete=extend_schema(description="Delete a team member"),
+)
+class GetUpdateDestroyTeamMemberView(generics.RetrieveUpdateDestroyAPIView):
```
suricactus, line 182:
```suggestion
class GetUpdateDestroyTeamMemberView(generics.RetrieveDestroyAPIView):
```

No need for patching the team member.

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Done

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Done

docker-app/qfieldcloud/core/views/teams_views.py
tdrivas: extend_schema_view also updated

docker-app/qfieldcloud/core/tests/test_team.py
suricactus: Note this applies to all tests below too.

docker-app/qfieldcloud/core/serializers.py
tdrivas: Solved (for real)

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Obviously :)

docker-app/qfieldcloud/core/models.py
tdrivas: Done

docker-app/qfieldcloud/core/serializers.py
tdrivas: Done.. 

Regarding docstring, do you prefer to do it this way? 
  
    **Retrieve and serialize teams associated with a given organization.

    Args:
        obj (Organization): The organization object for which blah blah...

    Returns:
        blah blah blah**

docker-app/qfieldcloud/core/serializers.py
tdrivas: This comments is answering all the comments above related to function internal_value:

The idea is the following:
1. check if the username of the member that is going to be added already exists in another member
2. check if member is indeed a user (this might be overkill due to an existing check via permissions in the view(???)
3. otherwise use the fast_search to obtain the user

```
  try:
            existing_user = User.objects.fast_search(email_or_username)
            
            if TeamMember.objects.filter(member=existing_user):
                raise serializers.ValidationError(
                        {
                            "member": f"A member with username '{email_or_username}' already exists."
                        }
                )
            
            validated_data["member"] = existing_user
        
        except User.DoesNotExist:  
            
            raise serializers.ValidationError(
                        {
                            "member": f"A user with username '{email_or_username}' does not exists."
                        }
            )
        
        return validated_data
```

docker-app/qfieldcloud/core/serializers.py
tdrivas: Check the final comment

docker-app/qfieldcloud/core/serializers.py
tdrivas: Changed

tdrivas:
> Can you please rebase on the latest master and squash all the changes in a single commit?
> 
> Please add typing on all the methods that you add in this PR, except the tests.
> 
> Also, unfortunately I see some of our verbal conversation feedback was not addressed.


Do you want to add typing also in variables or is it overkill?

for example `email_or_username: str = data.get("member", "")`

docker-app/qfieldcloud/core/serializers.py
suricactus: What happens if we don't explicitly handle the `DoesNotExist` error, or we attempt to break the uniqueness of the members in the organization? Do we get a proper error or just error 500? What is the HTTP response body?

If we get 500, then we need to handle it. Your proposal is good.
 
Note that the messages are weird, you can get a validation message like: "A member with username user@email.com already exists". Just say 'The team member "user@email.com" already exists."  We always put placeholders in double quotes in user strings.

docker-app/qfieldcloud/core/serializers.py
suricactus: See my final answer https://github.com/opengisch/QFieldCloud/pull/1031#discussion_r1868518703. Closing this thread.

suricactus:
> > Please add typing on all the methods that you add in this PR, except the tests.
> 
> Do you want to add typing also in variables or is it overkill?
> 
> for example `email_or_username: str = data.get("member", "")`

Add typing only when the type cannot be guessed by the type checker. In the case you mentioned, the typechecker has no way to get what is the type at key "member", so it will be `Any`.

But if you have:

```
def foo() -> int: ...
bar = foo()
```

There is no need to set the type of `bar`, as we already set that `foo` is always returning an `int`.

docker-app/qfieldcloud/core/serializers.py
suricactus: Ideally it should be like this, yes. All args and return value should be described.

Sometimes for smaller functions it is acceptable to have single line docstring. This example is a good one -> the function is very short, we have single parameter and clear return value, and the function is a callback that is very well known within DRF. Even if you add the `Args` and `Returns` there is not much added value.

tdrivas: Fair enough!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -180,6 +184,13 @@ def get_members(self, obj):
             ).values("member__username")
         ]
 
+    def get_teams(self, obj: Organization) -> list[str]:
+        """Implementation of `SerializerMethodField` for `teams`. Returns list of team names."""
+        return [
+            t.team_name
```
suricactus, line 190:
```suggestion
            t.teamname
```

Note that the suggestions I make are not always typo-free/compilable. The property is actually `teamname` instead of `team_name`.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +546,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+
+    class Meta:
+        model = Team
+        fields = ("username", "organization")
+
+
+class TeamMemberSerializer(serializers.ModelSerializer):
+    member = serializers.CharField()
+
+    def to_internal_value(self, data: Dict[str, Any]) -> Dict[str, Any]:
```
suricactus, line 562:
```suggestion
    def to_internal_value(self, data: dict[str, Any]) -> dict[str, Any]:
```

Since python 3.9 you can use the collections directly, check https://peps.python.org/pep-0585/

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +546,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+        validated_data = super().to_internal_value(data)
+        email_or_username = data.get("member")
```
suricactus, line 564:
```suggestion
        email_or_username: str = data["member"]
```

We can safely assume the value will always be present in the dict, as the Serializer will automatically check for it. I might be wrong, to be verified.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +546,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+
+        try:
+            existing_user = User.objects.fast_search(email_or_username)
+
+            if TeamMember.objects.filter(member=existing_user):
+                raise serializers.ValidationError(
+                    {
+                        "member": f"A member with username '{email_or_username}' already exists."
```
suricactus, line 572:
```suggestion
                        "member": _('Team member "{}" already exists.').format(email_or_username)
```

IMO the messages from the `ValidationError` should be translated. It might not be the case already in most of the code, but this is the direction.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +546,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
+                    }
+                )
+
+            validated_data["member"] = existing_user
+
+        except User.DoesNotExist:
+            raise serializers.ValidationError(
+                {
+                    "member": f"A user with username '{email_or_username}' does not exist."
```
suricactus, line 581:
```suggestion
                    "member": _('User "{}" does not exist.').format(email_or_username),
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,410 @@
...
+        response = self.client.get(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        payload = response.json()
+
+        self.assertEqual(payload["username"], self.t1.username)
+        self.assertEqual(payload["organization"], self.o1.username)
```
suricactus, line 91:
```suggestion
        self.assertJsonEqual(
            response.content.decode(),
            {
                "username": self.t1.username,
                "organization": self.o1.username,
            }
        )
```

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,410 @@
...
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.delete(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 0)
```
suricactus, line 244:
```suggestion
    def test_admin_user_can_delete_team(self):
        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 1)

        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)

        response = self.client.delete(
            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/"
        )

        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
        self.assertEqual(Team.objects.filter(username=self.t1.username).count(), 0)
        
    def test_unauthorized_user_cannot_list_teams(self):
        """Test that an unauthorized user cannot list teams."""
        response = self.client.get(f"/api/v1/organizations/{self.o1.username}/teams/")

        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```

Just swapping the order of the tests, all "deletes" together, all "lists" together. Makes it easier to read.

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,410 @@
...
+        response = self.client.get(f"/api/v1/organizations/{self.o1.username}/teams/")
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        payload = response.json()
+        team_usernames = sorted([team["username"] for team in payload])
+
+        self.assertEqual(len(team_usernames), 2)
+        self.assertEqual(self.t1.username, team_usernames[0])
+        self.assertEqual(self.t2.username, team_usernames[1])
```
suricactus, line 76: I see that down there in `test_non_admin_user_can_list_team_members` and other you use "assertJsonEqual". It is fine to use it, but here you properly catch the ordering that is not static. Let's assume the ordering will be fine for now and change the code to use `assertJsonEqual`.

What you can do is just sort the JSON properly:

```
teams = sorted(response.json(), key=lambda t: t["username"])

self.assertListEqual(teams, [
  {
      "username": self.t1.username,
  },
  {
      "username": self.t2.username,
  },
])
```

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +546,44 @@ class LatestPackageSerializer(serializers.Serializer):
...
     packaged_at = serializers.DateTimeField()
     data_last_updated_at = serializers.DateTimeField()
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    organization = serializers.StringRelatedField(source="team_organization")
+
+    class Meta:
+        model = Team
+        fields = ("username", "organization")
```
suricactus, line 556: I guess we want to list the `Team.teamname` here, instead of the full `username`.

docker-app/qfieldcloud/core/serializers.py
tdrivas: So, The _() function marks strings for translation, correct? Thus, it makes the string translatable to other languages.. Fair enough!

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: indeed!

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Does not find assertJsonEqual in APITestCase class. Shall i try to inherit from the TestCase?

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: Does not find assertJsonEqual in APITestCase class. Shall i try to inherit from the TestCase?

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: done

docker-app/qfieldcloud/core/serializers.py
tdrivas: Issue here with _() as mypy does not seems to recognize it. Solved with `from django.utils.translation import gettext as _
`

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: My mistake. Works well.

docker-app/qfieldcloud/core/serializers.py
tdrivas: I will check it!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +547,53 @@ class LatestPackageSerializer(serializers.Serializer):
...
+    organization = serializers.StringRelatedField(source="team_organization")
+    teamname = serializers.CharField(
+        required=True, allow_blank=False, source="username"
+    )
+
+    class Meta:
+        model = Team
+        fields = ("teamname", "organization")
+
+    def to_representation(self, instance):
```
suricactus, line 562:
```suggestion
    def to_representation(self, instance: Team) -> dict[str, Any]:
```
Final round of typing.

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,422 @@
...
+                    "organization": self.o1.username,
+                },
+                {
+                    "teamname": self.t2.teamname,
+                    "organization": self.o1.username,
+                },
+            ],
+        )
+
+    def test_authorized_user_can_retrieve_team_detail(self):
```
suricactus, line 87: It is.not just authorized user, I think u3 was a member?

docker-app/qfieldcloud/core/tests/test_team.py
suricactus, line 1: There are 4 user scenarios to be tested for each endpoint:
- non authenticated access (which is different from authorized). All others are authenticated.
- non authorized, or when the user is authenticated but they don't have permissions. Generally speaking, when the user is not an organization owner or member. All others might be authorized.
- organization owner (which is different from "admin user", as one might think it is Django's superuser) - can do pretty much anything with teams.
- organization member role admin - has permissions to do pretty much anything.
- organization member role member - has permissions to read but not to edit. Calling them non admin is oversimplification.

suricactus: This is a proposal for consistent naming.

test_non_authenticated
test_non_member
test_org_owner
test_org_member_role_admin
test_org_member_role_member

suricactus (changes requested):

duke-nyuki:
Task linked: [QF-4536 Add API endpoint to manage teams](https://app.clickup.com/t/2192114/QF-4536)

docker-app/qfieldcloud/core/tests/test_team.py
tdrivas: ok, i will rewrite/rename/add the tests with this logic.

tdrivas: Tests are grouped per user category!

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,422 @@
...
+        """Test that team members can list other members of their team."""
+        response = self.client.get(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertJSONEqual(
+            response.content.decode(),
+            {"code": "not_authenticated", "message": "Not authenticated"},
+        )
```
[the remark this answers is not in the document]
tdrivas: removed!

docker-app/qfieldcloud/core/tests/test_team.py
```diff
@@ -0,0 +1,422 @@
...
+        self.assertEqual(TeamMember.objects.filter(team=self.t1).count(), 2)
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        response = self.client.get(
+            f"/api/v1/organizations/{self.o1.username}/teams/{self.t1.teamname}/members/"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertJSONEqual(
```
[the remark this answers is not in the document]
tdrivas: indeed!

suricactus (commented):
Good! We are ready to merge this PR. One last renaming and we are good to merge!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -535,3 +548,53 @@ class LatestPackageSerializer(serializers.Serializer):
     package_id = serializers.UUIDField()
     packaged_at = serializers.DateTimeField()
     data_last_updated_at = serializers.DateTimeField()
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    organization = serializers.StringRelatedField(source="team_organization")
+    teamname = serializers.CharField(
```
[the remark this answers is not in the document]
tdrivas: Done!

suricactus (approved):
We are there! Thanks a lot!

Please squash all the changes into a single commit and merge the commit.

docker-app/qfieldcloud/core/serializers.py
suricactus: Are you going to check it :P

Issue #1067, opened 2024-11-29
DuplicateObject


shapley:
Hi,

I'm trying to create a new database connection in 'QFieldCloud admin' after deleting a previous connection. I'm now unable to enter a new 'Username' postgres because it already exists in the db. How can I remove the duplicate role 'postgres' in order to create a new connection or will it require entering the container in order to remove the role.

![duplicate_object](https://github.com/user-attachments/assets/741f5dc2-66f6-4e77-99dc-5c8cf2d218b8)

suricactus:
It is hard to say what happened, if you provide steps to reproduce and the full error stack will greatly help.

Note that GeoDB will soon be removed from QFieldCloud and it is not recomended to be used.

I am closing this issue as it is related to a module that will be not be exposed for usage from next version onwards.

PR #1068, opened 2024-11-29
Kill GeoDB admin

suricactus:
GeoDB was an experiment that is hard to manage and should not be in the core app. It is still used in the tests, so the functionality must stay.

See https://github.com/opengisch/QFieldCloud/issues/1067


duke-nyuki:
Task linked: [QF-3814 Remove all Geodb related code](https://app.clickup.com/t/2192114/QF-3814)

PR #1077, opened 2024-12-12
[refactor] Add error message for the `FileNotFoundError` instead of just showing the filename

suricactus:
This is only visible in the `Job.feedback_json`, but makes much more sense to the user and the developer when debugging.

### DIFF
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -229,7 +229,7 @@ def open_qgis_project(
     logging.info(f'Loading QGIS project "{project_filename}"…')
 
     if not Path(project_filename).exists():
-        raise FileNotFoundError(project_filename)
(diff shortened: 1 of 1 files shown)


docker-qgis/qfc_worker/utils.py
```diff
@@ -229,7 +229,7 @@ def open_qgis_project(
     logging.info(f'Loading QGIS project "{project_filename}"…')
 
     if not Path(project_filename).exists():
-        raise FileNotFoundError(project_filename)
+        raise FileNotFoundError(f"File not found: {project_filename}")
```
tdrivas, line 232: Consider of using  .format? So to be compatible with Internationalization, if needed.

suricactus: This code is running within the worker. There is very low probability we will ever translate the messages there. There is no translation framowork available at all.

PR #1065, opened 2024-11-26
[WIP] Store files metadata in database

suricactus:

PR #1078, opened 2024-12-12
[refactor] Make functions in `utils2.storages` us the `Project` instance, not only the `project_id`

suricactus:
This simplifies some of the future code that is coming in QF-2760.

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -451,6 +451,14 @@ def upload_project_file(
 
 
 def delete_all_project_files_permanently(project_id: str) -> None:
+    """Deletes all project files permanently.
+
+    Args:
+        project_id (str): the project which files shall be deleted. Note that the `project_id` might be a of a already deleted project which files are still dangling around.
+
+    Raises:
+        RuntimeError: if the produced Object Storage key to delete is not in the right format
+    """
     prefix = f"projects/{project_id}/"
 
     if not re.match(r"^projects/[\w]{8}(-[\w]{4}){3}-[\w]{12}/$", prefix):
@@ -466,7 +474,9 @@ def delete_project_file_permanently(
 ):  # noqa: F821
     logger.info(f"Requested delete (permanent) of project file {filename=}")
 
-    file = qfieldcloud.core.utils.get_project_file_with_versions(project.id, filename)
+    file = qfieldcloud.core.utils.get_project_file_with_versions(
+        str(project.id), filename
+    )
 
     if not file:
         raise Exception(
@@ -510,7 +520,7 @@ def delete_project_file_permanently(
 
 
 def delete_project_file_version_permanently(
-    project: qfieldcloud.core.models.Project,  # noqa: F821
+    project: qfieldcloud.core.models.Project,
     filename: str,
     version_id: str,
     include_older: bool = False,
@@ -526,7 +536,8 @@ def delete_project_file_version_permanently(
     Returns:
         int: the number of versions deleted
     """
-    file = qfieldcloud.core.utils.get_project_file_with_versions(project.id, filename)
+    project_id = str(project.id)
+    file = qfieldcloud.core.utils.get_project_file_with_versions(project_id, filename)
 
     if not file:
         raise Exception(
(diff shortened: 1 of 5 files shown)


duke-nyuki:

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -591,7 +602,8 @@ def delete_project_file_version_permanently(
     return versions_to_delete
 
 
-def get_stored_package_ids(project_id: str) -> set[str]:
+def get_stored_package_ids(project: qfieldcloud.core.models.Project) -> set[str]:
+    project_id = project.id
```
suricactus, line 606: I use the `project_id = str(project.id)` to minimize the diff.

These funcions will be gone in 3 months or so anyways.

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -526,7 +536,8 @@ def delete_project_file_version_permanently(
     Returns:
         int: the number of versions deleted
     """
-    file = qfieldcloud.core.utils.get_project_file_with_versions(project.id, filename)
+    project_id = str(project.id)
+    file = qfieldcloud.core.utils.get_project_file_with_versions(project_id, filename)
 
     if not file:
```
tdrivas, line 542: why not adding a specific Exception such as 
`raise FileNotFoundError(f"File '{filename}' not found for project '{project_id}'`

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -526,7 +536,8 @@ def delete_project_file_version_permanently(
...
         raise Exception(
```
tdrivas, line 543: if not file:
    raise FileNotFoundError(f"File '{filename}' not found for project '{project_id}'

tdrivas, line 543: Why not adding a specific exception such as 
`raise FileNotFoundError(f"File '{filename}' not found for project '{project_id}'`

docker-app/qfieldcloud/core/utils2/storage.py
tdrivas: Is there any added value to add a warning about this deprecation ?

docker-app/worker_wrapper/wrapper.py
```diff
@@ -431,7 +430,7 @@ def after_docker_run(self) -> None:
                 if package_id in job_ids:
```
tdrivas, line 430: this can be moved to the if above:

` if package_id ==str(self.job.project.last_package_job_id) or package_id in job_ids:
`

docker-app/worker_wrapper/wrapper.py
```diff
@@ -408,8 +408,7 @@ def after_docker_run(self) -> None:
         )
 
         try:
-            project_id = str(self.job.project.id)
-            package_ids = storage.get_stored_package_ids(project_id)
+            package_ids = storage.get_stored_package_ids(self.job.project)
             job_ids = [
```
tdrivas, line 412: we can use sets and and values_lists with flat to make it more efficient

docker-app/qfieldcloud/core/utils2/storage.py
suricactus: We did not change this line in this PR, is it a must to change it here?

docker-app/qfieldcloud/core/utils2/storage.py
suricactus: There is, check the #1065 PR.

suricactus: The current PR is not proposing the deprecation yet. But in the PR that actually introduces the deprecation I have added a deprecation warning.

https://github.com/opengisch/QFieldCloud/blob/QF-2760-files/docker-app/qfieldcloud/core/utils2/storage.py#L29-L47

docker-app/worker_wrapper/wrapper.py
suricactus: It is not clear to me what you mean by "make it more efficient" by using sets?

I would suggest that we don't change too much on these functions as they were serving the codebase well so far.

What  I see as potential improvement here is indeed using `value_list("id", flat=True)`, but this is out of scope of this PR. Personally I would prefer to keep it as it is. But I will also accept a follow-up PR that changes things in the proposed way.

docker-app/worker_wrapper/wrapper.py
suricactus: This is not a change of this specific PR, is it a must to apply this change?

Also, I would argue that small individual `if`s are much more readable than a bunch of conditions in one place.

suricactus (commented):
Hi, @tdrivas, thanks for the review!

Do you see some very scary things in the surrounding code that is not part of this patch? Most of your comments here are about code that is not part of this PR, so I would propose you add the respective tasks in the backlog and we just move on with the changes that are actually introduced in this PR.

PR #1079, opened 2024-12-12
[tests] Fix hardcoded file name when the package files are downloaded

suricactus:
Seems when this function was first created, someone forgot to make the filename dynamic.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_packages.py
+++ b/docker-app/qfieldcloud/core/tests/test_packages.py
@@ -160,7 +160,7 @@ class QfcTestCase(APITransactionTestCase):
                 if tempdir:
                     for filename in expected_files:
                         response = self.client.get(
-                            f"/api/v1/packages/{self.project1.id}/latest/files/project_qfield.qgs/"
+                            f"/api/v1/packages/{self.project1.id}/latest/files/{filename}/"
                         )
                         local_file = os.path.join(tempdir, filename)
 


docker-app/qfieldcloud/core/tests/test_packages.py
```diff
@@ -160,7 +160,7 @@ def check_package(
                 if tempdir:
                     for filename in expected_files:
                         response = self.client.get(
-                            f"/api/v1/packages/{self.project1.id}/latest/files/project_qfield.qgs/"
+                            f"/api/v1/packages/{self.project1.id}/latest/files/{filename}/"
```
tdrivas, line 163: We can also use less hardcoding approach by introducing the base url some lines above:

API_BASE_URL = "/api/v1/packages/{project_id}/latest/files/"

tdrivas, line 163:
```suggestion
                            f"/api/v1/packages/{self.project1.id}/latest/files/{filename}/"
                            
                            response = self.client.get(API_BASE_URL.format(project_id=self.project1.id, filename=filename))
                            
```

tdrivas:
```suggestion
                            response = self.client.get(API_BASE_URL.format(project_id=self.project1.id, filename=filename))
```

suricactus: Isn't it exactly the same amout of hardcoding by moving the f-string to string that we call `.format` on it?

IMO it is always better if we keep the code as explicit as possible as close as possible to the place it is actually used and adding a variable will just obscure what is behind the suggested `API_BASE_URL`.

tdrivas: as it is used only once in the code, yes, its overkill. agreed!

PR #1085, opened 2024-12-19
QFieldCloud release v0.30.0

dddpt:


suricactus:
@dddpt usually we do this locally without PR, but it is ok.

PR #1087, opened 2024-12-19
QFieldCloud release v0.30.0

dddpt:


suricactus:
@dddpt I am a bit lost how we get two identical PRs with https://github.com/opengisch/QFieldCloud/pull/1085 ? I am missing where is the difference?

PR #1088, opened 2024-12-20
Move `core.test_job` to `subscription.test_job_creation`

gounux:
The test is checking whether the user is able to create a job based on the current subscription status (activity,storage,etc...). So its belong to subscription rather than core.


duke-nyuki:
Task linked: [QF-4081 Add test to prevent regressions if `BadLayerHandler` does not work during `packaging` or `process_projectfile`](https://app.clickup.com/t/2192114/QF-4081)

suricactus (approved):

PR #1089, opened 2024-12-20
Add test to prevent regressions if `BadLayerHandler` does not work during `packaging` or `process_projectfile`

gounux:
This PR will :

- [X] Create a `simple_bumblebees_wrong_localized.qgs` QGIS project file with 3 layers :
  - 1 (invalid) layer with a datasource with non existing filename
  - 1 (invalid) localized layer with a datasource with non existing filename
  - 1 (valid) localized layer with a datasource with existing filename (QFC should consider it as invalid because it is localized)
- [X] Test that a `PROCESS_PROJECTFILE` type job parses the layer data correctly
- [x] Test that a `PACKAGE` type job parses the layer data correctly

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_jobs.py
+++ b/docker-app/qfieldcloud/core/tests/test_jobs.py
@@ -1,0 +1,211 @@
+import logging
+
+from qfieldcloud.authentication.models import AuthToken
+from qfieldcloud.core.models import (
+    Job,
+    PackageJob,
+    Person,
+    Project,
+)
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.test import APITransactionTestCase
+
+from .utils import (
+    set_subscription,
+    setup_subscription_plans,
+    testdata_path,
+    wait_for_project_ok_status,
+)
+
+logging.disable(logging.CRITICAL)
+
+
+class QfcTestCase(APITransactionTestCase):
+    def setUp(self):
+        setup_subscription_plans()
+
+        # Create a user
+        self.u1 = Person.objects.create_user(username="u1", password="abc123")
+        self.t1 = AuthToken.objects.get_or_create(user=self.u1)[0]
+
+        # Create a project
+        self.p1 = Project.objects.create(name="p1", is_public=False, owner=self.u1)
+
+    def upload_file(
+        self,
+        project: Project,
+        local_filename: str,
+        remote_filename: str,
+    ) -> Response:
+        """Upload a file to QFieldCloud using API.
+
+        Args:
+            project (Project): project that should contain the file.
+            local_filename (str): name of the local file to upload, should be in `testdata` folder.
+            remote_filename (str): name of the uploaded file.
+
+        Returns:
+            Response: response to the POST HTTP request.
+        """
+        file_path = testdata_path(local_filename)
+        response = self.client.post(
+            f"/api/v1/files/{project.id}/{remote_filename}/",
+            {
+                "file": open(file_path, "rb"),
+            },
+            format="multipart",
+        )
+        return response
+
+    def assertLayerData(
+        self, layer_data: dict, is_valid: bool, is_localized: bool, error_code: str
+    ) -> None:
+        self.assertEquals(layer_data["is_valid"], is_valid)
+        self.assertEquals(layer_data["is_localized"], is_localized)
+        self.assertEquals(layer_data["error_code"], error_code)
+
+    def test_bad_layer_handler_values_for_process_projectfile_job(self):
+        # Test that BadLayerHandler is parsing data properly during process projectfile jobs
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.t1.key)
+
+        # Push the QGIS project and datasource GPKG files
+        response = self.upload_file(self.p1, "bumblebees.gpkg", "bumblebees.gpkg")
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        response = self.upload_file(
+            self.p1, "simple_bumblebees_wrong_localized.qgs", "project.qgs"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        wait_for_project_ok_status(self.p1)
+        processprojectfile_job = Job.objects.filter(
+            project=self.p1, type=Job.Type.PROCESS_PROJECTFILE
+        ).latest("updated_at")
+
+        self.assertEqual(processprojectfile_job.status, Job.Status.FINISHED)
+        self.assertIsNotNone(processprojectfile_job.feedback)
+
+        self.p1.refresh_from_db()
+
+        self.assertIsNotNone(self.p1.project_details)
+
+        # extract layer data from job
+        processfile_layers = processprojectfile_job.get_feedback_step_data(
+            "project_details"
+        )["returns"]["project_details"]["layers_by_id"]
+
+        # "valid" localized layer -> QFC considers it as invalid
+        self.assertLayerData(
+            processfile_layers["valid_localized_point_layer_id"],
+            is_valid=False,
+            is_localized=True,
+            error_code="localized_dataprovider",
+        )
+
+        # invalid localized layer
+        self.assertLayerData(
+            processfile_layers["invalid_localized_polygon_layer_id"],
+            is_valid=False,
+            is_localized=True,
+            error_code="localized_dataprovider",
+        )
+
+        # invalid layer (datasource does not exist)
+        self.assertLayerData(
+            processfile_layers["invalid_point_layer_id"],
+            is_valid=False,
+            is_localized=False,
+            error_code="invalid_dataprovider",
+        )
+
+    def test_bad_layer_handler_values_for_package_job(self):
+        # Test that BadLayerHandler is parsing data properly during package jobs
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.t1.key)
+
+        # Push the QGIS project and datasource GPKG files
+        response = self.upload_file(self.p1, "bumblebees.gpkg", "bumblebees.gpkg")
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        response = self.upload_file(
+            self.p1, "simple_bumblebees_wrong_localized.qgs", "project.qgs"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # Create Package job and wait for it to finish
+        set_subscription(
+            self.u1,
+            "keep_10",
+            storage_mb=10,
+            is_premium=True,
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-4081 Add test to prevent regressions if `BadLayerHandler` does not work during `packaging` or `process_projectfile`](https://app.clickup.com/t/2192114/QF-4081)

suricactus (commented):
@gounux in case you did not use already, please install https://plugins.qgis.org/plugins/trackable_project_files/ to make the `.qgs` file git trackable.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,129 @@
...
+            is_premium=True,
+            is_external_db_supported=True,
+        )
+        package_job = PackageJob.objects.create(
+            type=Job.Type.PROCESS_PROJECTFILE,
+            project=self.p1,
+            created_by=self.u1,
+        )
+        self.assertEqual(package_job.status, Job.Status.PENDING)
+        wait_for_project_ok_status(self.p1)
```
gounux, line 88: @suricactus wondering if this test should be split in two : one for the `ProcessProjectfileJob` job, one for the `PackageJob` job :thinking:

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,129 @@
...
+        ]:
+            file_path = testdata_path(local_filename)
+            response = self.client.post(
+                f"/api/v1/files/{self.p1.id}/{upload_filename}/",
+                {
+                    "file": open(file_path, "rb"),
+                },
+                format="multipart",
+            )
+            self.assertTrue(status.is_success(response.status_code))
```
suricactus, line 56:
```suggestion
        # Push the QGIS project and datasource GPKG files
        for local_filename, upload_filename in [
            ("bumblebees.gpkg", "bumblebees.gpkg"),
            ("simple_bumblebees_wrong_localized.qgs", "project.qgs"),
        ]:
            file_path = testdata_path(local_filename)
            response = self.client.post(
                f"/api/v1/files/{self.p1.id}/{upload_filename}/",
                {
                    "file": open(file_path, "rb"),
                },
                format="multipart",
            )
            self.assertTrue(status.is_success(response.status_code))
```

You can move this to a separate function. Also check for `status.HTTP_201_CREATED` instead of `is_success`.

docker-app/qfieldcloud/core/tests/test_jobs.py
suricactus: Yes, better split in two.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,129 @@
...
+            # invalid localized layer
+            invalid_localized_layer = layers[INVALID_LOCALIZED_LAYER_KEY]
+            self.assertFalse(invalid_localized_layer["is_valid"])
+            self.assertTrue(invalid_localized_layer["is_localized"])
+            self.assertEquals(
+                invalid_localized_layer["error_code"], "localized_dataprovider"
+            )
+
+        # invalid layer is not present in qfield layers
+        assert INVALID_LAYER_KEY not in qfield_layers
```
suricactus, line 119: use the `self.assert` methods instead of `assert` keyword.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+        setup_subscription_plans()
+
+        # Create a user
+        self.u1 = Person.objects.create_user(username="u1", password="abc123")
+        self.t1 = AuthToken.objects.get_or_create(user=self.u1)[0]
+
+        # Create a project
+        self.p1 = Project.objects.create(name="p1", is_public=False, owner=self.u1)
+
+    def upload_file(self, project_id: str, local: str, remote: str) -> Response:
```
suricactus, line 40:
```suggestion
    def upload_file(
  	    self,
  	    project: Project,
  	    local_filename: str,
  	    remote_filename: str,
  	) -> Response:
```

Better pass the whole project object. Not that your typing was wrong, since you set the `project_id: str`, but `isinstance(project.id, UUID)`, you have to explicitly convert to `str`. Check if you can configure your editor to check this for you.

As for local/remote, it is not clear what they are, avoid using adjectives as variable names, add a noun whenever possible.

Last comment, I would make the `remote_filename` optional and use the `local_filename` value by default. Will save a few bytes of writing.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        return response
+
+    def test_bad_layer_handler_values_for_process_projectfile_job(self):
+        # Test that BadLayerHandler is parsing data properly during process projectfile jobs
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.t1.key)
+
+        # Push the QGIS project and datasource GPKG files
+        response = self.upload_file(self.p1.id, "bumblebees.gpkg", "bumblebees.gpkg")
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
suricactus, line 67:
```suggestion

        self.assertEqual(response.status_code, status.HTTP_201_CREATED)

```

Always add an empty line around asserts to highlight their presence. You can group multiple asserts without a newline between them.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Create ProcessProjectfileJob and wait for it to finish
+        processprojectfile_job = ProcessProjectfileJob.objects.create(
+            type=Job.Type.PROCESS_PROJECTFILE,
+            project=self.p1,
+            created_by=self.u1,
+        )
+        self.assertEqual(processprojectfile_job.status, Job.Status.PENDING)
+        wait_for_project_ok_status(self.p1)
```
suricactus, line 80: Why do you explicitly start a `ProcessProjectfileJob`, isn't there one automatically started for you once you upload the .qgs file?

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(processprojectfile_job.status, Job.Status.FINISHED)
+        self.assertIsNotNone(processprojectfile_job.feedback)
+
+        self.p1.refresh_from_db()
+        self.assertIsNotNone(self.p1.project_details)
+
+        # extract layer data from job
+        for step in processprojectfile_job.feedback["steps"]:
+            if step["id"] == "project_details":
+                processfile_layers = step["returns"]["project_details"]["layers_by_id"]
```
suricactus, line 92: You better add `Job.get_feedback_step_data(self, step_name: str) -> dict[str, Any] | None` for this, no?

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+            if step["id"] == "qgis_layers_data":
+                qgis_layers = step["returns"]["layers_by_id"]
+            if step["id"] == "qfield_layer_data":
+                qfield_layers = step["returns"]["layers_by_id"]
+
+        for layers in [qgis_layers, qfield_layers]:
+            # "valid" localized layer -> QFC considers it as invalid
+            valid_localized_layer = layers[VALID_LOCALIZED_LAYER_KEY]
+            self.assertFalse(valid_localized_layer["is_valid"])
+            self.assertTrue(valid_localized_layer["is_localized"])
```
suricactus, line 156: Is there a way to avoid the for loop? You better write the code twice, but be more explicit in the tests.

suricactus (changes requested):
A few comments on the test.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,173 @@
...
+    setup_subscription_plans,
+    testdata_path,
+    wait_for_project_ok_status,
+)
+
+logging.disable(logging.CRITICAL)
+
+VALID_LOCALIZED_LAYER_KEY = "apiary_853c29cb_ead8_4932_a60a_5cba33140b3d"
+INVALID_LAYER_KEY = "apiary_e7d1b542_1b1c_422d_ac75_c14d2b54e472"
+INVALID_LOCALIZED_LAYER_KEY = "area_a7b70aff_23f2_4876_9d4e_fa63036df0b2"
```
suricactus, line 26: You can change the IDs in the .qgs file to have human readable name, instead of the UUID. If you do so, then the constants are not required anymore.

from `apiary_853c29cb_ead8_4932_a60a_5cba33140b3d` to `valid_localized_point_layer_id`.

docker-app/qfieldcloud/core/tests/test_jobs.py
gounux: You're right, let me delete and replace that part

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,203 @@
...
+        # invalid localized layer
+        invalid_localized_layer = processfile_layers[
+            "invalid_localized_polygon_layer_id"
+        ]
+
+        self.assertFalse(invalid_localized_layer["is_valid"])
+        self.assertTrue(invalid_localized_layer["is_localized"])
+        self.assertEquals(
+            invalid_localized_layer["error_code"], "localized_dataprovider"
+        )
```
suricactus, line 108: If you define:
```
    def assertLayerData(self, layer_data: dict, is_valid: bool, is_localized: bool, error_code: str):
        self.assertEquals(layer_data["is_valid"], is_valid)
        self.assertEquals(layer_data["is_localized"], is_localized)
        self.assertEquals(layer_data["error_code"], error_code)
```
You can simplify the code and improve readability to something like:

```suggestion
        self.assertLayerData(
	        processfile_layers["invalid_localized_polygon_layer_id"],
	        False,
	        True,
	        "localized_dataprovider",
        )
```

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # invalid layer (datasource does not exist) (QGIS)
+        self.assertLayerData(
+            qgis_layers["invalid_point_layer_id"],
+            is_valid=False,
+            is_localized=False,
+            error_code="invalid_dataprovider",
+        )
+
+    def assertLayerData(
```
suricactus, line 206: Can you move this to the start of the class, right before the first `test_` method?

suricactus (approved):
One last small comment, please force push with the change and feel free to merge once addressed.

suricactus (approved):

PR #1090, opened 2024-12-21
Add `djangorestframework-stubs` for better typing

tdrivas:
This PR introduces the following updates to enhance type checking in our project:

- [x] Integrate mypy by adding `mypy_drf_plugin` to **mypy.ini** for better Django REST Framework type-checking.
- [x] Integrate pre-commit by updating **.pre-commit-config.yaml** to ensure `mypy` runs with `pre-commit` using the new plugin and settings.
- [x] Add djangorestframework-stubs to the **requirements.in** file in order to regenerate the appropriate **requirements.txt** 
- [x] Fix various typing issues such as type mismatches in exception, and incorrect annonations

Examples of issues:

1. `from rest_framework.views import View` --> The mypy-drf-plugin is strict about ensuring attributes or classes explicitly exported by a module. In simply words, mypy complains because it’s not officially listed in the __all__ attribute of the rest_framework.views module.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -9,3 +9,6 @@ ignore_errors: True
 # Turn off mypy for unit tests
 [mypy-*.tests.*]
 ignore_errors: True
+
+plugins =
+    mypy_drf_plugin.main
--- a/docker-app/qfieldcloud/core/drf_utils.py
+++ b/docker-app/qfieldcloud/core/drf_utils.py
@@ -1,6 +1,7 @@
 from django.db.models import QuerySet
-from django.http import HttpRequest
-from rest_framework import filters, viewsets
+from rest_framework import filters, views
+from rest_framework.request import Request
+from typing import Iterable
 
 
 class QfcOrderingFilter(filters.OrderingFilter):
@@ -15,7 +16,7 @@ class QfcOrderingFilter(filters.OrderingFilter):
     TOKENS_LIST_SEPARATOR = ","
     TOKENS_VALUE_SEPARATOR = "="
 
-    def _get_query_field(self, fields: list[str], term: str) -> str | None:
+    def _get_query_field(self, fields: Iterable[str], term: str) -> str | None:
         """Searches a term in a query field list.
 
         The field list elements may start with "-".
@@ -74,9 +75,9 @@ class QfcOrderingFilter(filters.OrderingFilter):
     def remove_invalid_fields(
         self,
         queryset: QuerySet,
-        fields: list[str],
-        view: viewsets.ModelViewSet,
-        request: HttpRequest,
+        fields: Iterable[str],
+        view: views.APIView,
+        request: Request,
     ) -> list[str]:
         """Process ordering fields by parsing custom field expression.
 
@@ -86,8 +87,8 @@ class QfcOrderingFilter(filters.OrderingFilter):
 
         Args:
             queryset (QuerySet): Django's ORM queryset of the same model as the one used in view of the `ModelViewSet`
-            fields (list[str]): ordering fields passed to the HTTP querystring
-            view (ModelViewSet): DRF view instance
+            fields (Iterable[str]): ordering fields passed to the HTTP querystring
+            view (APIView): DRF view instance
             request (HttpRequest): DRF request instance
         Returns :
             list[str]: parsed ordering fields where aliases have been replaced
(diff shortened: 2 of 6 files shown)


duke-nyuki:
Task linked: [QF-4916 Add `djangorestframework-stubs` for better typing](https://app.clickup.com/t/2192114/QF-4916)

.mypy.ini
```diff
@@ -9,3 +9,6 @@ ignore_errors: True
 # Turn off mypy for unit tests
 [mypy-*.tests.*]
 ignore_errors: True
+
+plugins =
+    mypy_drf_plugin.main
```
suricactus, line 14: Is this required? What happens if you don't have it?

tdrivas: In the original repo, the following is mentioned: "To make mypy aware of the plugin, you need to add to add it in the mypy.ini".  I guess even if the plugin is listed as a dependency in the pre-commit hook, it won’t be used unless activated via the mypy.ini. Or if it activated it might perform less accurate type checking. 

I tried to delete it from mypy.ini and I catch the same errors. Still investigating it.

suricactus: Would you mind sending a link to the "original repo"?

tdrivas: Here you are: https://github.com/typeddjango/djangorestframework-stubs/tree/master

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -21,7 +22,7 @@ class QFieldCloudException(Exception):
 
     code = "unknown_error"
     message = "QFieldcloud Unknown Error"
-    status_code = None
+    status_code: Optional[int] = None
```
suricactus, line 25:
```suggestion
    status_code: int | None = None
```

See https://peps.python.org/pep-0604/ .

We use Optional only for python that runs on python 3.9 or older.

docker-app/qfieldcloud/core/drf_utils.py
```diff
@@ -111,7 +114,7 @@ def remove_invalid_fields(
 
             definition_name, attrs = self._parse_definition(field_name)
             alias = attrs.get("alias", definition_name)
-            query_field_name = self._get_query_field(fields, alias)
+            query_field_name = self._get_query_field(list(fields), alias)
```
suricactus, line 117: You better change the type of the `_get_query_field` to accept `Iterable` than explicitly converting to `list` here.

docker-app/qfieldcloud/core/drf_utils.py
```diff
@@ -74,8 +76,8 @@ def _parse_definition(self, definition: str) -> tuple[str, dict[str, str]]:
     def remove_invalid_fields(
         self,
         queryset: QuerySet,
-        fields: list[str],
-        view: viewsets.ModelViewSet,
+        fields: Iterable[str],
```
suricactus, line 79: Don't forget to update the docstring too.

docker-app/qfieldcloud/core/drf_utils.py
```diff
@@ -74,8 +76,8 @@ def _parse_definition(self, definition: str) -> tuple[str, dict[str, str]]:
...
+        view: views.APIView,
         request: HttpRequest,
```
suricactus, line 81: Are we sure this is actually a `HttpRequest` and not `drf.Request`? I have a guess it will most probably be a wrong type here with the `HttpRequest`.

.mypy.ini
suricactus: Good, let's assume it is mandatory :)

docker-app/qfieldcloud/core/drf_utils.py
tdrivas: Updated!

docker-app/qfieldcloud/core/drf_utils.py
tdrivas: Correct, in order to increase the flexibility of the code.

docker-app/qfieldcloud/core/exceptions.py
tdrivas: Indeed, thank you! The operator | is also better for us, type less!

docker-app/qfieldcloud/core/drf_utils.py
tdrivas: The class `QfcOrderingFilter` inherits from `filters.OrderingFilter`, with the later to be part of Django REST Framework. Taking into account that the `OrderingFilter` works with DRF's Request, the most likely is the `request` a DRF one.

tdrivas: Changed to drf.request

suricactus (approved):
Thanks! Please squash all the commits into a single one and rebase on latest master. Then I will merge after second check.

suricactus (approved):
Thanks!

PR #1094, opened 2024-12-29
[optimization] Speed-up opening the authentication.authtoken admin by indexing `created_at` column

suricactus:
The reason is the default ORDER BY `created_at` , which makes a sequence scan on the whole table.

On the local database the difference is significant with and without that index: before: Execution Time: 175.196 ms
after: Execution Time: 0.562 ms


duke-nyuki:
Task linked: [QF-5021 Speed-up opening the authentication.authtoken admin by indexing `created_at` column](https://app.clickup.com/t/2192114/QF-5021)

PR #1096, opened 2024-12-29
[optimization] Migrate `django-auditlog` records to `jsonb` datatype

suricactus:
Remove the code needed for the two step migration, as it was already ran on app.qfield.cloud


duke-nyuki:
Task linked: [QF-4957 Migrate `django-auditlog` records to `jsonb` datatype](https://app.clickup.com/t/2192114/QF-4957)

PR #1097, opened 2024-12-29
Fix an error from `drf-spectacular` that returns 500 on GET /swagger.yml

suricactus:
The error was due to reference within a class for a class member without namespacing with the  class itself, or reference with `Status` instead of `Project.Status`.


duke-nyuki:
Task linked: [QF-5023 [sentry] Cannot open the swagger documentation](https://app.clickup.com/t/2192114/QF-5023)

PR #1101, opened 2025-01-07
Fix getting the list of available teams per organization

suricactus:
The problem comes from the fact that `teamname` is a runtime property of the `Team` class and not an actual database  field. In addition we were using `Team.object.values()` which returns an `Iterable[dict]`, instead of `Iterable[Team]`. So even if it returned the right type, it would have been terribly slow, due to N+1 situation.

This problem made the `/user/<username>/organization` return an error, therefore prevented QFieldSync from selecting the available organizations for the team.

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -187,10 +187,13 @@ class OrganizationSerializer(serializers.ModelSerializer):
 
     def get_teams(self, obj: Organization) -> list[str]:
         """Implementation of `SerializerMethodField` for `teams`. Returns list of team names."""
-        return [
-            t.teamname
-            for t in Team.objects.filter(team_organization=obj).values("username")
-        ]
+        team_qs = (
+            Team.objects.filter(team_organization=obj)
+            .select_related("team_organization")
+            .only("username", "team_organization__username")
+        )
+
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-5030 The `/users/<username>/organization` endpoint returns error 401 no matter what](https://app.clickup.com/t/2192114/QF-5030)

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -187,10 +187,13 @@ def get_members(self, obj):
...
-        return [
-            t.teamname
-            for t in Team.objects.filter(team_organization=obj).values("username")
-        ]
+        team_qs = (
+            Team.objects.filter(team_organization=obj)
+            .select_related("team_organization")
+            .only("username", "team_organization__username")
+        )
+
```
gounux, line 195: Quick question : do you consider this empty line as necessary ? (l195)

gounux (approved):

docker-app/qfieldcloud/core/serializers.py
suricactus: Usually I keep `keyword` lines separated from the rest of the code, not super hard enforced, but IMO makes it slightly easier to find where the `return` is.

PR #1102, opened 2025-01-07
Fix missing timezones

suricactus:
Basically we need to install all of them with `zoneinfo` package on
Ubuntu. It is also present as pip package, but I believe the Debian guys
are better at handling this.


duke-nyuki:
Task linked: [QF-5031 Fix missing timezones](https://app.clickup.com/t/2192114/QF-5031)

gounux (approved):

PR #1084, opened 2024-12-18
Bump django-allauth to 65.3.1

gounux:
This PR upgrades `django-allauth` dependency version to a newer one :

- [x] bumps `django-allauth` to `65.3.1` in requirements
- [x] adds `allauth.account.middleware.AccountMiddleware` to settings' `MIDDLEWARE` (see [Quickstart guide](https://docs.allauth.org/en/dev/installation/quickstart.html))
- [ ] ~removes `django-axes` from the requirements~ (edit : choice has been made to keep the `django-axes` rate limiter for the moment, since it provides a convenient way to re-enable a locked-out user, in the admin interface)
- [x] does not add any regression (check `authentication` module tests as well)

PR #1103, opened 2025-01-08
Bump `django-axes` to v7.0.1 and `django-allauth` to 65.3.1 

gounux:
This PR bumps the dependency to its newest version.

`axes`'s changelog: https://github.com/jazzband/django-axes/blob/master/CHANGES.rst#701-2024-12-02

Notable breaking changes: https://django-axes.readthedocs.io/en/latest/2_installation.html#version-7-breaking-changes-and-upgrading-from-django-axes-version-6

To be tested :

- try 6 unsuccessful logins with a correct username and a fake password -> custom QFC lockout page should appear, displaying username :

image

- try 6 unsuccessful logins with previous user's email address and a fake password -> custom QFC lockout page should appear, displaying email address:

image

- delete the access attempts in the admin page (or wait for 30min), then:
  - try again a login with username and correct password -> should be logged in
  - log out and try again a login with email address and correct password -> should be logged in

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -372,10 +372,10 @@ ACCOUNT_LOGOUT_ON_GET = True
 ###########################
 # The integer number of login attempts allowed before a record is created for the failed logins. Default: 3
 AXES_FAILURE_LIMIT = 5
-# If True, only lock based on username, and never lock based on IP if attempts exceed the limit. Otherwise utilize the existing IP and user locking logic. Default: False
-AXES_ONLY_USER_FAILURES = True
+# Configures the limiter to handle username only (see https://github.com/jazzband/django-axes/blob/4e89d72b92db044ff3f6b23ea2ab2e681211c98e/docs/2_installation.rst#version-6-breaking-changes-and-upgrading-from-django-axes-version-5)
+AXES_LOCKOUT_PARAMETERS = ["username"]
 # If set, defines a period of inactivity after which old failed login attempts will be cleared. If an integer, will be interpreted as a number of hours. Default: None
-AXES_COOLOFF_TIME = timedelta(minutes=30)
+AXES_COOLOFF_TIME = lambda request: timedelta(minutes=30)  # noqa: E731
 # If True, a successful login will reset the number of failed logins. Default: False
 AXES_RESET_ON_SUCCESS = True
 
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -375,7 +375,7 @@ AXES_FAILURE_LIMIT = 5
 # Configures the limiter to handle username only (see https://django-axes.readthedocs.io/en/latest/2_installation.html#version-7-breaking-changes-and-upgrading-from-django-axes-version-6)
 AXES_LOCKOUT_PARAMETERS = ["username"]
 # If set, defines a period of inactivity after which old failed login attempts will be cleared. If an integer, will be interpreted as a number of hours. Default: None
-AXES_COOLOFF_TIME = lambda request: timedelta(minutes=30)  # noqa: E731
+AXES_COOLOFF_TIME = lambda _request: timedelta(minutes=30)  # noqa: E731
 # If True, a successful login will reset the number of failed logins. Default: False
 AXES_RESET_ON_SUCCESS = True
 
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-4848 Bump django-axes to 6.0.0+](https://app.clickup.com/t/2192114/QF-4848)

docker-app/qfieldcloud/settings.py
```diff
@@ -372,10 +372,10 @@ def before_send(event, hint):
 ###########################
 # The integer number of login attempts allowed before a record is created for the failed logins. Default: 3
 AXES_FAILURE_LIMIT = 5
-# If True, only lock based on username, and never lock based on IP if attempts exceed the limit. Otherwise utilize the existing IP and user locking logic. Default: False
-AXES_ONLY_USER_FAILURES = True
+# Configures the limiter to handle username only (see https://django-axes.readthedocs.io/en/latest/2_installation.html#version-7-breaking-changes-and-upgrading-from-django-axes-version-6)
+AXES_LOCKOUT_PARAMETERS = ["username"]
```
suricactus, line 376: Added a bit of space to improve readability.
```suggestion

# Configures the limiter to handle username only.
# See https://django-axes.readthedocs.io/en/latest/2_installation.html#version-7-breaking-changes-and-upgrading-from-django-axes-version-6
AXES_LOCKOUT_PARAMETERS = ["username"]

```

Why don't we ratelimit on email login too?

docker-app/requirements/requirements.txt
```diff
@@ -77,7 +77,7 @@ django-appconf==1.0.6
     # via django-cryptography
 django-auditlog==3.0.0
     # via -r /requirements/requirements.in
-django-axes==5.40.1
+django-axes==7.0.1
```
suricactus, line 80: I assume you updated this file using `pipcompile`?

docker-app/qfieldcloud/settings.py
gounux: Interesting : `axes` rate limiter's lockout allowed parameters seem to be only `username`, `ip` and `useragent`

I get this error when trying to add somehow email to the `AXES_LOCKOUT_PARAMETERS` : 

```
'email' lockout parameter is not allowed. Allowed parameters: username, ip_address, user_agent
```

[This comment](https://github.com/jazzband/django-axes/issues/1133#issuecomment-1794649544) looks interesting : I tried setting the `AXES_USERNAME_FORM_FIELD = "email"`

-> the rate limiter works fine with email :

image

But, `axes` rate limiter then doesn't work with username, and it looks like `allauth`'s rate limiter is then activated :

image

Would it mean that `axes`'s rate limiter works only with `username` OR `email`, and not both of them simultaneously ?

Am going to burn a candle during this sunday morning catholic mass

suricactus: I think it works per unique login string. So if you have a `username` and `email@email.com`, they are two separate login strings and will be coutned separately. Is this what you observed?

Btw, why the `allauth` appears here, isn't it disabled?

docker-app/qfieldcloud/settings.py
```diff
@@ -372,10 +372,10 @@ def before_send(event, hint):
...
 # If set, defines a period of inactivity after which old failed login attempts will be cleared. If an integer, will be interpreted as a number of hours. Default: None
-AXES_COOLOFF_TIME = timedelta(minutes=30)
+AXES_COOLOFF_TIME = lambda request: timedelta(minutes=30)  # noqa: E731
```
suricactus, line 378:
```suggestion
AXES_COOLOFF_TIME = lambda _request: timedelta(minutes=30)  # noqa: E731
```

Just to indicate we do not want to use the param.

suricactus (approved):
Please squash and force push all in a single commit and we are ready to merge then.

docker-app/qfieldcloud/settings.py
gounux: > Ok, so no email in that list then.

Alright, so we go with username as lockout parameter.

> I think it works per unique login string. So if you have a username and email@email.com, they are two separate login strings and will be counted separately. Is this what you observed?

Correct : it actually creates 2 differents records in `Access attempts` objects:

image


> Btw, why the allauth appears here, isn't it disabled?

The `allauth` rate limiter is disabled in the other branch (#1084), after merging it into this one it should be fine

gounux:
> Please squash and force push all in a single commit and we are ready to merge then.

From other PR (#1084) which has been merged into this one :

> ideally be one per dependency with very strict commits and messages related to tme, so they can be easily reverted when needed

Wouldn't be willing to to mess things up here, and would like to ask you how to proceed with this one to get only a single commit _per_ dependency bump, assuming squashing all of them on this branch would regroup both bumps when merged on master later

suricactus (approved):

PR #1104, opened 2025-01-10
Introduce `filestorage` app, support for multiple file backends and store file metadata in the database

suricactus:
This PR introduces a new enhanced way to store filedata. Before just dumped everything to the Object Storage. This made everything clumsy, slow and not flexible.

The plan is to quickly migrate all projects to the new storage and delete the legacy approach.

I have closed #1065, which was the same PR.


duke-nyuki:

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,328 @@
...
+        project: Project,
+        filename: str,
+        content: ContentFile,
+        file_type: File.FileType,
+        uploaded_by: User,
+        uploaded_at: datetime | None = None,
+        created_at: datetime | None = None,
+        version_id: UUID | None = None,
+        package_job_id: UUID | None = None,
+    ) -> FileVersion:
```
gounux, line 156: I get this error when migrating project storage on my local (through `docker compose run app python manage.py migrateprojectstorage --all`):

```
  File "/usr/src/app/manage.py", line 20, in <module>
    main()
  File "/usr/src/app/manage.py", line 16, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/usr/src/app/qfieldcloud/filestorage/management/commands/migrateprojectstorage.py", line 100, in handle
    migrate_project_storage(project, "default", force)
  File "/usr/local/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/src/app/qfieldcloud/filestorage/migrate_project_storage.py", line 224, in migrate_project_storage
    raise err
  File "/usr/src/app/qfieldcloud/filestorage/migrate_project_storage.py", line 139, in migrate_project_storage
    _file_version = FileVersion.objects.add_version(
  File "/usr/local/lib/python3.10/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
TypeError: FileVersionQueryset.add_version() got an unexpected keyword argument 'id'
```

.env.example
```diff
@@ -38,17 +38,38 @@ LETSENCRYPT_RSA_KEY_SIZE=4096
...
+            "OPTIONS": {
+                "access_key": "minioadmin",
+                "secret_key": "minioadmin",
+                "bucket_name": "qfieldcloud-local",
+                "region_name": "",
+                "endpoint_url": "http://172.17.0.1:8009"
+            },
+            "QFC_IS_LEGACY": false
+        }
+    }'
```
gounux, line 72: Just to be sure, when migrating thus making old storage and new storage live at the same time on a local instance, should the `STORAGE` variable be something like :

```
STORAGES='{
    "legacy_storage": {
        "BACKEND": "qfieldcloud.filestorage.backend.QfcS3Boto3Storage",
        "OPTIONS": {
            "access_key": "minioadmin",
            "secret_key": "minioadmin",
            "bucket_name": "qfieldcloud-local",
            "region_name": "",
            "endpoint_url": "http://172.17.0.1:8009"
        },
        "QFC_IS_LEGACY": true
    },
    "default": {
        "BACKEND": "qfieldcloud.filestorage.backend.QfcS3Boto3Storage",
        "OPTIONS": {
            "access_key": "minioadmin",
            "secret_key": "minioadmin",
            "bucket_name": "qfieldcloud-local-something-new",
            "region_name": "",
            "endpoint_url": "http://172.17.0.1:8009"
        },
        "QFC_IS_LEGACY": false
    }
}'
```

.env.example
```diff
@@ -38,17 +38,38 @@ LETSENCRYPT_RSA_KEY_SIZE=4096
...
-STORAGE_SECRET_ACCESS_KEY=minioadmin
-STORAGE_BUCKET_NAME=qfieldcloud-local
-STORAGE_REGION_NAME=
-
-# URL to the storage endpoint either minio, or external (e.g. S3).
-# The URL must be reachable both from within docker and from the host, the default value is the `bridge` docker URL.
-# Read more on https://docs.docker.com/network/network-tutorial-standalone/ .
-# NOTE: to use minio on windows/mac, change the value to "http://host.docker.internal:8009"
-# DEFAULT: http://172.17.0.1:8009
-STORAGE_ENDPOINT_URL=http://172.17.0.1:8009
```
gounux, line 51: These env variables seem to be mandatory in order to set up `minio` container on local / standalone instances :

```
environment:
      MINIO_ROOT_USER: ${STORAGE_ACCESS_KEY_ID}
      MINIO_ROOT_PASSWORD: ${STORAGE_SECRET_ACCESS_KEY}
      MINIO_BROWSER_REDIRECT_URL: http://${QFIELDCLOUD_HOST}:${MINIO_BROWSER_PORT}
```

Should the `environment` key of the `minio` service be changed in the `docker-compose.override.standalone.yml` ?

gounux (commented):
First review focused on `docker compose` and environment / migrations considerations.

Did not look at the python code that much yet.

.env.example
suricactus: correct!

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -0,0 +1,227 @@
...
+
+                _file_version = FileVersion.objects.add_version(
+                    project=project,
+                    filename=file_version.name,
+                    content=django_thumbnail_file,
+                    file_type=File.FileType.PROJECT_FILE,
+                    uploaded_at=file_version.last_modified,
+                    uploaded_by=project.owner,
+                    created_at=now,
+                    id=file_version.id,
```
gounux, line 147:
```suggestion
                    version_id=file_version.id,
```

Tried to change this locally, then I get this error :

```
  File "/usr/src/app/manage.py", line 20, in <module>
    main()
  File "/usr/src/app/manage.py", line 16, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.10/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/usr/src/app/qfieldcloud/filestorage/management/commands/migrateprojectstorage.py", line 100, in handle
    migrate_project_storage(project, "default", force)
  File "/usr/local/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/src/app/qfieldcloud/filestorage/migrate_project_storage.py", line 224, in migrate_project_storage
    raise err
  File "/usr/src/app/qfieldcloud/filestorage/migrate_project_storage.py", line 131, in migrate_project_storage
    from_storage_bucket.download_fileobj(
  File "/usr/local/lib/python3.10/site-packages/boto3/s3/inject.py", line 843, in bucket_download_fileobj
    return self.meta.client.download_fileobj(
  File "/usr/local/lib/python3.10/site-packages/boto3/s3/inject.py", line 801, in download_fileobj
    return future.result()
  File "/usr/local/lib/python3.10/site-packages/s3transfer/futures.py", line 103, in result
    return self._coordinator.result()
  File "/usr/local/lib/python3.10/site-packages/s3transfer/futures.py", line 266, in result
    raise self._exception
  File "/usr/local/lib/python3.10/site-packages/s3transfer/tasks.py", line 269, in _main
    self._submit(transfer_future=transfer_future, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/s3transfer/download.py", line 354, in _submit
    response = client.head_object(
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 569, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 1023, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (405) when calling the HeadObject operation: Method Not Allowed
```

suricactus: Correct, sorry! I changed this several times and have not ran this for a while.

docker-app/qfieldcloud/filestorage/models.py
suricactus: Related to and fixed: https://github.com/opengisch/QFieldCloud/pull/1104#discussion_r1919730926

.env.example
suricactus: Yes, I need to address these.

suricactus:
@gounux apologies for the force push, I first rebased and then realized I had to force push.

docker-app/qfieldcloud/core/views/package_views.py
```diff
@@ -224,3 +249,228 @@ def post(self, request, project_id, job_id, filename):
...
+        description="Get all the files in a project package with files.",
+        responses={200: LatestPackageSerializer()},
+    ),
+)
+class LatestPackageView(views.APIView):
+    permission_classes = [permissions.IsAuthenticated, PackageViewPermissions]
+
+    def get(self, request, project_id):
+        """Get last project package status and file list."""
+        project = Project.objects.get(id=project_id)
```
gounux, line 265:
```suggestion
        project = get_object_or_404(Project, id=project_id)
```

This seem to be used in the view just below

docker-app/qfieldcloud/core/views/package_views.py
```diff
@@ -224,3 +249,228 @@ def post(self, request, project_id, job_id, filename):
...
+
+    def post(
+        self, request: Request, project_id: UUID, job_id: UUID, filename: str
+    ) -> Response:
+        """Upload the package files."""
+        # Only one file allowed to be uploaded at once
+        if len(request.FILES.getlist("file")) > 1:
+            raise exceptions.MultipleContentsError()
+
+        # project = get_object_or_404(Project, id=project_id)
```
gounux, line 373: Why is this commented ?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -0,0 +1,5 @@
+from storages.backends.s3boto3 import S3Boto3Storage
+
+
+class QfcS3Boto3Storage(S3Boto3Storage):
+    pass
```
gounux, line 5: Is this empty class mandatory ? Seems to be the one to use in the `STORAGE` env setting

docker-app/qfieldcloud/core/models.py
```diff
@@ -1022,6 +1069,8 @@ class Meta:
             )
         ]
 
+    files: "models.QuerySet[File]"
```
gounux, line 1072: Why is this between "" ? To avoid circular imports maybe ?

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,328 @@
...
+        )
+        qs = qs.prefetch_related("versions")
+        return qs
+
+    def get_by_name(self, filename: str):
+        return self.get(name=filename)
+
+
+class File(models.Model):
+    versions: QuerySet[FileVersion]
```
gounux, line 42: Just curious : is this the recommended way to get the possibility to fetch some reverse side objects from a OneToMany / FK relationship ?

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,328 @@
...
+
+    objects = FileQueryset.as_manager()
+
+    project = models.ForeignKey(
+        Project,
+        on_delete=models.CASCADE,
+        editable=False,
+        related_name="files",
+    )
+    # organization_id = models.ForeignKey(Organization, on_delete=models.CASCADE, editable=False)
```
gounux, line 61: Why is this commented ?

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,328 @@
...
+
+    package_job_id: int
+    package_job = models.ForeignKey(
```
gounux, line 64: Why storing the `id` and the `package_job` in two different attributes ?

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,328 @@
...
+    def __repr__(self) -> str:
+        return f'File({self.pk}) in "{self.project_id}" "{self.name}"'  # type: ignore
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+
+class FileVersionQueryset(models.QuerySet):
+    @transaction.atomic()
+    def add_version(
```
gounux, line 145: If I get this right, this is the only way to create a version right ?
No `FileVersion.objects.create(...)` ?

gounux (commented):
2nd review round, diving deeper into it.
Mainly curiosity, sometimes naive questions to understand better.
3rd one will come soon

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -101,6 +101,14 @@ class MultipleContentsError(QFieldCloudException):
     status_code = status.HTTP_400_BAD_REQUEST
 
 
+class ExplicitDeletionOfLastFileVersionError(QFieldCloudException):
+    """Raised when the only file version is explicitly requested for deletion."""
+
+    code = "explicit_deletion_of_last_version"
+    message = "Explicit deletion of last file version is not allowed!"
+    status_code = status.HTTP_400_BAD_REQUEST
```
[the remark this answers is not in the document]
suricactus: Because I think custom HTTP codes are not great practice. Some HTTP clients might not work great with unknown statuses and those provided by default are already good enough. We have the custom information in the `code` property which the client can interpret.

docker-app/qfieldcloud/core/fields.py
```diff
@@ -0,0 +1,38 @@
...
+
+        storage_name = instance._get_file_storage_name()
+        if not storage_name:
+            raise EmptyStorageNameError(instance=instance)
+
+        self.storage = storages[storage_name]
+
+
+class DynamicStorageFileField(models.FileField):
+    attr_class = DynamicStorageFieldFile
```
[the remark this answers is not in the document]
suricactus: Confusing, huh? I was there too. Luckily Django has nice docs about them: https://docs.djangoproject.com/en/5.1/ref/models/fields/#filefield-and-fieldfile

docker-app/qfieldcloud/filestorage/models.py
suricactus: I will remove it for now. The next task in this area is to make files available within the organization.

docker-app/qfieldcloud/filestorage/models.py
suricactus: This is just to make the typing happy. The `versions` attribute is defined here: https://github.com/opengisch/QFieldCloud/blob/467ad6141750e6dd9ced6ed2356becb035af59ff/docker-app/qfieldcloud/filestorage/models.py#L259

docker-app/qfieldcloud/filestorage/models.py
suricactus: The `package_job_id` is just to make the typing happy. You see there is no value assigned to it on purpose.

This is how Django FK works: https://docs.djangoproject.com/en/5.1/ref/models/fields/#database-representation

docker-app/qfieldcloud/filestorage/models.py
suricactus: You should not use `FileVersion.objects.create`, correct.

Thinking if we can highlight this more somehow.

docker-app/qfieldcloud/core/migrations/0080_file_storage_project_and_more.py
```diff
@@ -0,0 +1,81 @@
...
+            model_name="project",
+            name="file_storage",
+            field=models.CharField(
+                help_text="Which file storage provider should be used for the storing the project related files.",
+                max_length=100,
+                validators=[qfieldcloud.core.validators.file_storage_name_validator],
+                verbose_name="File storage",
+                default=qfieldcloud.core.models.get_project_file_storage_default,
+            ),
+        ),
```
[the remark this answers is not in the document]
suricactus: Unfortunately, no, it is because of the default value.

We first create a temporary default value, that populates the current "legacy storage" name. Note that the `preserve_default=False`, which means that there will be no default value permanently set.
Then we set the default to be `qfieldcloud.core.models.get_project_file_storage_default`, which is permanent.

docker-app/qfieldcloud/core/models.py
suricactus: Correct. I am shifting towards `from __future__ import annotations` but I want the whole module to do it at once. Sicne this module is quite big, I leave it for a future commit.

docker-app/qfieldcloud/filestorage/backend.py
suricactus: Yes, it is on purpose. In the `STORAGE` we define it as a string. Which means if we have to change it, we might change it to something completely different. Also the `STORAGE` comes from `.env` and we donn't want to change the `.env` every release.

Using an empty `QfcS3Boto3Storage` we allow ourselves the add more functionality transparently without any other modifications.

docker-app/qfieldcloud/core/views/package_views.py
suricactus: Because we no longer use the `Project` instance. But I uncommented for the validation purposes.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1028,7 +1031,7 @@ def get_project_thumbnail_upload_to(instance: "Project", _filename: str) -> str:
     return f"projects/{instance.id}/meta/thumbnail.png"
 
 
-class Project(models.Model):
+class Project(models.Model, FileStorageNameModelProtocol):
```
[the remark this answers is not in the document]
suricactus: Should be fixed, sorry. I was getting this too.

docker-app/qfieldcloud/core/migrations/0080_file_storage_project_and_more.py
```diff
@@ -0,0 +1,81 @@
...
+        ("core", "0079_organizationmember_created_at_and_more"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="project",
+            name="is_locked",
+            field=models.BooleanField(
+                default=False,
+                help_text="If set to true, the project is temporarily locked. Locking is internal QFieldCloud mechansism related to file storage migration or other file operations.",
```
gounux, line 26:
```suggestion
                help_text="If set to true, the project is temporarily locked. Locking is internal QFieldCloud mechanism related to file storage migration or other file operations.",
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1121,6 +1180,24 @@ class Meta:
...
+        ),
+        max_length=100,
+        validators=[validators.file_storage_name_validator],
+        default=get_project_file_storage_default,
+    )
+
+    is_locked = models.BooleanField(
+        _("Is locked"),
+        help_text=_(
+            "If set to true, the project is temporarily locked. Locking is internal QFieldCloud mechansism related to file storage migration or other file operations."
```
gounux, line 1196:
```suggestion
            "If set to true, the project is temporarily locked. Locking is internal QFieldCloud mechanism related to file storage migration or other file operations."
```

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,327 @@
...
+    def whatever(self):
+        qs = self.annotate(
+            last_version_pk=(
+                FileVersion.objects.filter(file=OuterRef("pk"))
+                .order_by("-uploaded_at")
+                .values("pk")[:1]
+            )
+        )
+        qs = qs.prefetch_related("versions")
+        return qs
```
gounux, line 35: Is this strictly mandatory or did you use this for testing ?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1092,8 +1141,18 @@ class Meta:
...
-        _("Thumbnail Picture URI"), max_length=255, blank=True
+    legacy_thumbnail_uri = models.CharField(
+        _("Legacy Thumbnail Picture URI"), max_length=255, blank=True
+    )
+
+    thumbnail = DynamicStorageFileField(
+        _("Thumbnail Picture"),
+        upload_to=get_project_thumbnail_upload_to,
+        # the s3 storage has 1024 bytes (not chars!) limit: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+        max_length=1024,
```
gounux, line 1152: Then why use 1024 characters and not the `STORAGE_FILENAME_MAX_CHAR_LENGTH` settings variable ?

docker-app/qfieldcloud/core/views/package_views.py
```diff
@@ -224,3 +249,228 @@ def post(self, request, project_id, job_id, filename):
...
+        last_job = project.last_package_job
+        if last_job.feedback.get("feedback_version") == "2.0":
+            layers = last_job.feedback["outputs"]["qgis_layers_data"]["layers_by_id"]
+        else:
+            steps = last_job.feedback.get("steps", [])
+            layers = (
+                steps[1]["outputs"]["layer_checks"]
+                if len(steps) > 2 and steps[1].get("stage", 1) == 2
+                else None
+            )
```
gounux, line 293: Find it hard to understand what this is supposed to be doing.
Would a comment be welcome ?

gounux (commented):
3rd review round.

I faced an error when creating a project with the basic OSM `.qgs` file :

- project exists in the database
- in the database, the project has `file_storage` value to `default`, but the files are stored in the legacy `bucket`
- running the `migrateprojectstorage` command throw this error : `Exception: Cannot migrate to storage "default", project is already stored there!`

After looking at the `Jobs` in QFC Admin, the `Process QGIS Project File` have `Failed` status with `UNKNOWN` Error type :

image

The `Output pre` display the error `Local files list for project "UUID": empty!` :

image

`Feedback pre` shows a `ProjectFileNotFoundException` thrown :

docker-app/qfieldcloud/core/models.py
suricactus: Good question!

Let me try to argue why, even though myself I am not convinced.

1) If you have a value that is dynamically set via a variable, it makes life much harder in Django migrations.
2) 1024 is actually quite long already, the `STORAGE_FILENAME_MAX_CHAR_LENGTH` is more of a windows limitation. I can imagine one day when bindows is dead and people only use filesystems that can handle arbirtrary length for paths.
3) the `STORAGE_FILENAME_MAX_CHAR_LENGTH` is the enduser length, relative to their base dir. However, we do some extra prefixes in our path, e.g. `projects/<project_id>/files/<user_path>/v20250202_hfsdjfhs`, so we add 100 or so characters on top.

Actually writing this, I think 3) is quite strong argument.

docker-app/qfieldcloud/filestorage/models.py
suricactus: Thanks for spotting. We need to kill this. It is super inefficient too.

docker-app/qfieldcloud/core/models.py
suricactus: Fixed independently.

docker-app/qfieldcloud/core/migrations/0080_file_storage_project_and_more.py
suricactus: Fixed independently.

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -0,0 +1,327 @@
...
+            uploaded_at__lt=self.uploaded_at,
+        ).order_by("-uploaded_at")
+
+        return file_version_qs.first()
+
+    @property
+    def next_version(self) -> "FileVersion | None":
+        file_version_qs = FileVersion.objects.filter(
+            file=self.file,
+            uploaded_at__lt=self.uploaded_at,
```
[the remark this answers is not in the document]
suricactus: Thanks for spotting! Fixed.

docker-app/qfieldcloud/core/views/package_views.py
suricactus: This is some ancient code. If you look, it is repeated twice in the file.

You are right, it is not great, so I tried to simplify a bit.

docker-app/qfieldcloud/filestorage/serializers.py
```diff
@@ -0,0 +1,121 @@
...
+
+@drf_spectacular.utils.extend_schema_serializer(
+    deprecate_fields=(
+        "last_modified",
+        "sha256",
+    )
+)
+class FileVersionSerializer(serializers.ModelSerializer):
+    md5sum = serializers.SerializerMethodField()
+    sha256 = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
suricactus: Why not? Each version has it's own sha and md5. It is required by the clients.

In the future, we will most probably stick to only one of the three:
- sha256
- md5
- s3 etag

I can see we go with the third options, as it is universal, easily reproducible and works in chunked file upload, while the other two require the whole file to be available.

docker-app/qfieldcloud/core/models.py
gounux: Indeed, thanks for the explanation and the arguments.

docker-app/qfieldcloud/filestorage/urls.py
```diff
@@ -1,6 +1,10 @@
 from django.urls import path
 
-from .views import compatibility_file_crud_view, compatibility_file_list_view
+from .views import (
```
gounux, line 3: QFC parliament voted: absolute imports! :D

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -188,24 +189,44 @@ def download_project_file_version(
...
+    request: Request,
+    field_file: FieldFile,
+    filename: str | None = None,
+    as_attachment: bool = False,
+) -> HttpResponseBase:
+    if not filename:
+        filename = field_file.name
+
+    if not filename:
+        raise Exception("Missing filename in `download_field_file`!")
```
gounux, line 210: Is this error supposed to be displayed to the user somehow ?

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -188,24 +189,44 @@ def download_project_file_version(
...
-            parameters={
-                "ResponseContentType": "application/force-download",
+            parameters={  # type: ignore
+                "ResponseContentType": "application.force-download",
                 "ResponseContentDisposition": f'attachment;filename="{filename}"',
             },
             # keep it a low number, in seconds
-            expire=600,
-            http_method="GET",
+            expire=600,  # type: ignore
```
gounux, line 228: Why not using a constant for this expiration time ?

gounux (commented):
Another quick review round. Thanks @suricactus

docker-app/qfieldcloud/filestorage/urls.py
suricactus: I have failed to enable this. Feel free to open a PR doing this, we can shift all the imports to absolute at once.

docker-app/qfieldcloud/filestorage/view_helpers.py
suricactus: In theory we can, but this number has never changed since it was first introduced.

docker-app/qfieldcloud/filestorage/view_helpers.py
suricactus: Added a comment:

```
    # While we should always have a filename, either as a parameter, or obtained from the uploaded file itself,
    # we can be paranoid and check if there is a one.
    # This scenario is more to prevent developers from mistakes, rather than real world situation.
```

docker-app/qfieldcloud/filestorage/view_helpers.py
gounux: You convinced me ;)

suricactus:
@gounux thanks for the review. I think I have fixed the last bit here. Lesson learne, you should always put a filename to a `ContentFile`.

PR #1111, opened 2025-01-23
Introduce `Project.the_qgis_file_name` and `Project.has_the_qgis_file`

suricactus:
This is a preparation PR for the bright future when the QGIS project file is not stored as a simple string, but it will be a foreign key from Project -> File. Most probably we will introduce a new field, that is called `Project.the_qgis_file`.

See PR #1065 .


duke-nyuki:
Task linked: [QF-5070 Rename the `Project.project_filename` to `Project.the_qgis_file`](https://app.clickup.com/t/2192114/QF-5070)

PR #1113, opened 2025-01-28
Store and display jobs' qgis app version

gounux:
This PR makes QFieldCloud store and display the QGIS app version used by Jobs

- [x] store the QGIS app version in a job's feedback
- [x] display the Job's QGIS version in admin

image

**Note** : since the QGIS app version is stored in job's feedback, it will not be possible to get and display "old" jobs this way. Extracting it from Job's `Output pre` with a bit of regexp could be considered, if old job's QGIS version must be fetched anyway :

image

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -943,6 +943,7 @@ class JobAdmin(QFieldCloudModelAdmin):
         "status",
         "error_type",
         "type",
+        "qgis_version",
         "created_at",
         "updated_at",
         "started_at",
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1764,6 +1764,30 @@ class Job(models.Model):
                 "The job ended in unknown state. Please verify the project is configured properly, try again and contact QFieldCloud support for more information."
             )
 
+    @property
+    def qgis_version(self) -> str | None:
+        """Returns QGIS app version used for the job.
+
+        The QGIS version is the one coming from the instanciated worker QGIS app.
+        The version number would be in `Major.Minor.Patch-NAME` format, e.g. 3.40.2-Bratislava
+
+        Returns:
+            str | None: QGIS version if found else None.
+        """
+        if not self.feedback:
+            return None
+
+        feedback_step_data = self.get_feedback_step_data("start_qgis_app")
+
+        if not feedback_step_data:
+            return None
+
+        if "qgis_version" in feedback_step_data["returns"]:
+            qgis_version = feedback_step_data["returns"]["qgis_version"]
+            return qgis_version
+
+        return None
+
     def check_can_be_created(self):
         from qfieldcloud.core.permissions_utils import (
             check_supported_regarding_owner_account,
--- a/docker-qgis/entrypoint.py
+++ b/docker-qgis/entrypoint.py
@@ -172,6 +172,7 @@ def cmd_package_project(args: argparse.Namespace):
                 id="start_qgis_app",
                 name="Start QGIS Application",
                 method=qfc_worker.utils.start_app,
+                return_names=["qgis_version"],
             ),
             Step(
                 id="download_project_directory",
@@ -250,6 +251,7 @@ def cmd_apply_deltas(args: argparse.Namespace):
                 id="start_qgis_app",
                 name="Start QGIS Application",
                 method=qfc_worker.utils.start_app,
+                return_names=["qgis_version"],
             ),
             Step(
                 id="download_project_directory",
@@ -308,6 +310,7 @@ def cmd_process_projectfile(args: argparse.Namespace):
                 id="start_qgis_app",
                 name="Start QGIS Application",
                 method=qfc_worker.utils.start_app,
+                return_names=["qgis_version"],
             ),
             Step(
                 id="download_project_directory",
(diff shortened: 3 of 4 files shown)


duke-nyuki:
Task linked: [QF-4970 Show information in the Web UI what is the current QGIS version used by QFieldCloud jobs](https://app.clickup.com/t/2192114/QF-4970)

docker-qgis/entrypoint.py
```diff
@@ -172,6 +172,7 @@ def cmd_package_project(args: argparse.Namespace):
                 id="start_qgis_app",
                 name="Start QGIS Application",
                 method=qfc_worker.utils.start_app,
+                return_names=["qgis_app", "qgis_app_version"],
```
suricactus, line 175: Ideally we should never pass complex objects between the steps in the workflow. I think we can make `start_app` to return just the qgis version and nothing else. We use the global  `QGISAPP` anyways.

```suggestion
                return_names=["qgis_version"],
```

docker-app/qfieldcloud/core/admin.py
```diff
@@ -924,6 +924,7 @@ class JobAdmin(QFieldCloudModelAdmin):
         "type",
         "status",
         "error_type",
+        "qgis_app_version",
```
suricactus, line 927: I don't think we need this info at this level. But we can show it as a readonly property when the Job form is opened.

suricactus (commented):
Two small changes before being ready to merge.

docker-qgis/entrypoint.py
gounux: Noted, thanks.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -943,6 +943,7 @@ class JobAdmin(QFieldCloudModelAdmin):
         "status",
         "error_type",
         "type",
+        "qgis_version",
```
gounux, line 946: @suricactus am wondering if this is the right place for this new `qgis_version` field, in the "list"

suricactus: Looks like.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1005,6 +1006,9 @@ def error_type(self, instance):
 
         return None
 
+    def qgis_version(self, instance: Job) -> str | None:
+        return instance.qgis_version
+
```
suricactus, line 1011: Is this required to exist? I don't think so, but might be wrong.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1764,6 +1764,27 @@ def fallback_output(self) -> str:
                 "The job ended in unknown state. Please verify the project is configured properly, try again and contact QFieldCloud support for more information."
             )
 
+    @property
+    def qgis_version(self) -> str | None:
+        """Returns QGIS app version used for the job.
+
```
suricactus, line 1770:
```suggestion
The QGIS version in formatted in ...
```

I think would be nice to give an example.

docker-qgis/qfc_worker/utils.py
```diff
@@ -158,9 +158,8 @@ def start_app():
 
         Returns
         -------
-        QgsApplication
-
-        A QgsApplication singleton
+        QgsApplication: A QgsApplication singleton
```
suricactus, line 161: This is no longer correct.

docker-app/qfieldcloud/core/admin.py
gounux: Actually I don't know and believe there might be some ways to avoid it, even if I like the verbosity of it.

docker-qgis/qfc_worker/utils.py
gounux: Indeed.

docker-app/qfieldcloud/core/admin.py
gounux: Okay actually it seems to work without it, thanks for the "suggestion".

Marking this conversation as resolved.

gounux:
@suricactus took the liberty to squash the commits. Ready for final review.

docker-app/qfieldcloud/core/models.py
suricactus:
```suggestion
The version number would be in `Major.Minor.Patch-NAME` format, e.g. 3.40.2-Bratislava
```

> I think would be nice to give an example.

suricactus (approved):
Ready to merge with a small comment

docker-app/qfieldcloud/core/models.py
```diff
@@ -1764,6 +1764,29 @@ def fallback_output(self) -> str:
                 "The job ended in unknown state. Please verify the project is configured properly, try again and contact QFieldCloud support for more information."
             )
 
+    @property
+    def qgis_version(self) -> str | None:
+        """Returns QGIS app version used for the job.
+        The QGIS version is the one coming from the instanciated worker QGIS app.
+        It is displayed in the job admin.
+        The version number would be in `Major.Minor.Patch-NAME` format, e.g. 3.40.2-Bratislava
```
suricactus, line 1772:
```suggestion

        The QGIS version is the one coming from the instanciated worker QGIS app.
        The version number would be in `Major.Minor.Patch-NAME` format, e.g. 3.40.2-Bratislava

```

I would add a bit of air around the sentences for readability and ignore where it is being shown, we might show it in other places too.

suricactus (approved):
Ready to be merged, small comment, no strong opinion.

suricactus (approved):

PR #1115, opened 2025-02-06
Add `.env` file CI step before generating Django strings for Transifex

gounux:
This PR is intended to fix the Transifex synchronization CI GitHub workflow for QFieldCloud Django strings.

Check this successful manually triggered Transifex sync Action : https://github.com/opengisch/QFieldCloud/actions/runs/13183139676.

### DIFF
--- a/.github/workflows/sync_translations.yml
+++ b/.github/workflows/sync_translations.yml
@@ -27,6 +27,15 @@ jobs:
             curl -OL https://github.com/transifex/cli/releases/download/v1.6.17/tx-linux-amd64.tar.gz
             tar -xvzf tx-linux-amd64.tar.gz
 
+      - name: Copy .env file
+        run: |
+          cp .env.example .env
+
+      - name: Perform docker pre operations
+        run: |
+            docker compose pull
+            docker compose build
+
       - name: Generate Django translation po files
         run: |
             docker compose run --user root app python manage.py makemessages -l es
@@ -35,7 +44,7 @@ jobs:
         run: ./tx push --source
 
       - name: Pull from Transifex
-        run: ./tx pull --all ---minimum-perc 0 --force
+        run: ./tx pull --all --minimum-perc 0 --force
 
       - name: Add and commit new translations
         uses: EndBug/add-and-commit@v9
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [DEV-1640 Translate QFieldCloud Admin](https://app.clickup.com/t/2192114/DEV-1640)

.github/workflows/sync_translations.yml
```diff
@@ -27,6 +27,11 @@ jobs:
             curl -OL https://github.com/transifex/cli/releases/download/v1.6.17/tx-linux-amd64.tar.gz
             tar -xvzf tx-linux-amd64.tar.gz
 
+      - name: Export the env variables file
+        run: |
+          cp .env.example .env
+          eval $(egrep "^[^#;]" .env | xargs -d'\n' -n1 | sed -E 's/(\w+)=(.*)/export \1='"'"'\2'"'"'/g')
```
gounux, line 33: Copy-pasted from the `test.yml` workflow.

I would highly appreciate some more explanations about this line, limits of my obscure linux/regex knowledge, I guess.

suricactus: I am not sure if the expression bash is needed.

.github/workflows/sync_translations.yml
```diff
@@ -27,6 +27,11 @@ jobs:
             curl -OL https://github.com/transifex/cli/releases/download/v1.6.17/tx-linux-amd64.tar.gz
             tar -xvzf tx-linux-amd64.tar.gz
 
+      - name: Export the env variables file
```
suricactus, line 30:
```suggestion
      - name: Copy .env file
```

If you do not export, better name the step like this.

suricactus (commented):
I would add a new separate "build docker images", or even better, use the cached ones from the github container repository, before calling "Generate Django translation po files" step. This is nicer because now the "generate" part is lost in thousands of lines of docke compose non-sense.

suricactus (commented):
Note the three dashes before the `---minimum-perc`.

suricactus (approved):
Great! IMO the pull and then build is kinda extra, but does not hurt.

PR #1118, opened 2025-02-10
Add S3 config for QFC on AWS hosts

faebebin:
When hosted on AWS botocore automatically uses the IAM auth and env settings for S3, therefore no further env/settings need to be defined and passed to boto3 s3.


suricactus:
@faebebin allowed myself to switch the base branch.

faebebin:
makes sense, thanks :)

lukasgraf (approved):
LGTM! 👍

PR #1121, opened 2025-02-14
Migrate avatar storage using django-storages instead of raw s3 queries

suricactus:
Migrate the avatars to the new storage too.


duke-nyuki:

PR #1122, opened 2025-02-19
[WIP] Use trial period on regular plan instead of trial plan

tdrivas:
**Current Approach**
Currently, we create two separate subscriptions:

- A Trial Subscription (is_trial=True) that lasts for TRIAL_PERIOD_DAYS.
- A Paid Subscription that is created after the trial expires.

**Problems with this approach:**

- Handling multiple subscriptions for one user
- Extra logic to fetch, cancel or update trial and regular subscriptions separately.

**New Approach**

- User asks for a trial subscription
- if the user has no remaining organization trials,  no subscription is created
- Otherwise, a subscription object is being created including, among others, a `trial_plan`, `a regular plan` and a default number of `trial days`.
- The trial subscription becomes active until the end of the trial. 
- When the trial ends and depending on the user input (payment method), the subscription update its status (ACTIVE_ * / INACTIVE_ *)


duke-nyuki:
Task linked: [QF-2912 Use trial period on regular plan instead of trial plan](https://app.clickup.com/t/2192114/QF-2912)

docker-app/qfieldcloud/subscription/admin.py
```diff
@@ -225,6 +229,14 @@ def save_model(self, request, obj, form, change):
 
         return super().save_model(request, obj, form, change)
 
+    @admin.display(description="Trial Active")
+    def is_trialing_display(self, instance):
+        return instance.is_trialing
+
+    @admin.display(description="Trial Ends At")
+    def trial_ends_at_display(self, instance):
+        return instance.trial_ends_at if instance.is_trialing else "-"
```
suricactus, line 238: I don't think this is needed. It is just a date, we better show it without any additional logic. If itis from the past, ok.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -438,12 +438,20 @@ class UpdateSubscriptionKwargs(TypedDict):
 
     uuid = models.UUIDField(default=uuid.uuid4, editable=False, db_index=True)
 
-    plan = models.ForeignKey(
+    regular_plan = models.ForeignKey(
```
suricactus, line 441: Could you please add some docs for each and every new model field you are adding? Especially these two, they need to be explained in detail.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -819,70 +841,45 @@ def create_subscription(
...
+        current_time = timezone.now()
+
+        subscription_data = {
+            "account": account,
+            "created_by": created_by,
+            "status": SubscriptionStatus.ACTIVE_PAID,
+            "active_since": current_time,
+            "active_until": None,  # Stripe will set this
+            "regular_plan": regular_plan,
+            "trial_plan": trial_plan,
```
suricactus, line 853:
```suggestion
            "trial_plan": None,
```
or
```suggestion
```

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -819,70 +841,45 @@ def create_subscription(
...
-            # be overwritten by a subscription plan change.
-            # But taking care of this would add quite some complexity.
-            if account.user.is_person:
-                account.user.remaining_trial_organizations = (
-                    regular_plan.max_trial_organizations
-                )
+                account.user.remaining_trial_organizations -= 1
                 account.user.save(update_fields=["remaining_trial_organizations"])
+            else:
+                logger.info(
```
suricactus, line 874:
```suggestion
                logger.warning(
```

suricactus: I am wondering if we want to throw an exception here?

docker-app/qfieldcloud/subscription/admin.py
```diff
@@ -117,11 +117,14 @@ class SubscriptionAdmin(QFieldCloudModelAdmin):
...
 
     fields = (
-        "plan",
+        "regular_plan",
+        "trial_plan",
         "account",
         "status",
         "active_since",
         "active_until",
+        "is_trialing_display",
```
suricactus, line 126:
```suggestion
        "is_trialing_display",
```

Can't we just use `is_trialing` without a custom rendering function?

docker-app/qfieldcloud/subscription/models.py
tdrivas: It should be considered an error. Thus, we can throw a custom exception (e.g. `NoRemainingTrialsException`) with a clear, user-friendly message and prevent further processing.

suricactus: Go for it!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Added!

docker-app/qfieldcloud/subscription/admin.py
tdrivas: Simplified!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Added!

docker-app/qfieldcloud/subscription/admin.py
tdrivas: Yes, as it is flagged as a property.

docker-app/qfieldcloud/subscription/admin.py
```diff
@@ -136,17 +139,19 @@ class SubscriptionAdmin(QFieldCloudModelAdmin):
         "id",
         "account__link",
         "account__user__email",
-        "plan__link",
+        "regular_plan__link",
+        "trial_plan__link",
```
suricactus, line 143: Don't you need to add the admin methods for these two, similar to `account__link`?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -819,70 +845,43 @@ def create_subscription(
...
-            )
-            trial_subscription = cls.objects.create(
-                plan=plan,
-                account=account,
-                created_by=created_by,
-                status=plan.initial_subscription_status,
-                active_since=active_since,
-                active_until=active_until,
+        current_time = timezone.now()
+
```
suricactus, line 849:
```suggestion
        now = timezone.now()
```

I think in pretty much every other case we name the variable `now`.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -819,70 +845,43 @@ def create_subscription(
...
-            # be overwritten by a subscription plan change.
-            # But taking care of this would add quite some complexity.
-            if account.user.is_person:
-                account.user.remaining_trial_organizations = (
-                    regular_plan.max_trial_organizations
-                )
+                account.user.remaining_trial_organizations -= 1
                 account.user.save(update_fields=["remaining_trial_organizations"])
+            else:
+                raise NoRemainingTrialOrganizationsError
```
suricactus, line 878:
```suggestion
                raise NoRemainingTrialOrganizationsError()
```

In the codebase we never throw a class, we always instantiate error classes. Valid also for JS.

docker-app/qfieldcloud/subscription/admin.py
tdrivas: My approach is for the `list_display` to show only the active plan. Since we already have the property `plan` in the model we can get only the active once. The reasons behind this approach is that it saves space, makes UI cleaner and still allows admins to differentiate trial status via the `is_trialing` column. If admin wants to see both the regular and trial plans separately, we could add filters.

docker-app/qfieldcloud/subscription/models.py
tdrivas: Indeed, thanks!

docker-app/qfieldcloud/subscription/admin.py
suricactus: ok, let's discuss it tomorrow. I thought if a column does not exist, as in our case `trial_plan__link` column does not exist, we need to write a `get_trial_plan__link` method, but I guess I am wrong.

suricactus:
Closing for now, it was a good excercise, but we are not there yet.

PR #1126, opened 2025-02-21
Reorder Organization admin tabs: members and teams should be neighbors

suricactus:
I believe the Organization members and Teams are very related and makes sense to be next to each other.

|before|after|
|-|-|
| image | image |

|before|after|
|-|-|
| image | image |


tdrivas (approved):
Maybe a short comment explaining why the order was change, can help future developers or even us.

suricactus:
@tdrivas updated the screenshot and added a bit of explanation.

PR #1134, opened 2025-03-02
Sign-up via invitation link process results in a broken `/accounts/profile` redirect

### LINKED ISSUE #1124: Sign up redirects to `/accounts/profile`
Hi,

I have deployed a self-instance with `v0.28.0` on a server. (I have also tested it on `v0.31.0`)

When I send an invitation, I can register a new user on the platform. When user is registered, the application redirects to `/accounts/profile` which is the default value of `LOGIN_REDIRECT_URL` https://docs.djangoproject.com/en/5.1/ref/settings/#login-redirect-url

Since there is no such url in the project, I have an error 404 Page not found which is not user-friendly for a user who would want to use the platform.

Solutions could be:
* implement a basic template for url `/accounts/url`
* change `LOGIN_REDIRECT_URL` value by exposing it in `settings.py` and read an env var

Am I missing a parameter here ? Which solution should be the best for QFC ?

Thanks

Benoît

suricactus:
Fixes #1124

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -310,6 +310,7 @@ EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 SITE_ID = 1
 
 LOGIN_URL = "account_login"
+LOGIN_REDIRECT_URL = "index"
 
 # Sentry configuration
 SENTRY_DSN = os.environ.get("SENTRY_DSN", "")


duke-nyuki:
Task linked: [QF-5302 Sign-up via invitation link process results in a broken `/accounts/profile` redirect](https://app.clickup.com/t/2192114/QF-5302)

PR #1135, opened 2025-03-02
[sentry] Fix comparison between of a constance setting and "Content-Length" header

tdrivas:
This PR resolves a TypeError that raised in attach_keys middleware when comparing `request.headers["Content-Length"]` (integer type) with `config.SENTRY_REQUEST_MAX_SIZE_TO_SEND` (string).

**Changes Implemented**

- Cast config.SENTRY_REQUEST_MAX_SIZE_TO_SEND to an integer before performing the comparison.
- Added error handling via defaulting max size to zero, if the value is invalid.

### DIFF
--- a/docker-app/qfieldcloud/core/middleware/requests.py
+++ b/docker-app/qfieldcloud/core/middleware/requests.py
@@ -17,15 +17,20 @@ def attach_keys(get_response):
     """
 
     def middleware(request):
+        try:
+            sentry_max_size = int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
+        except (ValueError, TypeError):
+            logger.error(
+                "Invalid SENTRY_REQUEST_MAX_SIZE_TO_SEND value. Defaulting to 0."
+            )
+            sentry_max_size = 0
+
         # add a copy of the request body to the request
         if (
             settings.SENTRY_DSN
             and request.method == "POST"
             and "Content-Length" in request.headers
-            and (
-                int(request.headers["Content-Length"])
-                < config.SENTRY_REQUEST_MAX_SIZE_TO_SEND
-            )
+            and (int(request.headers["Content-Length"]) < sentry_max_size)
         ):
             logger.info("Making a temporary copy for request body.")
 
--- a/docker-app/qfieldcloud/core/middleware/requests.py
+++ b/docker-app/qfieldcloud/core/middleware/requests.py
@@ -17,20 +17,15 @@ def attach_keys(get_response):
     """
 
     def middleware(request):
-        try:
-            sentry_max_size = int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
-        except (ValueError, TypeError):
-            logger.error(
-                "Invalid SENTRY_REQUEST_MAX_SIZE_TO_SEND value. Defaulting to 0."
-            )
-            sentry_max_size = 0
-
         # add a copy of the request body to the request
         if (
             settings.SENTRY_DSN
             and request.method == "POST"
             and "Content-Length" in request.headers
-            and (int(request.headers["Content-Length"]) < sentry_max_size)
+            and (
+                int(request.headers["Content-Length"])
+                < int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
+            )
         ):
             logger.info("Making a temporary copy for request body.")
 


duke-nyuki:
Task linked: [QF-5306 [sentry] Fix comparison between of a constance setting and "Content-Length" header](https://app.clickup.com/t/2192114/QF-5306)

docker-app/qfieldcloud/core/middleware/requests.py
```diff
@@ -17,15 +17,20 @@ def attach_keys(get_response):
...
 
     def middleware(request):
+        try:
+            sentry_max_size = int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
+        except (ValueError, TypeError):
+            logger.error(
+                "Invalid SENTRY_REQUEST_MAX_SIZE_TO_SEND value. Defaulting to 0."
+            )
+            sentry_max_size = 0
+
```
suricactus, line 27: Few things things to improve there:

- The logger does not show what is the invalid value.
- renaming the config to `sentry_max_size` make the code below harder to read, you better call again `int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)`
- setting `sentry_max_size = 0`  actually hides the error. We better want to fix it either via finding the root cause why the expected number is not a number, or just assume that somehow we get `"5"` which can be converted using the `int` function. Even if it throws again, we can then get the logs and fix the issue.

docker-app/qfieldcloud/core/middleware/requests.py
```diff
@@ -17,15 +17,20 @@ def attach_keys(get_response):
...
         # add a copy of the request body to the request
         if (
             settings.SENTRY_DSN
             and request.method == "POST"
             and "Content-Length" in request.headers
-            and (
-                int(request.headers["Content-Length"])
-                < config.SENTRY_REQUEST_MAX_SIZE_TO_SEND
-            )
+            and (int(request.headers["Content-Length"]) < sentry_max_size)
```
suricactus, line 33:
```suggestion
            and (
                int(request.headers["Content-Length"])
                < int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
            )
```
Even if you insist keeping the code in the previous hunk, this is the preferable change, so it is clear what is clear where `SENTRY_REQUEST_MAX_SIZE_TO_SEND` is coming from.

If I were you, I would drop the previous hunk and keep stick the PR only to the change proposed here.

docker-app/qfieldcloud/core/middleware/requests.py
tdrivas: > * The logger does not show what is the invalid value.
changed

> * renaming the config to `sentry_max_size` make the code below harder to read, you better call again `int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)`
ok, cool. 

> * setting `sentry_max_size = 0`  actually hides the error. We better want to fix it either via finding the root cause why the expected number is not a number, or just assume that somehow we get `"5"` which can be converted using the `int` function. Even if it throws again, we can then get the logs and fix the issue.

I guess the error is raised due to this (coming as string instead of int) or because it comes as a JSON string (`'{"__type__": "default", "__value__": aValue}'`, which means that we need to parse the JSON and extract "__value__".

docker-app/qfieldcloud/core/middleware/requests.py
tdrivas: I would suggest a more defensive way such as

```
        try:
            if (
            settings.SENTRY_DSN
            and request.method == "POST"
            and "Content-Length" in request.headers
            and (int(request.headers["Content-Length"]) < int(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND))
            ):
                logger.info("Making a temporary copy for request body.")

                input_stream = io.BytesIO(request.body)
                output_stream = io.BytesIO()
                shutil.copyfileobj(input_stream, output_stream)
                request.body_stream = output_stream
        except (ValueError, TypeError) as e:
            logger.error(
                f"Invalid SENTRY_REQUEST_MAX_SIZE_TO_SEND value: {config.SENTRY_REQUEST_MAX_SIZE_TO_SEND} "
                f"({type(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND).__name__}). Error: {e}"
            )
            raise 
```

docker-app/qfieldcloud/core/middleware/requests.py
suricactus: Is this an observation or a guess? The format you mentioned is the way the data is stored in the database, see https://github.com/jazzband/django-constance/blob/22bdb011db0f57562ff14203209a08dbcb2a5194/constance/codecs.py#L46

tdrivas: It is an indirect observation as I had the same problem with another config `TRIAL_PERIOD_DAYS`.

docker-app/qfieldcloud/core/middleware/requests.py
suricactus: I think this reduces the readability of the code. We have two paths for this error:

1) fix the root cause of the error, because constance is aware this must be an int but somehow it isn't.
2) just assume constance is temporarily failng to give the value as an int. If the value is not parsable int at all, then we need to fix 1) for sure.

suricactus: Just tried locally:
```
In [3]: from constance import config

In [4]: config.SENTRY_REQUEST_MAX_SIZE_TO_SEND
Out[4]: 1000

In [5]: type(config.SENTRY_REQUEST_MAX_SIZE_TO_SEND)
Out[5]: int
```

suricactus: Same on prod (different value)

tdrivas: Let's go with option 2 while staying alert for any errors. In case the value is not a parsable int at all, we will dig into it :)

> I think this reduces the readability of the code. We have two paths for this error:
> 
> 1. fix the root cause of the error, because constance is aware this must be an int but somehow it isn't.
> 2. just assume constance is temporarily failng to give the value as an int. If the value is not parsable int at all, then we need to fix 1) for sure.

docker-app/qfieldcloud/core/middleware/requests.py
suricactus: The value you received for `TRIAL_PERIOD_DAYS` was in the same JSON format as above, or you don't know?

tdrivas: I did encounter an error and inspected the returned object, which was in the JSON format described above. However, this turned out to be temporary as during the next test, the expected value loaded correctly.

suricactus (approved):
I believe we can give it a try and see if the errors persist.

docker-app/qfieldcloud/core/middleware/requests.py
suricactus: This is interesting, any way to reproduce?

PR #1136, opened 2025-03-02
Add some extra default settings to the cache configuration

suricactus:
Hopefully should fix all the `KeyError` and `OSError` exceptions.

This is the recommended setting from Django, with some inflated value for the number of pool members.
See: https://docs.djangoproject.com/en/5.1/topics/cache/#cache-arguments


duke-nyuki:
Task linked: [QF-5310 [sentry] Add enhanced cache configuration hopefully reducing the noise on Sentry](https://app.clickup.com/t/2192114/QF-5310)

gounux (approved):
Thanks! Code looks good to me, maybe we could add [a link to the documentation](https://docs.djangoproject.com/en/5.1/topics/cache/#cache-arguments) in the PR description ?

PR #1137, opened 2025-03-03
When reversing a URL, we need to pass it as `kwargs`, not just a dict

suricactus:
Causing a weird error:
```
TypeError: unhashable type: 'dict'
```

Follow-up of #1065


duke-nyuki:

gounux (approved):

suricactus:
@gounux I would request a second review, as there are some more changes required to make this run.

1) We need to return the absolute URL to the avatar in the REST API. Otherwise we are not longer backwards compatible.
2) Also add a smoke detection test if tests are failing when avatar is present. Not great test, but at least gives us some hints what is going on.
3) I made the tests less strict, but improved the error messages, which in this case I value more.

PR #1141, opened 2025-03-05
Do not raise when subscription overlap in the admin

tdrivas:
This PR handles the issue raised when the admin creates an overlaping subscription. 

**Before**
Previously, the system allowed admin users to create subscriptions without checking for overlaps. This resulted in conflicting active subscriptions, leading to IntegrityErrors due to PostgreSQL's exclusion constraint (subscription_subscription_prevent_overlaps).

**After**
This PR introduces strict validation to prevent overlapping subscriptions before saving them in the admin panel. It prevents overlapping subscriptions by checking:

- If the new subscription’s active_since falls between an existing subscription’s active_since and active_until.
- if active_until is NULL (meaning the subscription is still ongoing), the new subscription cannot start.
- ~~Ensures only ACTIVE_PAID, ACTIVE_PAST_DUE, TRIALING subscriptions are checked for conflicts.~~
- Raises a clear ValidationError in the admin panel instead of causing an IntegrityError and prevents object to be saved. 
 
Please see the attached screenshots demonstrating the behavior when overlapping subscriptions are attempted: 
 
 **Approach 1:**
image

**Approach 2:**
image

### DIFF
--- a/docker-app/qfieldcloud/subscription/admin.py
+++ b/docker-app/qfieldcloud/subscription/admin.py
@@ -9,6 +9,8 @@ from django.utils.translation import gettext as _
 from qfieldcloud.core.admin import QFieldCloudModelAdmin, model_admin_url
 
 from .models import PackageType, Plan, Subscription
+from django.db.models import Q
+from django.core.exceptions import ValidationError
 
 
 class PlanAdmin(admin.ModelAdmin):
@@ -220,6 +222,23 @@ class SubscriptionAdmin(QFieldCloudModelAdmin):
         return instance.account.user.email
 
     def save_model(self, request, obj, form, change):
+        conflicting_subscriptions = (
+            Subscription.objects.filter(
+                account=obj.account,
+                status__in=["active_paid", "active_past_due", "trialing"],
+            )
+            .filter(
+                Q(active_since__lte=obj.active_since)
+                & (Q(active_until__isnull=True) | Q(active_until__gte=obj.active_since))
+            )
+            .exclude(id=obj.id)
+        )
+
+        if conflicting_subscriptions.exists():
+            raise ValidationError(
+                "This account already has an active subscription. Please cancel it before adding a new one."
+            )
+
         if not change:
             obj.created_by_id = obj.created_by_id or request.user.id
 
--- a/docker-app/qfieldcloud/subscription/admin.py
+++ b/docker-app/qfieldcloud/subscription/admin.py
@@ -9,7 +9,6 @@ from django.utils.translation import gettext as _
 from qfieldcloud.core.admin import QFieldCloudModelAdmin, model_admin_url
 
 from .models import PackageType, Plan, Subscription
-from django.core.exceptions import ValidationError
 
 
 class PlanAdmin(admin.ModelAdmin):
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-3352 Do not raise when subscription overlap in the admin](https://app.clickup.com/t/2192114/QF-3352)

suricactus:
At least for me it still raises an ugly error, at least we have verified the dates are correct.

image

Can you paste a screenshot from your Django admin?

Few comment here:
- no need to check for the status, it should never happen
- no need to add greater/less than EQUAL. the equal is actually wrong here, we do allow equal values without causing overlaps.
- we can add a test that a value error is thrown.

tdrivas:
> At least for me it still raises an ugly error, at least we have verified the dates are correct.
> 
> image
> 
> Can you paste a screenshot from your Django admin?
Code has been updated, please check the description.

> Few comment here:
> 
> * no need to check for the status, it should never happen
Code has been updated, please check the description.

> * no need to add greater/less than EQUAL. the equal is actually wrong here, we do allow equal values without causing overlaps.
Fixed!

> * we can add a test that a ValidationError is thrown.
WIP

docker-app/qfieldcloud/subscription/admin.py
```diff
@@ -223,7 +223,7 @@ def save_model(self, request, obj, form, change):
         if not change:
             obj.created_by_id = obj.created_by_id or request.user.id
 
-        return super().save_model(request, obj, form, change)
```
suricactus, line 226: Why you removed the return?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -889,6 +889,37 @@ def create_subscription(
 
         return trial_subscription_obj, regular_subscription_obj
 
+    def clean(self):
+        """
+        Validates that the subscription's active period does not overlap with
+        any other active subscriptions for the same account.
+        """
+        # If there is no active_since, nothing to check.
```
suricactus, line 897:
```suggestion
        # If there is no `active_since`, nothing to check.
```

tdrivas: Fixed!

docker-app/qfieldcloud/subscription/admin.py
tdrivas: Accidentally.. Fixed now!

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -889,6 +889,37 @@ def create_subscription(
...
+    def clean(self):
+        """
+        Validates that the subscription's active period does not overlap with
+        any other active subscriptions for the same account.
+        """
+        # If there is no `active_since`, nothing to check.
+        if not self.active_since:
+            return
+
+        conflicts = (
```
suricactus, line 901:
```suggestion
        conflicting_subscriptions_qs = (
```

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -889,6 +889,37 @@ def create_subscription(
...
+            .filter(
+                Q(active_since__lt=self.active_since)
+                & (Q(active_until__isnull=True) | Q(active_until__gt=self.active_since))
+            )
+        )
+
+        if conflicts.exists():
+            raise ValidationError(
+                {
+                    "active_since": "This account already has an active subscription that overlaps with this period. Please cancel the existing subscription or choose a different date range."
```
suricactus, line 915:
```suggestion
                    "active_since": _("This account already has an active subscription that overlaps with this period. Please cancel the existing subscription or choose a different time range.")
```

We need to translate. Also we talk about time range, not only date.

docker-app/qfieldcloud/subscription/models.py
tdrivas: Better!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Yes, indeed, thanks!

suricactus (approved):

PR #1146, opened 2025-03-10
Fix case sensitivity issue in username handling

tdrivas:
This PR resolves an issue where Django Allauth was treating all usernames as lowercase during authentication, preventing users with capital letters in their usernames from logging in.

**The problem**
Usernames were correctly stored in the database with their original case (e.g., Bob), but authentication failed if users entered their names with uppercase letters.

**Solution**
Updated `ACCOUNT_PRESERVE_USERNAME_CASING` to `True` so to ensure usernames are not altered during authentication and are matched exactly as stored in the database.

**Impact**
- Users can now login using the their, exact case, username as stored in the db.
- No impact on existing stored usernames as only the authentication behavior has been adjusted.

**Test new functionality (Manual tests)**
- Tested logging in with usernames having uppercase letters 
- Ensured that new users can sign up and log in as expected.
- Verified password reset still works with case-sensitive usernames

### DIFF
--- a/docker-app/qfieldcloud/authentication/tests/test_authentication.py
+++ b/docker-app/qfieldcloud/authentication/tests/test_authentication.py
@@ -97,17 +97,7 @@ class QfcTestCase(APITransactionTestCase):
         self.assertTokenMatch(tokens[0], response.json())
         self.assertGreater(tokens[0].expires_at, now())
 
-    def test_prevent_duplicate_case_insensitive_registration(self):
-        with self.assertRaises(django.db.utils.IntegrityError):
-            Person.objects.create_user(username="USER1", password="abc123")
-
-        with self.assertRaises(django.db.utils.IntegrityError):
-            Person.objects.create_user(username="uSeR1", password="abc123")
-
-        users = Person.objects.filter(username__iexact="user1")
-        self.assertEqual(users.count(), 1)
-
-    def test_case_sensitive_user_login_with_session(self):
+    def test_login_with_session_case_insensitive(self):
         self.login_url = reverse("account_login")
 
         response = self.client.post(
@@ -192,6 +182,16 @@ class QfcTestCase(APITransactionTestCase):
         self.assertNotEqual(tokens[0], tokens[1])
         self.assertGreater(tokens[0].expires_at, now())
         self.assertGreater(tokens[1].expires_at, now())
+
+    def test_case_insensitive_username_uniqueness(self):
+        with self.assertRaises(django.db.utils.IntegrityError):
+            Person.objects.create_user(username="USER1", password="abc123")
+
+        with self.assertRaises(django.db.utils.IntegrityError):
+            Person.objects.create_user(username="uSeR1", password="abc123")
+
+        users = Person.objects.filter(username__iexact="user1")
+        self.assertEqual(users.count(), 1)
 
     def test_client_type(self):
         # QFIELDSYNC login
--- a/docker-app/qfieldcloud/authentication/tests/test_authentication.py
+++ b/docker-app/qfieldcloud/authentication/tests/test_authentication.py
@@ -63,7 +63,7 @@ class QfcTestCase(APITransactionTestCase):
         return response
 
     def test_login_logout(self):
-        response = self.login("UsEr1", "abc123")
+        response = self.login("user1", "abc123")
         tokens = self.user1.auth_tokens.order_by("-created_at").all()
 
         self.assertEqual(len(tokens), 1)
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-4962 Cannot login to admin if the username starts with a Capital letter](https://app.clickup.com/t/2192114/QF-4962)

suricactus (commented):
I am only afraid what will be the implications of such change. Let us discuss in the next meeting.

Btw, would you mind adding the docs in settings file since you are on it?

docker-app/qfieldcloud/settings.py
```diff
@@ -394,7 +394,7 @@ def before_send(event, hint):
 # Choose one of "mandatory", "optional", or "none".
 # For local development and test use "optional" or "none"
 ACCOUNT_EMAIL_VERIFICATION = os.environ.get("ACCOUNT_EMAIL_VERIFICATION")
-ACCOUNT_PRESERVE_USERNAME_CASING = False
+ACCOUNT_PRESERVE_USERNAME_CASING = True
```
suricactus, line 397:
```suggestion

# This setting determines whether the username is stored in lowercase (False) or whether its casing is to be preserved (True). 
# Note that when casing is preserved, potentially expensive __iexact lookups are performed when filter on username.
# For now, the default is set to True to maintain backwards compatibility.
# See https://docs.allauth.org/en/dev/account/configuration.html 
ACCOUNT_PRESERVE_USERNAME_CASING = True

```

suricactus:
@tdrivas let us know what tests you have performed to ensure no regressions are there.

tdrivas:
> @tdrivas let us know what tests you have performed to ensure no regressions are there.

Check updated description please!

suricactus:
> Test new functionality (Manual tests)
> Tested logging in with usernames having uppercase letters
> Ensured that new users can sign up and log in as expected.
> Verified password reset still works with case-sensitive usernames

A few other manual tests on top of the ones you performed:
- ensure usernames with lowercase letters can login
- ensure that username "SURICACTUS" or similar can access their profile no matter if they are written in lowercase or uppercase in the login form for: 
- a) Django admin
- b) app.qfiled.cloud login
- ensure that after login, one can load `host.com/a/SURICACTUS` and `host.com/a/suricactus` with the same result.

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -61,7 +63,7 @@ def login(self, username, password, user_agent="", success=True):
         return response
 
     def test_login_logout(self):
-        response = self.login("user1", "abc123")
+        response = self.login("UsEr1", "abc123")
```
suricactus, line 66: Why do you change existing tests?

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -80,6 +82,67 @@ def test_login_logout(self):
...
+        self.login_url = reverse("account_login")
+
+        response = self.client.post(
+            self.login_url, {"login": "user1", "password": "i_am_wrong"}, follow=True
+        )
+        # As we use a TemplateResponse we cannot check status_code 302 redirect,
+        # because it renders a template instead of returning an HTTP redirect response.
+        self.assertEqual(response.status_code, 200)
+        self.assertNotIn("_auth_user_id", self.client.session)
+        self.assertContains(response, "The username and/or password you specified are")
```
suricactus, line 120: Can you check for the full string? Or at least put a comment where it comes?

tdrivas: Added!

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -80,6 +82,67 @@ def test_login_logout(self):
...
+            follow=True,
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("_auth_user_id", self.client.session)
+
+        response = self.client.post(
+            self.login_url,
+            {
+                "login": "USER1",  # Correct case
```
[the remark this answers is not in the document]
tdrivas: "correct case" means that by using these credentials user must be authenticated. However, it is a bit confusing, so comments are deleted.

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -80,6 +82,67 @@ def test_login_logout(self):
...
+                "password": "abc123",
+            },
+            follow=True,
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("_auth_user_id", self.client.session)
```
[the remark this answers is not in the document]
tdrivas: Yes, thanks. Added and therefore no need to check  `        self.assertIn("_auth_user_id", self.client.session)` as it is now redundant

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -80,6 +82,70 @@ def test_login_logout(self):
...
+        with self.assertRaises(django.db.utils.IntegrityError):
+            Person.objects.create_user(username="USER1", password="abc123")
+
+        with self.assertRaises(django.db.utils.IntegrityError):
+            Person.objects.create_user(username="uSeR1", password="abc123")
+
+        users = Person.objects.filter(username__iexact="user1")
+        self.assertEqual(users.count(), 1)
+
+    def test_case_sensitive_user_login_with_session(self):
```
suricactus, line 110:
```suggestion
    def test_login_with_session_case_insensitive(self):
```

I believe we test the case insesitivity here, not sensitivity? I have suggested a name that is similar to `test_login_case_insensitive` above.

docker-app/qfieldcloud/authentication/tests/test_authentication.py
```diff
@@ -80,6 +82,70 @@ def test_login_logout(self):
...
+        self.assertGreater(tokens[0].expires_at, now())
+
+        response = self.login("USER1", "abc123")
+        tokens = self.user1.auth_tokens.order_by("-created_at").all()
+
+        self.assertEqual(len(tokens), 2)
+        self.assertTokenMatch(tokens[0], response.json())
+        self.assertGreater(tokens[0].expires_at, now())
+
+    def test_prevent_duplicate_case_insensitive_registration(self):
```
suricactus, line 100:
```suggestion
    def test_username_uniqueness_is_case_insensitive(self):
```

I would suggest we drop "registration" here, because we don't follow the full process of registering. I am ok if you adjust the suggested name further.

suricactus (commented):
Last two bits and we are ready.

docker-app/qfieldcloud/authentication/tests/test_authentication.py
tdrivas: Great!

docker-app/qfieldcloud/authentication/tests/test_authentication.py
tdrivas: Agree!

suricactus (approved):

suricactus:
I will leave this one for the next release. Not huge risk, but still sounds like an imporant settings we change.

PR #1147, opened 2025-03-10
Managing cross-project files (localized datasets)

tdrivas:
This PR introduces support for a new localized_datasets project, which serves as a centralized place to manage shared, localized files. This allows QFieldCloud users to reference common datasets across multiple projects without duplicating files. As a result, the updated logic ensures better synchronization, helps to avoid redundant uploads and provides QField with all the required information to efficiently access and fetch localized layers.

**Current Situation**
- The process_projectfile job only handled project files without checking for cross-project dependencies.
- Projects continued processing even when required datasets were missing.

**Actions to be implemented**
- [x] Added method `get_localized_datasets_project_id()` to retrieve the ID of the localized_datasets project for a given owner.
- [x] The ProjectSerializer now returns `localized_datasets_project_id` in the corresponding endpoint for a project.
- [x] Added method `get_missing_localized_layers()` to get the layers that are not found in the localized_datasets project, but are required for the project. 
- [x] Add tests

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1515,7 +1515,7 @@ class Project(models.Model):
                             "solution": solution,
                         }
                     )
-
+                # the layer is missing a primary key, warn it is going to be read-only
                 elif layer_data.get("layer_type_name") in ("VectorLayer", "Vector"):
                     if layer_data.get("qfc_source_data_pk_name") == "":
                         problems.append(
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1516,24 +1516,21 @@ class Project(models.Model):
                 layer_name = layer_data.get("name")
 
                 if layer_data.get("error_code") != "no_error":
-                    description = _(
-                        'Layer "{}" has an error with code "{}": {}'
-                    ).format(
-                        layer_name,
-                        layer_data.get("error_code"),
-                        layer_data.get("error_summary"),
-                    )
-                    solution = _(
-                        'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
-                    )
-
                     problems.append(
                         {
                             "layer": layer_name,
                             "level": "warning",
                             "code": "layer_problem",
-                            "description": description,
-                            "solution": solution,
+                            "description": _(
+                                'Layer "{}" has an error with code "{}": {}'
+                            ).format(
+                                layer_name,
+                                layer_data.get("error_code"),
+                                layer_data.get("error_summary"),
+                            ),
+                            "solution": _(
+                                'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
+                            ),
                         }
                     )
                 # the layer is missing a primary key, warn it is going to be read-only
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -29,7 +29,6 @@ from django.db.models.aggregates import Count, Sum
 from django.db.models.fields.json import JSONField
 from django.urls import reverse, reverse_lazy
 from django.utils.functional import cached_property
-from django.utils.safestring import SafeString
 from django.utils.translation import gettext as _
 from model_utils.managers import InheritanceManager, InheritanceManagerMixin
 from timezone_field import TimeZoneField
@@ -1243,6 +1242,27 @@ class Project(models.Model):
         default=False,
     )
 
+    def missing_localized_datasets(self) -> list[str]:
+        if not self.project_details:
+            return []
+
+        try:
+            localized_project = Project.objects.get(name="localized_datasets")
+            available_filenames = {
+                layer.get("filename") for layer in localized_project.localized_layers
+            }
+
+            missing = []
+            for layer in self.localized_layers:
+                filename = layer.get("filename")
+                if filename and filename not in available_filenames:
+                    missing.append(filename)
+
+            return missing
+
+        except Project.DoesNotExist:
+            return []
+
     @property
     def has_the_qgis_file(self) -> bool:
         return bool(self.the_qgis_file_name)
@@ -1435,80 +1455,61 @@ class Project(models.Model):
             )
 
         elif self.project_details:
-            localized_datasets_project = Project.objects.get(name="localized_datasets")
-            localized_datasets_project_layers = (
-                localized_datasets_project.localized_layers
-            )
-            available_localized_layers_filenames = [
-                localized_layer.get("filename")
-                for localized_layer in localized_datasets_project_layers
-            ]
+            try:
+                for missing_filename in self.missing_localized_datasets():
+                    problems.append(
+                        {
+                            "layer": missing_filename,
+                            "level": "warning",
+                            "code": "missing_localized_file",
+                            "description": _(
+                                'Localized Layer "{}" is missing in the centralized dataset project.'
+                            ).format(missing_filename),
+                            "solution": _(
+                                "Upload the missing file to the 'localized_datasets' project or update the layer to point to an available file."
+                            ),
+                        }
+                    )
+            except Project.DoesNotExist:
+                problems.append(
+                    {
+                        "layer": None,
+                        "level": "warning",
+                        "code": "missing_localized_project",
+                        "description": _(
+                            "Could not find the 'localized_datasets' project."
+                        ),
+                        "solution": _(
+                            "Ensure the shared dataset project exists and is correctly named."
+                        ),
+                    }
+                )
 
             for layer_data in self.project_details.get("layers_by_id", {}).values():
                 layer_name = layer_data.get("name")
 
                 if layer_data.get("error_code") != "no_error":
-                    solution: str | SafeString
+                    description = _(
+                        'Layer "{}" has an error with code "{}": {}'
+                    ).format(
+                        layer_name,
+                        layer_data.get("error_code"),
+                        layer_data.get("error_summary"),
+                    )
+                    solution = _(
+                        'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
+                    )
 
-                    if layer_data.get("error_code") == "localized_dataprovider":
-                        try:
-                            filename = layer_data.get("filename")
+                    problems.append(
+                        {
+                            "layer": layer_name,
+                            "level": "warning",
+                            "code": "layer_problem",
+                            "description": description,
+                            "solution": solution,
+                        }
+                    )
 
-                            if (
-                                filename
-                                and filename not in available_localized_layers_filenames
-                            ):
-                                problems.append(
-                                    {
-                                        "layer": filename,
-                                        "level": "warning",
-                                        "code": "missing_localized_file",
-                                        "description": _(
-                                            'Localized Layer "{}" is missing in the centralized dataset project.'
-                                        ).format(filename),
-                                        "solution": _(
-                                            "Upload the missing file to the 'localized_datasets' project or update the layer to point to an available file."
-                                        ),
-                                    }
-                                )
-
-                        except Project.DoesNotExist:
-                            problems.append(
-                                {
-                                    "layer": None,
-                                    "level": "warning",
-                                    "code": "missing_localized_project",
-                                    "description": _(
-                                        "Could not find the 'localized_datasets' project."
-                                    ),
-                                    "solution": _(
-                                        "Ensure the shared dataset project exists and is correctly named."
-                                    ),
-                                }
-                            )
-
-                    else:
-                        description = _(
-                            'Layer "{}" has an error with code "{}": {}'
-                        ).format(
-                            layer_name,
-                            layer_data.get("error_code"),
-                            layer_data.get("error_summary"),
-                        )
-                        solution = _(
-                            'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
-                        )
-
-                        problems.append(
-                            {
-                                "layer": layer_name,
-                                "level": "warning",
-                                "code": "layer_problem",
-                                "description": description,
-                                "solution": solution,
-                            }
-                        )
-                # the layer is missing a primary key, warn it is going to be read-only
                 elif layer_data.get("layer_type_name") in ("VectorLayer", "Vector"):
                     if layer_data.get("qfc_source_data_pk_name") == "":
                         problems.append(
@@ -1518,15 +1519,12 @@ class Project(models.Model):
                                 "code": "layer_problem",
                                 "description": _(
                                     'Layer "{}" does not support the `primary key` attribute. The layer will be read-only on QField.'
-                                ).format(
-                                    layer_name,
-                                ),
+                                ).format(layer_name),
                                 "solution": _(
                                     "To make the layer editable on QField, store the layer data in a GeoPackage or PostGIS layer, using a single column for the primary key."
                                 ),
                             }
                         )
-
         else:
             problems.append(
                 {
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -7,7 +7,7 @@ import uuid
 from datetime import datetime, timedelta
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import django_cryptography.fields
 from deprecated import deprecated
@@ -1065,7 +1065,7 @@ class Project(models.Model):
     @property
     def localized_layers(self) -> list[dict[str, Any]]:
         """
-        Retrieve all layers from project_details that are marked as localized.
+        Retrieve all layers from `Project.project_details` that have their `is_localized` flag set to `True`.
 
         Returns:
             A list of layer detail dictionaries where each dict has 'is_localized' == True.
@@ -1076,13 +1076,12 @@ class Project(models.Model):
 
         layers_by_id = self.project_details.get("layers_by_id", {})
 
-        localized = []
+        localized_layers = []
         for layer_detail in layers_by_id.values():
-            is_local = layer_detail.get("is_localized", False)
-            if is_local:
-                localized.append(layer_detail)
+            if layer_detail.get("is_localized", False):
+                localized_layers.append(layer_detail)
 
-        return localized
+        return localized_layers
 
     def _get_file_storage_name(self) -> str:
         """Returns the file storage name where all the files are stored. Used by `DynamicStorageFileField` and `DynamicStorageFieldFile`."""
@@ -1239,13 +1238,13 @@ class Project(models.Model):
         default=False,
     )
 
-    def get_localized_datasets_project(self) -> Optional[Project]:
+    def get_localized_datasets_project(self) -> Project | None:
         """
         Returns the 'localized_datasets' Project instance for the same owner,
         or None if no such project exists.
         """
         try:
-            project: Project = Project.objects.only("id").get(
+            project = Project.objects.get(
                 name="localized_datasets",
                 owner=self.owner,
             )
@@ -1279,7 +1278,7 @@ class Project(models.Model):
             )
         )
 
-        missing_layers = []
+        missing_localized_layers = []
         for layer in self.localized_layers:
             if "filename" not in layer:
                 continue
@@ -1287,9 +1286,9 @@ class Project(models.Model):
             filename = layer["filename"].split("localized:")[-1]
 
             if filename not in available_filenames:
-                missing_layers.append(layer)
+                missing_localized_layers.append(layer)
 
-        return missing_layers
+        return missing_localized_layers
 
     @property
     def has_the_qgis_file(self) -> bool:
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1063,6 +1063,17 @@ class Project(models.Model):
         QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    def get_localized_layers(self) -> list[dict]:
+        """
+        Returns a list of all localized layers (as dictionaries)
+        from project_details['layers_by_id'].
+        """
+        layers = (
+            self.project_details.get("layers_by_id", {}) if self.project_details else {}
+        )
+
+        return list(filter(lambda ld: ld.get("is_localized", False), layers.values()))
+
     def _get_file_storage_name(self) -> str:
         """Returns the file storage name where all the files are stored. Used by `DynamicStorageFileField` and `DynamicStorageFieldFile`."""
         return self.file_storage
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -29,7 +29,7 @@ from django.db.models.aggregates import Count, Sum
 from django.db.models.fields.json import JSONField
 from django.urls import reverse, reverse_lazy
 from django.utils.functional import cached_property
-from django.utils.safestring import SafeString, mark_safe
+from django.utils.safestring import SafeString
 from django.utils.translation import gettext as _
 from model_utils.managers import InheritanceManager, InheritanceManagerMixin
 from timezone_field import TimeZoneField
@@ -1074,6 +1074,17 @@ class Project(models.Model):
 
         return list(filter(lambda ld: ld.get("is_localized", False), layers.values()))
 
+    @property
+    def localized_layers(self):
+        layers_by_id = self.project_details.get("layers_by_id", {})
+
+        if self.name == "localized_datasets":
+            return list(layers_by_id.values())
+
+        return list(
+            filter(lambda ld: ld.get("is_localized", False), layers_by_id.values())
+        )
+
     def _get_file_storage_name(self) -> str:
         """Returns the file storage name where all the files are stored. Used by `DynamicStorageFileField` and `DynamicStorageFieldFile`."""
         return self.file_storage
@@ -1419,21 +1430,57 @@ class Project(models.Model):
                     ),
                 }
             )
+
         elif self.project_details:
+            localized_datasets_project = Project.objects.get(name="localized_datasets")
+            localized_datasets_project_layers = (
+                localized_datasets_project.localized_layers
+            )
+            available_names = [
+                localized_layer.get("name")
+                for localized_layer in localized_datasets_project_layers
+            ]
+
             for layer_data in self.project_details.get("layers_by_id", {}).values():
                 layer_name = layer_data.get("name")
 
                 if layer_data.get("error_code") != "no_error":
                     solution: str | SafeString
+
                     if layer_data.get("error_code") == "localized_dataprovider":
-                        description = _('Layer "{}" dataprovider is localized').format(
-                            layer_name
-                        )
-                        solution = mark_safe(
-                            _(
-                                'Make sure your <a href="https://docs.qfield.org/fr/how-to/outside-layers/">localized layer</a> is available on your QField device.'
+                        try:
+                            name = layer_data.get("name")
+
+                            if name and name not in available_names:
+                                problems.append(
+                                    {
+                                        "layer": name,
+                                        "level": "warning",
+                                        "code": "missing_localized_file",
+                                        "description": _(
+                                            'Localized Layer "{}" is missing in the centralized dataset project.'
+                                        ).format(name),
+                                        "solution": _(
+                                            "Upload the missing file to the 'localized_datasets' project or update the layer to point to an available file."
+                                        ),
+                                    }
+                                )
+
+                        except Project.DoesNotExist:
+                            problems.append(
+                                {
+                                    "layer": None,
+                                    "level": "warning",
+                                    "code": "missing_localized_project",
+                                    "description": _(
+                                        "Could not find the 'localized_datasets' project."
+                                    ),
+                                    "solution": _(
+                                        "Ensure the shared dataset project exists and is correctly named."
+                                    ),
+                                }
                             )
-                        )
+
                     else:
                         description = _(
                             'Layer "{}" has an error with code "{}": {}'
@@ -1446,15 +1493,15 @@ class Project(models.Model):
                             'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
                         )
 
-                    problems.append(
-                        {
-                            "layer": layer_name,
-                            "level": "warning",
-                            "code": "layer_problem",
-                            "description": description,
-                            "solution": solution,
-                        }
-                    )
+                        problems.append(
+                            {
+                                "layer": layer_name,
+                                "level": "warning",
+                                "code": "layer_problem",
+                                "description": description,
+                                "solution": solution,
+                            }
+                        )
                 # the layer is missing a primary key, warn it is going to be read-only
                 elif layer_data.get("layer_type_name") in ("VectorLayer", "Vector"):
                     if layer_data.get("qfc_source_data_pk_name") == "":
@@ -1473,6 +1520,7 @@ class Project(models.Model):
                                 ),
                             }
                         )
+
         else:
             problems.append(
                 {
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1068,14 +1068,17 @@ class Project(models.Model):
         Returns a list of all localized layers (as dictionaries)
         from project_details['layers_by_id'].
         """
-        layers = (
-            self.project_details.get("layers_by_id", {}) if self.project_details else {}
+        if self.project_details:
+            layers_by_id = self.project_details.get("layers_by_id", {})
+        else:
+            layers_by_id = {}
+
+        return list(
+            filter(lambda ld: ld.get("is_localized", False), layers_by_id.values())
         )
 
-        return list(filter(lambda ld: ld.get("is_localized", False), layers.values()))
-
     @property
-    def localized_layers(self):
+    def localized_layers(self) -> list[dict[str, Any]]:
         layers_by_id = self.project_details.get("layers_by_id", {})
 
         if self.name == "localized_datasets":
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1383,7 +1383,7 @@ class Project(models.Model):
 
     @property
     def files_count(self):
-        """
+        """*
         Todo:
             * Delete with QF-4963 Drop support for legacy storage
         """
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -7,7 +7,7 @@ import uuid
 from datetime import datetime, timedelta
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Dict, List, cast
 
 import django_cryptography.fields
 from deprecated import deprecated
@@ -1062,30 +1062,14 @@ class Project(models.Model):
         QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
-    def get_localized_layers(self) -> list[dict]:
-        """
-        Returns a list of all localized layers (as dictionaries)
-        from project_details['layers_by_id'].
-        """
-        if self.project_details:
-            layers_by_id = self.project_details.get("layers_by_id", {})
-        else:
-            layers_by_id = {}
-
-        return list(
-            filter(lambda ld: ld.get("is_localized", False), layers_by_id.values())
-        )
-
     @property
     def localized_layers(self) -> list[dict[str, Any]]:
+        if not self.project_details:
+            return []
+
         layers_by_id = self.project_details.get("layers_by_id", {})
 
-        if self.name == "localized_datasets":
-            return list(layers_by_id.values())
-
-        return list(
-            filter(lambda ld: ld.get("is_localized", False), layers_by_id.values())
-        )
+        return [ld for ld in layers_by_id.values() if ld.get("is_localized", False)]
 
     def _get_file_storage_name(self) -> str:
         """Returns the file storage name where all the files are stored. Used by `DynamicStorageFileField` and `DynamicStorageFieldFile`."""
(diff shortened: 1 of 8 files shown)


tdrivas:
@suricactus You can find my proposal about handling the localized datasets as a starting point for discussion. Questions, feedback and suggestions are highly appreciated!

tdrivas:
@suricactus First set of changes have been pushed. Please check whenever you can, thanks!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -95,6 +96,29 @@ def validate(self, data):
...
+                "id": layer.get("id"),
+                "name": layer.get("name"),
+                "datasource": layer.get("datasource"),
+                "filename": layer.get("filename"),
+            }
+            for layer in layers.values()
+            if layer.get("is_localized", False)
+        ]
+
+        return localized_datasets
```
suricactus, line 120:
```suggestion
        return filter(lambda ld: ld.get("is_localized", False),  layers.values())
```

Two suggestions:
- we avoid renaming the `layers_by_id` dictionary by simply returning those
- we slighthy pre-maturely optimize with a filter object instead of a list

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -37,7 +37,11 @@ def has_permission(self, request, view):
             return permissions_utils.can_create_project(user, owner_obj)
 
         projectid = permissions_utils.get_param_from_request(request, "projectid")
-        project = Project.objects.get(id=projectid)
+
+        try:
+            project = Project.objects.get(id=projectid)
+        except Project.DoesNotExist:
+            return False
```
suricactus, line 44: Just wondering if we can use `get_object_or_404` here, or it is not handled properly?

docker-app/qfieldcloud/core/urls.py
```diff
@@ -41,7 +41,23 @@
...
+        projects_views.ProjectViewSet.as_view(
+            {
+                "get": "retrieve",
+                "put": "update",
+                "patch": "partial_update",
+                "delete": "destroy",
+            }
+        ),
+        name="project-detail",
+    ),
```
suricactus, line 60: Why do we introduce these new API endpoints? What is the reason to kill the `projects/public` API? Also why do we explicitly pass the match between key/values? Finally, shall we really introduce a new naming convention for view names with `-` instead of `_` which is used along the whole codebase?

docker-app/qfieldcloud/core/serializers.py
tdrivas: Thanks for the suggestions:

- Moved the method to the Project model as proposed
- Used `filter()` with lambda to have a slight optimization
- Thanks for pointing this out, it's a great trade-off to consider, list or filter. My propose is to keep the list as it works better with serializers and we avoid any suprises.

docker-app/qfieldcloud/core/urls.py
tdrivas: You're absolutely right, I have restored the original URL structure. Thanks for pointing it out!

docker-app/qfieldcloud/core/models.py
```diff
@@ -1083,6 +1083,17 @@ class PackagingOffliner(models.TextChoices):
...
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    def get_localized_layers(self) -> list[dict]:
+        """
+        Returns a list of all localized layers (as dictionaries)
+        from project_details['layers_by_id'].
+        """
+        layers = (
+            self.project_details.get("layers_by_id", {}) if self.project_details else {}
+        )
```
suricactus, line 1093:
```suggestion
        if self.project_details:
            layers_by_id = self.project_details.get("layers_by_id", {}) 
        else:
            layers_by_id = {}
```
Avoid renaming keys. Also avoid inline ifs.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1083,6 +1083,17 @@ class PackagingOffliner(models.TextChoices):
...
+
+        return list(filter(lambda ld: ld.get("is_localized", False), layers.values()))
```
suricactus, line 1095:
```suggestion
        return filter(lambda ld: ld.get("is_localized", False), layers.values())
```
When you said "it works better with serializers" and "avoid any surprises", what did you mean?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1083,6 +1083,17 @@ class PackagingOffliner(models.TextChoices):
         QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    def get_localized_layers(self) -> list[dict]:
```
suricactus, line 1086:
```suggestion
    def get_localized_layers(self) -> list[LocalizedLayerDict]:
```
Can you please define a `LocalizedLayerDict(TypedDict)`, where you describe the fields with `total=False`?

suricactus:
```suggestion
    @property
    def localized_layers(self) -> list[dict]:
```
Another proposal would be make the method a dynamic property?

docker-app/qfieldcloud/core/models.py
tdrivas: `filter` lead to issues such as exhausted iterators. For example, if you return a filter object and then iterate over it twice, (e.g. once in a serializer's validation step and again in its representation step), then the first iteration will consume the iterator leading to an empty list during the second iteration.

suricactus: This is true, but what you have described is a property of any iterator, isn't it? I would leave the decision what to do with a returned iterator to the consumer of a method, rather than adding a performance penalty for consuming an iterator by default.

But if you find this confusing, I can live with the list, but in that case would be better if we rewrite it to:

```
localized_layers_data: list[LayerDataDict] = []

for layer_data in layers_by_id.values():
    if not layer_data.get("is_localized", False):
        continue

    localized_layers_data.append(layer_data)

return localized_layers_data
```

More imperative, more better.

docker-app/qfieldcloud/core/views/projects_views.py
tdrivas: So, the question is which is better in terms of logic: 403 ("Permission denied") coming from `return False` or 404 ("Not found") ? I guess the second.

docker-qgis/qfc_worker/utils.py
```diff
@@ -778,7 +792,7 @@ def get_layers_data(project: QgsProject) -> dict[str, dict]:
             "error_code": "no_error",
             "error_summary": error.summary() if error.messageList() else "",
             "error_message": layer.error().message(),
-            "filename": layer_source.filename,
+            "filename": filename,
```
tdrivas, line 795: @nirvn This field simplifies `datasource` and gives you the relative path along with the filename. 

e.g. localized:EMS/EMSR520_AOI01_DEL_PRODUCT_areaOfInterestA_r1_v1.kmz|layername=General Information|geometrytype=Polygon25D|uniqueGeometryType=yes  -->  EMS/EMSR520_AOI01_DEL_PRODUCT_areaOfInterestA_r1_v1.kmz

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -56,11 +56,25 @@ class Meta:
...
+
+
 class ProjectSerializer(serializers.ModelSerializer):
     owner = serializers.StringRelatedField()
     user_role = serializers.CharField(read_only=True)
     user_role_origin = serializers.CharField(read_only=True)
     private = serializers.BooleanField(allow_null=True, default=None)
+    localized_datasets = serializers.ListSerializer(
+        child=LayerSerializer(), source="get_localized_layers", read_only=True
+    )
```
tdrivas, line 77: @nirvn  No problem now to create a project :)

docker-app/qfieldcloud/core/views/projects_views.py
tdrivas: Changed!

docker-qgis/qfc_worker/utils.py
```diff
@@ -748,9 +749,22 @@ def get_layers_data(project: QgsProject) -> dict[str, dict]:
...
+            dom.setContent(layer.originalXmlProperties())
+            elements = dom.documentElement().elementsByTagName("datasource")
+
+            if not elements.isEmpty():
+                datasource = elements.at(0).toElement().text()
+                filename = datasource.split("localized:")[-1]
+
+                if "|" in filename:
+                    filename = filename.split("|")[0]
+
```
nirvn, line 766:
```suggestion
            datasource = bad_layer_handler.invalid_layer_sources_by_id.get(layer_id)
            filename = datasource.split("localized:")[-1]

            if "|" in filename:
                filename = filename.split("|")[0]

```

tdrivas: Replaced! Much cleaner and re-usable. Thanks both @suricactus  and @nirvn .

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -56,11 +56,27 @@ class Meta:
...
+    error_code = serializers.CharField(required=False)
+    error_summary = serializers.CharField(required=False)
+
+
 class ProjectSerializer(serializers.ModelSerializer):
     owner = serializers.StringRelatedField()
     user_role = serializers.CharField(read_only=True)
     user_role_origin = serializers.CharField(read_only=True)
     private = serializers.BooleanField(allow_null=True, default=None)
+    localized_datasets = serializers.ListSerializer(
```
nirvn, line 77: OK so on my end, in terms of what we get when getting a list of projects via `api/v1/projects/` , the list of files isn't really useful (we get them via /api/v1/packages/{project_id}/latest/ , which we call when we want to download a project and therefore is better suited there as we're in a download action)..

However, as discussed yesterday, what would be much more useful for QField is to get a localized_datasets_project_id value that points to that special project id, so we can fetch files via `/api/v1/files/{special project id}/my/{filename}` without having to fetch a list of projects to look for a project.name == "localized datasets" && project.owner == project_being_opened.owner match.

That's also a lot more future proof as it insures that the logic of determining what is the localized datasets project resides in QFC alone.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,31 @@ class PackagingOffliner(models.TextChoices):
...
+            layers_by_id = self.project_details.get("layers_by_id", {})
+        else:
+            layers_by_id = {}
+
+        return list(
+            filter(lambda ld: ld.get("is_localized", False), layers_by_id.values())
+        )
+
+    @property
+    def localized_layers(self) -> list[dict[str, Any]]:
```
suricactus, line 1080: Choose one of the two - either the @property or the function, but don't duplicate code.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1242,27 @@ class Meta:
         default=False,
     )
 
+    def missing_localized_datasets(self) -> list[str]:
+        if not self.project_details:
+            return []
+
+        try:
+            localized_project = Project.objects.get(name="localized_datasets")
```
suricactus, line 1250: This is not working, it get's all the `localized_dataset` projects? Where do we filter by user?


Also, I would suggest we either denormalize the database and add this on User/Useraccount model, or at least add a method that does it for us.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1242,27 @@ class Meta:
...
+            available_filenames = {
+                layer.get("filename") for layer in localized_project.localized_layers
+            }
+
+            missing = []
```
suricactus, line 1255: `missing` what?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1242,27 @@ class Meta:
         default=False,
     )
 
+    def missing_localized_datasets(self) -> list[str]:
```
suricactus, line 1245: We need a docstring on this function. I believe we can also improve the name, because 1) the name does not indicate this is a function as we are missing a verb; 2) we are not returning a list of "dataset", but returning a list of str?

Actually why don't we just return the layer data objects?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1242,27 @@ class Meta:
...
+        if not self.project_details:
+            return []
+
+        try:
+            localized_project = Project.objects.get(name="localized_datasets")
+            available_filenames = {
+                layer.get("filename") for layer in localized_project.localized_layers
```
suricactus, line 1252: I would rename the variable to `layer_data`, it's not the `layer` itself.

Also, what if there is no `filename` key, it will return `None`.

Finally, why do we get the `localized_project.localized_layers` from it? I guess the `localized_project.localized_layers` is always empty, no?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1242,27 @@ class Meta:
...
+            missing = []
+            for layer in self.localized_layers:
+                filename = layer.get("filename")
+                if filename and filename not in available_filenames:
+                    missing.append(filename)
+
+            return missing
+
+        except Project.DoesNotExist:
+            return []
```
suricactus, line 1264: I would prefer if we don't use try/catch in the logic, probably we can check for existance and do an early return?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1408,32 +1453,52 @@ def problems(self) -> list[dict]:
...
+                            "code": "missing_localized_file",
+                            "description": _(
+                                'Localized Layer "{}" is missing in the centralized dataset project.'
+                            ).format(missing_filename),
+                            "solution": _(
+                                "Upload the missing file to the 'localized_datasets' project or update the layer to point to an available file."
+                            ),
+                        }
+                    )
+            except Project.DoesNotExist:
```
suricactus, line 1473: Can you please elaborate when we have `Project.DoesNotExist` exception here, I could not follow in the code the scenario when it happens?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1408,32 +1453,52 @@ def problems(self) -> list[dict]:
...
         elif self.project_details:
+            try:
+                for missing_filename in self.missing_localized_datasets():
+                    problems.append(
+                        {
+                            "layer": missing_filename,
+                            "level": "warning",
+                            "code": "missing_localized_file",
+                            "description": _(
+                                'Localized Layer "{}" is missing in the centralized dataset project.'
```
suricactus, line 1466: Here the placeholder get's the filename, however reading the other examples in the file after the "layer" word usually we add the layer name in the project, which is different. Is it expected?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1408,32 +1453,52 @@ def problems(self) -> list[dict]:
...
-                            'Layer "{}" has an error with code "{}": {}'
-                        ).format(
-                            layer_name,
-                            layer_data.get("error_code"),
-                            layer_data.get("error_summary"),
-                        )
-                        solution = _(
-                            'Check the latest "process_projectfile" job logs for more info and reupload the project files with the required changes.'
-                        )
+                    description = _(
```
suricactus, line 1492: No need for the parasite variables if there are no if statements anymore.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1444,7 +1509,7 @@ def problems(self) -> list[dict]:
                             "solution": solution,
                         }
                     )
-                # the layer is missing a primary key, warn it is going to be read-only
```
suricactus, line 1447: Why did we remove this comment?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1454,9 +1519,7 @@ def problems(self) -> list[dict]:
                                 "code": "layer_problem",
                                 "description": _(
                                     'Layer "{}" does not support the `primary key` attribute. The layer will be read-only on QField.'
-                                ).format(
-                                    layer_name,
-                                ),
+                                ).format(layer_name),
```
suricactus, line 1522:
```suggestion
                                ).format(
                                    layer_name,
                                ),
```
The newline was on purpose to keep it very clear where the variables are coming from. If you check the whole `problems` function is following this for maintainability.

docker-app/qfieldcloud/core/serializers.py
suricactus: +1 for having the `localized_datasets` project ID easily available within the code and in the API. It should be on the user level on QFieldCloud. How do we expose it to the API - ideally it should follow the same pattern, but that would be inefficient, so I am ok that for each `Project` in GET /projects` we have additional key `localized_project_id`.

See one of my comments above that is suggesting the same in  `models.py`

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -298,8 +298,10 @@ def test_error_responses(self):
             "/api/v1/projects/{}/".format("a258db08-b1cb-4c34-a0cc-1e0e2a464f87")
         )
 
-        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
-        self.assertEqual(response.json()["code"], "object_not_found")
+        self.assertEqual(response.status_code, 404)
+        self.assertIn(
+            "Error 404 - Page Not Found", response.content.decode()
+        )  # Safer check
```
suricactus, line 304: Why this is a safer check? If we ask the API, we need to return an API response.

docker-qgis/qfc_worker/utils.py
```diff
@@ -748,9 +748,19 @@ def get_layers_data(project: QgsProject) -> dict[str, dict]:
...
 
-        if layer.dataProvider():
+        if layer_source.is_localized_path:
+            datasource = bad_layer_handler.invalid_layer_sources_by_id.get(layer_id)
+
+            if datasource and "localized:" in datasource:
+                filename = datasource.split("localized:")[-1]
+
+                if "|" in filename:
+                    filename = filename.split("|")[0]
```
suricactus, line 761: Ideally this logic should be moved to `libqfieldsync` , I would prefer we implement this logic only once. 

`LayerSource.localized_filename` or something like this?

Should be nearby this:
https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/layer.py#L899-L913

suricactus (commented):
Good progress, but there is a bit more work to be done before we have this one mergeable.

docker-app/qfieldcloud/core/tests/test_project.py
suricactus: I would not consider a string check "safer" or more "stable" than using the error code, or I am missing something? At least we need to check against `status.HTTP_404_NOT_FOUND` as we did before.

docker-app/qfieldcloud/core/serializers.py
suricactus: See my comment https://github.com/opengisch/QFieldCloud/pull/1147#discussion_r2045518741

tdrivas: Thanks for the input! I did update the API response for `/api/v1/projects/` to include a new field `localized_datasets_project_id` , which directly returns the special project ID associated with localized datasets for the same owner. Let me know if you need adjustments or additional metadata!

docker-app/qfieldcloud/core/models.py
tdrivas: @property won!

docker-app/qfieldcloud/core/models.py
tdrivas: Method was added!

docker-app/qfieldcloud/core/models.py
tdrivas: All suggestions approved!

docker-app/qfieldcloud/core/models.py
tdrivas: Changed so to retrieve the files from the localized_project via the File model!

docker-app/qfieldcloud/core/models.py
tdrivas: The `localized_project` is expected to be created at the time a localized data path is set via QGIS/QFieldSync. In case there is not such one, then we cannot get this project, as the command 
```
project = Project.objects.filter(
        name="localized_datasets", owner=self.owner
    ).first()

will give `None`. However, the exception `Project.DoesNotExist` is never reachable. I will kill it.

File: docker-app/qfieldcloud/core/models.py (hunk 18 above)
Comment (tdrivas) [reply docker-app/qfieldcloud/core/models.py:1472]: `Filename` gives the relative path along with the extension, whereas the `name` give us only the name without even the extension. 

e.g.
"name": "raster_crop_rotation_quinquennially_2023",                                                                                                                                                   "filename": "rasters/raster_crop_rotation_quinquennially_2023.tif",

File: docker-app/qfieldcloud/core/serializers.py (hunk 10 above)
Comment (nirvn) [reply docker-app/qfieldcloud/core/serializers.py]: @tdrivas , brilliant, just tested locally, much much better to me IMHO.

File: docker-app/qfieldcloud/core/tests/test_project.py (hunk 22 above)
Comment (tdrivas) [reply docker-app/qfieldcloud/core/tests/test_project.py:304]: What is happening is the following: The view related to the projects uses Django's built-in `get_object_or_404`, which raises a standard Django Http404 exception. This forces Django to respond with a default HTML error page rather than a JSON response. As a result, the API test expecting JSON fails with a content-type mismatch and thus, I had changed it. Fix in progress.

File: docker-app/qfieldcloud/core/models.py (hunk 21 above)
Comment (tdrivas) [reply docker-app/qfieldcloud/core/models.py:1528]: Changed!

File: docker-app/qfieldcloud/core/models.py (hunk 20 above)
Comment (tdrivas) [reply docker-app/qfieldcloud/core/models.py]: A funny little accident, but the patient pulled through!

File: docker-qgis/qfc_worker/utils.py (hunk 23 above)
Comment (tdrivas) [reply docker-qgis/qfc_worker/utils.py:761]: Can we finalize the PR without it or shall we include it in this PR ?

File: docker-app/qfieldcloud/core/tests/test_project.py (hunk 22 above)
Comment (tdrivas) [reply docker-app/qfieldcloud/core/tests/test_project.py]: Still no luck, as the view has to be updated and not retrieve the project_id inside the `has_permission` function. New PR or shall we re-construct the view?

Currently, I go with the `assertIn` test until we decide on the above.

**Comment by nirvn** [comment]
Updated QField build to work with the localized_datasets_project_id property: https://github.com/opengisch/QField/pull/6194

File: docker-app/qfieldcloud/core/models.py [hunk 24]
```diff
@@ -1218,6 +1226,49 @@ class Meta:
...
+    def get_localized_datasets_project_id(self) -> str | None:
+        """
+        Returns the ID of the 'localized_datasets' project for the same owner.
+        """
+        project = Project.objects.filter(
+            name="localized_datasets", owner=self.owner
+        ).first()
+        return project.id if project else None
+
+    def get_missing_localized_layers(self) -> List[Dict]:
```
suricactus, line 1238: Why on line 1066 you use `list[dict[...]]` and here you import and use `List[Dict[]]`? Please make sure you catch this kind of inconsistencies yourself.

Also, please always a bit of typing what to expect from the `dict[...]`.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1226,49 @@ class Meta:
...
+        """
+        Identifies and returns a list of localized layer metadata dictionaries
+        that are not present in the associated localized_datasets project storage.
+
+        Returns:
+            List[Dict]: A list of layer dictionaries missing from the
```
suricactus, line 1244:
```suggestion
            A list of layer dictionaries missing from the
```

No need anymore, see https://github.com/opengisch/QFieldCloud/pull/1213

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -56,11 +56,31 @@ class Meta:
...
+
+
 class ProjectSerializer(serializers.ModelSerializer):
     owner = serializers.StringRelatedField()
     user_role = serializers.CharField(read_only=True)
     user_role_origin = serializers.CharField(read_only=True)
     private = serializers.BooleanField(allow_null=True, default=None)
+    localized_datasets_project_id = serializers.SerializerMethodField(read_only=True)
+
+    def get_localized_datasets_project_id(self, obj):
```
suricactus, line 79: Why do we have the same code at multiple places? Can't we reuse the code that exist in the models?

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -299,8 +299,7 @@ def test_error_responses(self):
         )
 
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
-        self.assertEqual(response.json()["code"], "object_not_found")
-
+        self.assertIn("Not Found", response.content.decode())
```
suricactus, line 302: What is the motivation for this change?

docker-qgis/qfc_worker/utils.py
suricactus: We can, but please add a TODO describing the change and ideally also a link to a ClickUp task in the TODO comment.

docker-app/qfieldcloud/core/models.py
suricactus: Let me know when ready.

docker-app/qfieldcloud/core/models.py
suricactus: yes, and if I read the code correct, it will say:

>Localized Layer "/rasters/raster_crop_rotation_quinquennially_2023.tif" is missing...

But we always put the layer name in the message after the "Layer" word, so I would expect to have something like:

> Localized Layer "raster_crop_rotation_quinquennially_2023" is missing...

or if needed to rework the other words:

>Localized dataset stored at "/rasters/raster_crop_rotation_quinquennially_2023.tif" is missing...

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1226,49 @@ class Meta:
...
     )
 
+    def get_localized_datasets_project_id(self) -> str | None:
+        """
+        Returns the ID of the 'localized_datasets' project for the same owner.
+        """
+        project = Project.objects.filter(
+            name="localized_datasets", owner=self.owner
+        ).first()
+        return project.id if project else None
```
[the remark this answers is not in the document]
suricactus: Actually I am coming back to this method after a while in reviewing, why don't you make it `get_localized_datasets_project` and return the project instance itself?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1226,49 @@ class Meta:
...
+        available_filenames = set(
+            File.objects.filter(project_id=localized_project_id).values_list(
+                "name", flat=True
+            )
+        )
+        missing_layers = [
+            layer
+            for layer in self.localized_layers
+            if "filename" in layer
+            and layer["filename"].split("localized:")[-1] not in available_filenames
```
suricactus, line 1267: You do this `localized:` split in two places, so it seems we need to move it to a proper place a soon as possible. Please add a TODO.

Also, don't do a super long inline for loop with and if, please rewrite as imperative code.

docker-app/qfieldcloud/core/models.py
```diff
@@ -7,7 +7,7 @@
 from datetime import datetime, timedelta
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Dict, List, cast
```
suricactus, line 10: If you are writing for QFieldCloud and see a change that import `Dict` or `List` , it is already a flag you will get a review comment :)

suricactus (changes requested):
A few more changes required :)

docker-app/qfieldcloud/core/models.py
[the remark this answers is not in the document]
tdrivas: Yes, correct! Changed!

docker-app/qfieldcloud/core/models.py
tdrivas: Updated!

docker-app/qfieldcloud/core/serializers.py
tdrivas: Yes, now i use a single source of truth!

docker-app/qfieldcloud/core/models.py
tdrivas: Inline is simplified and TODO is added

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,15 @@ class PackagingOffliner(models.TextChoices):
...
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    @property
+    def localized_layers(self) -> list[dict[str, Any]]:
+        if not self.project_details:
+            return []
+
+        layers_by_id = self.project_details.get("layers_by_id", {})
+
+        return [ld for ld in layers_by_id.values() if ld.get("is_localized", False)]
```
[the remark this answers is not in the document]
tdrivas: Inline has been killed

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,15 @@ class PackagingOffliner(models.TextChoices):
         QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    @property
+    def localized_layers(self) -> list[dict[str, Any]]:
```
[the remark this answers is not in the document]
tdrivas: Added!

docker-app/qfieldcloud/core/models.py
tdrivas: Correct, thanks, changed!

docker-app/qfieldcloud/core/tests/test_project.py
tdrivas: Please see previous comments on this test file:

What is happening is the following: The view related to the projects uses Django's built-in get_object_or_404, which raises a standard Django Http404 exception. This forces Django to respond with a default HTML error page rather than a JSON response. As a result, the API test expecting JSON fails with a content-type mismatch and thus, I had changed it. Fix in progress.

(https://github.com/opengisch/QFieldCloud/pull/1147#discussion_r2051553148)
Still no luck, as the view has to be updated and not retrieve the `project_id` inside the has_permission function. New PR or shall we re-construct the view?
Currently, I go with the assertIn test until we decide on the above.

docker-app/qfieldcloud/core/models.py
tdrivas: Removed from imports and sticked to the list/dict!

docker-app/qfieldcloud/core/models.py
tdrivas: Changed!

docker-app/qfieldcloud/core/models.py
tdrivas: Done!

docker-app/qfieldcloud/core/models.py
tdrivas: Try/catch has been killed!

docker-app/qfieldcloud/core/tests/test_project.py
suricactus: We should return a proper JSON response somehow, if needed we can change from `get_object_or_404` to manually throwing the `object_not_found` error, but I believe we can fix this already on the error handler .

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,28 @@ class PackagingOffliner(models.TextChoices):
...
+        """
+        if not self.project_details:
+            return []
+
+        layers_by_id = self.project_details.get("layers_by_id", {})
+
+        localized = []
+        for layer_detail in layers_by_id.values():
+            is_local = layer_detail.get("is_localized", False)
+            if is_local:
```
suricactus, line 1082:
```suggestion
            if layer_detail.get("is_localized", False):
```

Do we really need the variable? If so, why don't we name it `is_localized` instead of renaming it to `is_local`?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,28 @@ class PackagingOffliner(models.TextChoices):
         QGISCORE = "qgiscore", _("QGIS Core Offline Editing (deprecated)")
         PYTHONMINI = "pythonmini", _("Optimized Packager")
 
+    @property
+    def localized_layers(self) -> list[dict[str, Any]]:
+        """
+        Retrieve all layers from project_details that are marked as localized.
```
suricactus, line 1068:
```suggestion
        Retrieve all layers from `Project.project_details` that have their `is_localized` flag set to `True`.
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1063,6 +1062,28 @@ class PackagingOffliner(models.TextChoices):
...
+        Returns:
+            A list of layer detail dictionaries where each dict has 'is_localized' == True.
+            If project_details is missing or empty, returns an empty list.
+        """
+        if not self.project_details:
+            return []
+
+        layers_by_id = self.project_details.get("layers_by_id", {})
+
+        localized = []
```
suricactus, line 1079:
```suggestion
        localized_layers = []

```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1239,58 @@ class Meta:
         default=False,
     )
 
+    def get_localized_datasets_project(self) -> Optional[Project]:
+        """
+        Returns the 'localized_datasets' Project instance for the same owner,
+        or None if no such project exists.
+        """
+        try:
+            project: Project = Project.objects.only("id").get(
```
suricactus, line 1248:
```suggestion
            project: Project = Project.objects.get(
```
Since now we return the whole project, I guess we don't need to select only the id?

suricactus: Also do we need the explicit typing for the `project` variable? I am surprised we don't have the correct time from `Project.objects.get()`.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1239,58 @@ class Meta:
...
+            # Return all layers if the project is missing
+            return self.localized_layers
+
+        available_filenames = set(
+            File.objects.filter(project_id=localized_project.id).values_list(
+                "name", flat=True
+            )
+        )
+
+        missing_layers = []
```
suricactus, line 1282:
```suggestion
        missing_localized_layers = []
```

docker-app/qfieldcloud/core/models.py
suricactus: Are you ok to kill the `description` and `solution` variables?

docker-app/qfieldcloud/core/tests/test_project.py
suricactus: I guess that was the reason why the permissions were using `        project = Project.objects.get(id=projectid)
` instead of the new change `        project = get_object_or_404(Project, id=projectid)`?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1218,6 +1239,58 @@ class Meta:
         default=False,
     )
 
+    def get_localized_datasets_project(self) -> Optional[Project]:
```
suricactus, line 1242:
```suggestion
    def get_localized_datasets_project(self) -> Project | None:
```
Unless Python version restricts us, as in `(lib)qfieldsync`, we always prefer to use the more modern syntax for type definition.

docker-app/qfieldcloud/core/models.py
tdrivas: Do you mean only in the following case  ??


```
 for layer_data in self.project_details.get("layers_by_id", {}).values():
                layer_name = layer_data.get("name")
                if layer_data.get("error_code") != "no_error":
```

docker-app/qfieldcloud/core/models.py
tdrivas: Changed!

docker-app/qfieldcloud/core/models.py
tdrivas: No need for this!

docker-app/qfieldcloud/core/models.py
tdrivas: Changed!

docker-app/qfieldcloud/core/models.py
tdrivas: Changed!

docker-app/qfieldcloud/core/models.py
tdrivas: Updated!

docker-app/qfieldcloud/core/tests/test_project.py
tdrivas: Yes, as `get_object_or_404` raises a Http404 immediately and thus skips DRF’s exception handling stack . I will get back to `Project.objects.get(id=projectid)` with a custom exception

docker-qgis/qfc_worker/utils.py
tdrivas: Clickup task has been created and linked to the TODO comment.

docker-app/qfieldcloud/core/tests/test_project.py
tdrivas: Back to the `Project.objects.get(id=projectid)` and `self.assertEqual(response.json()["code"], "object_not_found")`

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -300,8 +300,6 @@ def test_error_responses(self):
 
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
         self.assertEqual(response.json()["code"], "object_not_found")
-
-        # Get a project without a valid project id
```
suricactus, line 304: Why remove the comment?

suricactus (approved):
Looks much better now, I only one last comment.

docker-app/qfieldcloud/core/models.py
suricactus: @tdrivas can we close this comment?

docker-app/qfieldcloud/core/serializers.py
suricactus: @tdrivas can we close this comment?

docker-app/qfieldcloud/core/tests/test_project.py
tdrivas: Accidentally due to playing with the command above. Sorry for this. Restored.

suricactus: Please try to catch these by reviewing yourself. :)

docker-app/qfieldcloud/core/models.py
tdrivas: Sure, last suggestion has been applied!

tdrivas:
> Looks much better now, I only one last comment.

Thank you very much, @suricactus (and @nirvn) for the feedback and all the support throughout this process!  Comment has been addressed! 

Would it be okay to proceed with merging now?

suricactus:
We can proceed with merging, but we would need a follow-up PR that adds tests. Since this is a critical development, shipping without automated tests would be dangerous. Ideally the tests need to be part of the current PR.

I see a few interesting test cases;
- the `localized_datasets_project_id` is correctly set (or unset)
- the `localized_layers` properties returns properly
- the `missing_localized_layers`
- the `get_missing_localized_layers` returns properly

Also I have just checked the first message has outdated information.

tdrivas:
> We can proceed with merging, but we would need a follow-up PR that adds tests. Since this is a critical development, shipping without automated tests would be dangerous. Ideally the tests need to be part of the current PR.
> 
> I see a few interesting test cases;
> 
> * the `localized_datasets_project_id` is correctly set (or unset)
> * the `localized_layers` properties returns properly
> * the `missing_localized_layers`
> * the `get_missing_localized_layers` returns properly
> 
> Also I have just checked the first message has outdated information.

As I find it important to be on the current PR, additional tests have been introduced with the latest commit.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -300,7 +300,6 @@ def test_error_responses(self):
 
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
         self.assertEqual(response.json()["code"], "object_not_found")
-
```
suricactus, line 303: May we remove this empty line deletion from the diff?

suricactus (commented):
@tdrivas  I am afraid the tests are not pushed yet?

tdrivas:
> @tdrivas I am afraid the tests are not pushed yet?

just pushed :)

docker-app/qfieldcloud/core/tests/test_project.py
tdrivas: It is breathing now :)

suricactus (changes requested):
Just checked the tests. We usually do a full integration tests without faking the data, so please prepare the respective qgs project and files, upload them via API and test if everything works fine instead.

I guess adding a new file `test_localized_projects.py` is probably a good idea.

suricactus (approved):
So far so good! Thanks!

PR #1149, opened 2025-03-11
Use upstream `constance` database backend, since it no longer resets values

suricactus:
There was an issue a few versions ago, but now with constance 4.0.0 it was fixed.

The issue was in: https://github.com/jazzband/django-constance/issues/348
Fixed in: https://github.com/jazzband/django-constance/pull/538


gounux (commented):
Code looks good, thanks. What about removing the `qfieldcloud.core.constance_backends.py` file in this PR as well, since it does not seem to be needed anymore ?

PR #1151, opened 2025-03-12
Make `Project.status` a cached property to drastically reduce the number of DB queries

suricactus:
From 324 to 196 queries on the projects home screen.


duke-nyuki:
Task linked: [QF-5404 Cache the `Project.status` so there is no need for constant bombardment of the DB to calculate it](https://app.clickup.com/t/2192114/QF-5404)

gounux (commented):
CLG ;)

The `wait_for_project_ok_status` test seems to fail in the CI, would that be related ? :thinking:

PR #1155, opened 2025-03-12
Add the possibility to migrate file storage by exact username and support version ids that are not UUIDs

suricactus:
Basically the title, it seems besides minio version id that is a UUID , there are other providers that use other strings to represent version ids.


duke-nyuki:

PR #1156, opened 2025-03-12
For historical reasons we named the `etag` as `md5sum` in the API

suricactus:
If we don't fix that, all the clients will assume that files has changed their checksum, while they did not.


duke-nyuki:

PR #1157, opened 2025-03-13
Add more logs when deleting files

suricactus:


duke-nyuki:

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -469,12 +469,16 @@ def purge_old_file_versions_legacy(
...
         # Debug print
-        logger.debug(
+        logger.info(
             f'Purging {len(old_versions_to_purge)} out of {len(file.versions)} old versions for "{file.latest.name}"...'
         )
 
         # Remove the N oldest
         for old_version in old_versions_to_purge:
+            logger.info(
+                f'Purging {old_version.key=} {old_version.id=} as old version for "{file.latest.name}"...'
```
gounux, line 479:
```suggestion
                f'Purging {old_version.key} {old_version.id} as old version for "{file.latest.name}"...'
```

suricactus: Why do you prefer without the `=`? I always find it nice for these super technical logs that having extra context never hurts.

gounux: Alright, fair enough !

gounux (approved):

PR #1158, opened 2025-03-13
Move the default project setting `STORAGES_PROJECT_DEFAULT_STORAGE` to be available also in non-test environment

suricactus:
This way one can easily fallback to legacy storage if something goes wrong.


duke-nyuki:

.env.example
```diff
@@ -71,6 +71,11 @@ STORAGES='{
         }
     }'
 
+# Setting what is the default storage. If empty, it will use the `default` storage.
+# NOTE: The value must be a key of the `STORAGES` setting.
+# DEFAULT: ""
+# STORAGES_PROJECT_DEFAULT_STORAGE=
```
gounux, line 77: Why leave it commented ?

suricactus: It should not be used by default. We might even delete it in the future.

gounux (approved):

PR #1159, opened 2025-03-13
Add basic support for SSO using OpenID Connect

lukasgraf:
This adds basic support for SSO with OAuth2 / [OpenID Connect](https://en.wikipedia.org/wiki/OpenID#OpenID_Connect_(OIDC)) to QFC, using `django-allauth`'s socialaccount features.

TODO:
- [ ] Customize automatic username generation
- QGIS Authentication Middleware _(in a follow up PR)_
- API Endpoint for listing enabled providers _(in a follow up PR)_

---

With this, we set up the basic building blocks to support authentication using `django-allauth`'s socialaccount providers. No actual provider is configured to be used yet - so this PR should be safe to deploy, and not affect the default behavior yet. Everything will be in place though so that a specific provider can simply be added to `SOCIALACCOUNT_PROVIDERS` in `settings.py`.

Changes include:
- Install the `django-allauth[socialaccount]` extra to pull in necessary dependencies
- Prepare Google and Microsoft as potential providers (not configured yet though)
- Customize the admin login page so that social logins can be chosen, if they are configured. This is currently hidden behind an `sso=1` query string parameter while the feature is still in development.
- Show the `SocialAccount` related models in the admin panel
- Improve `django-allauth`'s error logging
- Configure various `SOCIALACCOUNT_` settings to get the login, signup and account linking flow to work the way we want

### Testing locally

To test these changes locally, add a `SOCIALACCOUNT_PROVIDERS` definition to your `settings.py`:

```python
SOCIALACCOUNT_PROVIDERS = {
    "google": {
        "APP": {
            "client_id": "<your-client-id>",
            "secret": "<your-client-secret>",
            "key": "",
        }
    }
}
```
Then visit the login page, and append `&sso=1` to the URL:
sso_login


docker-app/qfieldcloud/core/adapters.py
```diff
@@ -69,3 +76,30 @@ def send_confirmation_mail(
...
+    """
+
+    def on_authentication_error(
+        self,
+        request,
+        provider,
+        error=None,
+        exception=None,
+        extra_context=None,
+    ):
```
suricactus, line 94: Adding a bit of typing is highly appreciated :)

lukasgraf:
@suricactus do you want to have a final look over it, or may I 🚢 this?

docker-app/qfieldcloud/core/tests/test_username_generation.py
```diff
@@ -0,0 +1,47 @@
...
+            username="existing9928", password="abc123"
+        )
+
+    def test_generated_usernames(self):
+        random.seed(42)
+        expectations = [
+            # Bases username on localpart of email plus random 4-digit suffix
+            ("john@example.org", "john2824"),
+            # Collisions with existing usernames are avoided
+            ("existing@example.org", "existing99286"),
```
suricactus, line 27: Would you mind splitting this test into a separate test case? Now it's a bit hidden behind. Also you can kill the `existing_user` from the `setUp` and use it only in the separate test case.

suricactus, line 27: Would you mind splitting this test into a separate test case? Now it's a bit hidden behind. Also you can kill the `existing_user` from the `setUp` and use it only in the separate test case.

docker-app/qfieldcloud/core/templates/admin/login.html
```diff
@@ -0,0 +1,8 @@
+{% extends "admin/login.html" %}
+
+{% block content %}
+    {{ block.super }}
+    {% if request.GET.sso %}
```
suricactus, line 5:
```suggestion
    {% comment %}TODO add a configuration based mechanism to render the social login instead of depending on query param {% endcomment %}
    {% if request.GET.sso %}
```
I guess we want to add such TODO here?

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -69,3 +116,33 @@ def send_confirmation_mail(
...
+        logger.error(f"Provider: {provider!r}")
+        logger.error(f"Error: {error!r}")
+
+        if not extra_context:
+            extra_context = {}
+
+        # Make stack strace available in template context.
+        extra_context["formatted_exception"] = "\n".join(
+            traceback.format_exception(exception)
+        )
```
suricactus, line 145: The only thing I am missing and probably it's worth adding a comment is which "template context" we are talking about here? I am sure if I deep dive in the docs I can find it, just acting lazy. If screenshot is faster in explaining this, feel free.

suricactus (approved):
One of the best PRs I have ever had the pleasure to review and approve for merging.

I want to highly admire the clean code approach with the right amount of comments, detailed PR and commit messages, and not the least having the right tests.

I have two small comments and a question. It can also be merged at this stage.

docker-app/qfieldcloud/core/tests/test_username_generation.py
lukasgraf: Good point, this final collision avoidance is not actually performed by our customization, but the d.allauth superclass, so it deserves its own test case. Will do 👍

docker-app/qfieldcloud/core/adapters.py
lukasgraf: This just means that it would be available as one of the template variables in the top level namespace. So in the `socialaccount/authentication_error.htm` template, this could then be used as `{{ formatted_exception }}`.

_(The superclass implementation of `on_authentication_error()`, called right below, will stuff everything from `extra_context` into the `context` that is passed to the template)._

I have added a comment in the code to explain this a bit more. 👍

PR #1160, opened 2025-03-13
Add an assert when resending confirmation email

gounux:
This PR makes sure there is a `account_verified_email` key in the resend confirmation email view.


docker-app/qfieldcloud/core/views/accounts_views.py
```diff
@@ -35,6 +35,7 @@ def resend_confirmation_email(request: HttpRequest) -> HttpResponse:
     if request.method != "POST":
         return redirect_to_referer_or_view(request, "account_login")
 
+    assert "account_verified_email" in request.session
     email_address = request.session.get("account_verified_email")
```
suricactus, line 39:
```suggestion
    assert "account_verified_email" in request.session

    email_address = request.session["account_verified_email"]
```

If you assert it, no need to break the yping then.

PR #1161, opened 2025-03-13
Fix backwards compatibility with QField/QFieldSync for avatars

suricactus:
They expect that the avatar URL returned via API is always ending with a file extension, e.g. `.png`, `.svg` etc.

Now we do match this expectation by adding a new URL path:
```/files/avatars/<username>/avatar.<extension>/```

The `avatar.<extension>` does nothing, it is used just for the
backwards compatibility.

This is only used in the serializers.


duke-nyuki:

PR #1162, opened 2025-03-16
Remove trailing slash for avatars for backwards compatibility for QFieldSync

suricactus:
Fix the fix in #1161 .

The error we target with this PR is:
image


duke-nyuki:
Task linked: [QF-5307 The user's avatar is shown in QFieldSync](https://app.clickup.com/t/2192114/QF-5307)

PR #1173, opened 2025-03-22
Add STORAGE_ACCESS_KEY_ID and STORAGE_SECRET_ACCESS_KEY to .env.example

stcz:
When upgrading my QFieldCloud instance I noticed, that `STORAGE_ACCESS_KEY_ID` and `STORAGE_SECRET_ACCESS_KEY` has been removed from `.env.example`. As this is fine for development it is required for the standalone config and should also stay in the config to be able to change the credentials.

It is not perfect, that the credentials are now redundant and also used in `STORAGES` but I don't see a solution for that. 

Thanks for your great work :)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,39 +1,7 @@
-####################
-# Debug settings
-####################
-
 # Whether the environment is in debug mode (e.g. Django debug)
 # VALUES: 0 - non debug mode; 1 - debug mode
 # DEFAULT: 1
 DEBUG=1
-
-# Debugpy port used for the `app` service
-# DEFAULT: 5678
-DEBUG_DEBUGPY_APP_PORT=5678
-
-# Debugpy port used for the `worker_wrapper` service
-# DEFAULT: 5679
-DEBUG_DEBUGPY_WORKER_WRAPPER_PORT=5679
-
-# The Django development port. Not used in production.
-# DEFAULT: 8011
-DJANGO_DEV_PORT=8011
-
-####################
-# Deployment settings
-####################
-
-# Prefix used by docker compose for each of the containers, e.g. app will be `qfieldcloud_app`
-# DEFAULT: qfieldcloud
-COMPOSE_PROJECT_NAME=qfieldcloud
-
-# List of docker compose files
-# DEFAULT: docker-compose.yml:docker-compose.override.local.yml
-COMPOSE_FILE=docker-compose.yml:docker-compose.override.local.yml:docker-compose.override.standalone.yml
-
-# Separator in `COMPOSE_FILE` between filenames. Required for making COMPOSE_FILE above cross-platform (do not change)
-# DEFAULT: :
-COMPOSE_PATH_SEPARATOR=:
 
 # Environment where QFieldCloud runs
 # VALUES: `development` - local development; `staging` - staging server; `test` - tests; `production` - production server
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -76,6 +77,10 @@ STORAGES='{
 # DEFAULT: ""
 # STORAGES_PROJECT_DEFAULT_STORAGE=
 
+# Configuration for minio access. Needs to be the same as in the STORAGES setting in standalone config.
+STORAGE_ACCESS_KEY_ID=minioadmin
+STORAGE_SECRET_ACCESS_KEY=minioadmin
```
gounux, line 82:
```suggestion
# Local configuration for minio storage and standalone instances. Uncomment if necessary.
# Needs to be the same as in the STORAGES setting in standalone config.
# STORAGE_ACCESS_KEY_ID=minioadmin
# STORAGE_SECRET_ACCESS_KEY=minioadmin
```

gounux (commented):
Thanks @stcz, I'm wondering if it is absolutely necessary to have it in the `.env.example` file, maybe have it commented and give the possibility to uncomment in the active `.env` file ?

stcz:
Yes, it is not absolutely necessary, i think as a comment is a good solution

.env.example
```diff
@@ -76,6 +77,11 @@ STORAGES='{
 # DEFAULT: ""
 # STORAGES_PROJECT_DEFAULT_STORAGE=
 
+# Local configuration for minio storage and standalone instances. Uncomment if necessary.
+# Needs to be the same as in the STORAGES setting in standalone config.
+# STORAGE_ACCESS_KEY_ID=minioadmin
```
suricactus, line 82: I would name the variable `MINIO_ROOT_USER`, as it is used in the `minio` service. Also, would you mind if it follows the pattern for all newly added env variables:

```
# <description>
# DEFAULT: <value>
# <ENVVAR>=<value>
```

stcz: Done :)

.env.example
```diff
@@ -76,6 +77,12 @@ STORAGES='{
 # DEFAULT: ""
 # STORAGES_PROJECT_DEFAULT_STORAGE=
 
+# Local admin credentials configuration for minio storage and standalone instances. Uncomment if necessary.
+# NOTE: Needs to be the same as in the `STORAGES` setting in standalone config.
+# DEFAULT: MINIO_ROOT_USER=minioadmin, MINIO_ROOT_PASSWORD=minioadmin
+# MINIO_ROOT_USER=minioadmin
+# MINIO_ROOT_PASSWORD=minioadmin
```
suricactus, line 84: Would you mind ungrouping the envvars, as we do with any other? I can see the logic why you propose them merged, but I would like to keep one style. And slowly add this documentation for all the other envvars :)

.env.example
```diff
@@ -38,6 +38,7 @@ LETSENCRYPT_RSA_KEY_SIZE=4096
 # Set to 1 if you're testing your setup to avoid hitting request limits
 LETSENCRYPT_STAGING=1
 
+# Storage
```
suricactus, line 41:
```suggestion
####################
# Storage settings
####################
```

Thanks! I wanted to add this for long time.

And in the end of the section we can do:

```
####################
# /Storage settings
####################
```

.env.example
stcz: Done

.env.example
stcz: As a proposal, I added a little cleaned up structure for the .env.example: 75cd17343bb5dd72253941f36ee4aa6705f14a07

@suricactus If you whish we could work together on documenting this. Either in this PR or i create a new one and revert here to stay more structured.

suricactus: If you don't mind, let's keep this PR small and focused and make a follow-up PR with the documentation of the changes. I will not be reachable next few weeks, so I will leave you and @gounux to document them :)

stcz: Perfect. I will create a new PR as soon this one is merged, to avoid conflicts.

stcz:
Yes I like a nice git history. 

But can't you do this while merging? https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

I'm mainly working on gitlab so I'm not really experienced on github. 

If the only way is, that i'm doing it manually, i can do this :)

PR #1174, opened 2025-03-22
Update minio createbuckets

stcz:
When upgrading my QFieldCloud instance, I noticed, that minio wasn't updated in the whole project. 

It is not an issue, but it is better to use the same version, I think.

But I think it would be great to update the base image.

Here are different things possible:

1. use python as base image
2. use python as base image and use minio library to get rid of the mc subprocesses
3. stay at ubuntu as is

Further, I noticed that there is also a newer version of minio, which I cloud also update. 

Please let me know which of the possibilities you would prefer and I can implement it.

Thanks for your great work :)

### DIFF
--- a/docker-createbuckets/Dockerfile
+++ b/docker-createbuckets/Dockerfile
@@ -1,10 +1,10 @@
 FROM ubuntu:noble
 
 RUN apt-get update \
+    && apt-get upgrade -y \
     && apt-get install -y \
     curl \
-    python3 \
-    && rm -rf /var/lib/apt/lists/*
+    python3
 
 RUN curl https://dl.minio.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2025-02-18T16-25-55Z.fips -o /usr/bin/mc
 RUN chmod +x /usr/bin/mc
--- a/docker-createbuckets/Dockerfile
+++ b/docker-createbuckets/Dockerfile
@@ -1,12 +1,12 @@
-FROM ubuntu:noble-20241015
+FROM ubuntu:latest
 
 RUN apt-get update \
-    && apt-get upgrade -y \
     && apt-get install -y \
     curl \
-    python3
+    python3 \
+    && rm -rf /var/lib/apt/lists/*
 
-RUN curl https://dl.minio.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2024-11-17T19-35-25Z.fips -o /usr/bin/mc
+RUN curl https://dl.minio.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2025-02-18T16-25-55Z.fips -o /usr/bin/mc
 RUN chmod +x /usr/bin/mc
 
 COPY ./createbuckets.py /createbuckets.py
--- a/docker-createbuckets/Dockerfile
+++ b/docker-createbuckets/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:noble
 
 RUN apt-get update \
     && apt-get install -y \
--- a/docker-createbuckets/Dockerfile
+++ b/docker-createbuckets/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update \
     curl \
     python3
 
-RUN curl https://dl.minio.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2025-02-18T16-25-55Z.fips -o /usr/bin/mc
+RUN curl https://dl.min.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2025-03-12T17-29-24Z -o /usr/bin/mc
 RUN chmod +x /usr/bin/mc
 
 COPY ./createbuckets.py /createbuckets.py


suricactus:
Hey @stcz . First, thank you for taking care of people using the "standalone" version for more than testing! Made my day (and night) to see QFieldCloud has a healthy community!

Let try to cover some questions/comments from this PR:

- First, having the same version of `minio` and `mc` is a a good idea. Thanks for spotting!
- Second, if there is a new version, I am ok to update it (in a separate PR), tbh I don't know how it works with their compatibility and if/when some kind of migration is needed. This is the only reason we were using a super ancient version until last month, but if they are compatible, feel free to bump to latest every now and then.
- Regarding `rm -rf /var/lib/apt/lists/*`, I have checked on my Ubuntu 24.04 and it says:
```
$ man apt-get | grep -e '--list-cleanup' -A 3
       --list-cleanup
           This option is on by default; use --no-list-cleanup to turn it off. When it is on, apt-get will automatically manage the contents of /var/lib/apt/lists to ensure that obsolete files are erased. The only reason to turn it off is if you
           frequently change your sources list. Configuration Item: APT::Get::List-Cleanup.
```
So if I understand it right, the added line above is no longer needed. Correct me if I misunderstood something?
- finally, a bit of context how and why we ended up with a python script running subprocesses.  As you may know in previous versions this script was just a `mc` command in `docker-compose.yml`. Since we moved the configuration in `STORAGES` envvar, which has a JSON value, if we want to dynamically create and configure the buckets, we to parse the JSON somehow and loop through all the keys in there. So adding some kind of "real" language would make it easier than pure bash. Since qfc is mostly Python, I decided to use Python to parse the JSON and then run the same `mc` commands, since I was already familiar with them and I could copy/paste and experiment with the result.

With the information above, I want to highlight that for me keeping the `mc` copy/pastable commands is more valuable than using Python. If you can suggest not much longer and still readable bash to achieve the same, I am ok. If not, let's keep the subprocesses for now.

So back to your proposal:

> use python as base image

I think it is an overkill, I just used python since it is "cheap". I could have used Perl too.

> use python as base image and use minio library to get rid of the mc subprocesses

If I got you right, you suggest to rewrite the `mc` commands in Python? I would avoid that as I explained above.

> stay at ubuntu as is

I am good with this.

As for the ubuntu version, I would like to have a specific major version, as I do not want random breakages every April and October. So `ubuntu:noble` should be enough for now, as we would reuse the base image for the `qgis` container.

Thanks again for the proactiveness!

stcz:
Thanks for the compliment :)

I will have a look to minio docker compatibility. But as far I noticed up to now upgrading is no problem, downgrading is. But we should take care of the release notes.

stcz:
For the Dockerfile we should add `rm -rf /var/lib/apt/lists/*` as it is cleaning the apt cache. `--list-cleanup` is only looking for obsolete files. But it is not necessary to store the apt cache in a docker layer, its mostly waste of storage. You can refer to the [docker best pracrices](https://docs.docker.com/build/building/best-practices/).

I made a small experiment with the following Dockerfile with and without the last line with rm:
```
FROM ubuntu:latest
RUN apt-get update \
    && apt-get install -y \
    python3 \
    && rm -rf /var/lib/apt/lists/*
```
The resulting Image size without removing the apt cache is `171339444` (~163MB) and with cleaning with rm it is `125142338` (~119MB), so saving 44MB of storage.

stcz:
About the general approach: I'm fine with your suggestions. I just had different ideas and wanted to have some feedback before starting to work on it. 

About the ubuntu version: `latest` is pointing to the LTS for docker images, so only changes all two years in April. But we could also stick to `noble`.

But I would not recommend pinning the noble version, especially as currently we are running currently `apt-get upgrade` what produces unnecessary load on local instances and the ubuntu mirrors. Just using always the current baseimage is a much better approach to stay up to date and receive security updates.

suricactus:
> For the Dockerfile we should add `rm -rf /var/lib/apt/lists/*` as it is cleaning the apt cache. `--list-cleanup` is only looking for obsolete files. But it is not necessary to store the apt cache in a docker layer, its mostly waste of storage. You can refer to the [docker best pracrices](https://docs.docker.com/build/building/best-practices/).
> 
> The resulting Image size without removing the apt cache is `171339444` (~163MB) and with cleaning with rm it is `125142338` (~119MB), so saving 44MB of storage.

Great, thanks! I should have refreshed my memory since I last looked at this document! I guess it is not a bad idea to add the same on all Dockerfiles.

stcz:
I will create a new PR for applying this to all Dockerfiles, as it is unrelated to the minio update.

suricactus:
> But I would not recommend pinning the noble version, especially as currently we are running currently `apt-get upgrade` what produces unnecessary load on local instances and the ubuntu mirrors. Just using always the current baseimage is a much better approach to stay up to date and receive security updates.

Let's stick for now on `noble`. You were right I was exaggerating with breakage every 6 months, but I have bad experience in the last 3 ubuntu version updates using `latest`. I am usually enoying spirng in April rather than fixing random failures. Until next LTS release, `noble` and `latest` should be the same, aren't they (https://hub.docker.com/_/ubuntu/tags)?

I will be happy to open a PR the very next day after the LTS release though.

suricactus (approved):

stcz:
This PR should be ready

suricactus:
Seems the CI is not happy, ideas?

stcz:
The reason could be, that the tests are not waiting, until the bucket in minio is created, but should be unrelated to the MR. Can you just rerun the Pipeline?

stcz:
The Link was broken as minio doesn't have the same release names for server and client. I pinned now to the current, latest version.

suricactus (approved):
Looks good!

PR #1180, opened 2025-03-23
Move smtp4dev from standalone to local

stcz:
Hi everyone, working on qfieldcloud. 

I want to propose to move the smtp4dev configuration in the overrides from "standalone" to "local".

I my mind, the standalone override is for self-hosted standalone instances, where there is no requirement for smtp4dev.

smtp4dev is required for development, when the local override should be used for. 

I'm unsure if smtp4dev is used in any tests. 

If the change is fine, i would also add some more documentation for self-hosted instances.

### DIFF
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -47,6 +47,22 @@ services:
     ports:
       - "${MEMCACHED_PORT}:11211"
 
+  smtp4dev:
+    image: rnwood/smtp4dev:v3
+    restart: always
+    ports:
+      # Web interface
+      - ${SMTP4DEV_WEB_PORT}:80
+      # SMTP server
+      - ${SMTP4DEV_SMTP_PORT}:25
+      # IMAP
+      - ${SMTP4DEV_IMAP_PORT}:143
+    volumes:
+        - smtp4dev_data:/smtp4dev
+    environment:
+      # Specifies the server hostname. Used in auto-generated TLS certificate if enabled.
+      - ServerOptions__HostName=smtp4dev
+
   qgis:
     volumes:
       # allow local development for `docker-qgis`
@@ -56,3 +72,6 @@ services:
       - ./docker-qgis/libqfieldsync:/libqfieldsync:ro
       # allow local development for `qfieldcloud-sdk-python` if host directory present; requires `PYTHONPATH=/qfieldcloud-sdk-python:${PYTHONPATH}`
       - ./docker-qgis/qfieldcloud-sdk-python:/qfieldcloud-sdk-python:ro
+
+volumes:
+  smtp4dev_data:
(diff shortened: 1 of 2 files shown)


suricactus:
Unfortunately we need to keep it there, as it is hard requirement for QFieldCloud to function properly in tests and for especially for standalone installation with just copy/paste of the .env.example. If there are no functioning SMTP server, reset password, send invitations etc will not work. I would like to keep the a maling server required from code perspetive, as making it option within the code would introduce quite a lot of code burden.

> If the change is fine, i would also add some more documentation for self-hosted instances.

I think proposing some docs is always welcome, just keep it mind that `docker-compose.standalone.yml` is and will stay as an example and quickstart from our point of view. The "Caution" section here will not go away in observable future https://github.com/opengisch/QFieldCloud#launch-a-server-instance .

Not sure if docker compose profiles can resolve the issue? https://docs.docker.com/reference/compose-file/profiles/ .

stcz:
Thanks for the response.

I think it should be enough to move it to local, as it is also available there in the tests from the env.example and everything should still work as there is no change.

For a self-hosted version, it would then be nice, to only remove the `docker-compose.override.local.yml` from `COMPOSE_FILE=docker-compose.yml:docker-compose.override.local.yml:docker-compose.override.standalone.yml`. Currently, I need to override it again to get rid of smtp4dev. 

Profiles might be a solution, but it would introduce more complexity, what is currently not necessary in my mind.

As the Workflow passed the tests should be fine, or are there more hidden tests?

stcz:
Maybe we should also clearify what the override files are for:

### docker-compose.override.local.yml
Everything required for a local development setup

### docker-compose.override.prod.yml
QFieldCloud Production things (not interesting for community?)

### docker-compose.override.staging.yml
??

### docker-compose.override.standalone.yml
Everything required for a self-hosted qfieldcloud instance

### docker-compose.override.test.yml
Everything required for Workflows and tests

Sure, the "Caution" warning should not be removed, but i'm willing to share my knowledge about self-hosting and improve it as far as possible :)

suricactus:
> I think it should be enough to move it to local, as it is also available there in the tests from the env.example and everything should still work as there is no change.

Would you mind testing if `admin/invitations/invitation/add/` works fine for you with a deployment where there is no `local` and no intervention of `.env.example` used as `.env`?

suricactus:
@stcz did I managed to convince you of the purpose of `standalone`? I think it is still a good idea for your case to have a minimal wrapper around our docker-compose files. Here you can see ways to easily disabled services as you prefer: https://stackoverflow.com/questions/47718055/remove-service-on-docker-compose-through-overriding

I would suggest we close this PR.

stcz:
Yes, that's the way I'm doing it currently. I close the PR and will put what I have learned in a new PR maybe to the README or a new file, as I'm willing to share my self-hosting experience.

PR #1181, opened 2025-03-23
Apply docker best practices to Dockerfiles

stcz:
Starting form #1174 here a PR to apply the docker best practices for all docker files.

The following changes are included:
- Remove apt cache from layers
- Remove pip cache from layers


docker-app/Dockerfile
```diff
@@ -12,14 +12,20 @@ ARG PIP_DISABLE_PIP_VERSION_CHECK=1
...
 RUN pip install pip-tools
 
 # install pip dependencies
 COPY ./requirements/requirements.txt /requirements/requirements.txt
-RUN pip3 install -r requirements/requirements.txt
+RUN pip3 install -r requirements/requirements.txt \
+    && rm requirements/requirements.txt
 
 # save 60 MB of data... Safe to delete according to botocore contributor https://github.com/boto/botocore/issues/1629#issuecomment-451309885
 RUN ls -Q /usr/local/lib/python3.10/site-packages/botocore/data | grep -xv "endpoints.json" | xargs rm -rf
```
stcz, line 31: Where do we want to save the storage? As this is not done within the same layer, the data is still on local systems. This would only matter for images which are not used as far as I know.

docker-app/Dockerfile
```diff
@@ -3,7 +3,7 @@
 ##########################
 
 # pull a builder image, the same as the base
-FROM python:3.10.8-slim-buster AS build
+FROM python:3.10-slim AS build
```
suricactus, line 6:
```suggestion
FROM python:3.10-8-slim-buster AS build
```

Let's keep in this PR without such change and have a follow-up where we pin to `3.10-slim-bookworm`. I would like to stick to specific Debian version for the same reasons I have described in one of the other PRs.

Btw, we managed to update almost all dependencies and soon we will be able to move to python 3.12+.

docker-app/Dockerfile
suricactus: Github CI.

docker-app/Dockerfile
suricactus: Look at the CI:
```
 > [qgis  7/15] RUN if [ "1" = "1" ]; then apt install -y gdbserver valgrind; fi:
0.051 
0.054 Reading package lists...
0.067 Building dependency tree...
0.068 Reading state information...
0.070 Package valgrind is not available, but is referred to by another package.
0.070 This may mean that the package is missing, has been obsoleted, or
0.070 is only available from another source
0.070 
0.070 E: Unable to locate package gdbserver
0.070 E: Package 'valgrind' has no installation candidate
```

Exactly why I don't want to use `latest` tags :)

stcz: OK, i will just remove this from the PR

stcz: The valgrind was a mistake of mine and should be fixed now.

stcz: If everything passes, the PR should be ready now.

docker-app/Dockerfile
```diff
@@ -12,14 +12,20 @@ ARG PIP_DISABLE_PIP_VERSION_CHECK=1
 ARG PIP_NO_CACHE_DIR=1
 
 # install psycopg2 requirements
-RUN apt-get update && apt-get install -y libpq-dev python3-dev gcc
+RUN apt update \
```
suricactus, line 15: Why apt over apt-get? Looking at best practices article, they always use `apt-get`. https://docs.docker.com/build/building/best-practices/

Also, dockerfile linter (which was used years ago to cleanup the dockefiles and we can reintroduce as part of the CI) has a warning about it: https://github.com/hadolint/hadolint/wiki/DL3027

The warning actually matches my understading about apt vs apt-get.

docker-app/Dockerfile
```diff
@@ -12,14 +12,20 @@ ARG PIP_DISABLE_PIP_VERSION_CHECK=1
...
+    && apt install -y \
+    libpq-dev \
+    python3-dev \
+    gcc \
+    && rm -rf /var/lib/apt/lists/*
 
 # install `pip-compile` (as part of `pip-tools`)
 RUN pip install pip-tools
```
suricactus, line 23: Shall we make this one also pip3, I think it's the only place where we use pip over pip3?

docker-app/Dockerfile
```diff
@@ -85,7 +93,8 @@ ENTRYPOINT ["/entrypoint.sh"]
 FROM base AS webserver_runtime
 ENV LOGGER_SOURCE=app
 COPY ./requirements/requirements_runtime.txt requirements/requirements_runtime.txt
-RUN pip3 install -r requirements/requirements_runtime.txt
+RUN pip3 install -r requirements/requirements_runtime.txt \
+    && rm requirements/requirements_runtime.txt
```
suricactus, line 97: What do we win by removing the requirements.txt file?

docker-app/Dockerfile
stcz: You are completely right. apt is newer (what was the reason for me to use it) but apt-get is recommended.

docker-app/Dockerfile
stcz: A few bytes of free storage. :)

suricactus: Please sacrifice a few bytes for now :) there are actually a bunch of other files which can safely be removed too, but please don't. Open to revisit this in the future.

stcz: Done, i also squashed all commits

stcz:
The Error in the tests is from 5cf0f71798cf03427f345aca01459ab4c2e36172

PR #1182, opened 2025-03-23
Docker: Add memcached restart: unless-stopped policy

faebebin:
@suricactus seems when rebooting the server i.e. restarting docker daemon the memcached does not come up automatically on our infra and then the worker_wrapper hangs in restart loop

! This seems to help in our case at least but I let it up to you to check necessity for QFC classic

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -187,6 +187,7 @@ services:
 
   memcached:
     image: memcached:1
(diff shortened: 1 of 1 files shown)


suricactus (approved):
Looks good and does not hurt.

faebebin:
👍

PR #1183, opened 2025-03-24
Fix downloading package files from projects with multiple packages

suricactus:
Basiclly we need to filter not only by file type, name and version id, but when the file_type is PACKAGE, we also need to filter by package_job_id .


duke-nyuki:
Task linked: [QF-5531 Fix downloading package files from projects with multiple packages](https://app.clickup.com/t/2192114/QF-5531)

gounux (approved):
Code looks good, thanks!

PR #1186, opened 2025-03-25
Add QGIS authentication middleware

lukasgraf:
This middleware allows to authenticate to QFC using **QGIS auth configs** with an **OpenID Connect** Authorization Code flow.

---

The reason this middleware is needed is as follows:
QGIS already has support for performing an OAuth2 Authorization Code flow using its authentication configs.

However, the way that QGIS passes the result of that flow (an access_token) to QFC is very limited, and cannot really be controlled. QGIS just runs the flow for the user, and then attaches the `access_token` it gets to any subsequent requests in the `Authorization` header.

However, `django-allauth`, for a regular web based OIDC flow, expects the final request of the flow to hit e.g. `/accounts/google/login/callback`, then will log the user in and create a session.

Because QGIS doesn't hit *any* particular URL first, like a callback URL, but instead just attaches the token to every request, we need a middleware that looks for those tokens in any arbitrary requests, and authenticates them if necessary.

Once the authentication is performed (preferably using an OIDC ID token, or as a a fallback an OAuth2 access token), a session is created, and on subsequent requests the user is already authenticated (since QGIS sends back the session cookies it gets).

---

Relevant HTTP request headers:

| HTTP Header Name             | Example Value      | Meaning                                                                                                          |
| ---------------------------- | ------------------ | ---------------------------------------------------------------------------------------------------------------- |
| **X-QFC-IDP-ID**             | `google`           | A django-allauth socialaccount provider ID that uniquely identifies the provider that the token is intended for. |
| **X-QFC-ID-Token**           | `eyJhbOrtYGciO...` | OIDC ID token                                                                                                    |
| **Authorization**            | `Bearer eZO5el...` | OAuth2 access token                                                                                              |

A QFC client that uses QGIS to authenticate (QField, QFieldSync, ...) must set `X-QFC-IDP-ID` and at least one of the two token headers.

In addition, if session cookies are used, it will be necessary to set the Django `X-CSRFToken` header and the `Referer` header (see [Django CSRF docs](https://docs.djangoproject.com/en/5.1/howto/csrf/) for details).

### DIFF
--- a/docker-app/qfieldcloud/core/middleware/qgis_auth.py
+++ b/docker-app/qfieldcloud/core/middleware/qgis_auth.py
@@ -1,0 +1,130 @@
+import logging
+
+from allauth.socialaccount.adapter import get_adapter
+from allauth.socialaccount.helpers import complete_social_login
+from allauth.socialaccount.models import SocialToken
+from django.http import HttpRequest
+
+logger = logging.getLogger(__name__)
+
+
+class QGISAuthenticationMiddleware:
+    """Authentication middleware for OIDC authentication using QGIS.
+
+    This middleware allows to authenticate a user based on an OIDC ID token
+    that is passed in a custom HTTP header.
+
+    If a QGIS auth config is set up to pass the `id_token` in that header, and
+    has the proper OpenID Connect scopes configured, this allows to authenticate
+    to QFC using an OIDC flow performed in QGIS.
+
+    Alternatively, an OAuth2 access token (which QGIS passes in the
+    `Authorization` header by default) can be used.
+
+    This middleware MUST be listed after these:
+
+    - django.contrib.auth.middleware.AuthenticationMiddleware
+    - allauth.account.middleware.AccountMiddleware
+    - django.contrib.messages.middleware.MessageMiddleware
+
+    The reason why this middleware is required is this:
+
+    Normally, an OIDC flow using django-allauth's social account functionality
+    returns the user to a well defined callback URL, for example
+    /accounts/google/login/callback/. This URL is then handled by the callback
+    view that extracts the tokens from the request and completes the login.
+
+    QGIS however doesn't do this. It makes no distinction between an initial
+    "login request" and any subsequent requests. It just performs the OIDC flow
+    with the IDP, gets the access token / ID token, and attaches those to any
+    requests that use this auth config in the HTTP headers.
+
+    This middleware therefore looks for those tokens in those headers, extracts
+    them, and then delegates the actual authentication to django-allauth. Once
+    that succeeds, the user is logged in, a session is created, and the session
+    is persisted using the standard mechanisms (e.g. a session cookie).
+
+    On subsequent requests QGIS sends back those session cookies, and
+    authentication does not need to be performed again.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest):
+        """Detect and extract HTTP headers intended for SSO auth.
+
+        Because this code gets called for *every* single request, we need to
+        be quite defensive here. For reasons of performance and robustness.
+
+        We need to abort as early as possible if we can determine that the
+        request is not intended for SSO auth. And we must not cause any
+        unhandled exceptions that could interfere with normal requests.
+        """
+        if request.user.is_authenticated:
+            # User already has a session
+            return self.get_response(request)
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/middleware/qgis_auth.py
```diff
@@ -0,0 +1,129 @@
...
+    to QFC using an OIDC flow performed in QGIS.
+
+    Alternatively, an OAuth2 access token (which QGIS passes in the
+    `Authorization` header by default) can be used.
+
+    This middleware MUST be listed after these:
+
+    - django.contrib.auth.middleware.AuthenticationMiddleware
+    - allauth.account.middleware.AccountMiddleware
+    - django.contrib.messages.middleware.MessageMiddleware
```
gounux, line 27: Naive question: why must that be listed _after_ ?

docker-app/qfieldcloud/core/middleware/qgis_auth.py
```diff
@@ -0,0 +1,129 @@
...
+    def get_access_token(self, request):
+        auth_header = request.headers.get("Authorization")
+
+        if not (auth_header and auth_header.startswith("Bearer ")):
+            return
+
+        access_token = auth_header.split(" ")[1]
+        return access_token
+
+    def get_provider(self, request, idp_id):
```
gounux, line 122: Would it be possible to add typings ?

docker-app/qfieldcloud/core/middleware/qgis_auth.py
```diff
@@ -0,0 +1,129 @@
...
+        return self.get_response(request)
+
+    def get_id_token(self, request):
+        return request.headers.get("X-QFC-ID-Token")
+
+    def get_access_token(self, request):
+        auth_header = request.headers.get("Authorization")
+
+        if not (auth_header and auth_header.startswith("Bearer ")):
+            return
```
gounux, line 117:
```suggestion
            return None
```

Or maybe it's implicit ?

docker-app/qfieldcloud/core/middleware/qgis_auth.py
```diff
@@ -0,0 +1,129 @@
...
+        """Detect and extract HTTP headers intended for SSO auth.
+
+        Because this code gets called for *every* single request, we need to
+        be quite defensive here. For reasons of performance and robustness.
+
+        We need to abort as early as possible if we can determine that the
+        request is not intended for SSO auth. And we must not cause any
+        unhandled exceptions that could interfere with normal requests.
+        """
+        if request.user.is_authenticated:
```
gounux, line 63: Would that be the case for a QF/QFS client ?

gounux (commented):
Here is a review made of few questions, to get a better understanding of the great work. Thanks @lukasgraf, the comments / commit & PR descriptions are very insightful and highly appreciated!

docker-app/qfieldcloud/core/middleware/qgis_auth.py
lukasgraf: Because otherwise the user won't be authenticated yet when the request reaches our middleware and QGIS sends a valid session cookie.

Once the middleware has authenticated the user, the normal Django login process results in a session cookie being set. QGIS stores that session cookie, and attaches it to every future request. So the middleware only needs to perform authentication once, and all future requests are automatically authenticated via the session. But for this to work, the middleware that is responsible for authenticating a user via their session needs to come first. So that when the request hits our middleware, `request.user.is_authenticated` is true and we can abort early.

The `MessageMiddleware` is just something that is internally relied on in some of the `django-allauth` code, and therefore needs to process the request first, because otherwise things don't work.

docker-app/qfieldcloud/core/middleware/qgis_auth.py
lukasgraf: Yes, see the comment about the order of the middleware.

The `QGISAuthenticationMiddleware` will only perform authentication of the user the very first time. As a result of that authentication and logging in the user, the user gets a session (and the response contains session cookies). QGIS sends those session cookies with all future requests, and then the user is automatically authenticated using that session - so there is no need anymore to perform SSO auth.

lukasgraf:
@gounux thanks for the review! I have addressed all your points 😃

gounux (approved):
Thanks @lukasgraf !

PR #1192, opened 2025-03-28
Add API endpoint to list enabled auth providers

lukasgraf:
This introduces a new API endpoint `api/v1/auth/providers` that allows to **enumerate all the supported and enabled authentication providers** in QFC.

Implements [QF-5435](https://app.clickup.com/t/2192114/QF-5435)

This can be used by QFC clients (apps like QField or the QFieldSync plugin) to present the user with the supported options to log in. The information returned from this endpoint contains all the necessary details for a QFC client to build an appropriate **QGIS auth config** to log in using the selected provider.

The endpoint is obviously accessible without authentication, since it provides information about the authentication methods.

The `styles` property of each provider contains the necessary **style information** for QFC clients like QField or the QFieldSync plugin to dynamically get all the information they need to render the login buttons for the enabled providers in
a consistent way. 

A response from the endpoint looks like this:

```json
[
    {
        "type": "credentials",
        "id": "credentials",
        "name": "Username / Password"
    },
    {
        "type": "oauth2",
        "id": "google",
        "name": "Google",
        "grant_flow": 0,
        "scope": "openid profile email",
        "pkce_enabled": false,
        "token_url": "https://oauth2.googleapis.com/token",
        "refresh_token_url": "https://oauth2.googleapis.com/token",
        "request_url": "https://accounts.google.com/o/oauth2/v2/auth",
        "redirect_host": "localhost",
        "redirect_port": 7070,
        "redirect_url": "",
        "client_id": "736186946600-94a7l05bb9lg5rv0ganupr2oai2otubr.apps.googleusercontent.com",
        "extra_tokens": {
            "id_token": "X-QFC-ID-Token"
        },
        "styles": {
            "required": true,
            "light": {
                "logo": "https://localhost/staticfiles/sso/google.svg",
                "color_fill": "#FFFFFF",
                "color_stroke": "#747775",
                "color_text": "#1F1F1F"
            },
            "dark": {
                "logo": "https://localhost/staticfiles/sso/google.svg",
                "color_fill": "#131314",
                "color_stroke": "#8E918F",
                "color_text": "#E3E3E3"
            }
        }
    },
    {
        "type": "oauth2",
        "id": "keycloak",
        "name": "Keycloak",
        "grant_flow": 3,
        "scope": "openid profile email",
        "pkce_enabled": true,
        "token_url": "https://keycloak:7443/realms/ninjas/protocol/openid-connect/token",
        "refresh_token_url": "https://keycloak:7443/realms/ninjas/protocol/openid-connect/token",
        "request_url": "https://keycloak:7443/realms/ninjas/protocol/openid-connect/auth",
        "redirect_host": "localhost",
        "redirect_port": 7070,
        "redirect_url": "",
        "client_id": "qfc-pkce",
        "extra_tokens": {
            "id_token": "X-QFC-ID-Token"
        },
        "styles": {
            "required": false,
            "light": {
                "logo": "https://localhost/staticfiles/sso/keycloak.svg",
                "color_fill": "#FFFFFF",
                "color_stroke": "#747775",
                "color_text": "#1F1F1F"
            },
            "dark": {
                "logo": "https://localhost/staticfiles/sso/keycloak.svg",
                "color_fill": "#131314",
                "color_stroke": "#8E918F",
                "color_text": "#E3E3E3"
            }
        }
    }
]
```


docker-app/qfieldcloud/authentication/views.py
```diff
@@ -119,3 +125,90 @@ def get_queryset(self):
...
+
+class ListProvidersView(APIView):
+    """Lists the available authentication providers.
+
+    This will mostly be allauth SocialAccount providers, plus the
+    username/password login.
+    """
+
+    FLOW_AUTHORIZATION_CODE = 0
+    FLOW_AUTHORIZATION_CODE_PKCE = 3
```
nirvn, line 138: Maybe add a comment mentioning that these values are tied to this enum: We should mention somewhere here that this value reflects the following enum: https://github.com/qgis/QGIS/blob/master/src/auth/oauth2/core/qgsauthoauth2config.h#L45

docker-app/qfieldcloud/settings.py
```diff
@@ -428,6 +430,58 @@ def before_send(event, hint):
...
+            "color_text": "#1F1F1F",
+        },
+        "dark": {
+            "logo": "sso/google.svg",
+            "color_fill": "#131314",
+            "color_stroke": "#8E918F",
+            "color_text": "#E3E3E3",
+        },
+    },
+    "github": {
```
nirvn, line 468: Do we have to define colors for non-google providers? I'd rather have QField be in control of styling if we aren't legally forced otherwise.

lukasgraf: Great point, thank you! Will do 👍

docker-app/qfieldcloud/authentication/views.py
lukasgraf: Great point, thank you! Will do 👍

docker-app/qfieldcloud/settings.py
lukasgraf: So, I would consider these styles provided as as "suggested" or "safe" colors.

I would still define values for them server side so that a client can dynamically render a login button for a previously unknown provider in a way that has good contrast, and the colors don't immediately clash, but I don't view these as _prescriptive_. Rather some helpful info from the server that can be used as a fallback in clients so that they can render login buttons they haven't explicitly styled yet.

This should give us a bit more flexibility to add new providers without having to also coordinate updates to the clients every time.

In terms trademark legalities, I think Google is likely to be the most nitpicky one - but even then: if your pill buttons don't even have a stroke around them, simply ignore the stroke color. I can't see that being a real issue. This is also why I based those colors on Google's guidelines, and essentially just copy pasted the same colors to the other ones, which don't make as detailed recommendations.

So by all means, style them how it makes sense in QField, with as much or as little info from those server side suggested styles as you feel appropriate!

---

Having said that, please don't hesitate to say _"I think you've taken this too far"_ if that's what you feel 😛 I'm not emotionally attached to code I wrote, and I'm happy to slash it in half if it makes for a better solution.

suricactus: I am in favor of having recommended colors!

nirvn: @lukasgraf , OK, what about adding a "required": true/false property to each style, so we can differentiate recommendation vs. legally mandated styling?

docker-app/qfieldcloud/settings.py
```diff
@@ -428,6 +430,58 @@ def before_send(event, hint):
 SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT = True
 SOCIALACCOUNT_LOGIN_ON_GET = True
 
+QFIELDCLOUD_PASSWORD_LOGIN_DISABLED = (
+    os.environ.get("QFIELDCLOUD_PASSWORD_LOGIN_DISABLED", "true").lower() == "true"
```
suricactus, line 434: Would you mind switching to 0/1 for the envvar? I think most of the booleans are configured like this in .env.example? By the way, are the sample values with documentation added there too?

docker-app/qfieldcloud/authentication/sso/provider_styles.py
```diff
@@ -0,0 +1,25 @@
...
+
+class SSOProviderStyles:
+    """Helper class to get the style settings for a provider."""
+
+    def __init__(self, request: HttpRequest):
+        self.request = request
+
+    def get(self, provider_id: str) -> dict:
+        """Get the style settings for a (sub)provider."""
+        style_settings = settings.QFIELDCLOUD_SSO_PROVIDER_STYLES.get(provider_id, {})
```
suricactus, line 16: Not sure if I would prefer to just throw in this cases, so people would know what we support and what we dont?

docker-app/qfieldcloud/authentication/views.py
suricactus: I would suggest that you add at least a link to the QGIS docs/source for each of the QGIS variable below too.

docker-app/qfieldcloud/settings.py
```diff
@@ -428,6 +430,58 @@ def before_send(event, hint):
...
+)
+
+QFIELDCLOUD_SSO_PROVIDER_STYLES = {
```
suricactus, line 437: Ideally this info should somehow live as part of the supported adapter definition. I did not follow the latest developments, but if I remember right it could/should be in the database, so extra JSON database column would be perfect. Not strictly required as of this version, but if not too much work would be great to have it now.

docker-app/qfieldcloud/authentication/views.py
lukasgraf: @suricactus these are not really tied to QGIS though. If we were to pick `http://localhost:9999/banana` as the redirect URL, then QGIS would simply spawn the temporary webserver on port `9999`, instruct the webserver to expect the callback on the `/banana` subpath, and also pass that URL along to the IDP as the `redirect_url`. These values don't need to match anything that's defined in the QGIS source - QGIS will just honour them if the client puts them in the auth config, and that's that.

So it almost doesn't matter what values we use for those. I just went with the default values that QGIS puts in the form because why make them different than QGIS' defaults if there is no _need_ for them to be different, but they're pretty much arbitrary. So it's a bit easier when testing auth configs manually during development, and more recognizable. Except that I used `localhost` instead of `127.0.0.1` because many OAuth2 Apps don't allow IP addresses to be used as part of redirect URLs.

So maybe the naming I chose is a bit confusing: `QGIS_DEFAULT_REDIRECT_HOST` implies that this is the QGIS default. That's true for _PORT and _URL (not for _HOST though). But whatever QGIS's defaults are is not even relvevant here, it just uses what we tell it to.

What I rather wanted to express is that these are the defaults (on the QFC side, in the `ListProvidersView`) which are used to configure QGIS' dynamically spawned webserver that catches the redirect. And I put `QGIS` into the same to signify that these host / port settings apply to the QGIS webserver, not QFC. Technically, they're not even defaults though, but just fixed values that I didn't want to hardcode directly somewhere. So I see how that naming is confusing.

I would suggest to maybe change it like this:

```python
    # Address at which QGIS should spawn its temporary webserver to catch
    # the redirect. These mostly happen to match QGIS defaults, except we use
    # localost instead of 127.0.0.1 because many OAuth2 apps don't allow IPs.
    QGIS_REDIRECT_HOST = "localhost"
    QGIS_REDIRECT_PORT = 7070
    QGIS_REDIRECT_URL = ""
```

docker-app/qfieldcloud/settings.py
lukasgraf: I've been back and forth on this, and in the end I ended up with this because it has the lowest chance of accidentally disabling password login through misconfiguration.

Parsing of booleans from environment variables is actually quite tricky, and it's done in a few different ways in `settings.py`. The best approach for this I think would be [`values.BooleanValue()` from django-configurations](https://django-configurations.readthedocs.io/en/stable/values.html#overview). Does all the nasty parsing for your, behaves as expected, and handles defaults.

I'm very happy to revisit this once you're available to talk, but for now I'd like to keep it like this because there's some issues with parsing 0/1 (in a one-liner), and I've carefully tested all the failure modes of this particular implementation.

I have added the setting with its default value to the `.env.example` 👍

docker-app/qfieldcloud/settings.py
lukasgraf: I agree, that is where I would ideally want it too.

(I think you are saying it should be part of the social account providers configs, right? For example in [`SOCIALACCOUNT_PROVIDERS` like here](https://github.com/opengisch/QFieldCloud/compare/master...feat/oidc-qfield-testing#diff-fd0496d9f85dc710b6f459fdb77d77f42de7e96352116b1f3144e35eb049e57fR71-R91), or, alternatively, yes, in the database as d.allauth `SocialApp` models).

Unfortunately, django-allauth doesn't have a clean way to store additional info like this in its provider settings. Whatever solution we go with has to work with both `settings.py` settings and DB settings as `SocialApp`, which are the two equivalent mechanisms that django-allauth supports for provider configuration (unless we want to break that, and lose that flexibility, which I find quite nice).

I have tried just adding additional values directly to its provider settings in `settings.py`. That does work in the sense that django-allauth doesn't complain about those additional keys. But that's today, in its current implementation, and at some point it might reject provider "settings" that it doesn't recognize.

Worse, for the case where provider configs are stored in the DB, this info would end up in the [`SocialApp.settings` field](https://github.com/pennersr/django-allauth/blob/434c4c073642980b6777eab9b67f1161d6c2f522/allauth/socialaccount/models.py#L71), which is already a JSONField. This might work right now, but I could very well imagine that this could for example lead to any of django-allauths future migrations to break, if it finds unexpected settings, or remove our additional data during a migration.

I believe this would best be solved by making a PR upstream, to add a dedicated `extra_data` field or something like that to the provider configs. This aspect of d.a is not very easily customizable, unfortunately. While it has its account adapters, the way it reads its configuration is not really customizable from what I can tell (without monkey patching or anything like that).

docker-app/qfieldcloud/authentication/sso/provider_styles.py
lukasgraf: My idea here was to make it easy for clients to have robust parsing of these styles.
There will always be a `style` property, but possibly an empty dict. Anything below may or may not be there, it's style suggestions if we do have and want to give them. But I don't think a missing style definition for a newly configured provider should prevent login from working. We want as little as possible effort to support new providers, no? Ideally, we (or someone else) can just configure a new provider in the standard django-allauth way, and it will just work.

lukasgraf:
I'll take a break for now 😁 Sorry, I didn't mean to pull you guys in for reviews during the weekend! It just took longer to get this all ready yesterday than I thought, and I didn't want to delay opening the PR any further than I needed to.

docker-app/qfieldcloud/settings.py
suricactus: I think if we prefix with `QFC_` the risk of future collision would be quite low. However the migrations case is still there, but i think low risk. I am ok to accept the risk in favor of easier configuration. What is harder to me is when the settings are split in two places and one has to match both properly.

Polluting settings hashes is something that I have already done for STORAGES.

I believe of there is some kind of incompatibility, we will discover it during the development phase. Also it is worth seeing what migrations allauth guys already have.

I mean, feel free to merge like this, but it is highly likely I merge both settings hashes in the next few months.

docker-app/qfieldcloud/authentication/sso/provider_styles.py
suricactus: Depends how the client is going to implement it. If they can show sensible button and text, I am ok to leave this optional. But if we already put an expectation that this setting is mandatory, people will suffer a bit more when adding a new provider and a bit less on the client side and future qfc migrations.

I don't have the whole thing in my head right now, so might be missing something.

docker-app/qfieldcloud/settings.py
suricactus: Yes, as we discussed, one of the upcoming refactoring will be to switch to tha module. 

I am surprised that we already used different approaches, tbh i only remembe Rahat debug is using 1/0 approach.

duke-nyuki:
Task linked: [QF-5435 Add API endpoint that returns the currently avalable login methods](https://app.clickup.com/t/2192114/QF-5435)

nirvn:
PR looks good by me. @suricactus , do you feel this is OK to merge pending further discussions on some of the philosophical issues raised here? :)

suricactus:
Go ahead, I just put random comments now and then, but I will be back on 15th. Don't wait me until then.

PR #1198, opened 2025-04-02
Build and push createbuckets docker image

gounux:
Since the recent storage changes, the docker image used to create the buckets for local instances is not the `minio/mc` anymore, but a specific one that is built using a Dockerfile in the `docker-createbuckets` folder of the QFieldCloud repository.

This PR sets up the CI to build and push a `qfieldcloud-createbuckets` image to the hub, using the same way as the other built images (app, nginx, worker_wrapper, qgis).

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -45,6 +45,12 @@ jobs:
             docker_context: docker-nginx
             enable_dockerhub: true
 
+          -
+            service_name: createbuckets
+            dockerfile: docker-createbuckets/Dockerfile
+            docker_context: docker-createbuckets
+            enable_dockerhub: true
+
     steps:
       -
         name: Checkout


duke-nyuki:
Task linked: [QF-5622 [ci] build and push createbuckets image](https://app.clickup.com/t/2192114/QF-5622)

m-kuhn:
The code for the container image generation looks good.
Is a new container the right way to go here? I wonder if it would be possible to add that as a django management command instead https://github.com/opengisch/QFieldCloud/tree/master/docker-app/qfieldcloud/core/management/commands ?

suricactus:
See my related comment here, might give some insights on the approach

https://github.com/opengisch/QFieldCloud/pull/1174#issuecomment-2745915240

gounux:
Am aligned with the approach that keeps the `mc` commands running, for creating new buckets.

Questions : would we want to add the `minio`/`mc` dependencies to the app in order to keep doing so ?

suricactus:
I am wondering if we should really change the already existing approach of having a separate container?

gounux:
> I am wondering if we should really change the already existing approach of having a separate container?

+1, I'd also like to add to the discussion that this "script" for bucket creation exists only for local / standalone instances. IMHO, it makes sense to keep it separated from the "core" of QFC, also avoids us the task of developing it in python and adding the minio related dependencies to the QFC Django project.

m-kuhn:
No strong opinion here

suricactus:
Shall we then merge this PR?

m-kuhn (approved):

gounux:
Sure, thanks!

PR #1199, opened 2025-04-02
Fix incorrect key usage in kwargs.pop() for filename extraction

tdrivas:
This PR addresses a bug where `kwargs.pop("thumbnail.png")` was mistakenly using a literal string as a key in projects where `legacy_storage` is used.

**Impact** 
Using the wrong key caused the function to raise a KeyError, breaking the thumbnail processing logic. This fix ensures the correct file is handled.


duke-nyuki:
Task linked: [QF-5635 Failed to fetch thumbnail from legacy_storage related projects](https://app.clickup.com/t/2192114/QF-5635)

gounux (approved):
Code looks good ! Thanks for this fix.

suricactus:
Good catch! But I think there are some endpoints that have the kwargs with a key filename, no? It is the endpoint used by qfieldsync.

PR #1200, opened 2025-04-03
Make social account providers configurable via env var.

lukasgraf:
This is the final piece of the puzzle, besides some prettyfication, to be able to turn on desired SSO providers for any QFC deployment.

This will allow the **`SOCIALACCOUNT_PROVIDERS` dict to be set from an environment variable**, and therefore the enabled social providers for a particular deployment can simply be configured via its `.env`.

If a valid configuration for a provider is specified in this dict, it will automatically be offered as a login option on the web interface (*) as well as the `auth/providers` endpoint.

---

There is **no reasonable (non-empty) default value** for this, in my opinion. On-prem deployments will want to customize this anyway, and likely not offer any or all of the same providers as `app.qfield.cloud` does. Configs for staging vs. prod will likely differ. And for open source self-hosting users, we definitely don't want them to use _our_ OAuth2 / OIDC apps that are configured on Google, GitHub, etc..

SSO providers are tightly tied to a particular deployment, or at least organization, with their client ID, and sometimes client secret.

---

Implementation wise, this requires the env var to be set to a valid JSON string. I chose to follow the implementation for how the `STORAGES` config is loaded and parsed, which works in a similar way.

(*) For now, still hidden behind the `?sso=1` query string feature flag though, for a soft rollout


gounux (approved):
Looks good to me, thanks a lot!

PR #1203, opened 2025-04-09
Customize auth providers buttons by overriding allauth's template

gounux:
This PR customizes the display of the SSO providers buttons, in the login page.

image

Logic consists of overriding allauth's `provider.html` template, with style being configured in the `QFIELDCLOUD_SSO_PROVIDER_STYLES` settings variable (fill, stroke, text color + svg icon).

Since the QFieldCloud jazzmin UI is light, the `light` styles are being used at the moment.

Approach is to use a css file for non-dynamic style properties.

### DIFF
--- a/docker-app/qfieldcloud/core/templates/allauth/elements/provider.html
+++ b/docker-app/qfieldcloud/core/templates/allauth/elements/provider.html
@@ -1,0 +1,13 @@
+<li class="provider-button-parent">
+
+    <a title="{{ attrs.name }}" href="{{ attrs.href }}">
+        <div class="provider-button" style="border: 1px solid {{ attrs.styles.light.color_stroke|default:"grey" }}; background-color: {{ attrs.styles.light.color_fill|default:"white" }}">
+
+            <img src="{{ attrs.styles.light.logo }}" alt="{{ attrs.name }}">
+            <span>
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-5630 Customize SSO providers buttons](https://app.clickup.com/t/2192114/QF-5630)

lukasgraf (approved):
LGTM! 👍 🎉 

Nice, this is so much cleaner than my example! 😄

docker-app/qfieldcloud/core/templates/allauth/elements/provider.html
```diff
@@ -0,0 +1,13 @@
+<li class="provider-button-parent">
+
+    <a title="{{ attrs.name }}" href="{{ attrs.href }}">
+        <div class="provider-button" style="border: 1px solid {{ attrs.styles.light.color_stroke|default:"grey" }}; background-color: {{ attrs.styles.light.color_fill|default:"white" }}">
```
suricactus, line 4: @gounux a bit of nitpicking. If I were you, I would have moved the `border: 1px solid gray` into the CSS file and here kept only `border-color: {{ ... }}`. This way even if someone provides a color that is `25px` (obviously not a color), it will pick the default from the CSS. Same goes for the `background-color`.

gounux: Thanks for the remark @suricactus , should be refactored in #1208

PR #1205, opened 2025-04-10
Fix `thumbnail_url` for projects using legacy storage

gounux:
This PR fixes the `thumbnail_url` value for projects using legacy storage.

Those projects don't have a `thumbnail` attribute value, so I guess the condition after the `or` was always true.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1249,14 +1249,12 @@ class Project(models.Model):
         Todo:
             * Delete with QF-4963 Drop support for legacy storage
         """
-        if (
-            # legacy storage
-            self.uses_legacy_storage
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-5793 Fix thumbnail_url for projects using legacy storage](https://app.clickup.com/t/2192114/QF-5793)

tdrivas (approved):
Very good catch @gounux

suricactus:
If we split this in two separate ifs it might be more readable?

gounux:
Shall we merge this one ?

suricactus (approved):
Looks good. Another way to do it:
```
        if self.uses_legacy_storage and not self.legacy_thumbnail_uri:
            return ""

        if not self.uses_legacy_storage and not self.thumbnail:
            return ""
```

PR #1206, opened 2025-04-10
Fix downloading package files from projects with multiple packages

gounux:
Basically we need to filter not only by file type, name and version id, but when the file_type is PACKAGE, we also need to filter by package_job_id.

_Follow up of #1183 , which was merged then reverted._

### DIFF
--- a/docker-app/qfieldcloud/filestorage/view_helpers.py
+++ b/docker-app/qfieldcloud/filestorage/view_helpers.py
@@ -5,6 +5,7 @@ from uuid import UUID
 from django.conf import settings
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import transaction
+from django.db.models import Q
 from django.db.models.fields.files import FieldFile
 from django.http import FileResponse, HttpResponse
 from django.http.response import HttpResponseBase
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-5531 Fix downloading package files from projects with multiple packages](https://app.clickup.com/t/2192114/QF-5531)

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -162,23 +163,39 @@ def download_project_file_version(
...
     else:
-        file = File.objects.select_related("latest_version").get(
+        filters = Q(
             project_id=project_id,
             name=filename,
             file_type=file_type,
         )
 
+        if package_job_id:
+            filters &= Q(package_job_id=package_job_id)
```
gounux, line 195: What if the File is of type `PROJECT_FILE` ? Would it have a `package_job_id` value ?

Would something like this be some help ?

```python
filters &= Q(
  Q(package_job_id=package_job_id) | Q(file_type=File.FileType.PROJECT_FILE)
)
```

suricactus: When it is of `type=PROJECT_FILE` there should never be a `package_job_id`

PR #1207, opened 2025-04-10
Retain datasource of localized layers when generating the layers data

nirvn:
@tdrivas , as discussed; this change unlocks localized dataset in QField.

### DIFF
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -36,6 +36,7 @@ from qgis.core import (
     QgsZipUtils,
 )
 from qgis.PyQt import QtCore, QtGui
+from qgis.PyQt.QtXml import QDomDocument
 from tabulate import tabulate
 
 # Get environment variables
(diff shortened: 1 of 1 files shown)


suricactus (commented):
Would you mind explaining a bit further what we are trying to solve here?

I guess that if the datasource is invalid during project parsing, we lose the datasource value? Can't we use the bad file handler?

https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/utils/bad_layer_handler.py

tdrivas:
> Would you mind explaining a bit further what we are trying to solve here?
> 
> I guess that if the datasource is invalid during project parsing, we lose the datasource value? Can't we use the bad file handler?
> 
> https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/utils/bad_layer_handler.py


What we tried to solve with @nirvn is to retrieve the original datasource value for layers that are localized, in order to have the original relative filepath and filename. The solution is using the  `layer.originalXmlProperties()`, which allow us to access the raw XML snippet from the original project file, before QGIS tried to parse it. 

We still need to dig into the XML, in order to get full detail.

You can see the [localized data - PR](https://github.com/opengisch/QFieldCloud/pull/1147) for how this added detail is exploited afterwards.

nirvn:
@suricactus , yeah good idea.

PR #1208, opened 2025-04-15
Refactor styling for social login providers buttons

gounux:
No UI change, just css & style refactoring, with a defensive approach and default values. So in case the style color value gets somehow absurd, the default value is applied.

### DIFF
--- a/docker-app/qfieldcloud/core/staticfiles/css/sso.css
+++ b/docker-app/qfieldcloud/core/staticfiles/css/sso.css
@@ -6,6 +6,9 @@
   }
 
   .provider-button {
+    color: grey;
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/staticfiles/css/sso.css
```diff
@@ -6,6 +6,8 @@
   }
 
   .provider-button {
+    background-color: white;
```
suricactus, line 9:
```suggestion
    color: grey;
    background-color: white;
```

I guess we want to assure readable text over the white background too?

suricactus (approved):
Great,thanks!

PR #1213, opened 2025-04-18
Remove types from docstrings, we can get them from PEP484 typing and retire `List` etc

suricactus:
We don't need to repeat ourselves with:
```
def foo(n: int):
  """
  Args:
      n (int): number
  Returns:
      (int): result
  """
```
As modern tools are fine with just:

```
def foo(n: int) -> int:
  """
  Args:
      n: number
  Returns:
      result
  """
```


gounux (approved):
Code looks good, and should not have impacts I believe, since it's about docstrings and typing. Will note to prefer the `type | None` instead of `Optional[type]` :) thanks!

PR #1214, opened 2025-04-28
Add webdav storage backend

gounux:
This PR adds a Django storage backend for webdav, adapted and inspired [by this repository](https://github.com/marazmiki/django-webdav-storage).

Allows to use a new storage of type `qfieldcloud.filestorage.backend.QfcWebDAVStorage`, which implements basic methods for storing a project's files on it.

2 webdav server implementations tested:

- NextCloud's webdav.
- `bytemark/webdav` docker image (generic and basic webdav implementation), used for testing.

PR steps:

- [x] add Django webdav backend
- [x] add `bytemark/webdav` service to the test and standalone environment
- [x] add tests for uploading / downloading / deleting files of a project configured with the webdav storage

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,6 +64,36 @@ LETSENCRYPT_STAGING=1
 #         "QFC_IS_LEGACY": false
 #     }
 # }
+
+# WebDav storage setup with `bytemark/WebDAV` docker image:
+# NOTE: HTTP basic auth user and password must be the same as the WEBDAV_USERNAME and WEBDAV_PASSWORD env variables.
+# {
+#     "webdav": {
+#         "BACKEND": "qfieldcloud.filestorage.backend.QfcWebDavStorage",
+#         "OPTIONS": {
+#             "webdav_url": "http://qfc_webdav_user:qfc_webdav_pwd@webdav",
+#             "public_url": "http://webdav",
+#             "basic_auth": "qfc_webdav_user:qfc_webdav_pwd"
+#         },
+#         "QFC_IS_LEGACY": false
+#     }
+# }
+
+# WebDav storage setup with an external NextCloud server:
+# NOTE: USERNAME and PASSWORD refer to a user's credentials on the NextCloud server.
+# NOTE: NEXTCLOUD_SHARE_TOKEN refer to a public share token that can be generated from the NextCloud UI.
+# {
+#     "webdav_nextcloud": {
+#         "BACKEND": "qfieldcloud.filestorage.backend.QfcWebDavStorage",
+#         "OPTIONS": {
+#             "webdav_url": "https://USERNAME:PASSWORD@my.nextcloud.server/remote.php/dav/files/USERNAME",
+#             "public_url": "https://my.nextcloud.server/public.php/webdav",
+#             "basic_auth": "NEXTCLOUD_SHARE_TOKEN:"
+#         },
+#         "QFC_IS_LEGACY": false
+#     }
+# }
+
 STORAGES='{
         "default": {
             "BACKEND": "qfieldcloud.filestorage.backend.QfcS3Boto3Storage",
@@ -105,6 +135,26 @@ MINIO_API_PORT=8009
 # NOTE: Ignored if `minio` is not used.
 # DEFAULT: 8010
 MINIO_BROWSER_PORT=8010
+
+# Public port to the webdav server.
+# NOTE: Ignored if `webdav` docker service is not used.
+# DEFAULT: 8020
+WEBDAV_PUBLIC_PORT=8020
+
+# Username of the user in the webdav server.
+# NOTE: Ignored if `webdav` docker service is not used.
+# DEFAULT: qfc_webdav_user
+WEBDAV_USERNAME=qfc_webdav_user
+
+# Password of the user in the webdav server.
+# NOTE: Ignored if `webdav` docker service is not used.
+# DEFAULT: qfc_webdav_pwd
+WEBDAV_PASSWORD=qfc_webdav_pwd
+
+# Comma-separated list of values for the mini WebDAV server names.
+# NOTE: Ignored if `webdav` docker service is not used.
+# DEFAULT: "172.17.0.1,webdav"
+WEBDAV_SERVER_NAMES="172.17.0.1,webdav"
 
 WEB_HTTP_PORT=80
 WEB_HTTPS_PORT=443
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,6 +78,15 @@ jobs:
                   },
                   "QFC_IS_LEGACY": true
               },
+              "webdav": {
+                  "BACKEND": "qfieldcloud.filestorage.backend.QfcWebDavStorage",
+                  "OPTIONS": {
+                      "webdav_url": "http://qfc_webdav_user:qfc_webdav_pwd@webdav",
+                      "public_url": "http://webdav",
+                      "basic_auth": "qfc_webdav_user:qfc_webdav_pwd"
+                  },
+                  "QFC_IS_LEGACY": false
+              },
               "default": {
                   "BACKEND": "qfieldcloud.filestorage.backend.QfcS3Boto3Storage",
                   "OPTIONS": {
--- a/docker-app/qfieldcloud/core/tests/utils.py
+++ b/docker-app/qfieldcloud/core/tests/utils.py
@@ -1,12 +1,18 @@
 import io
 import os
 import tempfile
+import urllib
 from datetime import timedelta
 from time import sleep
 from typing import IO, Iterable
 
+from django.http import FileResponse, HttpResponse
+from django.urls import reverse
 from django.utils import timezone
+from rest_framework.response import Response
+from rest_framework.test import APITransactionTestCase
 
+from qfieldcloud.authentication.models import AuthToken
 from qfieldcloud.core.models import Job, Project, User
 from qfieldcloud.subscription.models import Plan, Subscription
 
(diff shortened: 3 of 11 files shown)


duke-nyuki:

nirvn:
@gounux , very cool

.env.example
```diff
@@ -64,6 +64,34 @@ LETSENCRYPT_STAGING=1
...
+#         "OPTIONS": {
+#             "webdav_url": "https://USERNAME:PASSWORD@my.webdav.server/remote.php/dav/files/USERNAME",
+#             "public_url": "https://my.webdav.server/public.php/webdav",
+#             "basic_auth": "NEXTCLOUD_SHARE_TOKEN:"
+#         },
+#         "QFC_IS_LEGACY": false
+#     }
+# }
+
+# NOTE: for using a storage using WebDAV (bytemark), that can be setup with:
```
suricactus, line 81: I immediately had the question what is `bytemark`?

```suggestion
# Storage setup with `bytemark/WebDAV` docker image with:
```

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -1,5 +1,203 @@
...
+from django.core.files.base import ContentFile
+from django.core.files.storage import Storage
+from django.http import HttpResponse
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
 class QfcS3Boto3Storage(S3Boto3Storage):
-    pass
+    def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
+        pass
```
suricactus, line 13: I am not 100% convinced this is the right way to go, but feel free to keep or remove it. If you keep it, please add description why this is needed and how to implement it.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -1,5 +1,203 @@
...
+    Adapted and inspired by this repository: https://github.com/marazmiki/django-webdav-storage
+    Copyright (c) 2020, Mikhail Porokhovnichenko
+    """
+
+    def __init__(self, **options):
+        self.requests = self.get_requests_session(**options)
+
+        self.webdav_url = options.get("webdav_url")
+        if not self.webdav_url:
+            raise NotImplementedError("Please define the webdav_url storage option")
```
suricactus, line 28:
```suggestion
            raise NotImplementedError("Please define the `webdav_url` storage option")
```

Since it is now in our codebase, it should follow our approach of documenting stuff.

Also, I am not sure if `NotImplementedError` is the best choice? I guess we can check how the other storage providers do it?

docker-app/qfieldcloud/filestorage/tests/test_webdav.py
```diff
@@ -0,0 +1,156 @@
...
+        # Create a user
+        self.u1 = Person.objects.create_user(username="u1", password="abc123")
+        self.t1 = AuthToken.objects.get_or_create(user=self.u1)[0]
+        self.p1 = Project.objects.create(
+            owner=self.u1,
+            name="p1",
+            file_storage="webdav",
+        )
+
+    def _upload_file(
```
suricactus, line 38: I think it is good time to add a mixin/base class that adds these functions, no?

docker-app/qfieldcloud/settings_utils.py
```diff
@@ -14,9 +14,15 @@ class S3StorageOptions(TypedDict):
     endpoint_url: str
 
 
+class WebDAVStorageOptions(TypedDict):
```
suricactus, line 17:
```suggestion
class WebDavStorageOptions(TypedDict):
```

Usually we kill the capitals from abbreviations.

docker-compose.override.standalone.yml
```diff
@@ -79,6 +79,19 @@ services:
     environment:
       STORAGES: ${STORAGES}
 
+  webdav:
```
suricactus, line 82: Don't forget to also update the qfield-docs architecture.md with the new changes once merged.

suricactus (commented):
Looks very promissing, mostly documentation stuff.

docker-app/qfieldcloud/filestorage/backend.py
gounux: Yes I agree, there are maybe more elegant ways to do so. If you have something in mind, feel free to suggest it.

docker-app/qfieldcloud/filestorage/tests/test_webdav.py
gounux: Here , or in a follow-up PR ?

docker-app/qfieldcloud/filestorage/backend.py
gounux: I guess `django-storage` is using `ImproperlyConfigured` exception type (https://github.com/jschneier/django-storages/blob/9de3cc9da4dbd67fcc56ecafc7cbf738af90136c/storages/base.py#L15C23-L15C43).

Thanks for spotting this, let's use it I guess.

suricactus: Thanks!

docker-app/qfieldcloud/filestorage/tests/test_webdav.py
suricactus: I am ok also here.

.env.example
```diff
@@ -106,6 +134,20 @@ MINIO_API_PORT=8009
...
 
+# Public port to the webdav server.
+# NOTE: Ignored if `webdav` is not used.
+# DEFAULT: 8020
+WEBDAV_PUBLIC_PORT=8020
+
+# Credentials of the user in the webdav server.
+# NOTE: Ignored if webdav is not used.
+WEBDAV_USERNAME=qfc_user
+WEBDAV_PASSWORD=qfc_pwd
```
suricactus, line 145: Makes it explicit where it is used just by looking at the values:
```suggestion
WEBDAV_USERNAME=qfc_webdav_user
WEBDAV_PASSWORD=qfc_webdav_pwd
```

suricactus: Also, add documentation before each envvar, do not group them like this. With a default value.

.env.example
```diff
@@ -106,6 +134,20 @@ MINIO_API_PORT=8009
 # DEFAULT: 8010
 MINIO_BROWSER_PORT=8010
 
+# Public port to the webdav server.
+# NOTE: Ignored if `webdav` is not used.
+# DEFAULT: 8020
+WEBDAV_PUBLIC_PORT=8020
+
+# Credentials of the user in the webdav server.
+# NOTE: Ignored if webdav is not used.
```
suricactus, line 143:
```suggestion
# NOTE: Ignored if `webdav` docker service is not used.
```
Update all related comments accordingly.

docker-app/qfieldcloud/filestorage/backend.py
suricactus: I don't have a strong opinion for now.

.github/workflows/test.yml
```diff
@@ -78,6 +78,15 @@ jobs:
                   },
                   "QFC_IS_LEGACY": true
               },
+              "webdav": {
+                  "BACKEND": "qfieldcloud.filestorage.backend.QfcWebDavStorage",
+                  "OPTIONS": {
+                      "webdav_url": "http://qfc_user:qfc_pwd@172.17.0.1:8020",
```
suricactus, line 84:
```suggestion
                      "webdav_url": "http://qfc_user:qfc_pwd@webdav:8020",
```
Does this work?

docker-app/qfieldcloud/filestorage/backend.py
gounux: Okay, let's add a comment why this is there then

docker-app/qfieldcloud/filestorage/tests/utils.py
suricactus, line 1: I guess moving this to `core` makes more sense?

docker-createbuckets/createbuckets.py
```diff
@@ -23,6 +23,10 @@ class DjangoStorages(TypedDict):
 
 
 for storage_name, storage_config in STORAGES.items():
+    if storage_config["BACKEND"] != "qfieldcloud.filestorage.backend.QfcS3Boto3Storage":
+        print(f"Skipping storage {storage_name}, no need to create S3 bucket.")
```
suricactus, line 27:
```suggestion
        print(f'Skipping storage "{storage_name}" with backend `{storage_config["BACKEND"]}`, no need to create S3 bucket.')
```

A bit more info.

.env.example
```diff
@@ -64,6 +64,34 @@ LETSENCRYPT_STAGING=1
...
+#     "webdav": {
+#         "BACKEND": "qfieldcloud.filestorage.backend.QfcWebDavStorage",
+#         "OPTIONS": {
+#             "webdav_url": "http://qfc_webdav_user:qfc_webdav_pwd@172.17.0.1:8020",
+#             "public_url": "http://172.17.0.1:8020",
+#             "basic_auth": "qfc_webdav_user:qfc_webdav_pwd"
+#         },
+#         "QFC_IS_LEGACY": false
+#     }
+# }
```
suricactus, line 93: I am not 100% sure why we need these two examples? Probably you can change the keys to `webdav_nextcloud` in case we need to keep that one and change the leading comment for both cases to be similar?

suricactus (approved):
Great work! I think we are good to merge after the last few bits of discussion I have just posted.

docker-app/qfieldcloud/core/tests/utils.py
```diff
@@ -148,3 +154,95 @@ def wait_for_project_ok_status(project: Project, wait_s: int = 30):
 
 def fail(msg):
     raise AssertionError(msg or "Test case failed")
+
+
+class QfcFilesTestCase(APITransactionTestCase):
```
suricactus, line 159: I suggest we change this to a mixin and leave the actual test user to choose what *TestCase class they want to use. Can be done in a follow-up PR too.

PR #1220, opened 2025-05-04
Add filtering capabilities to the /api/v1/projects/ endpoint to fetch project(s) details filtered by user name (and optionally a project name)

nirvn:
@suricactus , as discussed, the PR adds two GET parameters to /api/v1/projects/ to filter the list by owner and/or project name. E.g., `/api/v1/projects/?owner=ivan&name=kebab`.

This will allow us to provide easier permalinks to retrieve cloud project details via API within QField.

### DIFF
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -26,9 +26,16 @@ class ProjectFilterSet(django_filters.FilterSet):
         return queryset
 
     def filter_queryset(self, queryset):
+        include_public = "0"
+        if (
+            "include-public" in self.form.data
+            and self.form.data["include-public"] != ""
+        ):
+            include_public = self.form.data["include-public"]
+
         if "include_public" in self.form.cleaned_data:
-            if not self.form.cleaned_data["include_public"]:
-                self.form.cleaned_data["include_public"] = "0"
+            if self.form.cleaned_data["include_public"] == "":
+                self.form.cleaned_data["include_public"] = include_public
 
         return super().filter_queryset(queryset)
 
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -44,6 +44,7 @@ class ProjectFilterSet(django_filters.FilterSet):
         # since Django does not have a good support for hyphens in field names and it was previously used in the API.
         if "include-public" in self.form.data:
             self.form.cleaned_data["include_public"] = self.form.data["include-public"]
+
         return super().filter_queryset(queryset)
 
     class Meta:
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -1,0 +1,11 @@
+import django_filters
+
+from qfieldcloud.core.models import Project
+
+
+class ProjectFilter(django_filters.FilterSet):
+    owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
+
+    class Meta:
+        model = Project
+        fields = ["owner", "name"]
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -1,11 +1,37 @@
 import django_filters
+from django.db import models
 
-from qfieldcloud.core.models import Project
+from qfieldcloud.core.models import Project, ProjectQueryset
+
+
+class IncludePublicChoices(models.IntegerChoices):
+    EXCLUDE = 0, "Exclude"
+    INCLUDE = 1, "Include"
 
 
 class ProjectFilterSet(django_filters.FilterSet):
     owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
+    include_public = django_filters.ChoiceFilter(
+        label="Include public projects",
+        choices=IncludePublicChoices.choices,
+        method="filter_include_public",
+    )
+
+    def filter_include_public(self, queryset, name, value):
+        if value != "1":
+            queryset = queryset.exclude(
+                user_role_origin=ProjectQueryset.RoleOrigins.PUBLIC
+            )
+
+        return queryset
+
+    def filter_queryset(self, queryset):
+        if "include_public" in self.form.cleaned_data:
+            if not self.form.cleaned_data["include_public"]:
+                self.form.cleaned_data["include_public"] = "0"
+
+        return super().filter_queryset(queryset)
 
     class Meta:
         model = Project
-        fields = ["owner", "name"]
+        fields = ["name", "owner", "include_public"]
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -3,7 +3,7 @@ import django_filters
 from qfieldcloud.core.models import Project
 
 
-class ProjectFilter(django_filters.FilterSet):
+class ProjectFilterSet(django_filters.FilterSet):
     owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
 
     class Meta:
--- a/docker-app/qfieldcloud/core/filters.py
+++ b/docker-app/qfieldcloud/core/filters.py
@@ -10,14 +10,23 @@ class IncludePublicChoices(models.IntegerChoices):
 
 
 class ProjectFilterSet(django_filters.FilterSet):
-    owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
+    name = django_filters.CharFilter(
+        field_name="name", label="Project name", lookup_expr="iexact"
+    )
+    owner = django_filters.CharFilter(
+        field_name="owner__username",
+        label="Project owner username",
+        lookup_expr="iexact",
+    )
     include_public = django_filters.ChoiceFilter(
-        label="Include public projects",
+        label="Include public projects (can be provided with it's deprecated name `include-public`)",
         choices=IncludePublicChoices.choices,
         method="filter_include_public",
     )
 
-    def filter_include_public(self, queryset, name, value):
+    def filter_include_public(
+        self, queryset: models.QuerySet[Project], name: str, value: str
+    ) -> models.QuerySet[Project]:
         if value != "1":
             queryset = queryset.exclude(
                 user_role_origin=ProjectQueryset.RoleOrigins.PUBLIC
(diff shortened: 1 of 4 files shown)


docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -94,6 +94,21 @@ def get_queryset(self):
             if include_public_param and include_public_param.lower() == "1":
                 include_public = True
 
+            owner = self.request.query_params.get("owner")
+            if owner:
+                owner_account = User.objects.get(username=owner)
+                projects = projects.filter(owner=owner_account.id)
+                # When filtering projects by owner, include public projects by default
+                if not include_public_param:
+                    include_public = True
```
gounux, line 103: Am wondering, what happens if the `include_public_param` is there but the value set `0` ?

nirvn: @gounux , then public projects are excluded. I think if someone explicitly sets include-public=0&owner=gounux , that someone must have a reason to want public project exclusion :)

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -94,6 +94,21 @@ def get_queryset(self):
             if include_public_param and include_public_param.lower() == "1":
                 include_public = True
 
+            owner = self.request.query_params.get("owner")
```
suricactus, line 97:
```suggestion
            owner_username = self.request.query_params.get("owner")
```

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -94,6 +94,21 @@ def get_queryset(self):
...
+            if owner:
+                owner_account = User.objects.get(username=owner)
+                projects = projects.filter(owner=owner_account.id)
+                # When filtering projects by owner, include public projects by default
+                if not include_public_param:
+                    include_public = True
+
+            project = self.request.query_params.get("project")
```
suricactus, line 105:
```suggestion
            project_name = self.request.query_params.get("project")
```

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -94,6 +94,21 @@ def get_queryset(self):
...
+            if project:
+                projects = projects.filter(name=project)
+                # When filtering projects by project name, include public projects by default
+                if not include_public_param:
+                    include_public = True
```
suricactus, line 110: I would not put this logic neither here, not in the previous occurance. It is hard to document and understand. The client should know whether or not they want to search the public projects too.

suricactus (commented):
Thanks for the contribution!  Great that you are providing a test case!

A few small comments to address for the current approach.

However, ideally before we add any new filters, I would have preferr to switch to `django-filter` for such stuff:

- https://www.django-rest-framework.org/api-guide/filtering/#setting-filter-backends
- https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html

If this is blocking any client request, it is acceptable with the current approach. If not blocking, I would prefer we switch to `django-filter` first. Could be in the same PR.

tdrivas:
> Thanks for the contribution! Great that you are providing a test case!
> 
> A few small comments to address for the current approach.
> 
> However, ideally before we add any new filters, I would have preferr to switch to `django-filter` for such stuff:
> 
> * https://www.django-rest-framework.org/api-guide/filtering/#setting-filter-backends
> * https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html
> 
> If this is blocking any client request, it is acceptable with the current approach. If not blocking, I would prefer we switch to `django-filter` first. Could be in the same PR.

+1 for the django-filter

nirvn:
@suricactus , @tdrivas , done; the only one thing I dislike a tiny bit is how we can't alias owner__username , but everything is cleaner using django-filters :)

nirvn:
The one big benefit, its perfect integration with swagger listing:

image

tdrivas:
> @suricactus , @tdrivas , done; the only one thing I dislike a tiny bit is how we can't alias owner__username , but everything is cleaner using django-filters :)

Fast as the wind. 

Yes, you can create an alias by creating a custom filter.

More or less like this

```
class ProjectFilter(django_filters.FilterSet):
    owner   = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
    project = django_filters.CharFilter(field_name="name",           lookup_expr="exact")

And then use it in the view.

**Comment by tdrivas** [comment]
@nirvn 

The full working example: 

`filters.py`

```
@nirvn 

The full working example is the following:
```
import django_filters
from django_filters import rest_framework as filters
from qfieldcloud.core.models import Project

class ProjectFilter(django_filters.FilterSet):
    owner   = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
    project = django_filters.CharFilter(field_name="name",           lookup_expr="exact")

    class Meta:
        model  = Project
        fields = ["owner", "project"]
```

Then, the `ProjectViewSet` inside the `project_views.py` becomes

```
from qfieldcloud.core.filters import ProjectFilter
from django_filters.rest_framework import DjangoFilterBackend

class ProjectViewSet(viewsets.ModelViewSet):
    queryset           = Project.objects.all()
    serializer_class   = ProjectSerializer
    lookup_url_kwarg   = "projectid"
    permission_classes = [permissions.IsAuthenticated, ProjectViewSetPermissions]

    filter_backends    = [DjangoFilterBackend, QfcOrderingFilter]
    filterset_class    = ProjectFilter
    ordering_fields    = ["owner__username::alias=owner", "name", "created_at"]        
```

nirvn:
@tdrivas , thanks, that worked like a charm.

nirvn:
@suricactus , your requests have been addressed.

docker-app/qfieldcloud/core/filters.py
```diff
@@ -0,0 +1,44 @@
...
+
+class ProjectFilterSet(django_filters.FilterSet):
+    owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
+    include_public = django_filters.ChoiceFilter(
+        label="Include public projects",
+        choices=IncludePublicChoices.choices,
+        method="filter_include_public",
+    )
+
+    def filter_include_public(self, queryset, name, value):
```
suricactus, line 20:
```suggestion
    def filter_include_public(self, queryset: models.QuerySet[Project], name: str, value: str) -> models.QuerySet[Project]:
```

docker-app/qfieldcloud/core/filters.py
```diff
@@ -0,0 +1,44 @@
...
+            "include-public" in self.form.data
+            and self.form.data["include-public"] != ""
+        ):
+            include_public = self.form.data["include-public"]
+
+        if "include_public" in self.form.cleaned_data:
+            if self.form.cleaned_data["include_public"] == "":
+                self.form.cleaned_data["include_public"] = include_public
+
+        return super().filter_queryset(queryset)
```
suricactus, line 40:
```suggestion
    def filter_queryset(self, queryset: models.QuerySet[Project]) -> models.QuerySet[Project]:
        if self.form.cleaned_data.get("include_public") != "":
            return super().filter_queryset(queryset)

        # This is a workaround to make the filter work with the `include-public` parameter,
        # since Django does not have a good support for hyphens in field names and it was previously used in the API.
        self.form.cleaned_data["include_public"] = self.form.data["include-public"]

        return super().filter_queryset(queryset)
```

I think this is simpler to read and maintain and doing the same thing. Tried it locally, seems to work fine.

docker-app/qfieldcloud/core/filters.py
```diff
@@ -0,0 +1,44 @@
...
+from qfieldcloud.core.models import Project, ProjectQueryset
+
+
+class IncludePublicChoices(models.IntegerChoices):
+    EXCLUDE = 0, "Exclude"
+    INCLUDE = 1, "Include"
+
+
+class ProjectFilterSet(django_filters.FilterSet):
+    owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="exact")
```
suricactus, line 13:
```suggestion
    owner = django_filters.CharFilter(field_name="owner__username", lookup_expr="iexact")
```
Case insensitive search for owner, is it OPENGIS or openGIS :) They are case insestive unique anyways.

docker-app/qfieldcloud/core/filters.py
```diff
@@ -0,0 +1,44 @@
...
+    include_public = django_filters.ChoiceFilter(
+        label="Include public projects",
```
suricactus, line 15:
```suggestion
        label="Include public projects. Can be provided with it's deprecated name `include-public`.",
```

Or something better, you name it.

suricactus (commented):
Ok, man vs machine 0:1. Threre is an limitation deeply into Django forms and the implementation within the django_filters, that field names are evaluated twice. 

I think this is acceptable. A few small comments.

docker-app/qfieldcloud/core/filters.py
nirvn: Good.

docker-app/qfieldcloud/core/filters.py
nirvn: Done.

docker-app/qfieldcloud/core/filters.py
nirvn: Done.

docker-app/qfieldcloud/core/filters.py
nirvn: Done.

nirvn:
@suricactus , thanks for the review, I've applied the relevant changes.

One thing I've come to realize though is that we have to keep around the public project exclusion in the projects_views.py (not the parameter but the exclusion by default when no parameter is stated). With that it, all works smoothly. Thanks.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -90,21 +79,27 @@ class ProjectViewSet(viewsets.ModelViewSet):
...
-                include_public = True
+            if include_public_param and include_public_param != "":
+                force_exclude_public = False
+
+            include_public_param = self.request.query_params.get("include_public")
+            if include_public_param and include_public_param != "":
+                force_exclude_public = False
 
-            if not include_public:
+            if force_exclude_public:
```
suricactus, line 102: I would move this comments and stuff into `ProjectFilterSet.filter_include_public` with a small comment here pointing about "quirks in showing public projects".

suricactus: Nevermind, discussed over chat that this approach will not work. Thinking about something else.

nirvn:
@suricactus , any chance we can push this through the finish line? I'd really leverage this for QField 3.7.

suricactus (approved):
Let us merge this like it is now.

PR #1221, opened 2025-05-04
Fix internal server error 500 when calling /api/v1/projects/ from browser

nirvn:
Fixes server error 500 when calling /api/v1/projects/ within a browser. On app.qfield.cloud, the error shows as:

image

While on classic localhost installation, the error shows as:

image


suricactus:
Thanks for the PR. However, the problem is not where it seems to be, because this scenario was caught with the `if view.action == "list"` on the first line. It turned out that `OPTIONS` request was setting the `view.action` to be `None`.

Closing this PR as superseded by https://github.com/opengisch/QFieldCloud/compare/404_on_missing_project_in_browser?expand=1

PR #1223, opened 2025-05-05
Fix error 500 when using the DRF in-browser viewer on GET /projects

suricactus:
Turns out that there is a fake OPTIONS requests being processed, which sets the `view.action` to `None` instead of one of the known strings, e.g. the expected `list`.

Therefore we start returning `False` if the `view.action` is unknown as the most conservative approach.

NOTE most probably there are other list views that suffer from the same problem.

### DIFF
--- a/docker-app/qfieldcloud/core/views/projects_views.py
+++ b/docker-app/qfieldcloud/core/views/projects_views.py
@@ -18,7 +18,12 @@ User = get_user_model()
 
 
 class ProjectViewSetPermissions(permissions.BasePermission):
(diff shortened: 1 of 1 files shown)


nirvn (approved):
I had tried this option too, wasn't sure which one was right. As long as the list works, good by me ;)

suricactus:
This is the more correct one for sure, the other one is hiding a scenario.

nirvn:
@suricactus , good to hear.

gounux (approved):
Nice catch, works for me :)

PR #1224, opened 2025-05-05
Return ObjectNotFoundError when get_object_or_404 is used in the code

suricactus:
Literally the title, when there is 404, we need to handle the error into a JSON friendly `ObjectNotFoundError`.

### DIFF
--- a/docker-app/qfieldcloud/core/rest_utils.py
+++ b/docker-app/qfieldcloud/core/rest_utils.py
@@ -2,6 +2,7 @@ import logging
 
 from django.conf import settings
 from django.core import exceptions
+from django.http import Http404
 from rest_framework import exceptions as rest_exceptions
 from rest_framework.response import Response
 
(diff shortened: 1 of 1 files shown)


tdrivas:
Thanks for improving the exception handling, opening the door to use with safety `get_object_or_404` !

suricactus:
Thanks! Would you mind doing a code review?

docker-app/qfieldcloud/core/rest_utils.py
```diff
@@ -18,14 +19,18 @@ def exception_handler(exc, context):
...
         qfc_exc = qfieldcloud_exceptions.ObjectNotFoundError(detail=str(exc))
     elif isinstance(exc, exceptions.ValidationError):
         qfc_exc = qfieldcloud_exceptions.ValidationError(detail=str(exc))
     elif isinstance(exc, qfieldcloud_exceptions.QFieldCloudException):
         qfc_exc = exc
     elif isinstance(exc, rest_exceptions.APIException):
-        qfc_exc = qfieldcloud_exceptions.APIError(exc.detail, exc.status_code)
+        # Map DRF API exceptions to qfc exceptions. Note the `detail` attribute is always present but not properly typed in `APIException`
+        qfc_exc = qfieldcloud_exceptions.APIError(
+            detail=getattr(exc, "detail"),
```
tdrivas, line 31: Make the handler more robust in case of a custom exception (inherited from APIException) does not include `detail`: 

```suggestion
            detail=getattr(exc, "detail", str(exc))
```

tdrivas (approved):
Great improvement! Handling both `ObjectDoesNotExist` and `Http404` in the exception handler is much appreciated. Only one small suggestion for enhancing robustness.

docker-app/qfieldcloud/core/rest_utils.py
suricactus: Actually I tried (and seems I failed) to bring the atttention at the fact that `details` is always present and I did this change only to make the typing happy. We must throw if there is no `detail`, because something went very wrong.

tdrivas: Fair enough , thanks for the clarification!

suricactus: Just pushed a new commit here, hopefully now the comment says it all.

PR #1226, opened 2025-05-06
Allow individual files to be stored at different storage providers within the same project (attachments)

gounux:
This PR intends to bring the file storage to the `File` level, thus allowing files to be stored on different storage providers within the same projects.

There is a new `attachments_file_storage` field, on the `Project` level, that allows to set the storage provider of file attachments (the one located in the project's `attachment_dirs`).

Every uploaded file, if located in the project's `attachment_dirs`, will be stored on this `attachments_file_storage`.

This storage provider for file attachments can be changed in a project's form:

image

- [x] add new `file_storage` field on the `File` model and make it used by its `FileVersion`s' `DynamicStorageFieldFile` / `DynamicStorageFileField` fields
- [x] add new `attachments_file_storage` field in the `Project` model
- [x] create migrations in the `core` and `filestorage` modules
- [x] add successful file attachments upload / download / deletion tests


duke-nyuki:

docker-app/qfieldcloud/core/admin.py
```diff
@@ -762,6 +763,17 @@ class Meta:
         widgets = {"the_qgis_file_name": widgets.TextInput()}
         fields = "__all__"  # required for Django 3.x
 
+    def _storages_choices(self) -> list[tuple[str, str]]:
+        storages = list(settings.STORAGES.keys())[:-1]
+        print(storages)
+        return [(storage, storage) for storage in storages]
```
suricactus, line 769: Since this can be shown in multiple places, e.g. in the File or Project level (twice), shall we make it a util function that is available globally? `get_file_storage_choices` or something like this?

gounux: Great idea, let's do that!

docker-app/qfieldcloud/filestorage/tests/test_attachments.py
```diff
@@ -0,0 +1,125 @@
...
+            "DCIM/file3.name",
+            io.FileIO(testdata_path("DCIM/1.jpg"), "rb"),
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(self.p1.files.count(), 3)
+
+        file3 = self.p1.get_file("DCIM/file3.name")
+
+        self.assertEqual(file3.file_storage, "webdav")
```
gounux, line 125: If we look at what this test does, it means that _as is_, it would be possible, for a same project, to have attachments stored on several storages, if the `attachment storage` field of a project gets changed over time.

docker-app/qfieldcloud/core/migrations/0082_project_attachments_file_storage.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations, models
+
+import qfieldcloud.core.models
+import qfieldcloud.core.validators
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0081_file_storage_project_and_more"),
+    ]
```
gounux, line 12: @suricactus should we add a migration dependency to the `0082` from the core module here ?

docker-app/qfieldcloud/filestorage/tests/test_attachments.py
suricactus: Not sure if I follow, but let's discuss in the call later.

suricactus:
Would it be too hard to make the filestorage fields readonly if there are already some files?

docker-app/qfieldcloud/core/migrations/0082_project_attachments_file_storage.py
gounux: Actually I meant to make the `filestorage` migration brought by this PR dependent from the `core` one, apologies.

PR #1232, opened 2025-05-13
fix(nginx): refresh DNS when upstream app ip changes

toebivankenoebi:
## Summary
If the app container needs to be rebuilt or restarted it sometimes receives a new ip address (dynamically assigned by docker). If this address changes nginx didn't pick up on those changes since by default it caches the initial resolution forever. This lead to a server error and could be resolved by restarting the nginx service or container.

To counteract this we need to use docker DNS and make sure nginx resolves the name on a regular basis.

While doing this I also cleaned up the error message handling for 4xx and 5xx errors, made sure the image path is correct and they are shown properly.

## Reproduce race condition
To reproduce the race condition described above and being able to test, these were the steps taken.

- Make sure that the nginx container can successfully resolve the app backend dns name
```
docker inspect qfieldcloud-app-1 -f '{{json .NetworkSettings.Networks}}'
docker exec -it qfieldcloud-nginx-1 apt update
docker exec -it qfieldcloud-nginx-1 apt install dnsutils -y
docker exec -it qfieldcloud-nginx-1 dig app
```

- We need to force the app container to retrieve a different IP address and therefor disconnect it first
```
docker network disconnect qfieldcloud_default qfieldcloud-app-1
```

- Then create as many temp_containers as necessary until you take over the old ip of the app container and thus force the app container to get a new ip
```
docker run -d --name temp_container --network qfieldcloud_default alpine sleep infinity
docker inspect temp_container -f '{{json .NetworkSettings.Networks}}'
```

- Once the old ip is in use by the app container connect the app container again, it will receive another address now
```
docker network connect qfieldcloud_default qfieldcloud-app-1 --alias app
docker inspect qfieldcloud-app-1 -f '{{json .NetworkSettings.Networks}}'
```

- docker container is successfully able to resolve the name to the new ip
```
docker exec -it qfieldcloud-nginx-1 dig app
docker compose logs qfieldcloud-nginx-1
```

- Nginx now fails to connect to the upstream backend since it still assumes the old address
```
nginx-1  | 2025/05/13 14:15:08 [error] 37#37: *21 connect() failed (111: Connection refused) while connecting to upstream, client: 172.20.0.1, server: localhost, request: "GET /favicon.ico HTTP/1.1", upstream: "http://172.20.0.7:8000/favicon.ico", host: "localhost", referrer: "https://localhost/admin/"
```

- Remove all created temp docker containers again
```
for c in $(docker ps -f name=temp --format json | jq -r .Names); do docker stop $c; docker rm $c; done
```


docker-nginx/templates/default.conf.template
```diff
@@ -28,10 +28,17 @@ log_format json-logger escape=json
 
 
 upstream django {
+  # Defines a shared memory zone between worker processes, necessary for resolve keyword
```
suricactus, line 31:
```suggestion
  # Defines a shared memory zone between worker processes, necessary for `resolver` keyword
```

I guess you mean the `resolver` here? I guess the 64k is an arbitrary number, shall we also mention it in the comment?

suricactus (commented):
A few questions to clarify my understanding.

Basically we do cache the DNS resolution of the `app` container for 1s now (`resolver 127.0.0.11 valid=1s ipv6=off;`). I am wondering if this has any non-negligable performance penalty? I guess not, just trying to understand for myself.

As for the changes in `  server app:8000 max_fails=1 fail_timeout=1s resolve;`, what will happen if we fail to resolve `app:8000` in 1s and the retry fails too? I guess we get error 502 Bad Gateway?

docker-nginx/templates/default.conf.template
```diff
@@ -28,10 +28,17 @@ log_format json-logger escape=json
...
+  zone django 64k;
+
+  # Add Docker's DNS resolver with a short TTL and IPv6 turned off
+  resolver 127.0.0.11 valid=1s ipv6=off;
```
m-kuhn, line 35: Today we discussed a different issue with @suricactus which seems to be caused by ipv6. It is completely unrelated, so it would be interesting to know what the reason for disabling this here was.

docker-nginx/templates/default.conf.template
toebivankenoebi: The one second results in a docker internal dns lookup and is then shared between the workers for 1 second. This shouldn't have any performance impact.

When the app is not reachable, it fails after 1 second with either a 502 or 503 (I even saw a 500 error - it doesn't seem to consistent) which we then pickup to show a custom error message (see my newer commit).

Shared memory refers to the 'resolve' keyword that is reliant on it, clarified in a new commit.

docker-nginx/templates/default.conf.template
toebivankenoebi: As far as I understand it, docker internal DNS is via IPv4 and so is there default networking. Would be interested though what problems were caused by ipv6. Generally if we don't use it, we disable it, since it might cause delays we don't want (most probably wouldn't be the case here).

m-kuhn: nginx also needs to be able to proxy resources from external services (S3), we had problems with a system where ipv6 was not enabled but a hostname was still resolved to an ipv6. Disabling ipv6 in nginx was one workaround that did the job.

toebivankenoebi:
@suricactus , thanks for the questions, had me thinking a bit more about timeouts. Introduced the `proxy_connect_timeout` which gives a grace period of 5 seconds. Also now we handle the failing better by showing a loading page for the first minute until we really fail with a 500 error.

While I was at it I realised that we don't show the favicons in the error pages and fixed that.

suricactus (commented):
Thanks for the explanations, happy that my questions led to further improvements.

Would it make sense splitting the PR in two and keeping the `loading.html` related changes in another branch.

Very quick feedback on the javascript, can you please use `const` whenever instead of `let` and for sure instead of `var`? Thanks!

docker-nginx/templates/default.conf.template
```diff
@@ -25,13 +25,19 @@ log_format json-logger escape=json
...
 
-upstream django {
   # the number of keepalive connections comes from the number Django workers we have
   keepalive 32;
 
-  server app:8000 fail_timeout=0;
+  # Use Docker DNS resolution to dynamically resolve the app backend hostname
+  # max_fails=1 and fail_timeout=1s ensures Nginx retries on almost every request from loading.html
```
suricactus, line 39:
```suggestion
  # `max_fails=1` and `fail_timeout=1s` ensures Nginx retries on almost every request from `loading.html`
```
A small comment about styling. In general whenever we refer to something code-like, we always prefer to use backticks (like in Markdown). Probably not applied everywhere, but for sure all new code respects this convention.

toebivankenoebi:
> Thanks for the explanations, happy that my questions led to further improvements.
> 
> Would it make sense splitting the PR in two and keeping the `loading.html` related changes in another branch.

They are kind of together because once I reduce the timeouts it will pretty much instantly fail and display the old 500 error which does not recover (no refresh whatsoever). But I'm happy to do so, then I'd close this PR and create two separate ones, the first one with only the DNS resolving and timeouts (maybe add a meta refresh to the 500 error page and remove it in the second PR again?) and the second for error catching and displaying loading. @suricactus What do you think?

> Very quick feedback on the javascript, can you please use `const` whenever instead of `let` and for sure instead of `var`? Thanks!

Done.

suricactus:
No need to make extra work for the better git history. Good as it is.

docker-nginx/templates/default.conf.template
```diff
@@ -173,7 +194,6 @@ server {
     proxy_intercept_errors on;
 
     proxy_pass $redirect_uri;
-
```
suricactus, line 176: I guess this change was by accident?

Avoid random changes like this to reduce the unneccesary changes. Now it is fine, but especially if it is the only change in the PR it makes later git greping a bit of a mess.

docker-nginx/templates/default.conf.template
suricactus: @toebivankenoebi I will show you tomorrow what it was about. Basically line 156:
```diff
    # required DNS
-    resolver 8.8.8.8;
+    resolver 8.8.8.8 ipv6=off;
```

docker-nginx/pages/403.html
```diff
@@ -4,12 +4,13 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="icon" href="/pages/favicon.ico" type="image/x-icon">
```
suricactus, line 7: I think we should fix the `type` in some other places in the codebase, e.g. on app.qfield.cloud , feel free to take over the follow up after this PR :)

docker-nginx/pages/loading.html
```diff
@@ -0,0 +1,256 @@
...
+    (function() {
+        const refreshCountKey = 'qfcLoadingRefreshCount';
+        const maxRefreshes = 6; // 6 refreshes * 10 seconds per refresh = 60 seconds
+        const errorPageUrl = '/pages/500.html';
+        const retryUrl = '/'; // The URL to navigate to for a retry (e.g., the root)
+        const retryIntervalMs = 10000; // 10 seconds
+
+        const rawCount = sessionStorage.getItem(refreshCountKey);
+        let currentRefreshCount = parseInt(rawCount || '0', 10);
+        currentRefreshCount++;
```
suricactus, line 239:
```suggestion
        const incrementedRefreshCount = parseInt(rawCount || '0', 10) + 1;
```
Less mutable state, less problems :)

docker-nginx/templates/default.conf.template
suricactus: Thanks!

docker-nginx/pages/loading.html
suricactus: @toebivankenoebi you need to change the variable also a few line below where it is actually used.

This is a good note that I don't always propose commitablr code, I just give a sign what can/should be changed and it's up to the developer to pick the important bits and apply across the code.

toebivankenoebi: Will do and test the behaviour properly again before pushing.

toebivankenoebi: Done and successfully tested.

suricactus: Thanks!

suricactus (approved):
Thanks for the contribution and comments, ready to merge.

PR #1233, opened 2025-05-15
Allow individual files to be stored at different storage providers within the same project (attachments)

gounux:
This PR intends to bring the file storage to the `File` level, thus allowing files to be stored on different storage providers within the same projects.

There is a new `attachments_file_storage` field, on the `Project` level, that allows to set the storage provider of file attachments (the one located in the project's `attachment_dirs`).

Every uploaded file, if located in the project's `attachment_dirs`, will be stored on this `attachments_file_storage`.

This storage provider for file attachments can be changed in a project's form:

image

- [x] add new `file_storage` field on the `File` model and make it used by its `FileVersion`s' `DynamicStorageFieldFile` / `DynamicStorageFileField` fields
- [x] add new `attachments_file_storage` field in the `Project` model
- [x] create migrations in the `core` and `filestorage` modules
- [x] add successful file attachments upload / download / deletion tests

_Follow up of #1226 which was closed by accident, apologies._


duke-nyuki:

docker-app/qfieldcloud/core/migrations/0082_project_attachments_file_storage.py
```diff
@@ -0,0 +1,29 @@
...
+                default=qfieldcloud.core.models.get_project_file_storage_default,
+                help_text="Which file storage provider should be used for storing the project attachments files.",
+                max_length=100,
+                validators=[qfieldcloud.core.validators.file_storage_name_validator],
+                verbose_name="Attachments file storage",
+            ),
+        ),
+        migrations.RunSQL(
+            "UPDATE core_project SET attachments_file_storage = file_storage"
+        ),
```
suricactus, line 28:
```suggestion
        migrations.RunSQL(
            "UPDATE core_project SET attachments_file_storage = file_storage",
            migrations.RunSQL.noop,
        ),
```

I kind like the explicit noop whenever possible, it gives a sense that things are finished.

I guess writing in pure python is not worth it, hope we don't hate ourselves about this in the future.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -762,6 +764,14 @@ class Meta:
         widgets = {"the_qgis_file_name": widgets.TextInput()}
         fields = "__all__"  # required for Django 3.x
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["attachments_file_storage"] = forms.ChoiceField(
+            choices=get_file_storage_choices(), required=True
+        )
+        if self.instance.has_attachments_files:
```
suricactus, line 772: I was thinking to be even more conservative and count any files, not only attachments, but this should also work.

docker-app/qfieldcloud/filestorage/migrations/0002_file_file_storage.py
```diff
@@ -0,0 +1,59 @@
...
+    with connection.cursor() as cursor:
+        cursor.execute(
+            """
+            UPDATE filestorage_file ff
+            SET
+                file_storage = cp.file_storage
+            FROM core_project cp
+                WHERE cp.id = ff.project_id
+        """
+        )
```
suricactus, line 21: I tried in pure python, but it creates either a query that does not run:
```
File.objects.filter(id=-9).update(name=F('project__file_storage'))
```
with `FieldError: Joined field references are not permitted in this query`

or supper inefficient query:
```
File.objects.update(
    file_storage=Subquery(Project.objects.filter(id=OuterRef('project_id')).values('file_storage')[:1])
)
```

So it seems we have to live with the raw SQL .

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -190,10 +205,18 @@ def add_version(
                 package_job_id=package_job_id,
             )
         except File.DoesNotExist:
+            file_storage = project.file_storage
+
+            # if the new file is an attachment (i.e. in an attachment dir),
+            # it should be stored on the project's configured attachments storage.
+            if storage.get_attachment_dir_prefix(project, filename) != "":
+                file_storage = project.attachments_file_storage
```
suricactus, line 213:
```suggestion
            # if the new file is an attachment (i.e. in an attachment dir),
            # it should be stored on the project's configured attachments storage.
            if storage.get_attachment_dir_prefix(project, filename) != "":
                file_storage = project.attachments_file_storage
            else:
                file_storage = project.file_storage
```

suricactus (commented):
Looks good to me, nice tests. Added a few comments, but we are pretty much there.

suricactus (approved):
Great job! One small comment because of me and we are ready to merge.

suricactus (approved):
ready to merge!

PR #1234, opened 2025-05-15
Support `Range` in HTTP header during downloading files

gounux:
This PR intends to allow downloading only a byte range of a file, using QFieldCloud API, by providing a `Range` HTTP header. Can be used if a big file's first download has crashed for some reason, this new header allows then to download only the missing bytes.

**cURL usage example**:

```sh
curl -X GET -L "https://my.qfieldcloud.server/api/v1/files/<PROJECT_ID>/<FILE_NAME>" \
  -H "Authorization: Token <AUTH_TOKEN>" -H "Range: bytes=5-10"
```
 
**Examples of cURL HTTP headers that can be forwarded to the API, on the `/api/v1/files/<project_id>/<file_name>/` endpoint**:

- `-H "Range: bytes=2-7"` -> will download from byte number n°3 included to byte n°8 included.
- `-H "Range: bytes=3-"` -> will download from byte number n°4 included up to the last byte of the file.

**Steps**:

- [x] make API serve only a file's byte range when a `Range` HTTP header is provided, on a configured S3 storage
- [ ] make API serve only a file's byte range when a `Range` HTTP header is provided, on a configured webdav storage
- [x] add tests downloading only a file's specific byte ranges

### DIFF
--- a/docker-app/qfieldcloud/filestorage/tests/test_range.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_range.py
@@ -79,6 +79,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         self.assertIsNone(parse_range("bytes=1-1", file_size))
         # multiple ranges are not supported (yet), see https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.2-9.4.1
         self.assertIsNone(parse_range("bytes=1-5, 10-15", file_size))
+        self.assertIsNone(parse_range("bytes=1-5,10-15", file_size))
 
     def test_upload_file_then_download_range_succeeds(self):
         for project in [self.project_default_storage, self.project_webdav_storage]:
--- a/docker-app/qfieldcloud/filestorage/tests/test_range.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_range.py
@@ -60,9 +60,27 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
 
         self.assertIsNone(result)
 
-        result = parse_range("bytes=-5")
-
-        self.assertIsNone(result)
+        # start position cannot be greater than the end position
+        self.assertIsNone(parse_range("bytes=9-1"))
+        # suffix ranges are not supported (yet), see https://www.rfc-editor.org/rfc/rfc9110.html#rule.suffix-range
+        self.assertIsNone(parse_range("bytes=-5"))
+        # bytes should be numbers
+        self.assertIsNone(parse_range("bytes=one-two"))
+        # whitespaces are not accepted
+        self.assertIsNone(parse_range("bytes= 1-9"))
+        self.assertIsNone(parse_range("bytes=1 -9"))
+        self.assertIsNone(parse_range("bytes=1- 9"))
+        self.assertIsNone(parse_range("bytes=1-9 "))
+        self.assertIsNone(parse_range("bytes=1- "))
+        # typos in bytes
+        self.assertIsNone(parse_range("bites=0-9"))
+        self.assertIsNone(parse_range("starting bytes=0-9"))
+        self.assertIsNone(parse_range("bytes=0-9 closing bytes"))
+        # empty range
+        self.assertIsNone(parse_range("bytes=0-0"))
+        self.assertIsNone(parse_range("bytes=1-1"))
+        # multiple ranges are not supported (yet), see ....
+        ...
 
     def test_upload_file_then_download_range_succeeds(self):
         for project in [self.project_default_storage, self.project_webdav_storage]:
--- a/docker-app/qfieldcloud/filestorage/tests/test_range.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_range.py
@@ -13,6 +13,7 @@ from qfieldcloud.core.tests.mixins import QfcFilesTestCaseMixin
 from qfieldcloud.core.tests.utils import (
     setup_subscription_plans,
 )
+from qfieldcloud.filestorage.utils import parse_range
 
 logging.disable(logging.CRITICAL)
 
@@ -34,6 +35,34 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
             name="project_webdav_storage",
             file_storage="webdav",
         )
+
+    def test_parsing_range_function_succeeds(self):
+        start_byte, end_byte = parse_range("bytes=4-8")
+
+        self.assertEquals(start_byte, 4)
+        self.assertEquals(end_byte, 8)
+
+        start_byte, end_byte = parse_range("bytes=2-")
+
+        self.assertEquals(start_byte, 2)
+        self.assertIsNone(end_byte)
+
+    def test_parsing_wrong_invalid_range_function_succeeds(self):
+        result = parse_range("byte=4-8")
+
+        self.assertIsNone(result)
+
+        result = parse_range("bytes=-1-15")
+
+        self.assertIsNone(result)
+
+        result = parse_range("bytes=-10--15")
+
+        self.assertIsNone(result)
+
+        result = parse_range("bytes=-5")
+
+        self.assertIsNone(result)
 
     def test_upload_file_then_download_range_succeeds(self):
         for project in [self.project_default_storage, self.project_webdav_storage]:
--- a/docker-app/qfieldcloud/filestorage/tests/test_range.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_range.py
@@ -70,6 +70,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         self.assertIsNone(parse_range("bytes=1- 9", file_size))
         self.assertIsNone(parse_range("bytes=1-9 ", file_size))
         self.assertIsNone(parse_range("bytes=1- ", file_size))
+        self.assertIsNone(parse_range(" bytes=1-9", file_size))
         # typos in bytes
         self.assertIsNone(parse_range("bites=0-9", file_size))
         self.assertIsNone(parse_range("starting bytes=0-9", file_size))
--- a/docker-app/qfieldcloud/filestorage/tests/test_range.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_range.py
@@ -37,12 +37,12 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
 
     def test_upload_file_then_download_range_succeeds(self):
         for project in [self.project_default_storage, self.project_webdav_storage]:
-            with self.subTest(case=project):
-                # first upload of the file
-                response = self._upload_file(
-                    self.u1, project, "file.name", StringIO("abcdefghijkl")
-                )
+            # first upload of the file
+            response = self._upload_file(
+                self.u1, project, "file.name", StringIO("abcdefghijkl")
+            )
 
+            with self.subTest(case=project):
                 self.assertEqual(response.status_code, status.HTTP_201_CREATED)
                 self.assertEqual(project.files.count(), 1)
                 self.assertEqual(project.get_file("file.name").versions.count(), 1)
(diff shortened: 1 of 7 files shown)


duke-nyuki:
Task linked: [QF-5425 Support `Range` HTTP header in downloading files in QFieldCloud](https://app.clickup.com/t/2192114/QF-5425)

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +235,32 @@ def download_field_file(
...
+            range_end = int(range_match.group(2))
+        else:
+            range_end = file_size - 1
+
+        if (
+            range_start >= file_size
+            or range_end >= file_size
+            or range_end < range_start
+        ):
+            raise BadRequest("Invalid file range")
```
suricactus, line 260:
```suggestion
            raise BadRequest("Invalid value for `Range` HTTP header`")
```
Just to be aligned with the previous message.

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +235,32 @@ def download_field_file(
...
     https_port = http_host.split(":")[-1] if ":" in http_host else "443"
 
+    download_range = request.headers.get("Range", "")
+    if download_range:
+        range_match = re.match(r"bytes=(\d+)-(\d+)?", download_range)
+
+        if not range_match:
+            raise BadRequest(
+                "`Range` HTTP header must be formatted like `bytes=start-(end)`"
+            )
```
suricactus, line 245: Did you consider https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/416 ?

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +235,32 @@ def download_field_file(
     http_host = request.headers.get("host", "")
     https_port = http_host.split(":")[-1] if ":" in http_host else "443"
 
+    download_range = request.headers.get("Range", "")
+    if download_range:
+        range_match = re.match(r"bytes=(\d+)-(\d+)?", download_range)
```
suricactus, line 240: What if you move this to some of the utils, something like "def parse_range(str) -> tuple[int, int | None]"?

suricactus (commented):
Do we already pass https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Ranges ? If it is easy to add, we can do that. If not, feel free to add a backlog task with it.

A few small comments, seems we are prety much there.

docker-app/qfieldcloud/filestorage/utils.py
```diff
@@ -155,3 +156,27 @@ def to_uuid(value: Any) -> uuid.UUID | None:
         return uuid.UUID(value)
     except ValueError:
         return None
+
+
+def parse_range(input_range: str) -> tuple[int, int | None] | None:
```
suricactus, line 161: Would you mind adding a few unit tests for this? I think it is quite easy to test. 

I am also wondering what the RFC says when you send invalid request?

If one sends:
```
Range: third-10e6
```
You will ignore it because it will not match the regex, but if someone sends:
```
Range: 10-5
```

You will raise?

Shall you ignore the header also?

If yes, then you better move the `if` from:
`if (
            range_start >= file_size
            or range_end >= file_size
            or range_end < range_start
        ):
```
here, it feels more natural to me.

File: docker-app/qfieldcloud/filestorage/tests/test_range.py [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+            file_storage="webdav",
+        )
+
+    def test_upload_file_then_download_range_succeeds(self):
+        for project in [self.project_default_storage, self.project_webdav_storage]:
+            with self.subTest(case=project):
+                # first upload of the file
+                response = self._upload_file(
+                    self.u1, project, "file.name", StringIO("abcdefghijkl")
+                )
```
suricactus, line 44: I guess you can move this outside the `with` statement?

docker-app/qfieldcloud/filestorage/tests/test_range.py
```diff
@@ -0,0 +1,117 @@
...
+                self.assertEquals(r4.status_code, status.HTTP_206_PARTIAL_CONTENT)
+                self.assertEquals(r4.content, b"abcdefghijkl")
+
+    def test_upload_file_then_download_wrong_range_fails(self):
+        for project in [self.project_default_storage, self.project_webdav_storage]:
+            with self.subTest(case=project):
+                # first upload of the file
+                self._upload_file(
+                    self.u1, project, "file.name", StringIO("abcdefghijkl")
+                )
```
suricactus, line 85: Same.

suricactus (approved):
Code looks good, already mergeable, but I believe we can be a bit more paranoid on incoming data.

docker-app/qfieldcloud/filestorage/utils.py
gounux: > I am also wondering what the RFC says when you send invalid request?

A question I'm also asking myself, so far:

- multiple ranges are not supported at the moment
-  [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-range) says: "A server that supports range requests MAY ignore or reject a [Range](https://www.rfc-editor.org/rfc/rfc9110.html#field.range) header field that contains an invalid [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) ([Section 14.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#range.specifiers)), a [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) with more than two overlapping ranges, or a set of many small ranges that are not listed in ascending order, since these are indications of either a broken client or a deliberate denial-of-service attack ([Section 17.15](https://www.rfc-editor.org/rfc/rfc9110.html#overlapping.ranges))" -> not sure I'm reading this correctly, but it seems to me that each implementation may ignore or reject invalid range header

- [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.2-9.1.1) says: "The final 500 bytes (byte offsets 9500-9999, inclusive): bytes=-500" -> should we make the code compliant in this PR ?

gounux: > if someone sends:

> `Range: 10-5`

> You will raise?

> Shall you ignore the header also?

I think we should answer to that question : ignore or reject. Current state is that invalid requests are rejected, my personal opinion favors this approach, since the server will not get an overload in this case.

docker-app/qfieldcloud/filestorage/tests/test_range.py
```diff
@@ -0,0 +1,144 @@
...
+            owner=self.u1,
+            name="project_webdav_storage",
+            file_storage="webdav",
+        )
+
+    def test_parsing_range_function_succeeds(self):
+        start_byte, end_byte = parse_range("bytes=4-8")
+
+        self.assertEquals(start_byte, 4)
+        self.assertEquals(end_byte, 8)
```
suricactus, line 43: btw, this is more compact and IMO readable. No strong opinion, just a hint:

```suggestion
        self.assertEquals(parse_range("bytes=4-8"), (4, 8))
```

docker-app/qfieldcloud/filestorage/utils.py
suricactus: 1) multipart - totally fine if not supported, just add a test to confirm they are not supported. Also probably add it to the `parse_range` function doc string. You may also prepare the function to return a list of ranges already, you can assert in the code it returns only 1 range.
2) yes, I read the same, it's up to the implementation to decide whether to ignore or reject the header. I like being vocal rather than silent, so we better always return an error when we are not happy with the header.
3) I don't think it is needed, unless you say it is very easy to achieve so and it will work fine with Django content file. If someone wants this supported, we are open for PRs :)
4) Reject whenever possible.

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +238,36 @@ def download_field_file(
...
+
+        if range_match[1]:
+            range_end = range_match[1]
+        else:
+            range_end = file_size - 1
+
+        if (
+            range_start >= file_size
+            or range_end >= file_size
+            or range_end < range_start
```
suricactus, line 262: Can we move this specific line in the `parse_range` function?

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +238,36 @@ def download_field_file(
...
+                status=416,
+            )
+
+        file_size = field_file.size
+        range_start = range_match[0]
+
+        if range_match[1]:
+            range_end = range_match[1]
+        else:
+            range_end = file_size - 1
```
suricactus, line 257:
```suggestion
        file_size = field_file.size
        range_start, range_end = range_match[0]

		if range_end is None:
		    range_end = file_size - 1
```

It is safer in this case to explicitly check for `None`.

docker-app/qfieldcloud/filestorage/tests/test_range.py
suricactus: And a few more success cases:

```
# supports start position to be 0
self.assertEquals(parse_range("bytes=0-8"), (0, 8))
self.assertEquals(parse_range("bytes=0-"), (0, None))
# supports non single digit numbers
self.assertEquals(parse_range("bytes=10-999"), (10, 999))
```

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -233,6 +238,36 @@ def download_field_file(
...
+            range_start >= file_size
+            or range_end >= file_size
+            or range_end < range_start
+        ):
+            return HttpResponse(
+                "Invalid value for `Range` HTTP header`",
+                status=416,
+            )
+
+        range_length = range_end - range_start + 1
```
suricactus, line 269:
```suggestion
        range_length = range_end - range_start + 1
        
        if range_length < settings.QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH:
            raise InvalidRangeError("Requested range too small, expected at least {} but got {} bytes".format(settings.QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH, range_length))
```

A few suggestions in this comment:
1) We should add a minimum range of 1MB or so, otherwise someone can DDOS us without even realizing.
2) note the `InvalidRangeError`, which can be added to `qfieldcloud.core.exceptions` instead of manually returning `HttpResponse` here and there. Then the default handler as defined in `rest_utils` should convert it to the proper error.
3) Please do not hardcode the status to 416, use the `rest_framework.status.HTTP_416_REQUESTED_RANGE_NOT_SATISFIABLE` instead of the raw value.

docker-app/qfieldcloud/filestorage/utils.py
```diff
@@ -155,3 +156,36 @@ def to_uuid(value: Any) -> uuid.UUID | None:
...
+def parse_range(input_range: str, file_size: int) -> tuple[int, int | None] | None:
+    """Parses a range HTTP Header string.
+
+    Arguments:
+        range: string value of a HTTP range header to parse.
+        file_size: size of the file to get range for, in bytes.
+
+    Returns:
+        If compliant, a tuple with start and end (if any) bytes number.
+        If not, returns None.
```
suricactus, line 170:
```suggestion
        If not, returns `None`.
```

docker-app/qfieldcloud/testing.py
```diff
@@ -5,4 +5,5 @@
 class QfcTestSuiteRunner(DiscoverRunner):
     def __init__(self, *args, **kwargs):
         settings.IN_TEST_SUITE = True
+        settings.QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH = 1
```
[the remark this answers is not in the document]
gounux: Okay, interesting. That's probably smarter and more elegant.

suricactus (approved):
Ready to merge! Funny number #1234 :)

PR #1235, opened 2025-05-16
feat(secrets)!: add support for per user project secrets

suricactus:
So far QFieldCloud supported secrets that are stored on per-project basis.
So if there are multiple collaborators in one project, they all reuse the same secrets.

This commit allows in addition of the per project settings a user to be
able to set their own secrets.

A sample scenario when this is used:
An organization admin has 2 users in the same project with PostGIS layer
, but they need to see different subset of a particular offlined layer.
So far when both users will share the same credentials provided by the
organization admin as a PGSERVICE file.

With the changes introduced in this PR, the organization admin can
provide separate PGSERVICE file for each user, therefore automatic
filtering can happen already in the PostGIS database based on the DB user.


duke-nyuki:
Task linked: [QF-5423 Add user level secrets for each individual collaborator within the project](https://app.clickup.com/t/2192114/QF-5423)

docker-app/qfieldcloud/core/migrations/0082_alter_secret_options_and_more.py
gounux, line 1: This should be renamed to `0083` since another PR added the  `0082` one

suricactus: well could be, go ahead

docker-app/qfieldcloud/core/cron.py
```diff
@@ -106,6 +106,7 @@ def do(self):
             updated_at__gt=timezone.now() - timedelta(minutes=70),
             updated_at__lt=timezone.now() - timedelta(minutes=5),
         )
+        # TODO Include project jobs ids that are for individual users here
```
gounux: Facing some challenges to understand the sense and direction of this comment TBH.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2197,11 +2303,69 @@ def __str__(self):
...
+        secrets_qs = self.filter(project=project)
+        secrets_qs = secrets_qs.filter(
+            Q(assigned_to=user)
+            | Q(assigned_to__isnull=True)
+            | Q(organization=organization)
+        )
+        # remove the duplicates by name and get the given user overrides
+        secrets_qs = secrets_qs.order_by(
+            "name", "assigned_to", "project", "organization"
+        ).distinct("name")
```
gounux: Does that mean, that in order to take precedence, a secret **must** have the same name that a least-prioritized one ?

docker-app/qfieldcloud/core/migrations/0086_alter_secret_project.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name="secret",
+            name="project",
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="secrets",
+                to="core.project",
+            ),
+        ),
+    ]
```
gounux: @suricactus shall we move this migration block to the `0085` one ? In order to have one single migration for this PR ?

suricactus: Yes, I will ask you at some point to merge all the migrations in one file (if possible, sometimes it isn't).

docker-app/qfieldcloud/core/admin.py
```diff
@@ -671,14 +671,30 @@ def clean(self):
...
+        "assigned_to",
+        "organization",
+        "created_by__link",
+    )
     autocomplete_fields = ("project",)
 
     search_fields = (
         "name__icontains",
         "project__name__icontains",
+        "assigned_to__username__icontains",
```
suricactus, line 697: I guess we can add search by organization too?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1283,6 +1303,53 @@ def shared_datasets_project(self) -> Project | None:
...
+        return (
+            self.jobs.for_user(user)  # type: ignore[attr-defined]
+            .filter(
+                type=Job.Type.PACKAGE,
+            )
+            .order_by("-created_at")
+            .first()
+        )
+
+    def last_package_jobs(self) -> list[PackageJob]:
```
suricactus: Ideally here we should aim to return a querystring, rather than a list. Currently your implementation is very imperative and there is a bit of Django magic that might help you.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -154,7 +153,6 @@ class Meta:
             "data_last_packaged_at",
             "data_last_updated_at",
             "can_repackage",
-            "needs_repackaging",
```
suricactus: You need to ensure that `needs_repackaging` is delivered in the API, otherwise you will break QField.

docker-app/qfieldcloud/core/models.py
gounux: Indeed, correct, looking forward to that magic :)

docker-app/qfieldcloud/core/cron.py
```diff
@@ -124,9 +125,10 @@ def do(self):
                 package_ids = storage.get_stored_package_ids(project_id)
 
                 for package_id in package_ids:
-                    # keep the last package
-                    if package_id == str(project.last_package_job_id):
-                        continue
+                    # keep the last packages
```
suricactus: You made it work for the legacy storage, bug if you look at the else part of the `if project.uses_legacy_storage:`, we have the same issue. Luckily you have solved it already in `docker-app/worker_wrapper/wrapper.py`.

Which brings the question if we can abstract this in a separate function. If we cannot, we can live with it. If yes, then less change to fix one and forget the other. No pressure.

docker-app/qfieldcloud/core/cron.py
```diff
@@ -135,10 +137,18 @@ def do(self):
...
-                ).delete()
+                delete_count = (
+                    File.objects.filter(
+                        project=project,
+                        package_job_id__in=job_ids,
+                    )
+                    .exclude(
+                        package_job_id__in=project.last_package_jobs().values_list(
+                            "id", flat=True
+                        )
```
suricactus:
```suggestion
                        package_job_id__in=latest_project_package_jobs_id_qs,
```

I think this is slightly more redable if you move the qs to a variable.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1283,6 +1303,55 @@ def shared_datasets_project(self) -> Project | None:
         except Project.DoesNotExist:
             return None
 
+    def last_package_job_for_user(self, user: User) -> PackageJob | None:
```
```suggestion
    def latest_package_job_for_user(self, user: User) -> PackageJob | None:
```
No strong opinion, I just think it is a bit more correct English?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1283,6 +1303,55 @@ def shared_datasets_project(self) -> Project | None:
...
+        return (
+            self.jobs.for_user(user)  # type: ignore[attr-defined]
+            .filter(
+                type=Job.Type.PACKAGE,
+            )
+            .order_by("-created_at")
+            .first()
+        )
+
+    def last_package_jobs(self) -> JobQuerySet:
```
```suggestion
    def lastest_package_jobs(self) -> JobQuerySet:
```
No strong opinion, I just think it is a bit more correct English?

docker-app/qfieldcloud/core/models.py
```diff
@@ -2000,8 +2070,34 @@ def method(self):
...
+            The jobs for the user.
+        """
+        has_assigned_to_current_user_project_secrets = False
+        for project in user.projects.all():
+            if (
+                Secret.objects.for_user_and_project(user, project)  # type: ignore[attr-defined]
+                .filter(assigned_to__isnull=False)
+                .exists()
+            ):
+                has_assigned_to_current_user_project_secrets = True
```
suricactus: You have imperative code here (the `for` loop). This means:
1. first query to the database
2. then looping N times
3. each loop has a query to the database

No need to convince you more, we can make this in a single query. You already have warmed up with the querystrings :)

suricactus: Also if you for some reason still need to use `has_assigned_to_current_user_project_secrets`, I would try to shorten it a bit, or at least rename to `has_secrets_assigned_to_current_user_project`. What about `has_assigned_to_secrets`, and assume the reader knows `assigned_to` is a property is related to users?

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+
+        secrets = Secret.objects.for_user_and_project(self.u1, self.p1)
+
+        self.assertEqual(secrets.count(), 3)
+        self.assertEqual(secrets[0], s1_assigned)
+        self.assertEqual(secrets[1], s2_assigned)
+        self.assertEqual(secrets[2], s3_project)
+
+    def test_ind_project_secret_available_for_owner(self):
```
suricactus:
```suggestion
    def test_ind_project_secret_available_for_owner(self):
```
what `ind` stands for? I know it was originally from me, but if you can read the cipher and rename properly...

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,183 @@
...
+        s1_assigned = self._create_secret(
+            name="s1", project=self.p3, assigned_to=self.u1
+        )
+
+        secrets = Secret.objects.for_user_and_project(self.u1, self.p3)
+
+        self.assertEqual(secrets.count(), 1)
+        self.assertEqual(secrets[0], s1_assigned)
+
+    def test_add_secrets_to_projet_and_user_overrides(self):
```
suricactus:
```suggestion
    def test_add_secrets_to_project_and_user_overrides(self):
```

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,183 @@
...
+        p = Project.objects.create(name="project", owner=self.u1)
+        self._create_secret(name="s1", project=p)
+        user_project_secret = self._create_secret(
```
suricactus:
```suggestion
        s1_assigned = self._create_secret(
```

I just observed that in previous tests this would be the name. No strong opinion, just looking for consistency.

To me it is clear that it is a "project secret" based on the name of the test since there is not organization secret to distinguish with. 

Again, feel free to disagree.

suricactus (commented):
Good a few more comments. 

What we are still lacking are bunch of tests that we do not delete the package files by accident.

Sample test case would be:
1) user1 with assigned secret creates a package1 on project1
2) user2 with assgined secret creates a package2 on project1
3) the package1 and package2 should still be present at this point.

This should be tested for assigned and non-asigned secrets for project and organization.

docker-app/qfieldcloud/core/tests/test_secret.py
gounux: Oh, thanks ! "Excuse my french" ;)

docker-app/qfieldcloud/core/tests/test_secret.py
gounux: Probably a typo, I presume.

docker-app/qfieldcloud/core/models.py
gounux: :ok_hand:

docker-app/qfieldcloud/core/models.py
gounux: I guess so, thanks for spotting this.

docker-app/qfieldcloud/core/models.py
gounux: Not sure I'm following you.

Also, don't see changes in the gh suggestion.

suricactus: The two comments are unrelated, let us clarify over chat.

gounux: Alright, thanks.

Would the following be acceptable:

```python
has_assigned_to_current_user_project_secrets = Secret.objects.for_user_and_projects(
    user=user,
    projects=user.projects.all()
).filter(
    assigned_to__isnull=False
).exists()
```

There is still this `user.projects.all()` call as a parameter though.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -403,11 +406,10 @@ def before_docker_run(self) -> None:
     def after_docker_run(self) -> None:
         # only successfully finished packaging jobs should update the Project.data_last_packaged_at
         self.job.project.data_last_packaged_at = self.data_last_packaged_at
-        self.job.project.last_package_job = self.job
         self.job.project.save(
             update_fields=(
                 "data_last_packaged_at",
-                "last_package_job",
+                # "last_package_job",
```
suricactus: Shall we kill the comment?

gounux: For sure ! Thanks

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,119 @@
...
+from qfieldcloud.core.tests.mixins import QfcFilesTestCaseMixin
+from qfieldcloud.core.tests.utils import (
+    setup_subscription_plans,
+    testdata_path,
+    wait_for_project_ok_status,
+)
+
+
+class QfcTestCase(QfcFilesTestCaseMixin, APITestCase):
+    # def setUp(cls):
```
suricactus: comment :)

docker-app/qfieldcloud/core/utils2/db.py
suricactus, line 1: this is no longer used it seems, I guess we can keep it in theory, no strong opinion.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,119 @@
...
+            project=self.p1,
+            created_by=self.u1,
+            type=Job.Type.PACKAGE,
+        )
+        Job.objects.create(
+            project=self.p1,
+            created_by=self.u2,
+            type=Job.Type.PACKAGE,
+        )
+        # TODO FIXME: this should work
```
suricactus: I am not sure what this comment means? You do need to wait for packages process to finish, if this was your question.

Actually `latest_packages` should be empty if the jobs are not finished yet.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -687,7 +704,10 @@ def created_by__link(self, instance):
 
     @admin.display(ordering="project__name")
     def project__name(self, instance):
-        return model_admin_url(instance.project, instance.project.name)
+        if instance.project:
+            return model_admin_url(instance.project, instance.project.name)
+
+        return "-"
```
suricactus: If you return `None` I think you should get the placeholder? I might be wrong too.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,119 @@
...
+
+        return self.client.post(
+            "/api/v1/jobs/",
+            {
+                "project_id": project.id,
+                "type": Job.Type.PACKAGE.value,
+            },
+        )
+
+    def test_create_user_level_packages(self):
```
suricactus, line 64: Those are on project level, can we test on organization level too? 

I think we can add a few tests also for:
1) create project secret1 that is not assigned
2) package job created
3) verify the package is available
4) secret2 with the same name, but assigned to the user (org or project level)
5) no package is available now
5) create package job
6) verify the package is available
7) the package from 3) is no longer available as files

suricactus (commented):
A few more tests (can be follow-up after merging this). Consider it approved.

docker-app/qfieldcloud/core/utils2/db.py
gounux: I'd feel more confident if we remove it in a follow-up PR, since this one is already quite consequent.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: Thank you very much ! Stored on the ticketing tool.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: Yeah this is the challenge : I can not succeed getting job triggered during the tests, gat an `AssertionError: Still pending jobs after waiting for 30 seconds`. Same happens on this CI.

suricactus: I guess we have to look together at this. Ping me when you have time for this

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: > 5. no package is available now

@suricactus I'm not sure I get you there.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1283,6 +1303,55 @@ def shared_datasets_project(self) -> Project | None:
...
+        """Returns all the last package jobs for the users of the project.
+
+        Returns:
+            QuerySet of all the last package jobs.
+        """
+        if self.owner.is_organization:
+            # all the admin users including the organization owner
+            triggered_by_qs = Person.objects.for_organization(self.owner).filter(
+                organization_role=OrganizationMember.Roles.ADMIN,
+            )
```
gounux, line 1334: @suricactus are we sure about this filter here ?

Seems to make [some tests](https://github.com/opengisch/QFieldCloud/blob/86a836f2d61da2f3d565324e83c8f75cf8f8193b/docker-app/qfieldcloud/core/tests/test_packages.py#L581) fail [in the CI](https://github.com/opengisch/QFieldCloud/actions/runs/15850332095/job/44681962211?pr=1235),

docker-app/qfieldcloud/core/tests/test_secret_packages.py
suricactus: if you check what is the `latest_package` for that specific user at step 5, there should be 0 results, because the other package is no longer applicable.

gounux:
@suricactus all the tests are passing now, thanks !

Shall we go in a last review round ?

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -124,6 +124,13 @@ def validate(self, data):
 
         return data
 
+    def get_needs_repackaging(self, obj: Project) -> bool:
+        for collaborator in obj.direct_collaborators.all():
+            if obj.needs_repackaging(collaborator):
+                return True
+
+        return False
```
```suggestion
    def get_needs_repackaging(self, obj: Project) -> bool:
        return obj.needs_repackaging(self.request.user))
```

Actually we need to repackage only for this particular user.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -138,7 +145,6 @@ class Meta:
             "data_last_packaged_at",
             "data_last_updated_at",
             "can_repackage",
-            "needs_repackaging",
```
suricactus, line 141: Are you sure we need to remove it from here? If you check `avatar_url` in the next serializer, it is in both lists.

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,179 @@
...
+)
+from qfieldcloud.core.permissions_utils import (
+    UserOrganizationRoleError,
+    UserProjectRoleError,
+)
+from qfieldcloud.core.tests.utils import setup_subscription_plans
+
+
+class QfcTestCase(APITestCase):
+    # def setUp(cls):
```
suricactus: Do we need to keep this?

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus, line 1: Did you test the changes in this file?

docker-app/qfieldcloud/core/utils2/packages.py
suricactus, line 1: Do we have a test for `delete_obsolete_packages`? I think it is worth it at least one test, since it is deleting files.

docker-app/qfieldcloud/core/views/package_views.py
```diff
@@ -85,13 +85,16 @@ class LegacyLatestPackageView(views.APIView):
     def get(self, request, project_id):
         """Get last project package status and file list."""
         project = Project.objects.get(id=project_id)
+        last_package_job = project.latest_package_job_for_user(request.user)
```
suricactus:
```suggestion
        latest_package_job = project.latest_package_job_for_user(request.user)
```

Just to make sure we use the same name everywhere. I think it is worth it to grep for `last.*package` in the code, there might be a few other places where we are left with the old name.

docker-app/qfieldcloud/core/views/package_views.py
```diff
@@ -85,13 +85,16 @@ class LegacyLatestPackageView(views.APIView):
...
 
         # Check if the project was packaged at least once
-        if not project.last_package_job:
+        if not last_package_job:
             raise exceptions.InvalidJobError(
                 "Packaging has never been triggered or successful for this project."
             )
 
+        assert last_package_job
```
suricactus: Does it worth the assert since we have a `if not last_package_job: raise ...` above?

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self._create_secret(name="s1", organization=self.o1)
+        self._create_secret(name="s1", organization=self.o1, assigned_to=self.u1)
+        self._create_secret(name="s1", project=p)
+        s1_assigned = self._create_secret(name="s1", project=p, assigned_to=self.u1)
+
+        secrets = Secret.objects.for_user_and_project(self.u1, p)
+
+        self.assertEqual(secrets.count(), 1)
+        self.assertEqual(secrets[0], s1_assigned)
```
suricactus:
```suggestion
        secrets = Secret.objects.for_user_and_project(self.u1, p)
        
        s1_org = self._create_secret(name="s1", organization=self.o1)

        secrets.refresh_from_db()
        self.assertEqual(secrets.count(), 1)
        self.assertEqual(secrets[0], s1_org)

        s1_org_assigned = self._create_secret(name="s1", organization=self.o1, assigned_to=self.u1)

        secrets.refresh_from_db()
        self.assertEqual(secrets.count(), 1)
        self.assertEqual(secrets[0], s1_org_assigned)

        s1_project = self._create_secret(name="s1", project=p)

        secrets.refresh_from_db()
        self.assertEqual(secrets.count(), 1)
        self.assertEqual(secrets[0], s1_project)

        s1_project_assigned = self._create_secret(name="s1", project=p, assigned_to=self.u1)

        secrets.refresh_from_db()
        self.assertEqual(secrets.count(), 1)
        self.assertEqual(secrets[0], s1_project_assigned)
```

Since we have the test, I think we can check the priority for each step.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
+        latest_package_jobs_qs = self.p1.latest_package_jobs()
+
+        self.assertEquals(latest_package_jobs_qs.count(), 2)
+
+        latest_packages_users = latest_package_jobs_qs.values_list(
+            "created_by", flat=True
+        )
+
+        self.assertIn(self.u1.id, latest_packages_users)
+        self.assertIn(self.u2.id, latest_packages_users)
```
suricactus:
```suggestion
        u1_package_job = repackage(self.p1, self.u1)
        u2_package_job = repackage(self.p1, self.u2)
        
        # there must be two separate package jobs per user
        self.assertNotEqual(u1_package_job, u2_package_job)

        wait_for_project_ok_status(self.p1)
        self.p1.refresh_from_db()

        latest_package_jobs_qs = self.p1.latest_package_jobs()

        self.assertEquals(latest_package_jobs_qs.count(), 2)

        u1_latest_package_job = latest_package_jobs_qs.get(triggered_by=u1)
        u2_latest_package_job = latest_package_jobs_qs.get(triggered_by=u2)

        self.assertEquals(u1_package_job, u1_latest_package_job)
        self.assertEquals(u2_package_job, u2_latest_package_job)
```

I think this is a bit more strict in testing.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+        self._upload_file(
+            self.u1,
+            self.p1,
+            "simple_bumblebees.qgs",
+            io.FileIO(testdata_path("simple_bumblebees.qgs"), "rb"),
+        )
+
+        # upload some attachments files to the project.
```
suricactus: Just wondering do we need attachments at all? If yes, shall we assert something with them? If not, I believe we can remove them.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
```
suricactus: Just wondering do we need attachments at all? If yes, shall we assert something with them? If not, I believe we can remove them.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
```
suricactus: Just wondering do we need attachments at all? If yes, shall we assert something with them? If not, I believe we can remove them.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
+        latest_package_jobs_qs = self.p1.latest_package_jobs()
+
+        self.assertEquals(latest_package_jobs_qs.count(), 1)
+
+        # create a package for user u2
+        repackage(self.p1, self.u2)
+
+        latest_package_jobs_qs = self.p1.latest_package_jobs()
+
+        self.assertEquals(latest_package_jobs_qs.count(), 2)
```
suricactus, line 152: See my comment here for improved testing. Now you are just checking the count, but not the actual contents of the returning query. https://github.com/opengisch/QFieldCloud/pull/1235/files#r2167570659

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
+
+        package_job_2 = repackage(self.p1, self.u1)
+        wait_for_project_ok_status(self.p1)
+        self.p1.refresh_from_db()
+
+        latest_package_ids = self.p1.latest_package_jobs().values_list("id", flat=True)
+
+        self.assertEquals(len(latest_package_ids), 1)
+        self.assertIn(package_job_2.id, latest_package_ids)
+        self.assertNotIn(package_job_1.id, latest_package_ids)
```
suricactus:
```suggestion
        latest_package_jobs_qs = self.p1.latest_package_jobs()

        self.assertEquals(latest_package_jobs_qs.count(), 1)
        self.assertIn(package_job_2, latest_package_jobs_qs.first())
```

The same spirit as the previous comments.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -0,0 +1,245 @@
...
+        latest_package_ids = self.p1.latest_package_jobs().values_list("id", flat=True)
+
+        self.assertIn(package_job_1.id, latest_package_ids)
+
+        # create a user-level secret, assigned to user u1 for project p1
+        self._create_secret(
+            name="SECRET", assigned_to=self.u1, project=self.p1, value="u1"
+        )
+
+        package_job_2 = repackage(self.p1, self.u1)
```
suricactus, line 189:
```suggestion
        package_job_2 = repackage(self.p1, self.u1)

        self.assertNotEqual(package_job_1, package_job_2)
```

suricactus (commented):
I believe last round of review, great progress, very good work!

docker-app/qfieldcloud/core/tests/test_secret.py
gounux: I guess it's already gone :eyes:

docker-app/qfieldcloud/core/models.py
```diff
@@ -2136,6 +2241,10 @@ def get_feedback_step_data(self, step_name: str) -> dict[str, Any] | None:
         return None
 
 
+class PackageJobQuerySet(InheritanceQuerySet):
```
```suggestion
class PackageJobQuerySet(JobQuerySet):
```

It already inherits from `InheritanceQuerySet`. What you are missing in the child job classes is `objects = JobQuerySet.as_manager()`

docker-app/qfieldcloud/core/tests/test_secret.py
```diff
@@ -0,0 +1,197 @@
...
+    def test_check_secrets_type_priority(self):
+        p = Project.objects.create(name="project", owner=self.u1)
+
+        secrets = Secret.objects.for_user_and_project(self.u1, p)
+
+        s1_org = self._create_secret(name="s1", organization=self.o1)
+
+        secrets.refresh_from_db()
+        self.assertEqual(secrets.count(), 1)
+        self.assertEqual(secrets[0], s1_org)
```
gounux: @suricactus are we sure about the expected behaviour of those asserts ?

In the current state, the `for_user_and_project` method of SecretQuerySet does not seem consider to consider org-assigned secrets.

suricactus: This is now sorted I guess?

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus: ping @gounux

gounux: Not yet though. Didn't edit the file so far. Let me check.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1103,6 +1111,12 @@ class Meta:
 
     files: "FileQueryset"
 
+    # The secrets that are attached for a specific project.
+    secrets: "models.QuerySet[Secret]"
```
```suggestion
    secrets: "SecretQueryset"
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -1103,6 +1111,12 @@ class Meta:
...
+
+    # The jobs that are ran for a specific project.
+    jobs: "models.QuerySet[Job]"
```
```suggestion
    jobs: "JobQuerySet"
```

suricactus:
:green_circle:  APPROVED!

thembela66 (approved):

docker-app/qfieldcloud/core/serializers.py
suricactus: @gounux this change actually broke the API. As I suggested above, we should have kept the value here, otherwise it is no longer served. Adding the `get_needs_repackaging` is not enough. That's why `avatar_url` was listed there.

nirvn: I can confirm this leads to QField being unable to download projects.

suricactus: When addressing, please also add a basic test that ensures the presense and the value of this field.

gounux: Indeed, oops... Apologies, should be fixed by #1273

PR #1240, opened 2025-05-23
fix(nginx): disable ipv6 for the resolver in internal redirects for file download

suricactus:
For hosts that are not having ipv6 enabled we get an error. Unfortunately the resolver resolves to ipv6 by default.

### DIFF
--- a/docker-nginx/templates/default.conf.template
+++ b/docker-nginx/templates/default.conf.template
@@ -155,7 +155,7 @@ server {
     set $file_range "$upstream_http_file_range";
 
     # required DNS
-    resolver 8.8.8.8;
+    resolver 8.8.8.8 ipv6=off;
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-6109 Turn off ipv6 for the DNS resolver in the file download internal redirect for nginx](https://app.clickup.com/t/2192114/QF-6109)

suricactus:
@toebivankenoebi @m-kuhn that was the change we discussed briefly in our call and in the other PR #1232

m-kuhn:
This will prevent us from using an ipv6 only S3 backend IUC?

suricactus:
Correct.  Actually ipv6 only storage provider, not only s3, e.g. webdav.

I think it is acceptable limitation for now.

PR #1244, opened 2025-05-27
nginx: django stackstrace shown in debug mode

toebivankenoebi:
Since nginx handles 5xx errors and displays custom error pages we can't see the django stacktrace anymore if `DEBUG=1`.
Here we introduce an env variable in the nginx container and pick it up in the configuration.
If debugging is on, the django stacktrace will now be shown again.
If debugging is off, the custom 500 error page is shown.


docker-nginx/templates/default.conf.template
```diff
@@ -2,6 +2,11 @@ map "$time_iso8601 # $msec" $time_iso8601_ms {
   "~([^+]+)\+([\d:]+?) # \d+?\.(\d+)" "$1.$3+$2";
 }
 
+map "${DEBUG}" $debug_mode {
+    "1" "on";
+    default "off";
+}
```
suricactus, line 8: Is there any hidden value to rename $DEBUG values to `on`/`off` besides readability? Not sure if it is worth it to add a mapping when we can simply `if  ($DEBUG = '1')` below.

toebivankenoebi: The mapping and evaluation takes place when Nginx is loading the configuration while the error_page operates at the request level. I think it's the more robust way.

suricactus: If I understand you right, you mean that `if ($debug_mode = "on")` will be evaluated once on startup? If yes, this would be great performance boost. But I still do not understand why `if  ("$DEBUG" = "1")` will not behave the same, since it will be replaced even before startup with `if ("1" = "1")`.

Not saying one or the other approach is better, but so far I lack the understanding why, besides readability and maintainability, the `map` approach is more robust.

toebivankenoebi: As far as I can find out it's the preferred way in nginx of handling external vars, because:
1. Afaik It is evaluated when loading the configuration during startup (couldn't find a proper documentation) and is available then to all workers spawned in the future
2. If for whatever reason we're using the environment variable directly and it changes during nginx runtime a new worker might have a different variable available
3. [Env](https://nginx.org/en/docs/ngx_core_module.html#env) states that nginx by default removes all environment variables when spawning workers by default - so we'd have to add the `env`
4. It separates mapping from logic, I don't mind it.

docker-nginx/templates/default.conf.template
```diff
@@ -2,6 +2,11 @@ map "$time_iso8601 # $msec" $time_iso8601_ms {
   "~([^+]+)\+([\d:]+?) # \d+?\.(\d+)" "$1.$3+$2";
 }
 
+map "${DEBUG}" $debug_mode {
+    "1" "on";
+    default "off";
```
tdrivas, line 7: Naive question: Do we need the quotes? If nginx can live without them, then we can remove them, less noise and we protect the code from potential whitespaces.

docker-nginx/templates/default.conf.template
suricactus: I think I am missing something here. Again, I comment mostly for mutual understanding, not that I am strictly against using `map`.

Reading the dockerfile for debian, I see that we copy a few files:

https://github.com/nginx/docker-nginx/blob/7f1d49f6f222f7e588a9066fd53a0ce43c3466a5/mainline/debian/Dockerfile#L134-L139

Then the entrypoint file makes sure that first the files in `/docker-entrypoint.d` before the actually nginx process is started.

https://github.com/nginx/docker-nginx/blob/7f1d49f6f222f7e588a9066fd53a0ce43c3466a5/entrypoint/docker-entrypoint.sh

One of the those files is a shell script called `20-envsubst-on-templates.sh` which runs `envsubst` to generate static files with already replaced values. Again, this happens before nginx is even started. Therefore any change to a envvar during runtime will have zero effect in my understanding.

https://github.com/nginx/docker-nginx/blob/7f1d49f6f222f7e588a9066fd53a0ce43c3466a5/entrypoint/20-envsubst-on-templates.sh#L53

Based on the above, I might misread the statement made here:

> If for whatever reason we're using the environment variable directly and it changes during nginx runtime a new worker might have a different variable available

IMO the only chance to change an envvar in the nginx container would be to first change the envvar value passed to the container and then restart the container, any other intervention in already running nginx container will have no effect.

Let me know if you spot any flaw in my logic.

---

Actually this whole QnA in this thead started because I am wondering what shall we do with other envvars we use within the nginx config:

https://github.com/opengisch/QFieldCloud/blob/fix_nginx_500_django_debug/docker-nginx/templates/default.conf.template#L58-L80

Shall we all define them in the start of the quoted file in `set` directives? What I would like to keep is the consistency, we either set all variables above, or we use them just as placeholders in the envvar, but mixing both looks confusing for future contributors to me.

suricactus (approved):
I prefer to have this merged. Still interested about the questions above.

docker-nginx/templates/default.conf.template
toebivankenoebi: @suricactus: You're absolutely on point, I took another look at it since I ran into a problem today when DEBUG is not set in the docker-compose file. I wasn't aware of the envsubst-on-templates and will look at it again and create a new PR.

suricactus: Thanks! Also please consider unifying the way envvars are used in the template, one way or the other.

PR #1246, opened 2025-05-28
feat(project): add is_attachment_download_on_demand flag

tdrivas:
This PR introduces a new field to the Project model named `is_attachment_download_on_demand`. This boolean flag controls whether attachment files are downloaded only on-demand  or bundled with the project packaging.

**Changes**

- Added `is_attachment_download_on_demand` to the Project model.
- Made the field editable via the Django Admin.
- Exposed the field as read-only in the API through ProjectSerializer.

**Admin UI**
The field is added to the project form in the admin panel to allow toggling it, but it has been intentionally left out of the list view to avoid clutter, because it's not a commonly inspected attribute across many projects and also not to impact admin UI perfromance (to be reconsidered, if needed)


docker-app/qfieldcloud/core/migrations/0084_project_is_attachment_download_on_demand.py
```diff
@@ -0,0 +1,20 @@
...
+        ("core", "0083_project_is_sticky"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="project",
+            name="is_attachment_download_on_demand",
+            field=models.BooleanField(
+                default=False,
+                help_text="If enabled, attachments are downloaded on demand instead of being included in the main data package.",
```
suricactus, line 17:
```suggestion
                help_text="If enabled, it indicates to the client (e.g. QField) that the attachments may be downloaded on demand..",
```

suricactus (commented):
Thanks!  A few small adjustments on the comments around are needed, see my suggestion and apply to others too.

tdrivas:
> Thanks! A few small adjustments on the comments around are needed, see my suggestion and apply to others too.

Thanks! I have used your suggested `help_text` as it is and then, aligned the surrounding code comments to match this approach. Commits are also squashed!

suricactus (commented):
Last comment to be resolved and we are ready to fly!

PR #1248, opened 2025-05-30
fix(sentry): do not send reports about `InvalidRangeError` to sentry

suricactus:
The client sent the wrong/unsupported HTTP `Range` header that we cannot satisfy, we cannot do much.


gounux (approved):
Thanks, smart move ! Let's keep the noise low.

PR #1251, opened 2025-05-30
fix(filestorage): allow uploading files bigger than 2GB

suricactus:
When dealing with QFieldCloud projects, especially in relation with shared datasets, it is quite normal to deal with files bigger than 2GB, such as ortophotos.

We were on the safe side, as postgres will fail with `integer out of range`, but there is no technical limitation to go to infinity.

Not the new file size limit is 8192 petabytes, good luck!


nirvn:
>> the new file size limit is 8192 petabytes, good luck!

Challenge accepted! ;)

PR #1253, opened 2025-05-30
fix(localized): do not show shared datasets message for projects without localized layers

suricactus:


duke-nyuki:
Task linked: [QF-6185 Do not show localized warning for projects without localized datasets](https://app.clickup.com/t/2192114/QF-6185)

PR #1260, opened 2025-06-13
Refactor subscription base model: extract plan lookup and add frontend-editable flag

tdrivas:
This PR adds `get_plans_for_user(...)` into the `Plan` base model and adds the `is_frontend_user_editable = False` flag to `AbstractSubscription` so that by default subscriptions are considered not editable in the web UI.

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
-from django.db.models import Case, Q, When
+from django.db.models import Case, Q, QuerySet, When
 from django.db.models import Value as V
 from django.utils import timezone
 from django.utils.translation import gettext as _
@@ -85,6 +85,25 @@ class Plan(models.Model):
         result = cls.objects.order_by("-is_default").first()
 
         return cast(Plan, result)
+
+    @classmethod
+    def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"]:
+        """
+        Return all public plans of the given `user_type`, filtering out
+        trial plans for organizations when no trials remain.
+        """
+        filters = {
+            "is_public": True,
+            "user_type": user_type,
+        }
+
+        if (
+            user_type == User.Type.ORGANIZATION
+            and user.remaining_trial_organizations == 0
+        ):
+            filters["is_trial"] = False
+
+        return cls.objects.filter(**filters)
 
     # unique identifier of the subscription plan
     code = models.CharField(max_length=100, unique=True)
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/subscription/models.py
```diff
@@ -86,6 +86,22 @@ def get_or_create_default(cls) -> "Plan":
 
         return cast(Plan, result)
 
+    @classmethod
+    def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"]:
+        """
+        Return all public plans of the given user_type, filtering out
+        trial plans for organizations when no trials remain.
+        """
+        filters = {"is_public": True, "user_type": user_type}
```
```suggestion
        filters = {
            "is_public": True,
            "user_type": user_type,
        }
```

Avoid dicts with more than one key per line.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -86,6 +86,22 @@ def get_or_create_default(cls) -> "Plan":
 
         return cast(Plan, result)
 
+    @classmethod
+    def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"]:
+        """
+        Return all public plans of the given user_type, filtering out
```
```suggestion
        Return all public plans of the given `user_type`, filtering out
```
Whenever we refer to something more programattic, we always try to put it in backticks.

suricactus (approved):
Thanks!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Unfortunately, `pre-commit` and particulary `ruff` forces the dict to be in one line

suricactus: I am curious what started causing `ruff` to force the single line behaviour.

Can you check if by any change you have changed your ruff config file?

In `black`, `ruff` and `prettier` there is this setting https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma that controls single line behaviour.

I am also wondering why the CI is no failing, since we have so many examples like this, see:
- https://github.com/opengisch/QFieldCloud/blob/6cec344091a73db24e0ee9486e9638ca241711a7/docker-app/qfieldcloud/authentication/tests/test_list_auth_providers.py#L60-L62
- https://github.com/opengisch/QFieldCloud/blob/6cec344091a73db24e0ee9486e9638ca241711a7/docker-app/qfieldcloud/core/pagination.py#L35-L37
- https://github.com/opengisch/QFieldCloud/blob/6cec344091a73db24e0ee9486e9638ca241711a7/docker-app/qfieldcloud/core/templates/admin/project_files_widget.html#L149-L151
- https://github.com/opengisch/QFieldCloud-Nyuki/blob/bcb18920185aca5e9ed464fe203edf4db6c3156f/docker-app/qfieldcloud/billing/views/billing_views.py#L1256-L1258
and more.

Search in the repositorie(s) you contribute for "avoid kwargs and dicts on single line." and you will find at least one example (there are more, I just don't have the keywords to find them) where a reviewer commented for the same and you did not report any problem back then.

tdrivas: Thanks for catching this! It turns out that the dict will be reversed back to one line unless there’s a trailing comma after the last pair {key:value}. I have added the missing comma after `"user_type": user_type`, so the block stays in multiline form!

PR #1262, opened 2025-06-24
refactor(filestorage): rename files to all_files and add `project_files` filtered by `File.FileType.PROJECT_FILE`

suricactus:
Basically reduce the confusion when writing code which are the project files and which are the package files.

PR #1264, opened 2025-06-25
fix(filestorage): filter by file type when interacting with project files

suricactus:
`Project.files` is returning both PROJECT_FILE and PACKAGE_FILE

This PR is a temporary fix with minimal diff.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -39,7 +39,7 @@ from qfieldcloud.core.utils2 import storage
 from qfieldcloud.subscription.exceptions import ReachedMaxOrganizationMembersError
 
 if TYPE_CHECKING:
-    from qfieldcloud.filestorage.models import File
+    from qfieldcloud.filestorage.models import File, FileQueryset
 
 
 SHARED_DATASETS_PROJECT_NAME = "shared_datasets"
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/models.py
```diff
@@ -1101,7 +1101,7 @@ class Meta:
             )
         ]
 
-    files: "models.QuerySet[File]"
+    files: "FileQueryset"
```
gounux, line 1104: Naive question : if ones faces this generic, shall it be replaced _everywhere where it is found_ by the Model's QuerySet ?

gounux (approved):
Code looks good to me, thanks for the catch.

> This PR is a temporary fix with minimal diff.

Feeling curious about what the next step would be regarding this :wink:

suricactus:
> > This PR is a temporary fix with minimal diff.
> 
> Feeling curious about what the next step would be regarding this 😉

Probably something like this https://github.com/opengisch/QFieldCloud/pull/1262

docker-app/qfieldcloud/core/models.py
suricactus: This is the way to make typing aware that `Project().files` has the methods defined in the `FileQueryset` + everything that comes from `QuerySet[File]`.

PR #1265, opened 2025-06-25
fix(filestorage): show the correct number of project files instead of all associated files

suricactus:


duke-nyuki:
Task linked: [QF-6249 Fix project count that also counts packaged files](https://app.clickup.com/t/2192114/QF-6249)

PR #1266, opened 2025-06-26
feat(api): support search by project name or owner username in /projects

suricactus:
Verified that there shouldn't be any performance issues as indices are used.

The search is performed as substring search after converting to UPPER().

See https://www.django-rest-framework.org/api-guide/filtering/#searchfilter


duke-nyuki:
Task linked: [QF-4682 Add ability to search by substring for project owner's username, project's name or project's description](https://app.clickup.com/t/2192114/QF-4682)

gounux (approved):
Looks good to me, thanks.

PR #1269, opened 2025-06-27
Shared Datasets: do not warn about missing datasets when they do exist

tdrivas:
This PR resolved the issue with the warning about missing shared datasets in the project, even when the datasets were present and accessible.

**Scenario**
A project has been created in legacy storage mode
The dedicated `shared_datasets` project has been created in default storage mode

**Solution**
As the dedicated `shared_datasets` project is always created via default storage, all its files must be retrieved via the `file` model.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_localized_projects.py
+++ b/docker-app/qfieldcloud/core/tests/test_localized_projects.py
@@ -40,10 +40,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
             name=SHARED_DATASETS_PROJECT_NAME,
             is_public=False,
             owner=self.user1,
-        )
-
-        Project.objects.filter(pk=self.shared_datasets_project.pk).update(
-            file_storage="default"
+            file_storage="default",
         )
 
         self.shared_datasets_project.refresh_from_db()
@@ -139,9 +136,9 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
 
         # Localized layers found in the localized datasets project
         available_localized_filenames = set(
-            File.objects.filter(project_id=self.shared_datasets_project.id).values_list(
-                "name", flat=True
-            )
+            File.objects.filter(
+                project_id=self.shared_datasets_project.id,
+            ).values_list("name", flat=True)
         )
 
         self.assertEqual(len(available_localized_filenames), 1)
--- a/docker-app/qfieldcloud/core/tests/test_localized_projects.py
+++ b/docker-app/qfieldcloud/core/tests/test_localized_projects.py
@@ -10,7 +10,6 @@ from qfieldcloud.core.models import (
     Job,
     Person,
     Project,
-    utils,
 )
 from qfieldcloud.core.tests.mixins import QfcFilesTestCaseMixin
 from qfieldcloud.core.tests.utils import (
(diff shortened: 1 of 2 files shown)


suricactus (approved):
Good catch, thanks!

However, the CI is still failing?

docker-app/qfieldcloud/core/tests/test_localized_projects.py
```diff
@@ -37,9 +38,17 @@ def setUp(self):
...
 
         self.shared_datasets_project = Project.objects.create(
-            name=SHARED_DATASETS_PROJECT_NAME, is_public=False, owner=self.user1
+            name=SHARED_DATASETS_PROJECT_NAME,
+            is_public=False,
+            owner=self.user1,
         )
 
+        Project.objects.filter(pk=self.shared_datasets_project.pk).update(
+            file_storage=get_project_file_storage_default()
```
suricactus: I am a bit confused what is happeniing here:

1) why do you create the project 2 lines above and update right away? Why don't you create the project with the desired value already so there is no need to update?
2) why do you use `get_project_file_storage_default` which will return `"legacy_storage"` in the legacy storage tests? Shall we just hardcode to `"default"`?

suricactus (changes requested):
I think we are no there yet. Please pass `file_storage="default"` to the `create` of the shared datasets project as we discussed earlier, this shall make the tests pass.

docker-app/qfieldcloud/core/tests/test_localized_projects.py
tdrivas: Thanks for the comment!

1. To avoid running all  the model’s `save()` logic  (which may slow down the tests)
2. I had already pushed some changes to this direction with the hard-coded storage value.

suricactus: 1. Can you elaborate how adding a new field in the `save` method may slow down the tests?
2. ok, thanks!

docker-app/qfieldcloud/core/tests/test_localized_projects.py
```diff
@@ -38,9 +37,17 @@ def setUp(self):
...
         self.shared_datasets_project = Project.objects.create(
-            name=SHARED_DATASETS_PROJECT_NAME, is_public=False, owner=self.user1
+            name=SHARED_DATASETS_PROJECT_NAME,
+            is_public=False,
+            owner=self.user1,
         )
 
+        Project.objects.filter(pk=self.shared_datasets_project.pk).update(
+            file_storage="default"
+        )
```
suricactus:
```suggestion
            owner=self.user1,
            file_storage="default",
        )
```

docker-app/qfieldcloud/core/tests/test_localized_projects.py
```diff
@@ -131,20 +138,11 @@ def test_localized_layers_and_missing_localized_layers(self):
...
-        else:
-            available_localized_filenames = set(
-                File.objects.filter(
-                    project_id=self.shared_datasets_project.id
-                ).values_list("name", flat=True)
+        available_localized_filenames = set(
+            File.objects.filter(project_id=self.shared_datasets_project.id).values_list(
+                "name", flat=True
             )
+        )
```
suricactus:
```suggestion
        available_localized_filenames = set(
            File.objects.filter(
                project_id=self.shared_datasets_project.id,
            ).values_list("name", flat=True)
        )
```

In those cases we try to make the filter line by line.

docker-app/qfieldcloud/core/tests/test_localized_projects.py
tdrivas: 1. if we pass a value for `file_storage` it overrides the default value. That means all of the model logic fires, for every field, a series of action such as `clean`, extra queries etc. On the contrary, if we stick to the default value and then try to update it , it should only make a query directly to db without triggering the model’s `save()` method.

PR #1270, opened 2025-06-28
Improve attachment on demand help description

nirvn:
Trying to improve the wording a bit to ease understand. Using "should" instead of "may" to make it clear it's a behavior that will be enforced. I think we don't need the QField client bit, removing it makes for a shorter and clearer reading IMHO.

### DIFF
--- a/docker-app/qfieldcloud/core/migrations/0084_project_is_attachment_download_on_demand.py
+++ b/docker-app/qfieldcloud/core/migrations/0084_project_is_attachment_download_on_demand.py
@@ -14,7 +14,8 @@ class Migration(migrations.Migration):
             name="is_attachment_download_on_demand",
             field=models.BooleanField(
                 default=False,
-                help_text="If enabled, it indicates to the client (e.g. QField) that the attachments may be downloaded on demand.",
+                verbose_name="On demand attachment files download",
+                help_text="If enabled, attachment files should be downloaded on demand with QField.",
             ),
         ),
     ]
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1285,7 +1285,7 @@ class Project(models.Model):
     is_attachment_download_on_demand = models.BooleanField(
         default=False,
         help_text=_(
-            "If enabled, it indicates to the client (e.g. QField) that the attachments may be downloaded on demand."
+            "If enabled, attachment files should be downloaded on demand on QField."
         ),
     )
 
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1284,8 +1284,9 @@ class Project(models.Model):
     # projects with large or numerous attachments.
     is_attachment_download_on_demand = models.BooleanField(
         default=False,
+        verbose_name=_("On demand attachment files download"),
         help_text=_(
-            "If enabled, attachment files should be downloaded on demand on QField."
+            "If enabled, attachment files should be downloaded on demand with QField."
         ),
     )
 


nirvn:
@suricactus , will > should > may > might > meh :) 

I do understand that on demand will not always happen (as you said, old QField version), but I feel "may" was a bit too vague. I do think the word _should_ makes the intent clearer. While not all clients will, all clients should :)

Regarding the CLI, it's definitively not a priority there as you wouldn't meet desktop devices that need to care about storage space, etc.

Will edit the previous migration, no problem. I was wondering what to do here, good to know you prefer that.

nirvn:
@suricactus , should we do a bit of a poll internally? Otherwise, if you still feel you'd rather stick with the old wording, a 50/50 means we go for meritocracy, and your voice wins :)

suricactus:
Good with "should". 

> Will edit the previous migration, no problem. I was wondering what to do here, good to know you prefer that.

This is valid only to a very small subset of situations, like changing `help_text`, `verbose_name` and few others on a single field. The approach you took *should* be the default indeed. If you had multiple fields/models to migrate, and some of them affect the database, then no need to do the trick I asked you to save a migration.

---

Re-reading the new message, I would use the phrase "attachment files" +1 and drop "indicates" +0.5, as it gives little information:

> If enabled, attachment files should be downloaded on demand.

The above message is still a bit abstract for someone who is not fully familiar with the QField ecosystem, so I am slight +0.5 to put QField there:

> If enabled, attachment files should be downloaded on demand on QField.

or

> If enabled, attachment files should be downloaded only when explicitly requested on QField.

I don't think I can write a longer comment for a single sentence, so now I will be happy with anything you propose. :)

nirvn:
@suricactus , I went for:  _If enabled, attachment files should be downloaded on demand on QField_. Commit updated.

nirvn:
@suricactus , I went for "on demand attachment files download"

docker-app/qfieldcloud/core/migrations/0084_project_is_attachment_download_on_demand.py
```diff
@@ -12,9 +12,10 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name="project",
             name="is_attachment_download_on_demand",
+            verbose_name="On demand attachment files download",
             field=models.BooleanField(
                 default=False,
-                help_text="If enabled, it indicates to the client (e.g. QField) that the attachments may be downloaded on demand.",
+                help_text="If enabled, attachment files should be downloaded on demand with QField.",
             ),
         ),
```
suricactus:
```suggestion
        migrations.AddField(
            model_name="project",
            name="is_attachment_download_on_demand",
            field=models.BooleanField(
                default=False,
                help_text="If enabled, attachment files should be downloaded on demand with QField.",
                verbose_name="On demand attachment files download",
            ),
        ),
```

`verbose_name` should be within the field itself. I verified by looking at this one: https://github.com/opengisch/QFieldCloud/blob/25b43fcb25337957fc479eadc790f68dc1a48212/docker-app/qfieldcloud/core/migrations/0027_auto_20210408_0138.py#L18

docker-app/qfieldcloud/core/models.py
```diff
@@ -1284,8 +1284,9 @@ class Meta:
     # projects with large or numerous attachments.
     is_attachment_download_on_demand = models.BooleanField(
         default=False,
+        verbose_name="On demand attachment files download",
```
```suggestion
        verbose_name=_("On demand attachment files download"),
```

Translation missing.

docker-app/qfieldcloud/core/migrations/0084_project_is_attachment_download_on_demand.py
nirvn: TIL

docker-app/qfieldcloud/core/models.py
nirvn: @suricactus , _( == tr( ? TIL again :)

suricactus: `_` is quite common `gettext` alias in many languages, especially within Django/Python.

Basically Qt's `tr`, yes :)

suricactus:
Feel free to force push as single commit before or after the tests are happy again :)

PR #1271, opened 2025-06-28
Introduce concept of data directories

nirvn:
Since we now have on-demand attachment download, using the attachment directories' trick to ship additional project assets (images used in symbology, print/atlas layouts, and decorators as well as files used by project plugin) has become unreliable. If a given project is set to on demand, the necessary assets will not be downloaded by QField.

A good solution is to stop abusing attachment directories and come up with a new directory type, which I've named 'data'. This is how it looks in QFieldSync when configuring directories:

image

_For the QFieldSync side of things, see https://github.com/opengisch/qfieldsync/pull/685_

@suricactus , as discussed over chat a while back.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1464,9 +1464,6 @@ class Project(models.Model):
         Attachment dir is a special directory in the QField infrastructure that holds attachment files
         such as images, pdf etc. By default "DCIM" is considered a attachment directory.
 
-        TODO this function expects whether `attachment_dirs` key in project_details. However,
-        neither the extraction from the projectfile, nor the configuration in QFieldSync are implemented.
-
         Returns:
             A list configured attachment dirs for the project.
         """
@@ -1479,6 +1476,26 @@ class Project(models.Model):
             attachment_dirs = ["DCIM"]
 
         return attachment_dirs
+
+    @property
+    def data_dirs(self) -> list[str]:
+        """Returns a list of configured data dirs for the project.
+
+        Data dir is a special directory in the QField infrastructure that holds assets
+        used by the project symbology, layouts, or project plugins.
+
+        Unlike `attachmentDirs`, the `dataDirs` should always be served as a undivisible
+        part of project files.
+
+        Returns:
+            A list configured data dirs for the project.
+        """
+        data_dirs = []
+
+        if self.project_details and self.project_details.get("data_dirs"):
+            data_dirs = self.project_details.get("data_dirs", [])
+
+        return data_dirs
 
     @property
     def has_attachments_files(self) -> bool:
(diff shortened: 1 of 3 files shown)


docker-app/qfieldcloud/core/models.py
```diff
@@ -1480,6 +1480,26 @@ def attachment_dirs(self) -> list[str]:
...
 
+    @property
+    def data_dirs(self) -> list[str]:
+        """Returns a list of configured data dirs for the project.
+
+        Data dir is a special directory in the QField infrastructure that holds assets
+        used by the project symbology, layouts, or project plugins.
+
+        TODO this function expects whether `dataDirs` key in project_details. However,
+        neither the extraction from the projectfile, nor the configuration in QFieldSync are implemented.
```
suricactus: Not sure if I follow the TODO here? neither the one that you copied from the `attachmentDirs`, but does it make sense here?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1480,6 +1480,26 @@ def attachment_dirs(self) -> list[str]:
 
         return attachment_dirs
 
+    @property
+    def data_dirs(self) -> list[str]:
+        """Returns a list of configured data dirs for the project.
+
+        Data dir is a special directory in the QField infrastructure that holds assets
+        used by the project symbology, layouts, or project plugins.
```
suricactus, line 1485:
```suggestion
        Data dir is a special directory in the QField infrastructure that holds assets
        used by the project symbology, layouts, or project plugins.
        Unlike `attachmentDirs`, the `dataDirs` should always be served as a undivisible part of project files.
```

nirvn: @suricactus , good addition

docker-app/qfieldcloud/core/models.py
```diff
@@ -1464,7 +1464,7 @@ def attachment_dirs(self) -> list[str]:
         Attachment dir is a special directory in the QField infrastructure that holds attachment files
         such as images, pdf etc. By default "DCIM" is considered a attachment directory.
 
-        TODO this function expects whether `attachment_dirs` key in project_details. However,
+        TODO this function expects whether `attachmentDirs` key in project_details. However,
```
suricactus: Why the change? `attachment_dirs` is the name of the key in the `Project.project_details` JSON? See https://github.com/opengisch/QFieldCloud/pull/1271/files#diff-c5c429cde51e37e7ee4197d0dcf8d50d463ed5b560a6945dbb37504bddf2a74fR118

Actually I believe the TODO here is outdated.

nirvn: @suricactus , should I remove the TODO for both attachments and data dirs?

suricactus: Please do, yes. And squsah all into a single commit.

nirvn: @suricactus , TODOs removed.

nirvn:
@suricactus , squashed.

suricactus (approved):
Thanks!

nirvn:
@suricactus , thanks for the reviews, much appreciated.

PR #1273, opened 2025-07-02
fix(packaging): handle `needs_repackaging` on project API

gounux:
Following-up #1235

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -75,6 +75,7 @@ class ProjectSerializer(serializers.ModelSerializer):
     user_role_origin = serializers.CharField(read_only=True)
     private = serializers.BooleanField(allow_null=True, default=None)
     shared_datasets_project_id = serializers.SerializerMethodField(read_only=True)
+    needs_repackaging = serializers.SerializerMethodField()
 
     def get_shared_datasets_project_id(self, obj: Project) -> str | None:
         if obj.shared_datasets_project:
@@ -125,7 +126,12 @@ class ProjectSerializer(serializers.ModelSerializer):
         return data
 
     def get_needs_repackaging(self, obj: Project) -> bool:
-        return obj.needs_repackaging(self.request.user)  # type: ignore[attr-defined]
+        request = self.context.get("request")
+
+        if request:
+            return obj.needs_repackaging(request.user)  # type: ignore[attr-defined]
+
+        return False
 
     class Meta:
         fields = (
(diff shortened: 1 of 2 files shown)


nirvn:
@gounux , @suricactus , it did heal things locally.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -719,3 +719,17 @@ def test_get_missing_localized_layers_method(self):
             self.assertEqual(
                 missing_layers[0]["filename"], "localized:missing_layer.tif"
             )
+
+    def test_get_project_details_include_needs_repackaging(self):
```
nirvn: @gounux , in QField, we have this conditional block that needs passing to download a project when getting a project detail (i.e. /api/v1/projects/{project_uuid}/

```c++
    if (
      !projectData.value( "name" ).isString()
      || !projectData.value( "owner" ).isString()
      || !projectData.value( "description" ).isString()
      || !projectData.value( "user_role" ).isString()
      || !projectData.value( "is_public" ).isBool()
      || !projectData.value( "can_repackage" ).isBool()
      || !projectData.value( "needs_repackaging" ).isBool() )
    {
      emit dataRefreshed( reason, tr( "project(%1) trigger response refresh not contain all the expected keys: name(string), owner(string), description(string), user_role(string), is_public(bool), can_repackage(bool), needs_repackaging(bool)" ).arg( projectId ) );
      return;
    }
```

I'm wondering if we should have a test that covers all of these keys just to be super safe on that front.

gounux: @nirvn sure, I guess that's the purpose of the test. Let me add those keys.

nirvn: @gounux , indeed, it was the purpose :)

nirvn:
@gounux , nice, thanks for the tweaked test, it'll save us some unfortunate surprises down the line :)

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -125,7 +126,12 @@ def validate(self, data):
         return data
 
     def get_needs_repackaging(self, obj: Project) -> bool:
-        return obj.needs_repackaging(self.request.user)  # type: ignore[attr-defined]
+        request = self.context.get("request")
+
+        if request:
+            return obj.needs_repackaging(request.user)  # type: ignore[attr-defined]
```
suricactus, line 132: Is this type ignore still needed?

suricactus:
Merging, but still interested in the answer.

PR #1274, opened 2025-07-02
feat(auth): create adapters to determine if signup is open

gounux:
Not much more than the title : override allauth's `is_open_for_signup` with different classes overriding the previous adapter, determining if sign-up is open or not.

### DIFF
--- a/docker-app/qfieldcloud/core/adapters.py
+++ b/docker-app/qfieldcloud/core/adapters.py
@@ -134,6 +134,7 @@ class AccountAdapterSignUpClosed(AccountAdapter):
     """Account adapter for closed signup.
 
     This adapter is used when the signup is closed, i.e. when users cannot register themselves.
+    A user can still be added via Django admin.
     """
 
     def is_open_for_signup(self, request: HttpRequest) -> Literal[False]:


duke-nyuki:
Task linked: [QF-6264 Add a setting to disable sign-up to QFieldCloud on dedicated](https://app.clickup.com/t/2192114/QF-6264)

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -118,6 +121,15 @@ def send_confirmation_mail(
...
         super().send_confirmation_mail(request, email_confirmation, signup)
 
+    def is_open_for_signup(self, request: HttpRequest) -> Literal[True]:
+        """
+        Overrides allauth's default method for checking if signup is open.
+        Returns True if a new user can register, False otherwise.
+        """
+        model = apps.get_model(settings.QFIELDCLOUD_SUBSCRIPTION_MODEL)
+
+        return model.is_frontend_user_editable
```
suricactus: I would use something different here as a flag. If we keep your proposal, it will make all installations of this QFieldCloud inaccessible by the general public.

What if we expose `ACCOUNT_ADAPTER` as envvar and make two different classes out of `AccountAdapter`:

```
class AccountAdapterSingUpOpen: ...
class AccountAdapterSingUpClosed: ...
```

I think this is the cleanest solution?

gounux: Thanks for pointing this solution, I agree it's more convenient. Let me work on that.

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -119,6 +120,26 @@ def send_confirmation_mail(
...
+    """
+
+    def is_open_for_signup(self, request: HttpRequest) -> Literal[True]:
+        return True
+
+
+class AccountAdapterSignUpClosed(AccountAdapter):
+    """Account adapter for closed signup.
+
+    This adapter is used when the signup is closed, i.e. when users cannot register themselves.
```
suricactus, line 136:
```suggestion
    This adapter is used when the signup is closed, i.e. when users cannot register themselves.
    A user can still be added via Django admin.
```

suricactus (approved):

PR #1275, opened 2025-07-03
Improve the has_restricted_projectfiles display name and description

nirvn:
Because projectfiles is not an actual word :wink: 

Before (taken from app.qfield.cloud):
image

PR:
image

### DIFF
--- a/docker-app/qfieldcloud/core/migrations/0076_project_restrict_project_modification.py
+++ b/docker-app/qfieldcloud/core/migrations/0076_project_restrict_project_modification.py
@@ -14,7 +14,8 @@ class Migration(migrations.Migration):
             name="has_restricted_projectfiles",
             field=models.BooleanField(
                 default=False,
-                help_text="Restrict modifications of QGIS/QField projectfiles to managers and administrators.",
+                verbose_name="Restrict project files",
+                help_text="If enabled, modifications of QGIS project configuration (.qgs, qgz, qgd) and QField project plugins files will be restricted to managers and administrators.",
             ),
         ),
     ]
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1189,8 +1189,9 @@ class Project(models.Model):
 
     has_restricted_projectfiles = models.BooleanField(
         default=False,
+        verbose_name=_("Restrict project files"),
         help_text=_(
-            "Restrict modifications of QGIS/QField projectfiles to managers and administrators."
+            "If enabled, modifications of QGIS project configuration (.qgs, qgz, qgd) and QField project plugins files will be restricted to managers and administrators."
         ),
     )
 


docker-app/qfieldcloud/core/migrations/0076_project_restrict_project_modification.py
```diff
@@ -14,7 +14,8 @@ class Migration(migrations.Migration):
             name="has_restricted_projectfiles",
             field=models.BooleanField(
                 default=False,
-                help_text="Restrict modifications of QGIS/QField projectfiles to managers and administrators.",
+                verbose_name="Restrict project files",
+                help_text="If enabled, modifications of QGIS/QField project files (.qgs, qgz) will be restricted to to managers and administrators.",
```
gounux: @nirvn, nice words ! I believe to change this, rather than editing the existing migration, wouldn't we favor the creation of a new one ?

nirvn: @gounux , see https://github.com/opengisch/QFieldCloud/pull/1270#issuecomment-3015787242 for background on why I did it that way.

@suricactus , ping :)

suricactus: Yeah, for verbose name and help text changes like this we can afford modifying the existing migration.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1189,8 +1189,9 @@ class Meta:
 
     has_restricted_projectfiles = models.BooleanField(
         default=False,
+        verbose_name=_("Restrict project files"),
         help_text=_(
-            "Restrict modifications of QGIS/QField projectfiles to managers and administrators."
+            "If enabled, modifications of QGIS/QField project files (.qgs, qgz) will be restricted to to managers and administrators."
```
```suggestion
            "If enabled, modifications of QGIS/QField project files (.qgs, qgz) will be restricted to managers and administrators."
```

Double "to".

Also, this is a bit more than the label, see what it actually implies here: https://github.com/opengisch/QFieldCloud/blob/eb06ddda73f805b7502f216bcd521cd0ba137335/docker-app/qfieldcloud/filestorage/utils.py#L69-L108



```suggestion
            "If enabled, modifications of QGIS/QField funcitonal files (.qgs, .qgd) will be restricted to managers and administrators of the project."
```

nirvn: qmd ? you mean project plugin files? that'd be QML. 

I feel "functional files" would be a bit confusing here, @m-kuhn , do you have a better idea?

suricactus: Sorry, no ~qmd~, qgd. My bad, updated the comment above.

Happy to adjust "functional files" to something better indeed. I just find QGIS project files migth be confusing, as I am not sure how many people are aware that sidecar files are also included.

nirvn:
@suricactus , I've further adjusted the wording inspired by your suggestion. I also took the time to broaden the project plugin file check to include _all_ QML files (e.g. search locator integration plugins have at least two QML files) as well as Javascript files.

suricactus (approved):
Thanks! Shall we merge?

nirvn:
@suricactus , ya, let's do it, thanks for the review.

PR #1277, opened 2025-07-03
fix(core): switch collaborator field to SlugRelatedField using username

tdrivas:
This PR solved the issue of adding a collaborator, which caused a KeyError as it was reading directly from request.data["collaborator"]. 

**Changes**
`ProjectCollaboratorSerializer` has been updated so that the collaborator field is now a writable `SlugRelatedField` instead of a read-only `StringRelatedField`.


duke-nyuki:
Task linked: [QF-6257 Make sure the collaborator_views in Classic are using the serializer and not raw data](https://app.clickup.com/t/2192114/QF-6257)

suricactus (approved):
Great, thanks!

PR #1278, opened 2025-07-04
Fix file storage bytes help description in the project admin page

nirvn:
Just noticed that the project admin page has a wrong description string for the file storage bytes field:

image

This PR fixes the situation.


nirvn:
@suricactus , do we? It's defined here: https://github.com/opengisch/QFieldCloud/blob/ca65774876baa6580f90cc2a74353b0c44b08caa/docker-app/qfieldcloud/core/models.py#L1217-L1225

nirvn:
@suricactus , oh sorry, my bad. I was rushed by time, couldn't come up with an after screenshot.

PR #1279, opened 2025-07-07
fix(files): make files widget download the latest version when no version is selected

gounux:
Basically the title. Considering we'll always have a file version having the `is_latest` boolean flag to true.

image


suricactus (approved):
Ah, this is great catch. I am not sure about the longevity of the `is_latest` field,  but good for now!

PR #1280, opened 2025-07-08
fix(secrets): make the project field nullable in secret admin

gounux:
Secrets are not strongly related to a Project now, since there are also Organization-assigned Secrets. This PR makes the project nullable in the Admin form for Secrets.

| Before | After |
| --- | --- |
| image | image |

Migration-less approach considered here.


gounux:
> Can you now save a secret without a required organization field?

Yes, tested locally. I believe that is possible since there is the `blank=True` in the `organization` field, on Secret model level.

-> https://github.com/opengisch/QFieldCloud/blob/e46fffc984ec224f67ab6cc16427b4543fe52862/docker-app/qfieldcloud/core/models.py#L2481

PR #1281, opened 2025-07-09
feat(storage): ensure a FileVersion's storage depends on the related file storage

gounux:
This PR ensures a FileVersion's target storage is the one of the related File.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/tests/test_attachments.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_attachments.py
@@ -1,6 +1,8 @@
 import io
 import logging
 
+from django.conf import settings
+from django.core.files.storage import storages
 from django.http import FileResponse
 from rest_framework import status
 from rest_framework.test import APITransactionTestCase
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6330 Set a FileVersion's filestorage depending on if the related file is an attachment](https://app.clickup.com/t/2192114/QF-6330)

docker-app/qfieldcloud/filestorage/tests/test_attachments.py
```diff
@@ -37,6 +37,20 @@ def setUp(self):
...
+
+        Args:
+            file: The File instance to check.
+            file_storage_name: The name of the storage to check against.
+        """
+        self.assertEqual(file.file_storage, file_storage_name)
+
+        for version in file.versions.all():
+            self.assertEqual(version._get_file_storage_name(), file_storage_name)
+            self.assertTrue(version.content.storage.exists(version.content.name))
```
suricactus:
```suggestion
            self.assertEqual(version._get_file_storage_name(), file_storage_name)
            self.assertEqual(version.content.storage.name, file_storage_name)
            self.assertTrue(version.content.storage.exists(version.content.name))
```

I don't know the API exactly, but this is another level of paranoism which is good IMO.

docker-app/qfieldcloud/filestorage/tests/test_attachments.py
```diff
@@ -37,6 +37,20 @@ def setUp(self):
             attachments_file_storage="webdav",
         )
 
+    def _check_file_exists_on_storage(self, file: File, file_storage_name: str) -> None:
```
suricactus:
```suggestion
    def assertFileOnStorage(self, file: File, file_storage_name: str) -> None:
```

I think we can even do it like this?

docker-app/qfieldcloud/filestorage/tests/test_attachments.py
gounux: Also tried this ^^  
But I'm afraid there is no such method for the `Storage` class.

Maybe check the type ?  
Or add such a property to our Storage classes ?

suricactus: `for storage_name, storage_object in django.core.files.storage.storages.backends: print(storage_object == version.content.storage)`?

gounux: I don't think there is such things, but maybe I did not check all the places.

`self.assertIn(version._get_file_storage_name(), settings.STORAGES.keys())` ?

PR #1282, opened 2025-07-10
feat(auth): add `ClientTypeMiddleware` to set client_type

gounux:
This PR adds a Middleware that attaches the client type (e.g. QField, QFieldSync, Worker, SDK, Browser, etc.) to the session.


duke-nyuki:
Task linked: [QF-5865 [SSO] fix missing request.auth.client_type](https://app.clickup.com/t/2192114/QF-5865)

docker-app/qfieldcloud/core/middleware/client_type.py
```diff
@@ -0,0 +1,29 @@
...
+    This client_type is required by Views later on during the QField/Sync process.
+
+    Typically used for users connected with SSO, which don't have any `AuthToken`
+    attached to the session, but future requests still need to know the client type.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
```
lukasgraf: A type hint like `request: HttpRequest` could be used here (`from django.http import HttpRequest`).

docker-app/qfieldcloud/core/middleware/client_type.py
```diff
@@ -0,0 +1,29 @@
...
+    This middleware detects the client type based on the user agent
+    and sets it on the session object.
+
+    This client_type is required by Views later on during the QField/Sync process.
+
+    Typically used for users connected with SSO, which don't have any `AuthToken`
+    attached to the session, but future requests still need to know the client type.
+    """
+
+    def __init__(self, get_response):
```
suricactus: typing

gounux:
Thanks for the review @lukasgraf @suricactus, I guess we're not in a hurry to get this merged.

docker-app/qfieldcloud/authentication/views.py
```diff
@@ -62,6 +62,7 @@ def post(self, request, *args, **kwargs):
         self.token = create_token(
             self.token_model, validated_data["user"], self.serializer, self.request
         )
+        request.session["client_type"] = self.token.client_type
```
suricactus: @gounux if we have `ClientTypeMiddleware` enabled that sets `request.session["client_type"]` already, do we need this line?

gounux: Good question, thanks ! Extra safety here I presume, better stay with less diff, I agree ;)

Let me address this.

docker-app/qfieldcloud/core/middleware/client_type.py
```diff
@@ -0,0 +1,33 @@
...
+        if hasattr(request, "auth") and hasattr(request.auth, "client_type"):
+            request.session["client_type"] = request.auth.client_type
+
+            return self.get_response(request)
+
+        if "client_type" not in request.session:
+            user_agent = request.headers.get("user-agent", "")
+            request.session["client_type"] = AuthToken.guess_client_type(user_agent)
+
+        return self.get_response(request)
```
tdrivas:
```suggestion
    def __call__(self, request: HttpRequest) -> HttpResponse:
          if getattr(request, "auth", None) and hasattr(request.auth, "client_type"):
             client_type = request.auth.client_type
          else: 
              client_type = request.session.get("client_type", AuthToken.guess_client_type(request.headers.get("user-agent", ""))
      )
  
      if request.session.get("client_type") != client_type:
          request.session["client_type"] = client_type
  
      return self.get_response(request)    
          
```

Maybe something as the above in order to simplify the logic.

gounux: Thanks @tdrivas ! On line n°30 of the proposed suggestion, what happens if the session has no `client_type` ?

tdrivas: It gives you `None != client_type` where `client_type` does have value (even if it comes from `guess_client_type`) from the above if-else and therefore it stores the new value in the session.

gounux: Alright, thanks ! Addressed in next commit.

gounux:
Squashed and rebased on latest master, merging this. Thanks for the reviews :)

PR #1283, opened 2025-07-10
feat(status): add health check for database and configured storages

gounux:
This PR adds a health-check status for every configured storage in the `/status` endpoint, as well as one for the database.

Before, the server was responding a json like : `{"geodb":"ok","storage":"ok"}`

This PR makes QFC respond a json like : `{"database": "ok", "storage": "ok"}`

- No more `geodb` check
- Each storage is checked based on its type (S3, webdav, etc.), one single `storage` key keeps being answered by the API.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_api.py
+++ b/docker-app/qfieldcloud/core/tests/test_api.py
@@ -36,9 +36,16 @@ class QfcTestCase(APITransactionTestCase):
 
     def test_api_status(self):
         response = self.client.get("/api/v1/status/")
+
         self.assertTrue(status.is_success(response.status_code))
-        self.assertEqual(response.json()["storage"], "ok")
-        self.assertEqual(response.json()["geodb"], "ok")
+
+        data = response.json()
+
+        self.assertIn("database", data)
+        self.assertEqual(data["database"], "ok")
+
+        self.assertIn("storage", data)
+        self.assertEqual(data["storage"], "ok")
 
     def test_api_status_cache(self):
         tic = time.perf_counter()
--- a/docker-app/qfieldcloud/core/views/status_views.py
+++ b/docker-app/qfieldcloud/core/views/status_views.py
@@ -1,6 +1,7 @@
 from django.core.cache import cache
+from django.core.files.storage import storages
+from django.db import connections
 from drf_spectacular.utils import extend_schema, extend_schema_view
-from qfieldcloud.core import geodb_utils, utils
 from rest_framework import status, views
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
@@ -16,17 +17,32 @@ class APIStatusView(views.APIView):
         # Try to get the status from the cache
         results = cache.get("status_results", {})
         if not results:
-            results["geodb"] = "ok"
-            # Check geodb
-            if not geodb_utils.geodb_is_running():
-                results["geodb"] = "error"
-
-            results["storage"] = "ok"
-            # Check if bucket exists (i.e. the connection works)
+            # check database
+            results["database"] = "ok"
+            db_conn = connections["default"]
             try:
-                utils.get_s3_bucket()
+                with db_conn.cursor() as cursor:
+                    cursor.execute("SELECT 1")
             except Exception:
-                results["storage"] = "error"
+                results["database"] = "error"
+
+            # check storages
+            results["storage"] = "ok"
+            for storage_key in storages.backends.keys():
+                storage = storages[storage_key]
+
+                if not hasattr(storage, "check_status"):
+                    # Some storages may not have the check_status method.
+                    # e.g. ManifestStaticFilesStorage.
+                    continue
+
+                try:
+                    if not storage.check_status():
+                        results["storage"] = "error"
+                        break
+                except Exception:
+                    results["storage"] = "error"
+                    break
 
             # Cache the result for 10 minutes
             cache.set("status_results", results, 600)
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -1,5 +1,6 @@
 import base64
 import os
+from abc import ABC
 
 import requests
 from django.core.exceptions import ImproperlyConfigured
@@ -9,7 +10,42 @@ from django.http import HttpResponse
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
-class QfcS3Boto3Storage(S3Boto3Storage):
+class QfcBackendStorageMixin(ABC):
+    def check_status(self) -> bool:
+        """Checks if the storage is reachable.
+
+        Returns:
+            True if reachable, False if not.
+        """
+        raise NotImplementedError(
+            "Subclassing QFC specific storages must implement this method."
+        )
+
+    def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
+        """Patches a nginx redirect response for usage with the storage backend.
+        At the moment, does nothing.
+
+        Arguments:
+            response: HTTP redirect response to patch.
+        """
+        pass
+
+
+class QfcS3Boto3Storage(QfcBackendStorageMixin, S3Boto3Storage):
+    def check_status(self) -> bool:
+        """Checks if the S3 bucket is reachable.
+
+        Returns:
+            True if reachable, False if not.
+        """
+        try:
+            bucket_name = self.bucket.name
+            self.bucket.meta.client.head_bucket(Bucket=bucket_name)
+        except Exception:
+            return False
+
+        return True
+
     def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
         """Patches a nginx redirect response for usage with S3.
         At the moment, does nothing.
@@ -20,7 +56,7 @@ class QfcS3Boto3Storage(S3Boto3Storage):
         pass
 
 
-class QfcWebDavStorage(Storage):
+class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
     """
     Storage backend using WebDAV.
     Adapted and inspired by this repository: https://github.com/marazmiki/django-webdav-storage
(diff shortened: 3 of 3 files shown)


docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -16,17 +19,30 @@ def get(self, request):
         # Try to get the status from the cache
         results = cache.get("status_results", {})
         if not results:
+            # check geodb
             results["geodb"] = "ok"
-            # Check geodb
             if not geodb_utils.geodb_is_running():
                 results["geodb"] = "error"
```
gounux: Should we keep this ?

suricactus: Feel free to kill.

gounux: Thanks ! Addressed, farewell `geodb`.

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -16,17 +18,25 @@ def get(self, request):
...
+            # check storages
+            results["storages"] = {}
+            for storage_key in storages.backends.keys():
+                storage = storages[storage_key]
+
+                try:
+                    results["storages"][storage_key] = storage.check_status()
+                except Exception:
+                    # ManifestStaticFilesStorage may not have check_status method.
+                    continue
```
tdrivas: Instead of skipping silently, you can have an entry here so to wonder about a missing key. 

Something like that:

```suggestion
                    results["storages"][storage_key] = "error" 
```

gounux: Thanks ! Indeed, before your review, it meant that if a storage had an error, it would just go silent.

That being said, we don't want the [`ManifestStaticFilesStorage`](https://github.com/opengisch/QFieldCloud/blob/01cebd6baf7a97c5d36e1cab31b70c58b7de1e5b/docker-app/qfieldcloud/settings.py#L280) to be displayed in here, so maybe a solution could be :

```python
try:
  results["storages"][storage_key] = storage.check_status()
except AttributeError:
  # Some storages may not have check_status method.
  # e.g. ManifestStaticFilesStorage.
  continue
except Exception:
  results["storages"][storage_key] = "error"
```

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -36,9 +36,20 @@ def setUp(self):
...
+
+        self.assertIn("database", data)
+        self.assertEqual(data["database"], "ok")
+
+        self.assertIn("storages", data)
+        self.assertIsInstance(data["storages"], dict)
+
+        for storage_key, storage_status in data["storages"].items():
+            self.assertIn(storage_key, ["legacy_storage", "webdav", "default"])
+            self.assertEqual(storage_status, "ok")
```
tdrivas: A suggestion to have a cleaner code (and check firstly all the `storage_key`) could be:

```suggestion
    expected_storages = {"legacy_storage", "webdav", "default"}
    found_storages = set(data["storages"].keys())
    self.assertEqual(expected_storages , found_storages )

    for storage_key, storage_status in data["storages"].items():
        self.assertEqual(storage_status , "ok")
```

gounux: Thanks ! Addressing that.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -36,9 +36,23 @@ def setUp(self):
...
+
+        data = response.json()
+
+        self.assertIn("database", data)
+        self.assertEqual(data["database"], "ok")
+
+        self.assertIn("storages", data)
+        self.assertIsInstance(data["storages"], dict)
+
+        expected_storages = {"legacy_storage", "webdav", "default"}
```
suricactus: Can we actually get this values from the global `django...storages` value instead of hardcoding? I can imagine the `legacy_storage` to be gone quite soon.

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -36,9 +36,23 @@ def setUp(self):
 
     def test_api_status(self):
         response = self.client.get("/api/v1/status/")
+
         self.assertTrue(status.is_success(response.status_code))
-        self.assertEqual(response.json()["storage"], "ok")
```
suricactus, line 40: Can we actually do it a bit more backwards compatible:

`storage` is a global status - if all storages are ok, then it is ok, if any of the storages has error, then it is error.

And we keep the new key "storages"?

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -16,17 +18,28 @@ def get(self, request):
...
+            results["storages"] = {}
+            for storage_key in storages.backends.keys():
+                storage = storages[storage_key]
+
+                try:
+                    results["storages"][storage_key] = storage.check_status()
+                except AttributeError:
+                    # Some storages may not have the check_status method.
+                    # e.g. ManifestStaticFilesStorage.
+                    continue
```
suricactus: Can we do the other way around, instead of try/except with `AttributeError`, you do check if the storage class is instance of whatever class or `hasattr(storage, "check_status")`?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -10,6 +10,20 @@
 
 
 class QfcS3Boto3Storage(S3Boto3Storage):
+    def check_status(self) -> str:
```
suricactus:
```suggestion
    def check_status(self) -> Literal["ok", "error"]:
```

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -42,6 +56,19 @@ def __init__(self, **options):
         if not self.basic_auth:
             raise ImproperlyConfigured("Please define the `basic_auth` storage option")
 
+    def check_status(self) -> str:
```
suricactus:
```suggestion
    def check_status(self) -> Literal["ok", "error"]:
```

suricactus (commented):
A few small comments.

Mostly two suggestion: 
- do not kill the `storage` key for backwards compatibility
- be more conservative with error handling, accepting just any `AttributeError` is a bit risky to me
- if possible, we should not hardcode the available storages, because this test is practically unapplicable on local machines if they have a different set of configured storages.

docker-app/qfieldcloud/core/tests/test_api.py
gounux: Fine,

docker-app/qfieldcloud/core/tests/test_api.py
```diff
@@ -36,9 +37,30 @@ def setUp(self):
...
+        self.assertIn("storages", data)
+        self.assertIsInstance(data["storages"], dict)
+
+        expected_storages = storages.backends
+        # here we don't care about the configured ManifestStaticFilesStorage.
+        expected_storages.pop("staticfiles")
+
+        found_storages = data["storages"]
+
+        self.assertEqual(expected_storages.keys(), found_storages.keys())
```
suricactus: Shall we sort them or convert to set, otherwise if we have a different order of the keys the test will fail?

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -16,17 +17,31 @@ def get(self, request):
...
+            # check storages
+            results["storage"] = "ok"
+            results["storages"] = {}
+            for storage_key in storages.backends.keys():
+                storage = storages[storage_key]
+
+                if not hasattr(storage, "check_status"):
+                    # Some storages may not have the check_status method.
+                    # e.g. ManifestStaticFilesStorage.
+                    continue
```
suricactus, line 37: If I get it right, if we introduce a new storage provider tomorrow that is not implementing `check_status` method, then we never check for status.


Actually when I think about it, aren't we leaking some internals about what types of storages are currently available to the public?

I am wondering what is your opinon about:
1) do we really need the "storages" with it's keys in on-premise?
2) If we do, would it make sense to respond with the "storages" info only if a certain installation is with specific configuration enabled, such as "STORAGES_DETAILED_STATUS_ENABLED" or whatever name makes sense to name it? IMO this is a bit weird to be shared in the app.qfield.cloud case?

suricactus (approved):
Two more comments.

docker-app/qfieldcloud/core/views/status_views.py
gounux: > If I get it right, if we introduce a new storage provider tomorrow that is not implementing check_status method, then we never check for status.

That is a good point that the future us should remember, I believe.  
Maybe with a new `QfcBackendStorage` defined like this ?

```python
class QfcBackendStorage(ABC, Storage):
    def check_status(self) -> Literal["ok", "error"]:
        raise NotImplemented("check_status method must be implemented for QFieldCloud storage backends")
```

> Actually when I think about it, aren't we leaking some internals about what types of storages are currently available to the public?

This also came in my mind, and this evolution might not be super positive, I agree, so :

1. No.
2. IMHO a new envvar brings some overhead to the QFieldCloud stack.

So, I guess we can just stick to the minimal evolutions this PR can bring :

- remove the `geodb` check and key
- add the `database` check and key
- fix the `storage` check

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -10,6 +11,20 @@
 
 
 class QfcS3Boto3Storage(S3Boto3Storage):
+    def check_status(self) -> Literal["ok", "error"]:
```
gounux: @suricactus feels weird to my mind to have a method that can return only two values, as string...

Isn't there a special type for this in software development ?

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -16,17 +17,32 @@ def get(self, request):
...
             except Exception:
-                results["storage"] = "error"
+                results["database"] = "error"
+
+            # check storages
+            results["storage"] = "ok"
+            for storage_key in storages.backends.keys():
+                storage = storages[storage_key]
+
+                if not hasattr(storage, "check_status"):
```
suricactus, line 34: I guess there is a better way to check this, but let's merge it like this for now.

gounux: Alright, looking forward to enhance this.

PR #1284, opened 2025-07-11
feat(subscription): support per-subscription seats limit

tdrivas:
This PR introduces support for allowing seat limits in specific plans at the subscription level. To this direction a new field named `max_organization_members` is required for the `Subscription` model.


duke-nyuki:
Task linked: [QF-5474 Add support for seat based payment plans to the enduser](https://app.clickup.com/t/2192114/QF-5474)

gounux:
@tdrivas CI looks better now 👌

tdrivas:
> @tdrivas CI looks better now 👌

thanks! Drop and recreation of the views that are related to the migration updated fields was required.

docker-app/qfieldcloud/core/models.py
```diff
@@ -842,9 +842,13 @@ def clean(self) -> None:
         if self.organization.organization_owner == self.member:
             raise ValidationError(_("Cannot add the organization owner as a member."))
 
-        max_organization_members = self.organization.useraccount.current_subscription.plan.max_organization_members
+        default_max_organization_members = self.organization.useraccount.current_subscription.plan.default_max_organization_members
```
```suggestion
        plan_max_organization_members = self.organization.useraccount.current_subscription.plan.max_organization_members
```

Can we adjust the names here, "default_" prefix assumes that this comes as a default value and can be replaced, but the logic (I guess) is:
1) if the `plan.max_organization_members >= 0`, then there cannot be more than `plan.max_organization_members` organization members.
2) if the `plan.max_organization_members == -1` then check `subscription.max_organization_members`.
2.1) if the `subscription.max_organization_members > 1`, then there cannot be more than `subscription.max_organization_members` members.
2.2) if the `subscription.max_organization_members == -1` , then there can be unlimited members.

Actually I saw that 2.2 is checking for 0 instead of -1, which should be adjusted. We either have both fields to be -1 when they are inactive, or we change the logic to use 0, or we use null in such cases. I am in favor to keep the same already established logic everywhere with using -1.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -161,13 +167,16 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
...
+    is_seat_flexible = models.BooleanField(default=False)
+
     # The maximum number of organizations members that are allowed to be added per organization
     # This constraint is useful for public administrations with limited resources who want to cap
     # the maximum amount of money that they are going to pay.
     # Only makes sense when the user_type == User.Type.ORGANIZATION
     # If the organization subscription is changed from unlimited to limited organization members,
     # the existing members that are over the max_organization_members configuration remain active.
-    max_organization_members = models.IntegerField(
+    default_max_organization_members = models.IntegerField(
```
```suggestion
    max_organization_members = models.IntegerField(
```

As I stated above, we better not rename this as it is not really "default_", but more a different type of max org members limitation to the subscription one.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -161,13 +167,16 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
     # the plan is cancellable. If it True, the plan cannot be cancelled.
     is_storage_modifiable = models.BooleanField(default=True)
 
+    # has seats flexibility
+    is_seat_flexible = models.BooleanField(default=False)
```
suricactus, line 168: It is not clear what is the expected value here for the flex plan?

 Also we introduce a new term "seat" but actually everything else in the code is explicitly stating "organization_members". Let us discuss in chat what we can do there.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -466,6 +475,16 @@ class UpdateSubscriptionKwargs(TypedDict):
         related_name="+",
     )
 
+    # The maximum number of organization members (seats) allowed under this specific subscription.
+    # This value is set at subscription creation time, typically based on the quantity selected during checkout (for per-seat pricing).
+    max_organization_members = models.IntegerField(
+        default=0,
```
```suggestion
        default=-1,
```

Just to match the other `Plan.max_organization_members`

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -627,6 +656,45 @@ def active_users_count(self) -> int:
...
 
+    @property
+    def organization_members_count(self) -> int:
+        # if non-organization account, then it is always 1 user
+        if not self.account.user.is_organization:
+            return 1
+
+        return OrganizationMember.objects.filter(
+            organization_id=self.account.user.pk
+        ).count()
```
suricactus: This does not neccessarily incude the organization owner?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -524,10 +543,20 @@ class UpdateSubscriptionKwargs(TypedDict):
     @property
     @deprecated("Use `AbstractSubscription.active_storage_total_bytes` instead")
     def active_storage_total_mb(self) -> int:
+        if self.plan.is_seat_flexible:
+            return self.included_storage_mb + self.active_storage_package_mb
+
         return self.plan.storage_mb + self.active_storage_package_mb
```
suricactus: I guess we can either kill this method, or just `self.active_storage_total_bytes() / 1000 / 1000`

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -524,10 +543,20 @@ class UpdateSubscriptionKwargs(TypedDict):
...
     @property
     def active_storage_total_bytes(self) -> int:
+        if self.plan.is_seat_flexible:
+            return (
+                (self.included_storage_mb + self.active_storage_package_mb)
+                * 1000
+                * 1000
+            )
+
         return self.plan.storage_bytes + self.active_storage_package_bytes
```
suricactus: Can we move this logic within `Plans.included_storage_bytes`?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -627,6 +656,45 @@ def active_users_count(self) -> int:
...
+        # if non-organization account, then it is always 1 user
+        if not self.account.user.is_organization:
+            return 1
+
+        return OrganizationMember.objects.filter(
+            organization_id=self.account.user.pk
+        ).count()
+
+    @property
+    def included_storage_mb(self) -> int:
```
suricactus: I think we can kill `self.included_storage_mb` and use `self.included_storage_bytes` or at least reuse the logic there and just `self.included_storage_bytes() / 1000 / 1000`

suricactus (commented):
A few small comments mostly for clarity/deduplication of the code.

docker-app/qfieldcloud/subscription/models.py
tdrivas: No, `organization_members_count` only counts rows in the  `OrganizationMember ` table, thus the owner will not be included, unless he/she has also been added there as a member/collaborator.

docker-app/qfieldcloud/core/models.py
tdrivas: Thanks! Done!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Adjusted!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Second approach has been adapted! Thanks!

docker-app/qfieldcloud/subscription/models.py
tdrivas: I guess not, because we do need the `max_organization_members` from the `subscription` instance in order to calculagte the `included_storage_mb`

docker-app/qfieldcloud/subscription/models.py
tdrivas: Done!

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -911,3 +911,26 @@ def check_can_upload_file(project, client_type, file_size_bytes: int) -> bool:
...
+        or subscription.plan.max_organization_members == -1
+    ):
+        return False
+
+    if subscription.active_until:
+        return False
+
+    return user_has_organization_role_origins(
+        user,
+        subscription.account,
```
suricactus:
```suggestion
        subscription.account.organization,
```

I think this is the correct way of passing, since `Subscription.account` is a `UserAccount` and `user_has_organization_role_origins` expects an `Organization`?

docker-app/qfieldcloud/subscription/models.py
suricactus: I think we better add them in the count?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -523,11 +542,14 @@ class UpdateSubscriptionKwargs(TypedDict):
...
-        return self.plan.storage_mb + self.active_storage_package_mb
+    def active_storage_total_mb(self) -> float:
+        return self.active_storage_total_bytes / 1000 / 1000
 
     @property
     def active_storage_total_bytes(self) -> int:
+        if self.plan.is_seat_flexible:
+            return self.included_storage_bytes + self.active_storage_package_bytes
+
         return self.plan.storage_bytes + self.active_storage_package_bytes
```
```suggestion
        return self.included_storage_bytes + self.active_storage_package_bytes
```

Isn't it simpler like this?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -627,6 +649,38 @@ def active_users_count(self) -> int:
...
+            organization_id=self.account.user.pk
+        ).count()
+
+    @property
+    def included_storage_mb(self) -> float:
+        """
+        How much storage (in MB) this subscription comes with by default.
+        - For the 'flat' plan (1GB per seat) is calculated as seats x plan.storage_mb
+        - For any other plan, just plan.storage_mb
+        """
```
```suggestion
        """
        How much storage (in MB) this subscription comes with by default.
        - For the 'flat' plan (1GB per seat) is calculated as `Subscription.max_organization_members * Plan.storage_mb`
        - For any other plan, just use `Plan.storage_mb`
        """
```

Same applies to `included_storage_bytes`.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -523,11 +542,14 @@ class UpdateSubscriptionKwargs(TypedDict):
 
     @property
     @deprecated("Use `AbstractSubscription.active_storage_total_bytes` instead")
-    def active_storage_total_mb(self) -> int:
-        return self.plan.storage_mb + self.active_storage_package_mb
+    def active_storage_total_mb(self) -> float:
+        return self.active_storage_total_bytes / 1000 / 1000
```
suricactus: This is marked as deprecated, so I guess we can just adapt it. Also note you changed the API, it used to be an `int` and not it is a `float`. You can explicitly cast to `int`. If we want to be very paranoid, add an assert that the float and the int are not different after casting. Ideally this method should be dead.

docker-app/qfieldcloud/subscription/models.py
suricactus: Actually second though, can we skip adding this method at all, I saw that `active_storage_total_mb` was already marked as deprecated. In general there is a push to kill the megabytes and use only bytes in calculations, re int/float conversation from the previous comment.

suricactus (changes requested):
Thanks for the review address. I have a second round of things to be done/improved.

Also, I note a few fields that you have added to the models, which is fine, but they are not presented in the Django admin. Can we expose these fields to Django admin?

docker-app/qfieldcloud/subscription/models.py
suricactus: You are right, I meant the `Subscription.included_storage_bytes`, there is no `Plan.included_storage_bytes`.

docker-app/qfieldcloud/subscription/models.py
suricactus: Thanks for adapting to naming.

docker-app/qfieldcloud/subscription/models.py
tdrivas: Indeed!Thanks!

docker-app/qfieldcloud/subscription/models.py
tdrivas: if so, shall we introduce a new filter `bytes_to_gb` for the templates?

docker-app/qfieldcloud/core/permissions_utils.py
tdrivas: Correct, thanks!

docker-app/qfieldcloud/subscription/models.py
tdrivas: Sure!

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -627,6 +655,32 @@ def active_users_count(self) -> int:
...
+    def organization_members_count(self) -> int:
+        # if non-organization account, then it is always 1 user
+        if not self.account.user.is_organization:
+            return 1
+
+        return (
+            OrganizationMember.objects.filter(organization_id=self.account.user.pk)
+            .exclude(member_id=self.account.user.organization_owner_id)
+            .count()
+            + 1
```
suricactus, line 662:
```suggestion
            # +1 for the organization owner himself
            + 1
```

suricactus (commented):
Here we are at a good stage, no expected changes for now, except:

1) can you reduce the number of migrations to 1 or 2?
2) I am still not convinced with `is_seat_flexible` naming. Don't change it for now, we can live with it.
3) probably we can move `subscription.Plan.UsageType` to a more billing related place.

docker-app/qfieldcloud/subscription/models.py
tdrivas: Always good to explain, thanks!

tdrivas:
> Here we are at a good stage, no expected changes for now, except:
> 
> 1. can you reduce the number of migrations to 1 or 2?
> 2. I am still not convinced with `is_seat_flexible` naming. Don't change it for now, we can live with it.
> 3. probably we can move `subscription.Plan.UsageType` to a more billing related place.

Thanks for the fruitful review. 
1. Done
2. Yep, lets live with it for now
3. Finally, I moved the entire field to a more billing related place without breaking anything here.

suricactus:
@tdrivas can we make the CI happy again?

docker-app/qfieldcloud/subscription/migrations/0007_plan_is_seat_flexible_and_more.py
```diff
@@ -0,0 +1,36 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("subscription", "0006_auto_20230426_2222"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql=migrations.RunSQL.noop,
+            reverse_sql="""
+                DROP VIEW IF EXISTS current_subscriptions_vw;
```
suricactus, line 15: I am still not convinced this is the right way to go: two queries in one migration step.

If this is the only way to go, please add a
```
NOTE: while suboptimal to run two queries per migration step, this is the only working solution to recreate the `current_subscriptions_vw` on rollback so far.
```

suricactus (commented):
So far so good here, I will not push any more changes here.

docker-app/qfieldcloud/core/models.py
```diff
@@ -842,9 +842,13 @@ def clean(self) -> None:
...
+        plan_max_organization_members = self.organization.useraccount.current_subscription.plan.max_organization_members
+        max_organization_members = (
+            self.organization.useraccount.current_subscription.max_organization_members
+        )
+
         if (
-            max_organization_members > -1
+            plan_max_organization_members > -1
             and self.organization.members.count() >= max_organization_members
         ):
```
suricactus: Hm, I find the logic here a bit confusing. Basically we do the following:

check if the plan has `max_organization_members`
AND
check if the current organization has more than the subscription `max_organization_members`.

Would it make more sense to check for the `max_organization_members` in the same model in both conditions?

PR #1286, opened 2025-07-14
feat(secrets): use `django-fernet-encrypted-fields` for encrypted values

gounux:
This PR changes the way encrypted fields' value is stored, not using [django-cryptography](https://github.com/georgemarshall/django-cryptography)'s `encrypt` method anymore, but [django-fernet-encrypted-fields](https://github.com/jazzband/django-fernet-encrypted-fields)'s `EncryptedTextField`.

There is a migration that ensures the old value remains here.


duke-nyuki:
Task linked: [QF-6045 Migrate the dependency of encrypted fields from `django-cryptography` to `django-fernet-encrypted-fields`](https://app.clickup.com/t/2192114/QF-6045)

.env.example
```diff
@@ -14,6 +14,9 @@ DJANGO_ALLOWED_HOSTS="localhost 127.0.0.1 0.0.0.0 app nginx"
 
 SECRET_KEY=change_me
 
+# Key used for cryptographic operations on encrypted fields.
+# Change this value, for instance with `pwgen -sn 128`.
+SALT_KEY=0123456789abcdefghijklmnopqrstuvwxyz
```
suricactus: Link to docs will help too.

docker-app/qfieldcloud/core/migrations/0086_secret_new_value.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.2.19 on 2025-07-14 09:01
+
+from django.db import migrations, models
+from encrypted_fields.fields import EncryptedTextField
+
+
+def copy_secret_value(apps, schema_editor):
+    Secret = apps.get_model("core", "Secret")
+    # is this going to decrypt the value as we want ?
```
suricactus: Is it? Tbh I don't think so, please carefully test before we accidentally merge like this :)

docker-app/qfieldcloud/core/migrations/0086_secret_new_value.py
```diff
@@ -0,0 +1,41 @@
...
+        ),
+        migrations.AddField(
+            model_name="secret",
+            name="value",
+            field=EncryptedTextField(default=""),
+            preserve_default=False,
+        ),
+        migrations.RunPython(
+            copy_secret_value,
+            reverse_code=migrations.RunPython.noop,
```
suricactus, line 45: Hm, here we either should not make this migration reversible, OR we should make the `value` being transfered to `old_value`.

This way we don't lose data if we have to revert.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2495,7 +2496,10 @@ class Type(models.TextChoices):
     )
 
     created_at = models.DateTimeField(auto_now_add=True)
-    value = django_cryptography.fields.encrypt(models.TextField())
+
+    # encrypted value of the secret
+    old_value = django_cryptography.fields.encrypt(models.TextField())
+    value = EncryptedTextField()
```
```suggestion
    # TODO QF-XXX Remove legacy encrypted fields support
    # legacy field to store the encrypted value of the secret
    old_value = django_cryptography.fields.encrypt(models.TextField())
    
    # encrypted value of the secret
    value = EncryptedTextField()
```

Feel free to adjust the text as you wish, but be explicit what we are doing here.

docker-app/qfieldcloud/settings.py
```diff
@@ -35,6 +35,10 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
+# Key used for password hashing and other cryptographic operations.
```
suricactus: Shall we align the text here with the `.env.example`?

Also here you mention password hashing, but is it true? Don't we use this value ONLY for secrets?

gounux:
Thanks for the review @suricactus. Indeed, I guess it deserves more testing before merging.

gounux:
@suricactus thanks, indeed, we'll need to fine-tune the migration as well as non-nullable field values, `old_value` needing to be nullable and `value` non-nullable after the migration.

suricactus:
@gounux can you rebase to fix the conflict?

suricactus (approved):
I think tests passed now, so we can merge?

docker-app/qfieldcloud/core/models.py
```diff
@@ -2534,7 +2535,15 @@ class Type(models.TextChoices):
     )
 
     created_at = models.DateTimeField(auto_now_add=True)
-    value = django_cryptography.fields.encrypt(models.TextField())
+
+    # TODO Remove with QF-6361 & legacy encrypted fields support drop.
```
```suggestion
    # TODO Remove with QF-6361 Remove legacy `django_cryptography`'s encrypted field support
```

gounux:
> I think tests passed now, so we can merge?

Indeed, could migrate forward and backward, thanks !  
This can be merged.

PR #1287, opened 2025-07-14
clean(storage): study and enhance filestorage migration command

gounux:
This PR to study and optimize the file storage migration command for migrating projects storage.

Considering the projects stored on the `default` storage should not be migrated, this PR excludes them from ore-filtering based on CLI arguments.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
+++ b/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
@@ -4,10 +4,11 @@ from collections.abc import Collection
 
 from django.core.management.base import BaseCommand
 from qfieldcloud.core.models import Project
-
-from ...migrate_project_storage import migrate_project_storage
+from qfieldcloud.filestorage.migrate_project_storage import migrate_project_storage
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_STORAGE = "default"
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -37,9 +37,11 @@ def migrate_project_storage(
     from_storage = project.file_storage
 
     if from_storage == to_storage:
-        raise Exception(
-            f'Cannot migrate to storage "{to_storage}", project is already stored there!'
+        # skip if the project is already migrated, avoids the script to fail.
+        logger.error(
+            f'Cannot migrate to storage "{to_storage}", project {project.id} is already stored there!'
         )
+        return
```
gounux: This is maybe not needed anymore, since we exclude before the projects stored on the new storage.

Anyway, looking forward to discuss the approach here, if silent skipping or loud exception raising.  
Both have their pros and cons I guess, and we should also consider a usage on a large number of projects to migrate.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -84,7 +90,8 @@ def handle(self, *args, **options) -> None:
 
             exit(1)
 
-        if len(projects) == 0:
+        nb_projects = len(projects)
```
```suggestion
        projects_count = len(projects)
```

If you don't mind, most of the codebase does the numbering with `_count` suffix.

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus: Whatever makes you feel more confortable to run. I slightly prefer exceptions, but in this case I don't have a strong opinion.

As you said, since we prefilter, actually we should never enter this condition, but we better keep it. A note in the code about prefiltering and "this should never happen" would be good idea too.

suricactus (approved):

PR #1290, opened 2025-07-17
test(package): ensure repackaging is needed with secrets

gounux:
This PR ensures that the repackaging of projects is triggered for users and collaborators, when some User-level secrets are assigned.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1580,16 +1580,24 @@ class Project(models.Model):
         return True
 
     def needs_repackaging(self, user: User) -> bool:
+        latest_package_job_for_user = self.latest_package_job_for_user(user)
+
         if (
             # if has_online_vector_data is None (happens when the project details are missing)
             # we assume there might be
             self.has_online_vector_data is False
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/models.py
```diff
@@ -1580,16 +1580,19 @@ def can_repackage(self) -> bool:
...
             and self.data_last_updated_at
             and self.data_last_packaged_at
-            and self.latest_package_job_for_user(user)
+            and latest_package_job_for_user
         ):
-            # if all vector layers are file based and have been packaged after the last update, it is safe to say there are no modifications
-            return self.data_last_packaged_at < self.data_last_updated_at
+            # if all vector layers are file based and have been packaged for the user after the last update,
+            # it is safe to say there are no modifications.
+            return latest_package_job_for_user.updated_at < self.data_last_updated_at
```
suricactus: Hm, I think `Job.finished_at` is actually a better candidate to use here, no?

gounux: Indeed, yes. Addressed, thanks.

suricactus (approved):

PR #1294, opened 2025-07-24
feat(drf): ensure the login on drf browseable API page is also using allauth

suricactus:
Using the DRF browserable API, the users are invited to login but the login form that appears does not go through allauth. This should be fixed by now by explicitly redirecting users to the allauth alternatives. Unfortunately we still need to keep the `    path("auth/", include("rest_framework.urls")),`, otherwise the login button does not appear at all.

image


duke-nyuki:
Task linked: [QF-6412 Use only one login page from `allauth`](https://app.clickup.com/t/2192114/QF-6412)

PR #1295, opened 2025-07-30
fix(projects): remove org membership checks for projects not being public

gounux:
This was mainly causing results of `Project.objects.for_user` casting a too wide net, when getting Secrets for a User.


docker-app/qfieldcloud/core/models.py
```diff
@@ -2374,16 +2374,18 @@ def for_user_and_project(self, user: User, project: Project) -> SecretQueryset:
...
                 user, organization, OrganizationMember.ALL_ROLES
             )
         else:
             organization = None
-        # ensure the user has access to the project
-        check_user_has_project_roles(user, project, ProjectCollaborator.ALL_ROLES)
+
+        # ensure the user has access to the project if not public
+        if not project.is_public:
+            check_user_has_project_roles(user, project, ProjectCollaborator.ALL_ROLES)
```
```suggestion
        check_user_has_project_roles(user, project, ProjectCollaborator.ALL_ROLES)
```

Isn't it sufficient without the explicit if statement? The user should have by default a "reader" role?

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -749,3 +749,32 @@ def test_get_project_details_for_qfield(self):
         self.assertIsInstance(project_json["can_repackage"], bool)
         self.assertIn("needs_repackaging", project_json)
         self.assertIsInstance(project_json["needs_repackaging"], bool)
+
+    def test_seamless_projects_fetching_with_a_public_one(self):
```
suricactus: What if the project is public and owned by an organization?

suricactus (commented):
Two questions.

docker-app/qfieldcloud/core/models.py
gounux: Interesting, makes me consider changing the approach here.

Reason behind this if is this log line, spotted sometimes on public projects: `raise UserProjectRoleError(\nqfieldcloud.core.permissions_utils.UserProjectRoleError: User \"USER\" has not role \"['ADMIN', 'MANAGER', 'EDITOR', 'REPORTER', 'READER']\" on project PUBLIC_PROJECT."}`

gounux: Addressed, thanks.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2374,16 +2374,19 @@ def for_user_and_project(self, user: User, project: Project) -> SecretQueryset:
             check_user_has_project_roles,
         )
 
-        if project.owner.is_organization:
-            # ensure the user has access to the organization
+        # ensure the user has access to the organization if private project
+        if project.owner.is_organization and not project.is_public:
```
suricactus: Are you sure we need to check for this `is-public` here?

gounux: Good question, probably not. Thanks !

gounux: Addressed in next commit, thanks.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2374,15 +2374,16 @@ def for_user_and_project(self, user: User, project: Project) -> SecretQueryset:
             check_user_has_project_roles,
         )
 
+        # ensure the user has access to the organization if private project
         if project.owner.is_organization:
-            # ensure the user has access to the organization
             organization = Organization.objects.get(id=project.owner_id)
             check_user_has_organization_roles(
                 user, organization, OrganizationMember.ALL_ROLES
             )
```
```suggestion
    # NOTE if the project is public, then it is not neccessary for the user to be a member of the project's organization
    if project.is_public:
        # NOTE While this assertion is nice to have, it causes huge performance penalty on production, so we explicitly check if we are in non-production mode
        assert (
            settings.ENVIRONMENT != "production"
            and user_has_organization_roles(
                user,
                organization,
                OrganizationMember.ALL_ROLES,
            )
        )
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -2374,15 +2374,16 @@ def for_user_and_project(self, user: User, project: Project) -> SecretQueryset:
...
         else:
             organization = None
-        # ensure the user has access to the project
+
+        # ensure the user has access to the project if not public
         check_user_has_project_roles(user, project, ProjectCollaborator.ALL_ROLES)
```
```suggestion
        # NOTE While this assertion is nice to have, it causes huge performance penalty on production, so we explicitly check if we are in non-production mode
        assert (
            settings.ENVIRONMENT != "production"
            and check_user_has_project_roles(user, project, ProjectCollaborator.ALL_ROLES)
        )
```

gounux: Addressed, thanks.

docker-app/qfieldcloud/core/models.py
gounux: Addressed, thanks.

PR #1299, opened 2025-08-05
fix(secrets): make worker get a query set with all secrets

gounux:
We don't want to miss Org-level Secrets in the worker.


gounux:
> Can we add a test for this before we merge and confirm it fails when the suggested change is not present?

Before this PR, the issue results in a unavailable PostGIS layer, in the QField project. Let me work on that but that might be long to write.

suricactus:
I see it is a bit trickier to test this as it is in the worker wrapper. It's ok like this, just make sure there are no other instance of `project.secrets` in the code I guess.

suricactus (approved):

PR #1301, opened 2025-08-05
test(secret): ensure the worker is retrieving Org-level secrets

gounux:
Follow-up of #1299 with a test case.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_secret_packages.py
+++ b/docker-app/qfieldcloud/core/tests/test_secret_packages.py
@@ -1,8 +1,13 @@
 import io
 
+import psycopg2
+from django.conf import settings
 from rest_framework.test import APITransactionTestCase
 
+from qfieldcloud.core.geodb_utils import delete_db_and_role
 from qfieldcloud.core.models import (
+    Geodb,
+    Job,
     Organization,
     OrganizationMember,
     Person,
@@ -31,7 +36,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         self.o1 = Organization.objects.create(username="o1", organization_owner=self.u1)
 
         # Activate Subscriptions
-        set_subscription(self.o1, "default_org")
+        set_subscription(self.o1, is_external_db_supported=True)
 
         self.p1 = Project.objects.create(name="p1", owner=self.o1)
 
@@ -54,6 +59,22 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         ]
 
         self.p1.direct_collaborators.bulk_create(collaborators)
+
+        delete_db_and_role("test", self.u1.username)
+        self.geodb = Geodb.objects.create(
+            user=self.u1,
+            dbname="test",
+            hostname="geodb",
+            port=5432,
+        )
+
+        self.conn = psycopg2.connect(
+            dbname="test",
+            user=settings.GEODB_USER,
+            password=settings.GEODB_PASSWORD,
+            host=settings.GEODB_HOST,
+            port=settings.GEODB_PORT,
+        )
 
     def _create_secret(self, **kwargs) -> Secret:
         return Secret.objects.create(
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-6454 Make worker fetch all secrets for a user and project](https://app.clickup.com/t/2192114/QF-6454)

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -55,6 +59,20 @@ def setUp(self):
 
         self.p1.direct_collaborators.bulk_create(collaborators)
 
+        self.geodb, _ = Geodb.objects.get_or_create(
+            user=self.u1,
+            hostname="geodb",
+            port=5432,
+        )
```
gounux: Here the `name="test"` key has been removed, since it was making the test fail with an error message like `"test" database already exist`.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -206,3 +224,79 @@ def test_create_secrets_and_packages(self):
...
+
+        _secret_pgs2 = Secret.objects.create(
+            name="PG_SERVICE_GEODB2",
+            type=Secret.Type.PGSERVICE,
+            created_by=self.u1,
+            organization=self.o1,
+            value=(
+                "[geodb2]\n"
+                "dbname=test\n"
```
gounux, line 262: Are we sure we are targeting the database created in the same test case ? Or another `test` database created somwhen before in the tests ?

docker-app/qfieldcloud/core/tests/test_secret_packages.py
```diff
@@ -55,6 +59,20 @@ def setUp(self):
 
         self.p1.direct_collaborators.bulk_create(collaborators)
 
+        self.geodb, _ = Geodb.objects.get_or_create(
```
suricactus:
```suggestion
        self.geodb = Geodb.objects.create(
```

Why `get_or_create`?

docker-app/qfieldcloud/core/tests/test_secret_packages.py
suricactus: I guess that is the reason why this line exists here? https://github.com/opengisch/QFieldCloud/blob/960b7298066300e721327d76498ffeaa2250e9d9/docker-app/qfieldcloud/subscription/tests/test_external_db.py#L76

docker-app/qfieldcloud/core/tests/test_secret_packages.py
suricactus: See https://github.com/opengisch/QFieldCloud/pull/1301?notification_referrer_id=NT_kwDOACsJV7MxODAzMDYzOTY0MDoyODIwNDM5#discussion_r2255935636

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: Thanks ! Indeed this will help.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: #1301 ? This is the current one.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: Why not ? Not.

suricactus: Because you are assuming the presense of a database which should not be the case. For more deterministic tests, I would suggest we always create a database and use it instead of relying on the order and combination of the tests ran before, which is most probably the reason to have a test db already created.

gounux: Fully agree.

docker-app/qfieldcloud/core/tests/test_secret_packages.py
suricactus: I was refering to the previous comment here: https://github.com/opengisch/QFieldCloud/pull/1301#discussion_r2256412786 . But I think you already have the answer here?

suricactus (approved):

docker-app/qfieldcloud/core/tests/test_secret_packages.py
gounux: Exactly, thanks.

PR #1303, opened 2025-08-11
fix: prevent N+1 problem when listing projects with a lot of files

suricactus:
See individual commit messages and comments in the code.

We remove the default `storage` attribute added by the base class `FieldFile`, and we later add a cached property that computes the `storage` dynamically.
This is done to avoid the `storage` being calculated at each instantiation of the `DynamicStorageFieldFile`.
Using the cached and on-demand `storage` property calculation greatly optimizes the performance of the field, by saving N*3 database query calls.
The actual value of the `storage` property is not needed until the file is accessed.

Further optimize with `select_related` and `prefetch_related` on the files endpoint.

### DIFF
--- a/docker-app/qfieldcloud/core/fields.py
+++ b/docker-app/qfieldcloud/core/fields.py
@@ -3,10 +3,10 @@ from __future__ import annotations
 from collections.abc import Callable
 from typing import Any, Protocol, cast
 
-from django.core.files.storage import storages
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6474 Fix extremely slow GET files querie on non-legacy storage](https://app.clickup.com/t/2192114/QF-6474)

PR #1304, opened 2025-08-11
fix: after dynamically getting the file storage backend, make sure they are actually deleted

suricactus:
continuation of #1303 

Too impatient with the CI to show us the previous approach was not deleting permanently the files.

This should be now fixed.


duke-nyuki:
Task linked: [QF-6474 Fix extremely slow GET files querie on non-legacy storage](https://app.clickup.com/t/2192114/QF-6474)

PR #1305, opened 2025-08-15
fix: be explicit that the given assert is disabled on prod environment

suricactus:
I guess here we overlooked a test. Such changes should be done with greater care.


gounux:
Thanks @suricactus :)

PR #1293, opened 2025-07-24
feat(admin): add custom admin site that reuses the allauth login

suricactus:
Since login is a quite a critical part of every application, it would be great if it is always done via a single entrypoint. However, currently the login is handled either via allauth login, or via Django admin login for django admin.

Or in other words, there are `accounts/login` and `admin/login`. Suboptimal and fixed in this PR.

What is more, adding a custom admin site for QFieldCloud will allow us to do other "magical" stuff.

In a follow-up PR we need to clean-up the current available configuration on the admin template and provide the same on the allauth template.

PR #1310, opened 2025-08-19
feat(admin): add custom admin site that can be extended in the future

suricactus:
Preparation for #1293

### DIFF
--- a/docker-app/qfieldcloud/authentication/tests/test_authentication.py
+++ b/docker-app/qfieldcloud/authentication/tests/test_authentication.py
@@ -2,6 +2,7 @@ import logging
 from datetime import datetime
 
 import django.db.utils
+from django.conf import settings
 from django.core.files.base import ContentFile
 from django.urls import reverse
 from django.utils.timezone import now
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/admin.py
```diff
@@ -62,9 +63,23 @@
 from qfieldcloud.core.utils import get_file_storage_choices
 from qfieldcloud.core.utils2 import delta_utils, jobs, pg_service_file
 
-admin.site.unregister(LogEntry)
 
-Invitation = get_invitation_model()
+class QfcAdminSite(AdminSite):
+    site_header = _("QFieldCloud Admin")
+    site_title = _("QFieldCloud Admin")
```
gounux, line 69: What about getting those values from the `JAZZMIN_SETTINGS` value, in the settings ?

https://github.com/opengisch/QFieldCloud/blob/02a4d30eaaf21718a0d24424594ab9742d8e6bfa/docker-app/qfieldcloud/settings.py#L807-L808

suricactus: We cannot translate the values from Jazzmin.

PR #1312, opened 2025-08-20
feat: do not report `TooManyLoginAttemptsError` to Sentry

suricactus:
We cannot do much about it. Probably we can just monitor if there are way too many of them per hour or so in the monitoring.


gounux:
What if there is a surprisingly high number of login requests ? Will we still be available to see it go by ?

PR #1322, opened 2025-08-25
feat: added configuration to set incident info and status page url

suricactus:
Added constance fields:

- STATUS_PAGE_URL - sets the status page
- INCIDENT_IS_ACTIVE - whether there is a known incident or not
- INCIDENT_TIMESTAMP_UTC - indicates when the incident started. We need
two separate fields with `INCIDENT_IS_ACTIVE` and
`INCIDENT_TIMESTAMP_UTC` because `constance` does not support the latter
to be nullable (it must be a datetime instance).
- INCIDENT_MESSAGE - message that can be shown to the users

image

Also some tests improvements: moved tests about the GET status API in a dedicated file and test the newly added fields `status_page_url`, `incident_timestamp_utc` and `incident_message`
Addings tests for the `database == error` field is hard, as one cannot simply
overwrite the `DATABASES` settings. Django prevents this with:
> Overriding setting DATABASES can lead to unexpected behavior.

Finally, HTTP clients can enjoy cool stuff like:
```
{
    "database": "ok",
    "storage": "ok",
    "status_page_url": "https://status.qfield.cloud/",
    "incident_message": "QFieldCloud is currently experiencing stability issues. Our team is investigating the problem.",
    "incident_timestamp_utc": "2025-01-01T00:00:00"
}
```

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_status.py
+++ b/docker-app/qfieldcloud/core/tests/test_status.py
@@ -1,0 +1,170 @@
+import logging
+import time
+from datetime import datetime
+
+from constance import config
+from constance.test import override_config
+from django.core.cache import cache
+from django.test import override_settings
+from rest_framework import status
+from rest_framework.test import APITransactionTestCase
+
+from qfieldcloud.core.views.status_views import StatusValue
+
+logging.disable(logging.CRITICAL)
+
+
+class QfcTestCase(APITransactionTestCase):
+    def setUp(self):
+        # Empty cache value
+        cache.clear()
+
+        # # Create needed subscription relations
+        # setup_subscription_plans()
+
+        # # Set up a user to own projects
+        # self.user = Person.objects.create_user(username="user1", password="abc123")
+        # self.token = AuthToken.objects.get_or_create(user=self.user)[0]
+
+    def test_status_api(self):
+        response = self.client.get("/api/v1/status/")
+
+        self.assertTrue(status.is_success(response.status_code))
+
+        data = response.json()
+
+        self.assertIn("database", data)
+        self.assertEqual(data["database"], StatusValue.OK)
+
+        self.assertIn("storage", data)
+        self.assertEqual(data["storage"], StatusValue.OK)
+
+        self.assertIn("status_page_url", data)
+        self.assertEqual(data["status_page_url"], config.STATUS_PAGE_URL)
+
+        self.assertIn("incident_timestamp_utc", data)
+        self.assertIsNone(data["incident_timestamp_utc"])
+
+        self.assertIn("incident_message", data)
+        self.assertIsNone(data["incident_message"])
+
+    @override_settings(STORAGES={})
+    def test_status_storage_fails_with_no_storages(self):
+        response = self.client.get("/api/v1/status/")
+
+        self.assertTrue(status.is_success(response.status_code))
+
+        data = response.json()
+
+        self.assertIn("database", data)
+        self.assertEqual(data["database"], StatusValue.OK)
+
+        self.assertIn("storage", data)
+        self.assertEqual(data["storage"], StatusValue.ERROR)
+
+        self.assertIn("status_page_url", data)
+        self.assertEqual(data["status_page_url"], config.STATUS_PAGE_URL)
+
+        self.assertIn("incident_timestamp_utc", data)
+        self.assertIsNone(data["incident_timestamp_utc"])
+
+        self.assertIn("incident_message", data)
+        self.assertIsNone(data["incident_message"])
+
+    @override_settings(
+        STORAGES={
+            "default": {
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-6538 Add banner setting in case QFieldCloud is experiencing an incident](https://app.clickup.com/t/2192114/QF-6538)

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -1,50 +1,111 @@
...
+                cursor.execute("SELECT 1")
+
+            return StatusValue.OK
+        except Exception:
+            logger.error('Failed to connect to the database "default".', exc_info=True)
+
+            # just to be safe, we catch all exceptions and set the status to error again.
+            return StatusValue.ERROR
+
+    def _check_storages(self) -> StatusValue:
```
gounux, line 79: Should we somehow extract those 2 methods to a `utils*.py` file ?  
They'd also be used by the `status` Django command.

suricactus: I would rather kill the `status` command, I completely forgot it's existance. Or you actually use it?

suricactus: Killed here https://github.com/opengisch/QFieldCloud/pull/1326

gounux: I used it once, recently, but I guess it's a duplicate of the `/api/v1/status` API endpoint. I'm okay to drop it.

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -1,50 +1,111 @@
...
-                except Exception:
-                    results["storage"] = "error"
-                    break
+        # add info about ongoing incident if any
+        if config.INCIDENT_IS_ACTIVE:
+            logger.warning(
+                "Incident is active, reporting incident details in status API since %s with message: %s",
+                config.INCIDENT_TIMESTAMP_UTC,
+                config.INCIDENT_MESSAGE,
+            )
```
gounux, line 58: Do we need this log line ?

docker-app/qfieldcloud/core/views/status_views.py
```diff
@@ -1,50 +1,111 @@
...
-                with db_conn.cursor() as cursor:
-                    cursor.execute("SELECT 1")
-            except Exception:
-                results["database"] = "error"
-
-            # check storages
-            results["storage"] = "ok"
-            for storage_key in storages.backends.keys():
-                storage = storages[storage_key]
+    @method_decorator(cache_page(60))
```
gounux, line 38: Why is the `@method_decorator` used here ?

docker-app/qfieldcloud/core/tests/test_status.py
```diff
@@ -0,0 +1,170 @@
...
+    def setUp(self):
+        # Empty cache value
+        cache.clear()
+
+        # # Create needed subscription relations
+        # setup_subscription_plans()
+
+        # # Set up a user to own projects
+        # self.user = Person.objects.create_user(username="user1", password="abc123")
+        # self.token = AuthToken.objects.get_or_create(user=self.user)[0]
```
gounux:
```suggestion
```

gounux (approved):
Thanks ! Few understanding / clean-ups questions.

docker-app/qfieldcloud/core/views/status_views.py
suricactus: Because you cannot decorate a view method without a method decorator.  Please read the whole page and specifically https://docs.djangoproject.com/en/4.2/topics/class-based-views/intro/#decorating-the-class

As for caching, I readded it as it was already there before and 10mins was too much so I reduced it to 1min.

docker-app/qfieldcloud/core/views/status_views.py
suricactus: I think yes. We can setup an alarm in the monitoring in case someone accidentally sets the incident to true.

docker-app/qfieldcloud/core/views/status_views.py
gounux: Okay, thanks :) Was asking that since I saw this decorator several time in the codebase.

> As for caching, I readded it as it was already there before and 10mins was too much so I reduced it to 1min.

Alright, looks good,

docker-app/qfieldcloud/core/views/status_views.py
gounux: Oh indeed, nice idea !

docker-app/qfieldcloud/core/tests/test_status.py
suricactus: Ah, thanks!

PR #1323, opened 2025-08-25
fix(storage): if the project is using non-legacy storage, get it's size directly from the database

suricactus:
Also added a test with two different storage backends.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -514,13 +514,28 @@ class UserAccount(models.Model):
     @property
     def storage_used_bytes(self) -> float:
         """Returns the storage used in bytes"""
-        used_quota = (
-            self.user.projects.aggregate(sum_bytes=Sum("file_storage_bytes"))[
-                "sum_bytes"
-            ]
+        from qfieldcloud.filestorage.models import File, FileVersion
+
+        project_files_used_quota = (
+            FileVersion.objects.filter(
+                file__file_type=File.FileType.PROJECT_FILE,
+                file__project__in=self.user.projects.exclude(
+                    file_storage=settings.LEGACY_STORAGE_NAME
+                ),
+            ).aggregate(sum_bytes=Sum("size"))["sum_bytes"]
+            or 0
+        )
+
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6173 Use File and FileVersion to calculate the used storage amount for the projects of a given account](https://app.clickup.com/t/2192114/QF-6173)

docker-app/qfieldcloud/filestorage/tests/test_storage_usage.py
```diff
@@ -0,0 +1,85 @@
...
+        self.p1.refresh_from_db()
+        p2.refresh_from_db()
+        self.u1.useraccount.refresh_from_db()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        # p1 checks
+        self.assertEqual(self.p1.project_files.count(), 1)
+        self.assertEqual(self.p1.get_file("file.name").versions.count(), 2)
+        self.assertEqual(self.p1.file_storage_bytes, 13)
+        # p2 checks
```
gounux, line 81:
```suggestion
        # p2 checks # TODO: Delete with QF-4963 Drop support for legacy storage
```

For the future...

docker-app/qfieldcloud/core/models.py
```diff
@@ -514,14 +514,29 @@ def upcoming_subscription(self):
     @property
     def storage_used_bytes(self) -> float:
         """Returns the storage used in bytes"""
-        used_quota = (
-            self.user.projects.aggregate(sum_bytes=Sum("file_storage_bytes"))[
-                "sum_bytes"
-            ]
+        from qfieldcloud.filestorage.models import File, FileVersion
```
gounux, line 517: Naive question : why is the import here and not on top of the file ?

suricactus: Usually the answer is circular imports :(

docker-app/qfieldcloud/filestorage/tests/test_storage_usage.py
suricactus: Thanks!

PR #1324, opened 2025-08-25
feat(filestorage): disable the minimum bytes length in the HTTP Range header

suricactus:
Also added a test that to check if `QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH>0` still works.

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -680,7 +680,7 @@ CONSTANCE_CONFIG_FIELDSETS = {
 }
 
 # Minimum number of bytes to ask a range when requesting a file part, otherwise a HTTP 416 is returned. Set to 0 to allow any number of bytes in the range.
-QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH = 1000000
+QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH = 0
 
 # Name of the qgis docker image used as a worker by worker_wrapper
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6539 Disable the minimum bytes length in the HTTP Range header](https://app.clickup.com/t/2192114/QF-6539)

docker-app/qfieldcloud/settings.py
```diff
@@ -680,7 +680,7 @@ def before_send(event, hint):
 }
 
 # Minimum number of bytes to ask a range when requesting a file part, otherwise a HTTP 416 is returned. Set to 0 to allow any number of bytes in the range.
-QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH = 1000000
+QFIELDCLOUD_MINIMUM_RANGE_HEADER_LENGTH = 0
```
gounux, line 683: Would it make sense to extract it in the `.env` and docker compose file ?

nirvn: Question: is it common practice to enforce a minimum range to begin with?

suricactus: I don't see a reason to enable it for configuration any time soon, therefore not being exposed.

The reason for this was to protect ourselves from some (purposely) dumb clients requesting files 1 byte at a time.

gounux: Alright, thanks :)

PR #1325, opened 2025-08-25
refactor: rewrite the way how secrets are filtered in `latest_package_job_for_user`

suricactus:
Make it more explicit what we are filtering for.


duke-nyuki:
Task linked: [QF-6541 Be more conservative with secrets for organizations](https://app.clickup.com/t/2192114/QF-6541)

gounux (approved):
Thanks @suricactus, code looks good. Also, this change uses filter combinations of the 3 attributes (project, organization + assigned_to), which is a good thing I believe :)

PR #1326, opened 2025-08-26
feat: kill manage.py status command as very venerable and unused in practice

suricactus:
It did not work properly for quite some time already


README.md
```diff
@@ -67,11 +67,6 @@ And collect the static files (CSS, JS etc):
 
     docker compose run app python manage.py collectstatic --noinput
 
-You can check if everything seems to work correctly using the
-`status` command:
-
-    docker compose exec app python manage.py status
```
gounux, line 73: Wouldn't we keep a mention in the README regarding the status check ?

Maybe something like `curl http[s]://qfieldcloud.instance/api/v1/status/` ?

gounux (approved):
Apart from a little comment regarding a status mention in the README, I'm okay with that :)

README.md
suricactus: Added, see my latest commit.

PR #1328, opened 2025-08-27
feat(filestorage): Support `If-(None-)Match` and `If-(Un)Modified-Since` headers in file download endpoints

suricactus:
Since we do internal redirect to the Object storage API, we need to pass the headers to the Object storage and leave their provider to decide what to do.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/view_helpers.py
+++ b/docker-app/qfieldcloud/filestorage/view_helpers.py
@@ -1,4 +1,6 @@
 import logging
+from datetime import datetime
+from datetime import timezone as tz
 from pathlib import PurePath
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -270,7 +273,32 @@ def download_field_file(
         # this is the relative path of the file, including the containing directories.
         # We cannot use `ContentFile.path` with object storage, as there is no concept for "absolute path".
         storage_filename = field_file.name
-        parameters = {}
+        parameters: dict[str, str | datetime] = {}
+
+        if_match_etags = request.headers.get("if-match", "")
```
gounux, line 278: Quick thought from the readability point of view : what about putting the `if`s right after the variable declaration ?

e.g. :

```python
if_match_etags = request.headers.get("if-match", "")
if if_match_etags:
    parameters["IfMatch"] = if_match_etags

if_none_match_etags = request.headers.get("if-none-match", "")
if if_none_match_etags:
    parameters["IfNoneMatch"] = if_none_match_etags

...
```

Seems more readable to me, but maybe it's not that important though, I'm also fine with the code as-is.

suricactus: > I'm also fine with the code as-is

I will merge like this then :)

PR #1330, opened 2025-08-28
fix(deltas): fix AssertionError when Exception on Delta POST

gounux:
With this, faulty delta files are uploaded to S3 only when the storage is configured with a legacy, thus avoiding the assert to fail when direct access to Object Storage bucket.


suricactus (approved):
Nice and pragmatic one.  Follow-up coming.

PR #1333, opened 2025-09-02
feat(storage): add file storage migration command arguments

gounux:
This PR adds filestorage migration Django command arguments, in order to be able to filter which projects to migrate with more filtering.


duke-nyuki:
Task linked: [QF-6599 Add filestorage migration command arguments](https://app.clickup.com/t/2192114/QF-6599)

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -61,6 +88,18 @@ def handle(self, *args, **options) -> None:
                 return
 
             projects = project_qs.all()
+
+            if updated_until:
+                project_qs = project_qs.filter(updated_at__lt=dt_zoned)
```
gounux: @suricactus am wondering if the `updated_at` is the best candidate date to get the latest update of a project, maybe yes since the datetime auto-updates on save ? (auto_now=True)

suricactus: you can also check date last updated at, which might be more recent. But this should be good enough, if the difference is in hours, I don't see the point to change.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -25,6 +27,16 @@ def add_arguments(self, parser):
         group.add_argument("--all", action="store_true", default=None)
         group.add_argument("--owner-username-startswith", type=str)
         group.add_argument("--owner-username-matches", type=str)
+        group.add_argument(
+            "--updated-until",
+            type=str,
+            help="Migrate only projects updated until this date, using format YYYY-MM-DD. Active only if --all option is active.",
```
```suggestion
            help="Migrate only projects updated until this date, using format YYYY-MM-DD.",
```

I am not convinced if `--all` should be passed, because you can pass `--all` and forget to pass `--updated-until` and then to hate yourself for such accident. Can we just make it mutually exclusive as the rest? Or you really want to combine `--only-community` with `--updated-until`?

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -47,8 +61,21 @@ def handle(self, *args, **options) -> None:
             assert owner_username_startswith is None
             assert owner_username_matches is None
 
+            if updated_until:
+                dt_nozoned = datetime.strptime(updated_until, "%Y-%m-%d")
```
```suggestion
                dt_naive = datetime.fromisoformat(updated_until)
```

This should be fine, no? Also the term they use in Python is `naive`.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -61,6 +88,18 @@ def handle(self, *args, **options) -> None:
                 return
 
             projects = project_qs.all()
```
suricactus: I am a bit confused if here you get all the projects, then why the new variable is called `projects_qs`?

I think it was a general mistake that the variiable is not called `projects_qs` and there should not be any `projects` variable at all, no?

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
gounux: The initial idea was to combine both `--only-community` and `--updated-until`, to filter all the projects actually. Let me find something out.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
gounux: Thanks !

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
gounux: Alright, noted thanks. That's true if the difference is not that important, it should be good enough.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
gounux: I agree, let's stick to the query_set only.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
suricactus: Then you can add something like `--advanced-filter` and require at least one of `--only-community` and `--updated-until`?

gounux: Should be added by this commit : https://github.com/opengisch/QFieldCloud/pull/1333/commits/552c1385faf4cdc795253c7dd3f42c4814eef729

This `--advanced-filter` is applied when the `-all` argument is passed, hoping that what the logger prints is explicit enough to avoid self-hating.

suricactus: I am not sure if we are still there, currently the `--all` shall be passed too?

So imagine you have to write the command and you indend to write:

```
migratefilestorage --all --advanced-filter --only-community
```

But you slip and write just:

```
migratefilestorage --all
```

without the ability to finish your command. What my initial comment here was trying to prevent such situation.

So here we have a few options:

1) 
```
migratefilestorage --advanced-filter --only-community --updated-until DDDD
```

but then why do we even pass `--advance-filter`?

2)
```
migratefilestorage --only-community --updated-until DDDD
```

nice and simple

3)
```
migratefilestorage --all --no-filter
# OR
migratefilestorage --all --only-community --updated-until DDDD
```

so do not accept running just `--all` without any further argument, so the risk of mistake is close to zero. But still, why we need to pass `--all`.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -47,10 +78,43 @@ def handle(self, *args, **options) -> None:
             assert owner_username_startswith is None
             assert owner_username_matches is None
 
-            self.stderr.write(
-                "You are going to migrate all files on this installation."
-                "This will take a long time to finish and should be done only if you know what is going on!"
-            )
+            if not no_filter and not advanced_filter:
+                self.stderr.write(
+                    "Impossible to migrate all projects files without --no-filter or --advanced-filter options."
```
suricactus: I would just change the wording of "impossible" with something else :)

suricactus (approved):
Looks much better and safer now, thanks. One small comment on wording, feel free to merge when you are happy with it too.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
gounux: Option 3 selected, thanks.

PR #1334, opened 2025-09-03
Add dedicated model to store faulty deltafiles

lukasgraf:
This introduces a new model `FaultyDeltaFile` that is used to preserve invalid deltafiles that we failed to parse or process.

When an error happens during parsing a deltafile's JSON, or processing it, the exception is caught and logged. In addition, the contents of the offending deltafile are stored in a FaultyDeltaFile instance, together with the traceback and some metadata that should help debug it.

### Listing view
listing

### Detail view
detail


duke-nyuki:
Task linked: [QF-6600 Add a new model where problematic deltafiles are collected](https://app.clickup.com/t/2192114/QF-6600)

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +155,36 @@ def post(self, request, projectid):
 
         return Response()
 
+    def preserve_faulty_deltafile(self, request_file, project, request):
```
suricactus: Add typing for any new code. One day we may live in paradise and enable strict and have Django native support

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +155,36 @@ def post(self, request, projectid):
...
+        # Be defensive about figuring out the deltafile id - we might not
+        # even have a valid JSON file.
+        try:
+            deltafile_json = json.load(request_file)
+            deltafile_id = deltafile_json.get("id", "unknown")
+        except Exception:
+            deltafile_id = "unknown"
+
+        filename = f"{deltafile_id}.json"
+        user_agent = request.META.get("HTTP_USER_AGENT", "unknown")
```
suricactus: You better store `None` or `""` in case of empty user-agent IMO.
```suggestion
        user_agent = request.META.get("HTTP_USER_AGENT", None)
```

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +155,36 @@ def post(self, request, projectid):
...
+
+        # otherwise we upload an empty file
+        request_file.seek(0)
+
+        timestamped_fn = f"{datetime.now().isoformat()}-{filename}.json"
```
suricactus: Two things here:
1) why don't you not set the filename variable name above the same value as in this line??
2) unfortunately the `_fn` suffix is something that I always associate with `function`, so you better write `filename` to avoid confusion. Too much Javascript in my past life :(

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +155,36 @@ def post(self, request, projectid):
...
+        key = f"projects/{project.id}/deltas/{timestamped_fn}"
+
+        faulty_delta = FaultyDelta(
+            delta_contents=ContentFile(request_file.read(), filename),
+            project=project,
+            user_agent=user_agent,
+        )
+        faulty_delta.save()
+
+        logger.info(f'Invalid deltafile saved as "{key}"')
```
suricactus: I would avoid calculatign the `key` at all costs.

```suggestion
        logger.info(f'Invalid deltafile saved as "{faulty_delta.delta_contents.name}"')
```

docker-app/qfieldcloud/core/views/deltas_views.py
suricactus: 3. the `timestamped_fn` contents might be wrong, as the specific moment when you calculate the key is different here and in `get_faulty_delta_upload_to`, because `datetime.now()` is different time moment.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,46 @@ class Meta:
...
+class FaultyDelta(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    delta_contents = DynamicStorageFileField(
+        _("Faulty Delta"),
+        upload_to=get_faulty_delta_upload_to,
+        max_length=1024,
+        null=False,
+        blank=False,
+    )
```
```suggestion
    deltafile = DynamicStorageFileField(
        _("Deltafile"),
        upload_to=get_faulty_deltafile_upload_to,
        max_length=1024,
        null=False,
        blank=False,
    )
```

I think this is easier naming based on the existing codebase.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,46 @@ class Meta:
...
+        blank=True,
+    )
+
+    def _get_file_storage_name(self) -> str:
+        if self.project.uses_legacy_storage:
+            # Legacy storage - use default storage
+            return "default"
+        else:
+            # Non-legacy storage - use same storage as project
+            return self.project.file_storage
```
```suggestion
        # Legacy storage - use default storage
        if self.project.uses_legacy_storage:
            return "default"
        # Non-legacy storage - use same storage as project
        else:
            return self.project.file_storage
```

The majority of the commentary about if/else in the codebase is before the keyword, not within. No strong preference on one or the other, just a matter of consistency.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,46 @@ class Meta:
...
         ]
+
+
+def get_faulty_delta_upload_to(
+    instance: models.Model,
+    filename: str,
+) -> str:
+    instance = cast(FaultyDelta, instance)
+    key = f"{datetime.now().isoformat()}-{filename}"
+    return f"projects/{instance.project.id}/deltas/{key}"
```
```suggestion
    return f"projects/{instance.project.id}/deltafiles/{key}"
```

I just reralized we better name the prefix with `deltafile` instead of `delta` as we store the deltafile, not a single delta.

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -133,12 +133,8 @@ def post(self, request, projectid):
...
 
         except Exception as err:
-            if request_file and settings.LEGACY_STORAGE_NAME:
-                key = f"projects/{projectid}/deltas/{datetime.now().isoformat()}.json"
-                # otherwise we upload an empty file
-                request_file.seek(0)
-                utils.get_s3_bucket().upload_fileobj(request_file, key)
-                logger.info(f'Invalid deltafile saved as "{key}"')
+            if request_file:
+                self.preserve_faulty_deltafile(request_file, project_obj, self.request)
```
suricactus: Just curious if there is a need to explicitly seek back to 0 position as we did before?

docker-app/qfieldcloud/core/views/deltas_views.py
lukasgraf: Yep, this wasn't intended like that, somehow I accidentally duplicated deriving the key, I'll definitely rework this bit.

docker-app/qfieldcloud/core/models.py
lukasgraf: Yep. I think I will even go as far to also rename the model to `FaultyDeltaFile`. The difference between the two concepts is already quite subtle, no point in confusing people by using inconsistent terminology.

docker-app/qfieldcloud/core/views/deltas_views.py
lukasgraf: I'm quite certain we do.
At least, nothing has changed regarding what state the `request_file` file handle will be in when it reaches the code that preserves the deltafile. If there was an chance before where the file pointer could be at a non-0 position that chance still exists after my change. E.g. just if an attempt has been made to parse the JSON, valid or not, the file pointer will already have been moved at that point.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,48 @@ class Meta:
...
+    )
+
+    def _get_file_storage_name(self) -> str:
+        # Legacy storage - use default storage
+        # TODO Delete with QF-4963 Drop support for legacy storage
+        if self.project.uses_legacy_storage:
+            return "default"
+        # Non-legacy storage - use same storage as project
+        else:
+            return self.project.file_storage
```
```suggestion

        # Non-legacy storage - use same storage as project
        return self.project.file_storage
```

Generally we prefer early returns in the codebase. Especially with the above if planned to be deleted, make a minimal diff :)

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,48 @@ class Meta:
...
+    instance: models.Model,
+    filename: str,
+) -> str:
+    instance = cast(FaultyDeltaFile, instance)
+    key = f"{datetime.now().isoformat()}-{filename}"
+    return f"projects/{instance.project.id}/deltafiles/{key}"
+
+
+class FaultyDeltaFile(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
```suggestion
	# The deltafile id if parseable UUID value
    deltafile_id = models.UUIDField(editable=False, null=True)
```

Note that a faulty client might try to push the same delta more than once and then we may have the same UUID pushed more than once, therefore an exception in the except block when we create the FaultyDelta. As per model ID, we don't mind using the Django default one.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,48 @@ class Meta:
...
+
+    deltafile = DynamicStorageFileField(
```
suricactus, line 2631:
```suggestion
	# The deltafile contents as submitted from the client. It might contain non-valid JSON.
    deltafile = DynamicStorageFileField(
```
Whenever we add a new model, we always try to add some hints for the dev what each of the field means. Please add for the rest below and get inspired from across the codebase.

docker-app/qfieldcloud/core/models.py
```diff
@@ -2609,3 +2609,48 @@ class Meta:
...
+        _("Deltafile"),
+        upload_to=get_faulty_deltafile_upload_to,
+        max_length=1024,
```
suricactus, line 2635:
```suggestion
        # the s3 storage has 1024 bytes (not chars!) limit: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
        max_length=1024,
```

Just to highlight it is not an arbitrary choice of 1024.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1606,6 +1643,16 @@ class LogEntryAdmin(
     list_filter = ("action", QFieldCloudResourceTypeFilter)
 
 
+class FaultyDeltaFilesAdmin(admin.ModelAdmin):
```
suricactus, line 1611: The most interesting here for debugging would be to be able to see the faulty project and to be able to search by project name and project owner username.

You can add somethings like this (copied and modified from existing code):
```
    search_fields = (
        "project__id__iexact",
        "project__name__icontains",
        "project__owner__username__icontains",
    )
```

Also add the `project` column in the list display.

Btw, if you are not aware, check `list_select_related` as an admin configuration to make the whole thing fast :)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -904,6 +940,7 @@ class ProjectAdmin(QFieldCloudModelAdmin):
     inlines = (
         ProjectSecretInline,
         ProjectCollaboratorInline,
+        FaultyDeltaFilesInline,
```
suricactus: I am here to blame, but the more I think about it, the more I am convinced that `Inline` might cause a lot of troubles performance wise, as the project will get more and more faulty deltas, but the admin page does not support pagination on inline models and can downgrade the UX of the project admin page.

What we can do as quick switch between the project and the related faulty deltafiles to add a new link here: https://github.com/opengisch/QFieldCloud/blob/57c005c23ee03a516e458577834baae14a47cfa9/docker-app/qfieldcloud/core/templates/admin/project_change_form.html#L8
This approach is used in several places with great success.

suricactus: Visual hint:

Image

docker-app/qfieldcloud/core/admin.py
```diff
@@ -614,6 +616,40 @@ class ProjectCollaboratorInline(admin.TabularInline):
...
+    readonly_fields = (
+        "created_at",
+        "short_file_link",
+        "user_agent",
+    )
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def short_file_link(self, obj):
```
```suggestion
    @admin.display(description=_("Faulty deltafile"))
    def short_file_link(self, obj: FaultyDeltaFile) -> str:
```

Few things here:

- we always prefer to use `@admin.display` over `<property>.{short_descirption|ordering|...}`. Cleaner to read IMO and already established convention throughout the code.
- always try to translate stuff strings. If you find untraslated string, open a side PR that translates it. 
- we agreed on "DeltaFile" for pascal/camel case (ok to revise, but I don't think it is needed), but lowercase or human text it should be "deltafile" or "Deltafile". Why? Because established convention and just to reduce confusion (including QField and docs).
- any new code besides tests should have typing for in and out params, including django admin.

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +156,35 @@ def post(self, request, projectid):
...
+        # File contents might already have been partially read by JSON parser.
+        # Rewind to avoid uploading an empty file.
+        request_file.seek(0)
+
+        faulty_deltafile = FaultyDeltaFile(
+            deltafile=ContentFile(request_file.read(), filename),
+            project=project,
+            user_agent=user_agent,
+        )
+        faulty_deltafile.save()
```
suricactus:
```suggestion
        faulty_deltafile = FaultyDeltaFile.objects.create(
            deltafile=ContentFile(request_file.read(), filename),
            project=project,
            user_agent=user_agent,
        )
```

There is the philosophical discussion over `Model.save()` vs `Model.objects.create`. The existing codebase always prefers the latter, and the former is for very specific use cases.

docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -159,6 +156,35 @@ def post(self, request, projectid):
 
         return Response()
 
+    def preserve_faulty_deltafile(
+        self, request_file: IO, project: Project, request: HttpRequest
+    ):
```
suricactus:
```suggestion
    ) -> FaultyDeltaFile:
```

IMO it's always a good idea to return the thing that you have just created. Makes the life of future you much easier.

docker-app/qfieldcloud/core/models.py
suricactus: After I read the whole PR again, I discovered a misunderstanding the purpose of the field at the current state of the codebase from my side. That's why early reviews are dangerous.

If you want to store the `deltafile_id` here - ok to proceed.
If you want to add other than the default (big)int PK in the model, namely a UUID, you better not. First - more code, more maintanence. Second, it will be confusing with all the UUIDs involved - Project UUID, DeltaFile UUID, Delta UUID and now FaultyDeltaFile UUID.
It might be something else that I cannot forsee now, please excuse me in that case.

lukasgraf: Ok, gotcha, your second comment cleared up some confusion for me :P

So, my intent for the field in question was indeed for it to be the (UUID based) primary key for the `FaultyDeltaFile` model. I went with an `UUIDField` because that looked like a pattern that most models use, but if I understand you correctly, there's no need - just remove `id` alltogether and let Django hand out the default (big)int PK for `FaultyDeltaFile` instances.

In regards to storing the `deltafile_id` I had a bit of a brainfart :D I thought, no need to store it twice, we already have a ForeignKey / relationship. But we obviously don't (or can't, since deltafiles don't have their own model). So I'll add the `deltafile_id` like you suggested, and drop the custom `id`. 👍

suricactus: Great!

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1606,6 +1608,67 @@ class LogEntryAdmin(
...
+        if obj.deltafile:
+            filename = basename(obj.deltafile.name)
+            return format_html(
+                '<a href="{}" target="_blank">{}</a>',
+                obj.deltafile.url,
+                filename,
+            )
+        return "-"
+
+    @admin.display(ordering="project__owner")
```
lukasgraf, line 1660: I took the liberty of harmonizing the display of the traceback, project name and owner here in the same way that it's done for the `Job` admin listing view (with the difference that `project` is nullable here).

suricactus: Thanks!

docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -1167,13 +1191,15 @@ def get_delta_file_with_project_id(self, project, delta_filename):
             json_str = json.dumps(deltafile)
             return io.StringIO(json_str)
 
-    def upload_deltas(self, project, delta_filename):
+    def upload_deltas(self, project, delta_filename, headers=None):
```
suricactus: Since you touched this, please add some typings.

docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -1167,13 +1191,15 @@ def get_delta_file_with_project_id(self, project, delta_filename):
...
+        headers = headers if headers else {}
```
suricactus:
```suggestion
        if not headers:
            headers = {}

```

We discourage the inline if statements whenever possible. Here it is ok, it's short, but for other places we avoid them.

Here we can debate if we need explicit check for `headers is None` or not, in general we are not so strict and it is a matter of taste.

Also something you might have seen in most of the places, we try to add some "air" before and after blocks or keyswords. Just skim through some recent code (e.g. `filestorage` app) and you will spot the pattern.

docker-app/qfieldcloud/core/tests/test_delta.py
lukasgraf: I swear, at some point I will stop forgetting this 😜

docker-app/qfieldcloud/core/tests/test_delta.py
lukasgraf: I checked the signature of `client.post()`, and `headers` defaults to `None` there anyway, and therefore is handled accordingly. So I just removed the None-check alltogether.

suricactus:
@lukasgraf would you mind adding the same info to the public `README.md`?

I think we can actually reduce the contents of this README in favor of the public one, feel free to do as a side quest :)

PR #1335, opened 2025-09-04
edit(packaging): check and throw if latest packaging job is in progress

gounux:
This PR intends, when downloading latest package, to add a check if the latest package job is still ongoing.

Defensive approach proposed here by throwing an error if the latest package job is ongoing, fully open to discuss it and set up another approach.

### DIFF
--- a/docker-app/qfieldcloud/core/views/package_views.py
+++ b/docker-app/qfieldcloud/core/views/package_views.py
@@ -96,6 +96,16 @@ class LegacyLatestPackageView(views.APIView):
                 "Packaging has never been triggered or successful for this project."
             )
 
+        # Check if the latest package job is still in progress
+        if latest_package_job.status in [
+            PackageJob.Status.PENDING,
+            PackageJob.Status.QUEUED,
+            PackageJob.Status.STARTED,
+        ]:
+            raise exceptions.InvalidJobError(
+                "Packaging is still in progress, retry later when the packaging is finished."
+            )
+
         filenames = set()
         files = []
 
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/models.py
```diff
@@ -1337,22 +1337,25 @@ def latest_package_job_for_user(self, user: User) -> PackageJob | None:
...
-            jobs_qs.filter(
-                type=Job.Type.PACKAGE,
-            )
-            .order_by("-created_at")
-            .first()
+        return jobs_qs.filter(
+            type=Job.Type.PACKAGE,
         )
 
+    def latest_package_job_for_user(self, user: User) -> PackageJob | None:
```
suricactus, line 1372: Why don't we add latest finished package  job as a function as you already use it twice? Also do you use the current function somewhere?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1337,22 +1337,25 @@ def latest_package_job_for_user(self, user: User) -> PackageJob | None:
...
-
-        return (
-            jobs_qs.filter(
-                type=Job.Type.PACKAGE,
-            )
-            .order_by("-created_at")
-            .first()
+        return jobs_qs.filter(
+            type=Job.Type.PACKAGE,
         )
```
suricactus: I find it a bit weird that we filter about the type in the very end of the function, should it be among the first things we do in the function that is named `package_jobs_*`?

Also if you already have the queryset in a variable, always prefer to just return the variable instead of building a new expression to return.

gounux: Alright, thanks. Applied in upcoming commit.

docker-app/qfieldcloud/core/models.py
gounux: Indeed, `latest_finished_package_job_for_user` introduced in upcoming commit, that does the status filtering and ordering, thanks !

gounux: > do you use the current function somewhere?

The `needs_repackaging` is currently using this `latest_package_job_for_user` function, since it's now on the user level. During the `needs_repackaging` computation, there is no `Job.status` check, the only check so far is if the Job's `finished_at` is set. If not it basically says "yes this project needs repackaging for this user".  

So I presume that if the latest package job is ongoing and not finished yet, `needs_repackaging` will be true.  Is this what we want ?

suricactus: > So I presume that if the latest package job is ongoing and not finished yet, needs_repackaging will be true. Is this what we want ?

I think at that stage in `needs_repackaging` we have no idea if the status of the job will be success, so it is always safer to trigger more jobs rather than giving false promises.

PR #1336, opened 2025-09-04
edit(fs migration): update command error handling

gounux:
This PR intends to make the filestorage migration command run through all the filtered projects, by catching exceptions so that the command can be monitored.

ping @toebivankenoebi

### DIFF
--- a/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
+++ b/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
@@ -1,4 +1,5 @@
 import logging
+import traceback
 import uuid
 from collections.abc import Collection
 from datetime import datetime
@@ -21,6 +22,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument("--force", action="store_true", default=False)
         parser.add_argument("--accept", action="store_true", default=False)
+        parser.add_argument("--no-raise", action="store_true", default=False)
 
         group = parser.add_mutually_exclusive_group()
 
@@ -57,6 +59,7 @@ class Command(BaseCommand):
 
         force: bool = options.get("force", False)
         accept: bool = options.get("accept", False)
+        no_raise: bool = options.get("no_raise", False)
 
         # these should be alternative to each other
         project_id = options.get("project_id")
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -96,7 +97,9 @@ def migrate_project_storage(
         )
 
         if not project_files:
-            raise Exception("No files to migrate")
+            logger.warning(
+                f'No files to migrate for project "{project.name}" ({str(project.id)})!'
+            )
```
suricactus, line 102: If `no_raise` is not needed here?

gounux: That is a good question : do you think we should raise if there is no file to migrate ?

suricactus: probably not? I didn't pay attention in the code so much, so probably this is good.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -193,4 +196,14 @@ def handle(self, *args, **options) -> None:
                 f'⏳ {index}/{projects_count}: migrating storage for project "{project.name}" ({project.id}) from "{project.file_storage}" to "default".'
             )
 
-            migrate_project_storage(project, DEFAULT_STORAGE, force)
+            try:
+                migrate_project_storage(project, DEFAULT_STORAGE, force)
+            except Exception as e:
```
```suggestion
            except Exception as err:
```

Avoid single char names, `e` migth mean a lot of things :)

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -193,4 +196,14 @@ def handle(self, *args, **options) -> None:
...
+                self.stderr.write(
+                    f"Error when migrating project '{project.name}' ({project.id}): {e}"
+                )
+
+                if no_raise:
+                    self.stderr.write(traceback.format_exc())
+                else:
+                    raise e
```
suricactus: I have recently discovered, that some people even prefer not having which exception is being raised, and have just `raise` instead of `raise err`. I leave it to your personal taste for now.

suricactus (approved):

PR #1337, opened 2025-09-04
edit(delta): fix delta applied by several jobs when users pushing around the same time

gounux:
This PR should avoid too many `pending_deltas` hanging around, by making sure to exclude the ones which are already to be applied in ApplyDelta type jobs, for the project.


duke-nyuki:
Task linked: [QF-5316 Fix when a delta is applied by several jobs when different users are pushing around the same time](https://app.clickup.com/t/2192114/QF-5316)

suricactus (approved):
Thank you for spotting this.

Important reminder why careful reviews are required.

PR #1340, opened 2025-09-09
docs: add docs about dump restore, improve the instance launch, add info how to access the database

suricactus:
This is related to the overall improvement of the README file. Note this PR could be merged ignoring some upcoming documentation improvements about testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,37 +43,41 @@ To fetch upstream development, don't forget to update the submodules too:
 
 ### Launch a local instance
 
-Copy the `.env.example` into `.env` file and configure it to your
-desire with a good editor:
+1. Copy the `.env.example` into `.env` file:
 
     cp .env.example .env
-    emacs .env
 
+2. Change the `ENVIRONMENT` variable to `development`.
 
-Make sure the host's firewall allows port _8009_, required by the `minio` service. Failing to meet this requirement is likely to result in the service being unable to start.
+    ENVIRONMENT=development
 
-To build development images and run the containers:
+3. Build development images and run the containers:
 
     docker compose up -d --build
 
-It will read the `docker-compose*.yml` files specified in the `COMPOSE_FILE`
-variable and start a django built-in server at `http://localhost:8011`.
+The command will read the `docker-compose*.yml` files specified in the `COMPOSE_FILE` variable from the `.env` file. Then Django built-in server will be directly reachable at `http://localhost:8011` or through `nginx` at `https://localhost`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,6 +89,39 @@ If there is some kind of problem, first check the `nginx` and `app` logs, usuall
...
+    docker compose exec -it db psql -U qfieldcloud_db_admin -d qfieldcloud_db
+
+Or by creating `~/.pg_service.conf` in their user home directory and appending:
+
+    [localhost.qfield.cloud]
+    host=localhost
+    dbname=qfieldcloud_db
+    user=qfieldcloud_db_admin
+    port=5433
+    password=3shJDd2r7Twwkehb
```
lukasgraf, line 129: Just a nitpick:
Passwords only used for development / testing, that are _published_, should actually be something that makes them recognizable as such - something like `secret`, `demo` or `password`. Random looking passwords give a false sense of security, and it makes them easily confused with _real_ passwords that need to be kept secret. Also, they often get flagged by security checks that check for accidentally published credentials (which would ignore obvious placeholder passwords).

Though, since those are already defined in `.env.example`, maybe out of scope for this PR.

README.md
```diff
@@ -43,38 +43,42 @@ To fetch upstream development, don't forget to update the submodules too:
 
 ### Launch a local instance
 
-Copy the `.env.example` into `.env` file and configure it to your
-desire with a good editor:
+1. Copy the `.env.example` into `.env` file:
```
lukasgraf, line 46: The change to the numbered bullet points leads to the code snippets in these section not being highlighted automatically any more (from what I can tell from the [rich preview](https://github.com/opengisch/QFieldCloud/blob/1ac714e2de450dfb7094230402ab3c759e648e9c/README.md#launch-a-local-instance)).

Probably needs some fiddling with whitespace (indentation and/or newlines) to make the automagic happen, or explicit triple-backticks.

README.md
suricactus: Yeah, I do agree with this. If ok for you, let's just merge as it is and happy to have a follow-up with more obvious "fake" password. One thing I find confusing in some docs is when the example is:
1. `user=demo password=demo` - you don't know what `demo` refers to.
2. `password=password` - you don't know what the `pasword` refers to and which side to change.

README.md
suricactus: Nice spot! Sorry for that, my local markdown previewer did not show this. Fixing right away.

README.md
lukasgraf: Yep, totally fine. Also agree with your points - I usually like `secret` for that reason ;)

lukasgraf (approved):

PR #1342, opened 2025-09-09
feat: added django-stubs and drf-stubs

suricactus:
Make typing great again! In theory we should add the typing in a
separate requirements file for development, but we already did some
mess with boto3 and we better not try to fix it now.

@lukasgraf FYI to make your vscode experience better


lukasgraf (approved):
Yay, LGTM! 👍 🎉

PR #1345, opened 2025-09-10
edit(fs migration): use harmonised logger during fs migration

gounux:
This PR intends to make the migration command use harmonized logging for everything that needs to be logged during one run.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
+++ b/docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
@@ -82,27 +82,28 @@ class Command(BaseCommand):
             assert owner_username_matches is None
 
             if not no_filter and not advanced_filter:
-                self.stderr.write(
+                logger.error(
                     "Can not migrate all projects files without --no-filter or --advanced-filter options."
                 )
                 exit(1)
 
             if no_filter and advanced_filter:
-                self.stderr.write(
+                logger.error(
                     "Can not use --no-filter and --advanced-filter options together."
                 )
                 exit(1)
 
             if no_filter:
-                self.stderr.write(
+                logger.info(
+                    "Running migration on all projects with --no-filter set."
                     "You are going to migrate all files on this installation, without any project filtering!"
                     "This will take a long time to finish and should be done only if you know what is going on!"
                 )
 
             if advanced_filter:
                 if not updated_until and not only_community:
-                    self.stderr.write(
-                        "You are using advanced filtering, but no specific filters are set. Please use one."
+                    logger.error(
+                        "--advanced-filter set but mandatory options missing, aborting."
                     )
                     exit(1)
 
@@ -110,21 +111,15 @@ class Command(BaseCommand):
                     dt_naive = datetime.fromisoformat(updated_until)
                     dt_zoned = timezone.make_aware(dt_naive)
 
-                    self.stderr.write(
-                        f"You are going to migrate projects updated until '{dt_zoned}'."
-                    )
+                    logger.info(f"Migrating projects updated until '{dt_zoned}'.")
 
                 if only_community:
-                    self.stderr.write(
-                        "You are going to migrate projects owned by community accounts."
-                    )
+                    logger.info("Migrating projects owned by community accounts.")
 
             if not accept:
                 yes_no = input("Are you sure you want to continue? [y/n]\n")
                 if yes_no != "y":
-                    self.stderr.write(
-                        "The files migration will not happen, probably a good choice!"
-                    )
+                    logger.info("File storage migration aborted by user.")
                     return
 
             if advanced_filter:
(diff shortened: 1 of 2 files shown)


suricactus:
I leave this to be reviewed by @toebivankenoebi . I only have one question - why do we want to use the same logger in the command as in the other file? What is the benefit? IMO we can do two separate loggers and it will be easily distingueshed in the log file, but I might be missing something.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -9,7 +9,7 @@
 from qfieldcloud.core.models import Project
 from qfieldcloud.filestorage.migrate_project_storage import migrate_project_storage
 
-logger = logging.getLogger(__name__)
+fs_migration_logger = logging.getLogger(__name__)
```
suricactus: Well, then the question why it is called `fs_migration_logger` but not just `logger`?

gounux: True :) thanks for spotting this,

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -110,19 +110,19 @@ def handle(self, *args, **options) -> None:
...
+                    logger.info(
                         "You are going to migrate projects owned by community accounts."
                     )
 
             if not accept:
                 yes_no = input("Are you sure you want to continue? [y/n]\n")
                 if yes_no != "y":
-                    self.stderr.write(
+                    logger.info(
                         "The files migration will not happen, probably a good choice!"
```
toebivankenoebi:
```
File storage migration aborted by user.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -82,26 +82,26 @@ def handle(self, *args, **options) -> None:
...
             if no_filter and advanced_filter:
-                self.stderr.write(
+                logger.error(
                     "Can not use --no-filter and --advanced-filter options together."
                 )
                 exit(1)
 
             if no_filter:
-                self.stderr.write(
+                logger.info(
```
toebivankenoebi, line 97: Simplify the info message, e.g.
```
Running migration on all projects with "--no-filter" set.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -82,26 +82,26 @@ def handle(self, *args, **options) -> None:
...
                     "You are going to migrate all files on this installation, without any project filtering!"
                     "This will take a long time to finish and should be done only if you know what is going on!"
                 )
 
             if advanced_filter:
                 if not updated_until and not only_community:
-                    self.stderr.write(
+                    logger.error(
```
toebivankenoebi, line 105:
```
--advanced-filter set but mandatory options missing, aborting.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -110,19 +110,19 @@ def handle(self, *args, **options) -> None:
                     dt_naive = datetime.fromisoformat(updated_until)
                     dt_zoned = timezone.make_aware(dt_naive)
 
-                    self.stderr.write(
+                    logger.info(
```
toebivankenoebi:
```
Migrating projects updated until '{dt_zoned}'.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -110,19 +110,19 @@ def handle(self, *args, **options) -> None:
...
                         f"You are going to migrate projects updated until '{dt_zoned}'."
                     )
 
                 if only_community:
-                    self.stderr.write(
+                    logger.info(
```
toebivankenoebi:
```
Migrating projects owned by community accounts.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -162,48 +162,46 @@ def handle(self, *args, **options) -> None:
                 owner__username=owner_username_matches,
             )
         else:
-            self.stderr.write(
+            logger.error(
```
toebivankenoebi, line 160:
```
Filters "--project-id", "--all", "--owner-username-startswith", "--owner-username-matches" must be mutually exclusive, aborting.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -162,48 +162,46 @@ def handle(self, *args, **options) -> None:
...
                 "You must pass exactly one of filter arguments: --project-id, --all, --owner-username-startswith, or --owner-username-matches!"
             )
 
             exit(1)
 
         projects_count = project_qs.count()
         if projects_count == 0:
-            self.stderr.write("No projects match the passed filters!")
+            logger.error("No projects match the passed filters!")
```
toebivankenoebi:
```
No projects match the passed filters, aborting.
```

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -162,48 +162,46 @@ def handle(self, *args, **options) -> None:
...
 
             exit(1)
 
-        self.stderr.write(
-            f"The storage migration will affect {len(projects)} project(s):"
-        )
+        logger.info(f"The storage migration will affect {len(projects)} project(s):")
 
-        self.stderr.write("ID\tNAME")
+        logger.info("ID\tNAME")
```
toebivankenoebi: Would remove, no need for csv style header in log file.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -162,48 +162,46 @@ def handle(self, *args, **options) -> None:
...
         for project in project_qs:
-            self.stderr.write(f"{project.id}\t{project.owner.username}/{project.name}")
+            logger.info(f"{project.id}\t{project.owner.username}/{project.name}")
```
toebivankenoebi: Would choose a different separator than \t for log files.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -162,48 +162,46 @@ def handle(self, *args, **options) -> None:
...
 
         if not accept:
             yes_no = input("Are you sure you want to continue? [y/n]\n")
 
             if yes_no != "y":
-                self.stderr.write(
+                logger.info(
```
toebivankenoebi: Same as above.

toebivankenoebi (commented):
Would change log messages to be a bit more generic, less user focused.

docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
suricactus: Just wondering, was it on purpose that the change was not suggested using the GH suggestion mechanism:

```suggestion
            logger.error("No projects match the passed filters, aborting.!")
```

image

toebivankenoebi: Nope, that's much neater, wasn't aware, thanks!

suricactus: then I take the opportunity to suggest multiline comments with click, hold, and drag down to comment/suggest for multiple lines. :)

PR #1350, opened 2025-09-13
chore: add debugger config for vscode and update README.md

suricactus:
The debugger attaches to already running docker compose stack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,16 +253,17 @@ debugpy.wait_for_client()  # optional
 
 3. Or permanently change the command in `docker-compose.override.local.yml` to add the `--wait-for-client`.
 
-To add breakpoints in vendor modules installed via `pip` or `apt`, you need a copy of their source code.
+To add breakpoints in vendor modules installed via `pip` or `apt`, you need a copy of their source code on your host machine.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -280,23 +264,6 @@ The configuration for the vendor modules is the second object in the example `pa
```
lukasgraf, line 265: The unchanged sentence above 

> The configuration for the vendor modules is the second object in the example `pathMappings` above, as well as setting `justMyCode` to `false`.

now refers to an example that isn't in `.vscode/launch.json` any more.

Maybe add back those two parts to the provided `launch.json`? Though I suspect you had a good reason to remove them 😉 , so maybe it's the wording that needs adjusting.
```
                {
                    "localRoot": "${workspaceFolder}/docker-app/site-packages",
                    "remoteRoot": "/usr/local/lib/python3.10/site-packages/"
                },
```

and
```
"justMyCode": false,
```

suricactus: Thanks for spotting. No hidden agenda, pure overlook. I have added the example now, it get's a bit intense in `.vscode/settings.json` with all the comments, but I believe whoever looks there should be professional enough to orientate.

PR #1351, opened 2025-09-13
typing: added typing to dequeue.py

suricactus:


docker-app/qfieldcloud/authentication/models.py
```diff
@@ -28,7 +28,7 @@ class ClientType(models.TextChoices):
         UNKNOWN = "unknown", _("Unknown")
 
     @staticmethod
-    def guess_client_type(user_agent: str) -> ClientType:
+    def guess_client_type(user_agent: str) -> tuple[str, str]:
```
gounux: I'm wondering about this typing change, which I found more complicated to read then before, but maybe I'm missing something.  
What is the reason behind this ?

suricactus: It was a faulty version of mypy/ruff/whatever that pointed me in that direction. Reverting the change.

gounux (approved):
Code looks good, thanks !

PR #1356, opened 2025-09-16
feat(nginx): Cleaner nginx configuration, production ready

toebivankenoebi:
Proxy settings are DRY.
Static content /pages, /staticfiles, /mediafiles is directly served by nginx without try_files (performance.
API requests are passed to django, no custom error pages. 500 errors are handled based on debug variable (either show custom page or stacktrace).

Few things to consider:
- `@handle_500_error` still exists since an if statement is not allowed in `location /`
- `collectstatic` is now mandatory to run even if `debug=true` since `try_files` logic has been removed


toebivankenoebi:
@suricactus : What would be the easiest way to generate a 500 error for testing?
@lukasgraf : Do you want to take a look at the config and give your opinion?

suricactus:
@toebivankenoebi see the attached sentry error on clickup, it is actual way to reproduce.

toebivankenoebi:
@suricactus that is for /api/ and should be fixed with this PR since we're just passing everyhing directly to django and don't intercept / overwrite. I wanted to test the 500 error page behavior when using django admin. Any hints there?

docker-nginx/templates/default.conf.template
```diff
@@ -98,43 +98,31 @@ server {
...
-
-    # Proxy settings
-    proxy_connect_timeout 5s;
-    proxy_read_timeout 300;
-    proxy_send_timeout 300;
-    proxy_redirect off;
-
-    proxy_intercept_errors on;
+  location ~ ^/(staticfiles|pages|mediafiles)/ {
+    try_files $uri =404;
```
suricactus, line 120: We still need to proxy to Django for local development if the file is not available, otherwise we as devs are force to run `collectstatic` every time and also mount the volumes from `docker-compose.override.prod.yml` .

toebivankenoebi:
After discussing it with @suricactus in person, handling of staticfiles is still handled with try_files and forwarded to django (for ease of development).

Tested error 500 by adding 1/0 division in admin.py: 
- Custom error page shown successfully when debug=false
- Django stacktrace shown successfully when debug=true

Tested /api/ calls:
- /api/ still works as expected
- Gets routed directly to django app
- Returns proper json in case of error/not found, no custom error pages shown

toebivankenoebi:
Successfully tested in Nyuki as well.

PR #1357, opened 2025-09-16
Use variable storage key for thumbnails

lukasgraf:
Using a variable storage key that changes every time for thumbnails is an easy way to avoid thumbnail files exceeding the maximum allowed versions on object storages with versioning enabled.

Before, when a fixed storage key like `'projects/{proj}/meta/thumbnail.png'` was used, this created a new version on the object storage level for every new thumbnail. This has lead to situations where that thumbnail file then exceeded the maximum amount of 1000 versions, and caused issues.

By using a variable key we save every new thumbnails as a new object, with always just a single version. When `project.thumbnail` is updated, and the reference to the old thumbnail file is removed, the old file will be marked as deleted. These deleted files can in the future then be cleaned up with a script or a maintenance job.


docker-app/qfieldcloud/filestorage/tests/test_files_api.py
```diff
@@ -1055,3 +1055,14 @@ def test_delete_project_deletes_thumbnail_and_all_project_files(self):
...
+    def test_thumbnail_storage_key_is_variable(self):
+        self.p1.thumbnail = ContentFile("<svg />", "thumbnail.svg")
+        self.p1.save()
+        key_one = self.p1.thumbnail.file.name
+
+        self.p1.thumbnail = ContentFile("<svg />", "thumbnail2.svg")
+        self.p1.save()
+        key_two = self.p1.thumbnail.file.name
+
+        assert key_one != key_two
```
suricactus:
```suggestion
        self.assertNotEqual(key_one, "thumbnail.svg")
        self.assertNotEqual(key_two, "thumbnail2.svg")
        self.assertNotEqual(key_one, key_two)
```

A few small remarks here:

1. All the existing tests are using some of the `self.assert*` functions instead of the raw `assert` keyword.
2. Added two more conditions that assert the actual thumbnail is different from the one they were uploaded with, I think it makes sense to ensure the filename would always be different?
3. Question on uploading SVG, don't we always get a `png` in the end? Probably makes sense to test for `self.assertNotEqual(key_one, "thumbnail.png")` too.
4. Very little one on style, usually we suffix stuff with a number `1`, `2` e.g. `user1`, `user2` etc instead of text. I would also suggest to change to some more descriptive name than `key`, adding `thumbnail` in the name might be better. 

I once won a championship on nitpicking :) .

suricactus (approved):
Great and pragmatic fix, thanks!

docker-app/qfieldcloud/filestorage/tests/test_files_api.py
lukasgraf: Aaah, damn, too much switching between `pytest` based tests using the `assert` keyword, and classic `TestCase` based ones 🙈 Addressed that and the other points 👍

lukasgraf:
I think I will hold off on merging this one before the release, until we have decided how to identify and treat existing thumbnail files that need fixing (see [QF-6749](https://app.clickup.com/t/2192114/QF-6749)).

PR #1358, opened 2025-09-16
fix(admin): revive the django axes admin

suricactus:
Because we need to register it to the newly added `qfc_admin_site`.


duke-nyuki:
Task linked: [QF-6787 Revive django-axes in QFieldCloud admin](https://app.clickup.com/t/2192114/QF-6787)

suricactus:
I checked visually comparing with prod, nothing else was detected.

PR #1360, opened 2025-09-17
storage: make project's `file_storage` configurable in admin when there is no file

gounux:
Basically the title.

When creating a new "empty" project (with no files yet) -> file storage is editable through a ChoiceField :

image

When there is at least one file related to the project -> the ChoiceField is not editable :

image


duke-nyuki:
Task linked: [QF-6783 Make the project storage configurable when there is no file](https://app.clickup.com/t/2192114/QF-6783)

PR #1361, opened 2025-09-17
edit(fsmigration): use a project's `locked_at` instead of `is_locked`

gounux:
With this PR, when migrating a project's file storage, the Django command is making use of a `locked_at` datetime field rather than a `is_locked` boolean field. Thus giving more information about the locked state with when the "lock flag" was set. If a project's `locked_at` is not null, it's similar to the former `is_locked = True`.


docker-app/qfieldcloud/filestorage/management/commands/migratefilestorage.py
```diff
@@ -55,8 +54,6 @@ def add_arguments(self, parser):
         )
 
     def handle(self, *args, **options) -> None:
-        projects: Collection = []
```
suricactus, line 58: Why pre-commit didn't catch this?

suricactus (approved):
Looks good to me.

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -198,6 +208,7 @@ def migrate_project_storage(
                 django_thumbnail_file,
             )
             project.thumbnail = django_thumbnail_file  # type: ignore
+            project.legacy_thumbnail_uri = ""
```
gounux: Assuming here the `legacy_thumbnail_uri` field has to be empty after the migration.

suricactus: I don't think it should be. Do we win something by emptying it? It is scheduled for droping anyways.

If you empty it, it means we do not have a recovery plan for thumbnail, no?

gounux: > If you empty it, it means we do not have a recovery plan for thumbnail, no?

That's right, I agree.

It's just that I tried to delete such a migrated project (so on default project, and having `legacy_thumbnail_uri` not null), there was an error here :

https://github.com/opengisch/QFieldCloud/blob/28f3f1ee6c02e06c0a4afb8f3b09581f9db351ee/docker-app/qfieldcloud/core/models.py#L1856-L1857

That would probably make more sense then to check with `uses_legacy_storage` before manually deleting the thumbnail of a legacy storage project ?

Addressed by https://github.com/opengisch/QFieldCloud/pull/1361/commits/adf1f92a0ed46097f405375b226fa50be2fe6ced

suricactus: Yes, now looks better! But still needs to not reset the `legacy_thumbnail_url` field.

PR #1367, opened 2025-09-21
feat(worker): make qgis jobs defined in a single file

suricactus:
This PR addressed multiple problems:

- job definition requires changes in multiple files - `entrypoint.py`, custom file etc
- extending QFieldCloud with custom jobs is impossible unless `entrypoint.py` is overwritten
- testability of the qgis worker as a standalone module without going through the worker wrapper is cumbersome
- some files (e.g. apply_deltas.py) defined two CLI interfaces to be called - one via `entrypoint.py`, and one by running the file itself.

### DIFF
--- a/docker-app/worker_wrapper/wrapper.py
+++ b/docker-app/worker_wrapper/wrapper.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import Any, Iterable
 
 import docker
+import docker.errors
 import requests
 import sentry_sdk
 from constance import config
@@ -18,8 +19,6 @@ from django.core.files.base import ContentFile
 from django.db import transaction
 from django.forms.models import model_to_dict
 from django.utils import timezone
-from docker.client import DockerClient
-from docker.errors import APIError
 from docker.models.containers import Container
 from qfieldcloud.authentication.models import AuthToken
 from qfieldcloud.core.models import (
@@ -648,7 +647,7 @@ class ProcessProjectfileJobRun(JobRun):
 
 
 def cancel_orphaned_workers() -> None:
-    client: DockerClient = docker.from_env()
+    client: docker.client.DockerClient = docker.from_env()
 
     try:
         running_workers: list[Container] = client.containers.list(
@@ -677,6 +676,6 @@ def cancel_orphaned_workers() -> None:
             container.kill()
             container.remove()
             logger.info(f"Cancel orphaned worker {worker_id}")
-        except APIError:
+        except docker.errors.APIError:
             # Container already removed
             pass


docker-app/qfieldcloud/core/management/commands/dequeue.py
```diff
@@ -114,13 +114,16 @@ def handle(
             if single_shot:
                 break
 
-    def _run(self, job: Job) -> None:
-        job_run_classes: dict[Job.Type, type[JobRun]] = {
+    def get_job_mapping(self) -> dict[Job.Type, type[JobRun]]:
+        return {
             Job.Type.PACKAGE: PackageJobRun,
-            Job.Type.DELTA_APPLY: DeltaApplyJobRun,
+            Job.Type.DELTA_APPLY: ApplyDeltaJobRun,
```
suricactus, line 120: I promise one day we also change the `DELTA_APPLY` name with `APPLY_DELTA`. I am sorry for the dislexic name I introduced back then.

docker-qgis/qfc_worker/commands_base.py
```diff
@@ -0,0 +1,117 @@
...
+from pathlib import Path
+from typing import Any
+
+from qfc_worker.workflow import Workflow, run_workflow
+
+logger = logging.getLogger(__name__)
+
+
+parser = argparse.ArgumentParser(prog="COMMAND")
+subparsers = parser.add_subparsers(dest="cmd")
```
suricactus, line 14: In theory these two can also go to the metaclass below, but failed to do it at first go. Since it is an implementation detal, let them live as global vars.

docker-app/qfieldcloud/core/management/commands/dequeue.py
boardend: Noted that down for the follow up task to make the Job.Type extensible (so that custom jobs are correctly displayed in the admin)

docker-qgis/qfc_worker/commands_base.py
boardend: Since the QFC extending projects can define the args freely in the commands, I don't see a need for that. All good for now.

docker-qgis/qfc_worker/exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+
+class InvalidFileExtensionException(QfcWorkerException):
+    message = 'Project file "%(the_qgis_file_name)s" has unknown file extension "%(extension)s"'
+
+
+class InvalidXmlFileException(QfcWorkerException):
+    message = "Project file is an invalid XML document:\n%(xml_error)s"
+
+
+class FailedThumbnailGenerationException(QfcWorkerException):
```
boardend, line 25: Maybe command specific excpetion should be defined in the command?

suricactus: I encoutered a circular dependency at some point of the refactoring. Suboptimal as you noted, but pragmatic for now.

docker-app/qfieldcloud/core/management/commands/dequeue.py
suricactus: Do you already have CU task? Can you post a link here?

suricactus:
@boardend thanks for the review!

PR #1371, opened 2025-09-27
fix(admin): the FaultyDelta admin was taking forever to open

suricactus:
Faulty delta admin was optimistically listing the available projects and owners, which can be hundreds of thousands.

| before | after |
|-|-|
| image | image |


lukasgraf (approved):
LGTM, thank you! 👍 🎉

PR #1373, opened 2025-09-30
feat: check s3 etags before and after file storage migration

gounux:
Checking that ETags are the same before and after a file migration, using a `head_object` call with the boto library.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/migrate_project_storage.py
+++ b/docker-app/qfieldcloud/filestorage/migrate_project_storage.py
@@ -154,7 +154,7 @@ def migrate_project_storage(
                     },
                 )
 
-                _file_version = FileVersion.objects.add_version(
+                new_file_version = FileVersion.objects.add_version(
                     project=project,
                     filename=file_version.name,
                     content=django_content_file,
@@ -164,6 +164,23 @@ def migrate_project_storage(
                     created_at=now,
                     legacy_version_id=file_version.id,
                 )
+
+                # check that etags before and after are the same.
+                legacy_storage_etag = file_version.e_tag.strip('"')
+
+                new_file_version.content.open()
+                to_storage_etag = new_file_version.content._file.obj.e_tag.strip('"')
+                new_file_version.content.close()
+
+                if legacy_storage_etag != to_storage_etag:
+                    raise Exception(
+                        f"ETag mismatch: '{new_file_version}' on legacy has value {legacy_storage_etag} but new storage has {to_storage_etag} !"
+                    )
+
+                if legacy_storage_etag != new_file_version.etag:
+                    raise Exception(
+                        f"ETag mismatch: version object has ETag {new_file_version.etag} but remote systems have {to_storage_etag} !"
+                    )
 
         package_files = []
 
@@ -189,7 +206,7 @@ def migrate_project_storage(
                     django_content_file,
                 )
 
-                _file_version = FileVersion.objects.add_version(
+                new_file_version = FileVersion.objects.add_version(
                     project=project,
                     filename=package_file.name,
                     content=django_content_file,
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -165,6 +217,25 @@ def migrate_project_storage(
                     legacy_version_id=file_version.id,
                 )
 
+                # check that etags before and after are the same.
+                legacy_storage_etag = get_s3_object_etag(
```
suricactus: Is there a reason to prefer direct access to the S3 obejct storage instead of using the already available data in `file_version.e_tag`?

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -165,6 +217,25 @@ def migrate_project_storage(
...
+                    legacy_s3_client,
+                    from_storage_bucket.name,
+                    file_version.key,
+                    version_id=file_version.id,
+                )
+
+                default_storage_etag = get_s3_object_etag(
```
suricactus: Why don't you use `new_file_version.content._file.obj.e_tag`?

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -165,6 +217,25 @@ def migrate_project_storage(
...
+                    default_s3_client,
+                    to_storage_bucket.name,
+                    new_file_version.content.file.name,
+                )
+
+                if legacy_storage_etag != default_storage_etag:
+                    raise Exception(
+                        f"ETag mismatch for file '{file_version.key}' (version '{file_version.id}') !"
+                    )
```
suricactus:
```suggestion
                if legacy_storage_etag != default_storage_etag:
                    raise Exception(
                        f"ETag mismatch: '{new_file_version} on legacy has value {legacy_storage_etag} but new storage has {default_storage_etag}!"
                    )
```

I think this gives more information, but I might be missing something?

suricactus:
```suggestion
                if legacy_storage_etag != default_storage_etag:
                    raise Exception(
                        f"ETag mismatch for file '{file_version.key}' (version '{file_version.id}') !"
                    )

                if legacy_storage_etag != new_file_version.etag:
                    raise Exception(
                        f"ETag mismatch: version object has ETag {file_version.etag} but remote systems have {default_storage_etag}!"
                    )
```

Also make sure the etag on the storage is the same as the one in the database.

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus: Why this is prefixed as `default`? In theory it can be any other storage, no?

gounux: That's right, last commit makes it more generic by targetting the `to_storage`.

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
gounux: Correct, thanks ! Addressed.

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus: @gounux what about this one? You can easily kill the newly introduce functions then.

gounux: That is weird, the `new_file_version.content._file` seems to be None,

suricactus: what if you `open` it?

gounux: Indeed, thanks !

PR #1377, opened 2025-10-08
Fix unnecessary creation of storage_package as `PackageType`

Rakanhf:
At some point storage_package `PackageType` was renamed to `storage` but the auto creation of it was never updated.

steps to reproduce
1. fresh install
2. `manage.py migrate`
3. created a super user `manage.py createsuperuser`

Now we see an extra package type which has `code = storage_package`, it caused issues when syncing with stripe and when deleting in admin panel

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -310,12 +310,14 @@ class PackageType(models.Model):
             return cls.objects.get(type=cls.Type.STORAGE)
         except cls.DoesNotExist:
             return cls.objects.create(
-                code="storage_package",
+                code="storage",
+                display_name="Additional storage",
+                is_public=True,
                 type=cls.Type.STORAGE,
                 unit_amount=1000,
                 unit_label="MB",
                 min_quantity=0,
-                max_quantity=100,
+                max_quantity=50,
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6820 Fix unnecessary creation of storage_package as 'PackageType'](https://app.clickup.com/t/2192114/QF-6820)

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -310,12 +310,14 @@ def get_storage_package_type(cls) -> "PackageType":
             return cls.objects.get(type=cls.Type.STORAGE)
         except cls.DoesNotExist:
             return cls.objects.create(
-                code="storage_package",
+                code="storage",
+                display_name="Additional storage",
+                is_public=True,
                 type=cls.Type.STORAGE,
                 unit_amount=1000,
```
suricactus, line 317:
```suggestion
                unit_amount=3000,
```

Just to know that we changed this for our prod envs, but we don't need to change it here.

Rakanhf: Thx, I think we also need to change is in other places as well. will check it out in next commits

PR #1379, opened 2025-10-09
feat: kill the geodb once and forever

suricactus:
Geodb used to be a core functionality on pre-production days of QFC, where users can create their own postgis database dynamically and this could be sold as additional product along QFieldCloud.

This never took off due to complexities of managing a database for someone, but the testing framework was heavily dependent on the GeoDB already.

So long Geodb and thanks for the fish!

RELEASE NOTES:

- When starting the stack, make sure you restart it with `--remove-orphans` flag to delete the orphaned `geodb` service container. This is valid for both local development, staging and testing.
- Remove all `GEODB_*` envvars from `.env` file.
- Applying the change will permanently delete the `core_geodb` table. If you still somehow use Geodb, do not use this version.


docker-compose.override.test.yml
```diff
@@ -19,6 +25,14 @@ services:
     environment:
       POSTGRES_DB: test_${POSTGRES_DB}
 
+  test_postgis_db:
+    image: postgis/postgis:17-3.6-alpine
```
gounux:
```suggestion
    image: postgis/postgis:${POSTGIS_IMAGE_VERSION}
```

Would that make sense to use the same as in the standalone docker-compose override docker-compose file (`db` service)

suricactus: I had the same question myself... If you highlight it, then I will oblige :)

suricactus: Done

PR #1381, opened 2025-10-10
refactor: no need for the `WEB_HTTP(S)_PORT` in the `app` container

suricactus:
Not sure what the original author (me from the past) thought with the
direct check of the port, since `request.is_secure()` checks exactly
this and in the current docker compose stack we always use `nginx`.

If someone decides to change the front from nginx to something else,
then we will encounter a problem in the current or the latter case, so
there was no additional safety.


docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -259,11 +259,8 @@ def file_response(
     if version is not None:
         extra_params["VersionId"] = version
 
-    # check if we are in NGINX proxy
-    http_host = request.headers.get("host", "")
-    https_port = http_host.split(":")[-1] if ":" in http_host else "443"
-
-    if https_port == settings.WEB_HTTPS_PORT and not settings.IN_TEST_SUITE:
+    # Assume that if the request is secure, we are behind a nginx proxy and we can use `X-Accel-Redirect`
```
lukasgraf, line 262: This is definitely better than before, but I don't really like that it still ties the use of `X-Accel-Redirect` together with the use of HTTPS - those two things are unrelated.

An alternative would be to check for the presence of the `X-Forwarded-For` header in the request. If the request came in via nginx (or a different kind of correctly configured reverse proxy), that header should be present, and missing otherwise. That would then also allow developers to use nginx locally, but with HTTP, instead of HTTPS, to test X-Accel-Redirect downloads.

But I'd probably consider this out of scope for this PR - just a comment 😉. Happy to discuss this some other time in the future 😉

suricactus: Totally aggree here, I don't like it either. What I was actually looking for is `X-Proxy-Name: nginx`, but I didn't find any.

Because `X-Forwarded-For` is not ideal either, there are people in the wild that proxy QFC with a service from the host.

I came across this module https://github.com/jazzband/django-downloadview that pretty much does the same (I guess in more proper way).

But let's just assumed the https == nginx for now, happy to iterate over when we hit a wall again.

suricactus: See https://app.clickup.com/t/86c602x9g

PR #1384, opened 2025-10-15
chore: drop `django_cryptography` dependency

gounux:
[django-fernet-encrypted-fields](https://github.com/jazzband/django-fernet-encrypted-fields) is now used to store the Secrets' values, no need for `django_cryptography` dependency anymore.

I cleaned up a bit the `requirements.txt` file after running `pipcompile`, in order to keep only the dependencies change brought by the removal of this one.

Also used some "dummy" fields in older migration, since the dependency is not there anymore to create the field.

**RELEASE NOTES**

- There are some dependencies changes, so make sure the containers are rebuilt with the `--build` option when relaunching the containers.
- This PR adds a migration that will remove old values of Secrets. Make sure you run the Django migrations (`docker compose exec app python manage.py migrate`)


duke-nyuki:
Task linked: [QF-6361 Remove legacy `django_cryptography`'s encrypted field support](https://app.clickup.com/t/2192114/QF-6361)

suricactus (approved):
Lovely to see. Please add it to the release notes.

lukasgraf:
@gounux @suricactus I have created PR #773 that adds UPGRADES.md, but I don't believe these should go in there:
There is nothing out of the ordinary here. No manual steps are required:
- Containers are [already rebuilt by `upgrade.sh`](https://github.com/opengisch/QFieldCloud-Nyuki/blob/e05873a7f7f3b304d32485a5287d8219f3641c6a/scripts/upgrade.sh#L72)
- Migrations are (of course) also [applied by upgrade.sh](https://github.com/opengisch/QFieldCloud-Nyuki/blob/e05873a7f7f3b304d32485a5287d8219f3641c6a/scripts/upgrade.sh#L75)

So when just following the regular, generic release guide, there is nothing extra to be done here. Or did I miss something that actually *needs* to be done *manually*, in addition to the regular release guide?

suricactus:
Here I ment the RELEASE NOTES, not deployment notes. What I had in mind is:

- we modify existing migrations, which is always a bit shady and nice to be aknowledged;
- ~if you start a new setup, you migrate all up to 0086, adding a secret and then migrate 0086 most probably will fail when converting from old value to new value.~
- no, it will not fail, I had wrong preliminary thought.

suricactus:
Not sure if  "RELEASE NOTES" in the body is an official thing, it is just a way for me to do mental notes that are a bit less mental :) .

We need to find a way how these end up in the releases, let's bring this topic tomorrow in our dev meeting.

PR #1385, opened 2025-10-15
Remove hardcoding of the third-party auth related headers

gounux:
This PR makes the former hard-coded `X-QFC-IDP-ID` and `X-QFC-ID-Token` third-party auth related header names configurable in the settings.

### DIFF
--- a/docker-app/qfieldcloud/authentication/views.py
+++ b/docker-app/qfieldcloud/authentication/views.py
@@ -189,7 +189,8 @@ class ListProvidersView(APIView):
             "redirect_port": self.QGIS_REDIRECT_PORT,
             "redirect_url": self.QGIS_REDIRECT_URL,
             "client_id": provider.app.client_id,
-            "extra_tokens": {"id_token": "X-QFC-ID-Token"},
+            "extra_tokens": {"id_token": settings.QFIELDCLOUD_ID_TOKEN_HEADER_NAME},
+            "idp_id_header": settings.QFIELDCLOUD_IDP_ID_HEADER_NAME,
             "styles": SSOProviderStyles(request).get(provider_id),
         }
         return provider_data
--- a/docker-app/qfieldcloud/authentication/tests/test_list_auth_providers.py
+++ b/docker-app/qfieldcloud/authentication/tests/test_list_auth_providers.py
@@ -4,6 +4,7 @@ import requests_mock
 from allauth.socialaccount.models import SocialApp
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 from allauth.socialaccount.providers.oauth2.views import OAuth2Adapter
+from django.conf import settings
 from django.http import HttpRequest
 from django.test import override_settings
 from rest_framework.test import APITestCase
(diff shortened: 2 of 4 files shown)


docker-app/qfieldcloud/authentication/views.py
```diff
@@ -189,7 +189,7 @@ def get_provider_data(self, provider: Provider, request: HttpRequest) -> dict:
             "redirect_port": self.QGIS_REDIRECT_PORT,
             "redirect_url": self.QGIS_REDIRECT_URL,
             "client_id": provider.app.client_id,
-            "extra_tokens": {"id_token": "X-QFC-ID-Token"},
+            "extra_tokens": {"id_token": settings.QFIELDCLOUD_ID_TOKEN_HEADER_NAME},
```
lukasgraf, line 192: This view should now return an additional property with the key `idp_id_header` and the value `settings.QFIELDCLOUD_IDP_ID_HEADER_NAME`.

In qfieldsync, the name of the header will be [read from this dictionary here](https://github.com/opengisch/qfieldsync/blob/99f6d163879a18d0546333a8f0002d9bedceb5fd/qfieldsync/core/cloud_api.py#L200-L201), and the header name it looks for is [defined as "idp_id_header" here](https://github.com/opengisch/qfieldsync/blob/99f6d163879a18d0546333a8f0002d9bedceb5fd/qfieldsync/core/cloud_api.py#L64).

This will then make both of these headers completely configurable on the server side, and qfieldsync will use the appropriate header names dynamically. (What's left after that is the change in QField to also read the header names from this dict, instead of hard coding them).

lukasgraf:
Upgrade notes are perfect 👌 🎉!

Once opengisch/QFieldCloud-Nyuki#773 is merged, you could add them exactly like this, by just replacing `No special notes for this release.` in the `(next release)` section.

.env.example
```diff
@@ -267,6 +267,14 @@ QFIELDCLOUD_PASSWORD_LOGIN_IS_ENABLED=1
...
 SOCIALACCOUNT_PROVIDERS='{}'
 
+# Header name used to specify the Identity Provider (IDP) when authenticating with third-party.
+# DEFAULT: X-QFC-IDP-ID
+QFIELDCLOUD_IDP_ID_HEADER_NAME=X-QFC-IDP-ID
+
+# Header name used to provide the ID token when authenticating with third-party.
+# DEFAULT: X-QFC-ID-Token
+QFIELDCLOUD_ID_TOKEN_HEADER_NAME=X-QFC-ID-Token
+
```
suricactus: Do we actually need them as envvars for now? I would say no, not really.

It would be enough to be in settings.py and eventually expose them to the .env one day, no?

@lukasgraf @gounux ?

lukasgraf: Yes, I actually agree with this.
It's good to have a mechanism in place for them to be configurable on the server side, and to get clients to implement that mechanism before a broad rollout, but as of now I don't see a need to *actually* configure them to a different value. And it would keep the `.env` simpler until we actually need it.

suricactus: I would say we keep all like it is in this PR, we only revert the .env.example and docker-compose.yml files.

gounux: Alright, let's keep it only in the settings then.

docker-app/qfieldcloud/authentication/views.py
gounux: Thanks for pointing that out, indeed. Should be addressed by https://github.com/opengisch/QFieldCloud/pull/1385/commits/cb1037ab54ef9029db118e5fb8c92bb5696d79dd.

lukasgraf (approved):
Looks good to me! 👍 🎉 
Just the test in `test_list_auth_providers.py` needs a little update 😜

gounux:
Well, upgrade notes are not relevant on this one anymore :wink: 

Commits squashed, if that's okay for you and the CI tests pass, this can get merged.

PR #1386, opened 2025-10-16
Ensure legacy thumbnail file exists on S3 before migrating it

gounux:
Can happen that the thumbnail can not be found on legacy storage, which prevents the project from being migrated.  
So checking if the thumbnail file exists on S3 can spare us some errors for such projects.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/migrate_project_storage.py
+++ b/docker-app/qfieldcloud/filestorage/migrate_project_storage.py
@@ -8,6 +8,7 @@ from django.utils import timezone
 
 from qfieldcloud.core.models import Job, Project
 from qfieldcloud.core.utils import (
+    check_s3_key,
     get_project_files_with_versions,
     get_project_package_files,
 )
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-6850 Ensure there is a project thumbnail before migrating it](https://app.clickup.com/t/2192114/QF-6850)

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -239,16 +240,22 @@ def migrate_project_storage(
...
-            django_thumbnail_file = ContentFile(b"", project.legacy_thumbnail_uri)
-            from_storage_bucket.download_fileobj(
-                project.legacy_thumbnail_uri,
-                django_thumbnail_file,
-            )
-            project.thumbnail = django_thumbnail_file  # type: ignore
+            if not check_s3_key(project.legacy_thumbnail_uri):
+                logger.warning(
+                    f"Thumbnail '{project.legacy_thumbnail_uri}' does not exist in legacy storage!"
+                )
```
gounux, line 246:
```suggestion
            if not check_s3_key(project.legacy_thumbnail_uri):
                logger.warning(
                    f"Thumbnail '{project.legacy_thumbnail_uri}' does not exist in legacy storage!"
                )
                
            project.legacy_thumbnail_uri = ""
```

Question here if we should reset the `legacy_thumbnail_uri` value, since the file is not there.  
What do you think @suricactus ?

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
```diff
@@ -239,16 +240,22 @@ def migrate_project_storage(
...
 
-            django_thumbnail_file = ContentFile(b"", project.legacy_thumbnail_uri)
-            from_storage_bucket.download_fileobj(
-                project.legacy_thumbnail_uri,
-                django_thumbnail_file,
-            )
-            project.thumbnail = django_thumbnail_file  # type: ignore
+            if not check_s3_key(project.legacy_thumbnail_uri):
+                logger.warning(
+                    f"Thumbnail '{project.legacy_thumbnail_uri}' does not exist in legacy storage!"
```
suricactus:
```suggestion
                    f"Thumbnail '{project.legacy_thumbnail_uri}' does not exist in legacy storage, skip thumbnail migration!"
```

docker-app/qfieldcloud/filestorage/migrate_project_storage.py
suricactus: Not super sure. On one hand - yes, it would be nice.

On the other hand, the migrate storage will change stuff that is not directly responsible for. I leave it to you to have a better feeling.

gounux: That's right, it probably makes more sense to bring those changes out of the migration script, and stick to the necessary only here.

PR #1387, opened 2025-10-17
Use only one login page from `allauth`

Rakanhf:
Login and authentication are critical components of any application. This PR implements a complete custom UI overhaul for all django-allauth authentication pages, providing a consistent, polished, and user-friendly experience.

New Templates:
* `account/base.html` - Base template with branding, navbar, and footer.
* `login`, `signup`, `password_reset`, etc.

Configuration Updates:
* Added `sri` to `INSTALLED_APPS`
* Added `BOOTSTRAP4`, `AXES_LOCKOUT_TEMPLATE`
* Added `smooth_timedelta` filter for human-readable time durations for the `axes` locked out template.
* Add `context_processors.py`

Note that [PR 1293](https://github.com/opengisch/QFieldCloud/pull/1293) and [PR1294 ](https://github.com/opengisch/QFieldCloud/pull/1294) are related to should be merged after merging this PR!

## **Before**

Sign in Before

---

Sign up Before

---

Account Locked Before

---

## **After**

Sign in After

---

Sign up After

---

Account Locked After


suricactus (changes requested):
How this differes from #1293 and #1294? 

I tried this PR locally and the login page looks not so great if I only use this repository as a standalone QFieldCloud.

Rakanhf:
> How this differes from #1293 and #1294?
> 
> I tried this PR locally and the login page looks not so great if I only use this repository as a standalone QFieldCloud.

That was due to incomplete task details. The task has been updated now with the correct information.

suricactus:
Shall we keep the PR contents as it is? Or base it on #1293 instead and just add the templates here?

Then we merge first #1293, then #1294 and finally rebase this PR on `master` and merge it, all of these in a single day, so we don't break login on master for too long?

docker-app/qfieldcloud/core/templates/account/base.html
suricactus, line 1: Move this to `layouts/base.html` and make this minimal file, similar to your `axes/lockedout.html`.

Keep all other `accounts/*.html` files to extend from `accounts/base.html`.

docker-app/qfieldcloud/settings.py
```diff
@@ -533,7 +541,7 @@ def before_send(event, hint):
 
 # Django axes configuration
 # https://django-axes.readthedocs.io/en/latest/4_configuration.html
-###########################
+AXES_LOCKOUT_TEMPLATE = "axes/lockedout.html"
```
suricactus, line 547:
```suggestion
###########################
# Template for the page shown then the user is locked out of their account for too many failed attempts.
AXES_LOCKOUT_TEMPLATE = "axes/lockedout.html"
```

docker-app/qfieldcloud/settings.py
```diff
@@ -136,6 +136,7 @@
     "constance",
     "django_extensions",
     "bootstrap4",
+    "sri",
```
suricactus, line 139:
```suggestion
    "sri",
    # To ensure that exceptions inside other apps' signal handlers do not affect the integrity of file deletions within transactions, `django_cleanup` should be placed last in `INSTALLED_APPS`. See https://github.com/un1t/django-cleanup#configuration
```

docker-app/qfieldcloud/core/staticfiles/js/vendor/bootstrap-autocomplete.js
suricactus, line 1: I think we don't need this dependency here (and it's CSS probably).

I verified by running:

```
grep -r 'autoComplete=' . 
grep -r autoComplete . # only returns this file
```

suricactus (commented):
Thanks, pretty much it, a few small comments.

Note the CI obviously.

docker-app/qfieldcloud/core/templates/account/base.html
Rakanhf: I think that keeping them all in `account/base.html` will be better, since If I created a simpler version (`layouts/base.html`) It will cause compatibility issues if I changed `lockedout.html` to extend from `layouts/base.html` instead of account. 

So the only benefit that we will get is a simpler version of `account/base.html` but a nested extension for `lockedout.html` 

I believe that leaving it as is will be more straight forward. let me know what do you think,

docker-app/qfieldcloud/settings.py
Rakanhf: Done !

docker-app/qfieldcloud/settings.py
Rakanhf: Done !

docker-app/qfieldcloud/core/staticfiles/js/vendor/bootstrap-autocomplete.js
Rakanhf: I have removed the file, but I prefer just to keep the `vendor.css` as is (default).

docker-app/qfieldcloud/core/templates/account/base.html
suricactus: I didn't really follow here. Mostly my confusion comes from "It will cause compatibility issues if I changed lockedout.html to extend from layouts/base.html instead of account.".

My understanding is that `layouts/base.html` will contain 90+% of the current `accounts/base.html`. The other 10% or so will be defined in `accounts/base.html`. As for `axes/lockedout.html`, it will be exactly the same as the current one, but also add the `<div class="row">` wrapper from `accounts/base.html` and changing the base from `account/base.html` to `layouts/base.html`.

So basically, the only problem is that the `<div class="row">` from the current `accounts/base.html` will be repeated twice, which is fine with me.

It's also fine with me to merge with the current state of the PR.

docker-app/qfieldcloud/core/templates/account/base.html
```diff
@@ -0,0 +1,76 @@
...
+          {% endblocktrans %}
+        </div>
+      </nav>
+    </div>
+  </footer>
+
+  {% sri_static 'js/vendor/jquery.js' %}
+  {% sri_static 'js/vendor/popper.min.js' %}
+  {% sri_static 'js/vendor/bootstrap.min.js' %}
+  {% sri_static 'js/vendor/bootstrap-autocomplete.js' %}
```
suricactus: This fellow here?

docker-app/qfieldcloud/core/staticfiles/css/qfieldcloud.css
```diff
@@ -0,0 +1,75 @@
...
+  left: 0;
+  z-index: 2147483645;
+  box-sizing: border-box;
+  width: 100%;
+  background-color: rgba(230, 236, 234, 0.5);
+}
+
+.qfc-header-logo {
+  height: 1.5rem;
+}
```
suricactus: Duplicate as in line 38:40

docker-app/qfieldcloud/core/templates/account/base.html
```diff
@@ -0,0 +1,76 @@
...
+  {% block extrahead %}{% endblock extrahead %}
+</head>
+
+<body class="d-flex flex-column min-vh-100">
+
+  <!-- Navbar -->
+  <nav class="navbar navbar-expand-md navbar-dark nav-fill w-100 bg-primary">
+    <a href="{% url 'index' %}">
+      <img src="{% static 'logo_sidetext_white.svg' %}" alt="QFieldCloud" class="logo-nav qfc-header-logo ml-1"
+      >
```
suricactus:
```suggestion
      <img src="{% static 'logo_sidetext_white.svg' %}" alt="QFieldCloud" class="logo-nav qfc-header-logo ml-1">
```

docker-app/qfieldcloud/core/templates/account/base.html
```diff
@@ -0,0 +1,76 @@
...
+      <p class="text-center">
+      <img src="{% static 'logo_undertext.svg' %}" alt="QFieldCloud" class="logo-nav" style="height: 15rem">
+      </p>
+
+      {% block content %}{% endblock content %}
+
+    </div>
+  </div>
+</div>
+</div>
```
suricactus: Indent?

docker-app/qfieldcloud/core/templates/account/email_confirm.html
```diff
@@ -0,0 +1,50 @@
...
+        {{ user_display }}.
+        {% endblocktrans %}
+
+      {% else %}
+
+        {% url 'account_email' as email_url %}
+        {% blocktrans %}
+          This email confirmation link expired or is invalid.
+          Please contact support to get your email confirmation done.
+        {% endblocktrans %}
```
suricactus: The `email_url` is not used. We either use it, or remove it. I guess it would be nice to see it, but make sure you test how it looks locally first.

docker-app/qfieldcloud/core/templates/account/password_reset_done.html
```diff
@@ -0,0 +1,22 @@
...
+
+{% block content %}
+
+<div class="card">
+  <div class="card-body">
+    <h5 class="card-title">{% trans 'Reset Password Complete' %}</h5>
+    <p class="card-text">
+      {% url 'account_login' as login_url %}
+      {% blocktrans with login_url=login_url %}
+        We have sent you an email. Please contact us if you do not receive it within a few minutes. Proceed to the <a href="{{ login_url }}">sign-in form</a>.
```
suricactus, line 16: Are we sure this should be the text here? Isn't this page shown when the reset was successful already? Please verify and adapt the text

docker-app/qfieldcloud/core/templates/account/password_reset_from_key.html
```diff
@@ -1,55 +1,56 @@
...
+      </p>
+    {% else %}
+      {% if form %}
         <p class="card-text">
-          {% url 'account_reset_password' as passwd_reset_url %}
-          {% blocktrans %}
-            The password reset link was invalid, possibly because it has already been used. Please request a <a href="{{ passwd_reset_url }}">new password reset</a>.
+          {% url 'account_login' as login_url %}
+          {% blocktrans with login_url=login_url %}
+            Please enter you newly chosen secure password in the form below.
```
suricactus:
```suggestion
            Please enter your newly chosen secure password in the form below.
```

docker-app/qfieldcloud/core/templates/account/password_reset_from_key.html
```diff
@@ -1,55 +1,56 @@
...
+          {% csrf_token %}
 
+          {% bootstrap_form form %}
+
+          {% if redirect_field_value %}
+            <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}" />
+          {% endif %}
+
+          <div class="float-right">
+            <button class="btn btn-primary" type="submit" name="action">{% trans 'Reset Password' %}</button>
```
suricactus:
```suggestion
            <button class="btn btn-primary" type="submit" name="action">{% trans 'Change Password' %}</button>
```

docker-app/qfieldcloud/core/templates/axes/lockedout.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'account/base.html' %}
+{% load i18n %}
+{% load bootstrap4 %}
+{% load filters %}
+
+{% block title_contents %}{% trans 'Login' %}{% endblock %}
```
suricactus: I believe we can do better with the title here, probably "Account locked"?

docker-app/qfieldcloud/settings.py
```diff
@@ -534,6 +543,8 @@ def before_send(event, hint):
 # Django axes configuration
 # https://django-axes.readthedocs.io/en/latest/4_configuration.html
 ###########################
+# Template for the page shown then the user is locked out of their account for too many failed attempts.
```
suricactus:
```suggestion
# Template for the page shown when the user is locked out of their account for too many failed attempts.
```

docker-app/qfieldcloud/core/templates/account/email_confirm.html
Rakanhf: I removed it, we can add it later. just to not get out of scope of this PR, since we need to implement it's template also.

docker-app/qfieldcloud/core/templates/account/password_reset_done.html
Rakanhf: This page is Shown after reset email is sent, [password_reset_done.html source code](https://github.com/pennersr/django-allauth/blob/8a91a09d5c4543d420e05165b49ae223862b43e2/allauth/templates/account/password_reset_done.html#L16)

I have change the title to make more sense.

docker-app/qfieldcloud/core/templates/account/base.html
Rakanhf: You're right about the approach, but there's a compatibility issue: other applications that extend this codebase may already have their own layouts/base.html. Creating one here could cause template resolution conflicts.

So I will move forward with the current approach

docker-app/qfieldcloud/core/templates/account/email_confirm.html
suricactus: Sure!

PR #1389, opened 2025-10-19
fix: when the project where we try to upload a file to does not exist return a JSON formatted 404

suricactus:
Since we use a view function it was returning the regular 404 which is an HTML.


lukasgraf (approved):
LGTM 👍

PR #1391, opened 2025-10-20
fix: consider org member when getting who triggered latest_package_jobs

gounux:
This PR intends to also consider an organization's members (not only admins) when computing a Project's `latest_package_jobs`, when the owner of the project is an organization.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1342,10 +1342,8 @@ class Project(models.Model):
             QuerySet of all the last package jobs.
         """
         if self.owner.is_organization:
-            # all the admin users including the organization owner
-            triggered_by_qs = Person.objects.for_organization(self.owner).filter(
-                organization_role=OrganizationMember.Roles.ADMIN,
-            )
+            # all the users including the organization owner
+            triggered_by_qs = Person.objects.for_organization(self.owner)
         else:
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-6863 Fix not all the relevant packaged being part of a Project's `latest_package_jobs`](https://app.clickup.com/t/2192114/QF-6863)

suricactus (commented):
Is it possible to add a test for this case?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1342,9 +1342,12 @@ def latest_package_jobs(self) -> PackageJobQuerySet:
...
         """
         if self.owner.is_organization:
-            # all the admin users including the organization owner
+            # all the users including the organization owner
             triggered_by_qs = Person.objects.for_organization(self.owner).filter(
-                organization_role=OrganizationMember.Roles.ADMIN,
+                organization_role__in=[
+                    OrganizationMember.Roles.ADMIN,
+                    OrganizationMember.Roles.MEMBER,
+                ]
```
suricactus: Do we need the filter at all in that case? Any idea why we added this in the first place, I don't remember...

gounux: Indeed, I guess the filtering is not needed anymore.

suricactus (approved):

PR #1392, opened 2025-10-20
Replace TMP_DIRECTORY var with QFIELDCLOUD_QGIS_TMP_DIRECTORY and ensure consistent usage

spwoodcock:
## Issue

- I tested changing from  `TMP_DIRECTORY=/tmp` to  `TMP_DIRECTORY=/tmp/qfield`, but was receiving errors:

```
{"error": "[Errno 2] No such file or directory: '/tmp/tmpzubn2h7u/feedback.json'", "error_stack": ["  File \"/usr/src/app/worker_wrapper/wrapper.py\", line 177, in run\n    with open(self.shared_tempdir.joinpath(\"feedback.json\")) as f:\n"], "error_origin": "worker_wrapper", "container_exit_code": 0}
```

- Seems the QGIS worker container was bind mounting hardcoded `/tmp`:

https://github.com/opengisch/QFieldCloud/blob/e752324019be7264e146b3aa0d9839c3d5ec802f/docker-app/worker_wrapper/wrapper.py#L47

https://github.com/opengisch/QFieldCloud/blob/e752324019be7264e146b3aa0d9839c3d5ec802f/docker-app/worker_wrapper/wrapper.py#L63

https://github.com/opengisch/QFieldCloud/blob/e752324019be7264e146b3aa0d9839c3d5ec802f/docker-app/worker_wrapper/wrapper.py#L120-L124

## Solution

- The QField worker wrapper has the env var `TMP_DIRECTORY` present.
- We can use that in `wrapper.py`, so it's used by **both** the QGIS worker and the QField worker wrapper (they need to share data inside a unique tmp dir per project).

### Notes

- The `TMP_DIRECTORY` is not present in Django settings, so can't be used this way (instead we do an os.getenv). This is probably correct, as we don't use the variable in the QField app, and based on comments in the repo, the plan is to separate out config for app/worker_wrapper better.
- This approach of binding mounting obviously won't work in a distributed deployment, such as with Kubernetes, but that's a problem for another PR! (and a much bigger refactor)


docker-app/worker_wrapper/wrapper.py
```diff
@@ -44,7 +45,7 @@
 RETRY_COUNT = 5
 TIMEOUT_ERROR_EXIT_CODE = -1
 DOCKER_SIGKILL_EXIT_CODE = 137
-TMP_FILE = Path("/tmp")
+TMP_FILE = Path(os.getenv("TMP_DIRECTORY", "/tmp"))
```
suricactus: We avoid using the `os.env` outside the `settings.py` file. It just brings a lot of mess to handle.

I would sugggest you add the very same line to `settings.py` probably with some of the explanations that you have already in your PR. We should better avoid "renaming" the envvar in Django settings and call it `TMP_DIRECTORY` instead of `TMP_FILE`.

In this file you can kill the `TMP_FILE` variable (it has wrong name anyways) and replace it's usage with `settings.TMP_DIRECTORY`.

suricactus: Btw, I have enabled the CI on this PR, so we can see if all works fine already.

docker-compose.yml
```diff
@@ -166,7 +166,7 @@ services:
       - transformation_grids:/transformation_grids
       - /var/run/docker.sock:/var/run/docker.sock
       - ${LOG_DIRECTORY}:/log
-      - ${TMP_DIRECTORY}:/tmp
+      - ${QFIELDCLOUD_QGIS_TMP_DIRECTORY:-/tmp}:/tmp
```
spwoodcock, line 163: Added default as `/tmp` here, if the env var is not present

docker-app/qfieldcloud/settings.py
```diff
@@ -734,6 +734,10 @@ def before_send(event, hint):
 # Port to be used by `debugpy` to connect to the QGIS process inside the `qgis` container
 DEBUG_QGIS_DEBUGPY_PORT = os.environ.get("DEBUG_QGIS_DEBUGPY_PORT")
 
+# The temporary directory to share QGIS projects between the `worker_wrapper` and `worker` containers
+# FIXME this solution will not work in distributed deployments, such as Kubernetes, so perhaps needs to be refactored
+QFIELDCLOUD_QGIS_TMP_DIRECTORY = os.getenv("QFIELDCLOUD_QGIS_TMP_DIRECTORY", "/tmp")
```
spwoodcock: As a second safeguard, I also used `os.getenv` here to allow setting a default directly in the settings

spwoodcock:
Actually, probably no need for the `${QFIELDCLOUD_QGIS_TMP_DIRECTORY:-/tmp}` default, as we can rely on the os.getenv in the settings.

I don't think we have a convention set in the repo yet for env var defaults? I think we are currently relying on the .env.example values to set the defaults. In a few places we have defaults via `os.environ.get("STORAGES_PROJECT_DEFAULT_STORAGE") or "default"`

Other options could be `${VAR:-default}` inside compose, but probably preferred is defaults directly in the Django settings with os.getenv?

docker-app/qfieldcloud/settings.py
```diff
@@ -734,6 +734,10 @@ def before_send(event, hint):
 # Port to be used by `debugpy` to connect to the QGIS process inside the `qgis` container
 DEBUG_QGIS_DEBUGPY_PORT = os.environ.get("DEBUG_QGIS_DEBUGPY_PORT")
 
+# The temporary directory to share QGIS projects between the `worker_wrapper` and `worker` containers
+# FIXME this solution will not work in distributed deployments, such as Kubernetes, so perhaps needs to be refactored
+QFIELDCLOUD_QGIS_TMP_DIRECTORY = os.environ.get("QFIELDCLOUD_QGIS_TMP_DIRECTORY")
```
suricactus, line 739:
```suggestion
# FIXME: this solution will not work in distributed deployments, such as Kubernetes, so perhaps needs to be refactored
QFIELDCLOUD_QGIS_TMP_DIRECTORY = os.environ["QFIELDCLOUD_QGIS_TMP_DIRECTORY")
```

Let's raise a keyerror if the `TMP_DIRECTORY` key is not present.

docker-compose.yml
suricactus: I think it's a bit too dangerous to add a default here, would you mind removing it?


```suggestion
      - ${TMP_DIRECTORY}:/tmp
```

duke-nyuki:
Task linked: [QF-6830 Add documentation for each environment variable in the .env file](https://app.clickup.com/t/2192114/QF-6830)

docker-app/qfieldcloud/settings.py
suricactus: This is outdated by now. Please delete/resolve if you outdate your own comments :)

docker-compose.yml
```diff
@@ -73,6 +72,7 @@ services:
       QFIELDCLOUD_DEFAULT_LANGUAGE: ${QFIELDCLOUD_DEFAULT_LANGUAGE}
       QFIELDCLOUD_DEFAULT_TIME_ZONE: ${QFIELDCLOUD_DEFAULT_TIME_ZONE}
       QFIELDCLOUD_QGIS_IMAGE_NAME: ${QFIELDCLOUD_QGIS_IMAGE_NAME:-${COMPOSE_PROJECT_NAME}-qgis}
+      QFIELDCLOUD_QGIS_TMP_DIRECTORY: ${QFIELDCLOUD_QGIS_TMP_DIRECTORY:-/tmp}
```
suricactus, line 75:
```suggestion
      QFIELDCLOUD_QGIS_TMP_DIRECTORY: ${TMP_DIRECTORY:}
```

There will be a big renaming of envvars coming in the next two releases, after we deploy https://github.com/opengisch/QFieldCloud/pull/1395 . So let's not rename envvars yet. 

The upcoming change will follow the `QFIELDCLOUD_` prefix convention for this and many more envvars. We prefer to do it all at once, instead of bit by bit to annoy users :) .

suricactus: This means the PR title needs an update.

PR #1393, opened 2025-10-20
Fix flaky files API tests

lukasgraf:
Some of these tests occasionally failed in a flaky way:

```
ERROR: test_delete_all_file_versions_fails (qfieldcloud.filestorage.tests.test_files_api.QfcTestCase)
----------------------------------------------------------------------
  File "/usr/src/app/qfieldcloud/filestorage/tests/test_files_api.py", line 727, in test_delete_all_file_versions_fails
    response = self._delete_file(
  File "/usr/src/app/qfieldcloud/core/tests/mixins.py", line 115, in _delete_file
    token = AuthToken.objects.get_or_create(user=user)[0]
  File "/usr/local/lib/python3.10/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 640, in get
    raise self.model.MultipleObjectsReturned(
qfieldcloud.authentication.models.AuthToken.MultipleObjectsReturned: get() returned more than one AuthToken -- it returned 2!
```

 The reason was that methods in `QfcFilesTestCaseMixin` created AuthTokens using this pattern:

```python
AuthToken.objects.get_or_create(user=user)
```

This created a token for the user, with the client type `UNKNOWN`, because a user agent is never set in tests. This worked, unless during a test a worker job was triggered as a side effect.

Then the worker wrapper would issue a token for the same user, but with client_type `WORKER`. This would then cause `get_or_create(user=user)` to fail, because ([as per docs](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create)) it assumes that its keyword arguments act as a unique constraint for the object to be fetched.

Further discriminating the `get_or_create()` call by the `client_type` addresses this issue.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/mixins.py
+++ b/docker-app/qfieldcloud/core/tests/mixins.py
@@ -1,7 +1,6 @@
 import urllib
 from typing import IO
 
-from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.http import FileResponse, HttpResponse
 from django.urls import reverse
 from rest_framework.response import Response
@@ -17,12 +16,14 @@ class QfcFilesTestCaseMixin:
     """
 
     def _get_token_for_user(self, user: User) -> AuthToken:
-        try:
-            return AuthToken.objects.get(user=user)
-        except ObjectDoesNotExist:
-            return AuthToken.objects.create(user=user)
-        except MultipleObjectsReturned:
-            return AuthToken.objects.filter(user=user).first()
+        # We pass the client_type to prevent get_or_create() from failing with
+        # MultipleObjectsReturned if a worker token for the same user happens
+        # to be created during the test.
+        token = AuthToken.objects.get_or_create(
+            user=user,
+            client_type=AuthToken.ClientType.UNKNOWN,
+        )[0]
+        return token
 
     def _upload_file(
         self, user: User, project: Project, filename: str, content: IO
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/tests/mixins.py
```diff
@@ -17,12 +16,14 @@ class QfcFilesTestCaseMixin:
...
-            return AuthToken.objects.create(user=user)
-        except MultipleObjectsReturned:
-            return AuthToken.objects.filter(user=user).first()
+        # We pass the client_type to prevent get_or_create() from failing with
+        # MultipleObjectsReturned if a worker token for the same user happens
+        # to be created during the test.
+        token = AuthToken.objects.get_or_create(
+            user=user,
+            client_type=AuthToken.ClientType.UNKNOWN,
+        )[0]
```
suricactus, line 25:
```suggestion
        token, _is_created = AuthToken.objects.get_or_create(
            user=user,
            client_type=AuthToken.ClientType.UNKNOWN,
        )
```

Slight readability pattern that is usually followed in the current codebase. Basically if something returns a tuple, we always prefer to unset the tuple on the left side with unnamed prefix.

Reasons:

- highlights the meaning of the ignored tuple value.
- prevents accidental overlook on the `[N]` part.
- allows easier refactoring in case the `is_created` is somehow needed.

No strong opinion on this one, feel free to ignore.

suricactus (approved):
Thanks a lot!

PR #1394, opened 2025-10-20
fix: kill geodb envvars that are no longer needed

suricactus:
Follow-up of #1379 . Somehow I overlooked these envvars in my enthusiasm.


lukasgraf (approved):
LGTM 👍

PR #1395, opened 2025-10-20
feat: add docs to each and every envvar with links to docs, group them logically

suricactus:
Pretty much the title. Check the individual commits for some other cleanups on the way.

This will results in a bit of extra work on the current environments, but I believe for the good of the world.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -530,15 +530,3 @@ DEBUG_QGIS_WORKER_HOST_PATH=""
 # The Django development port. Not used in production.
 # DEFAULT: 8011
 DJANGO_DEV_PORT=8011
-
-# Memcached port.
-# DEFAULT: 11211
-MEMCACHED_PORT=11211
-
-# Deprecated, do not use.
-GEODB_HOST=geodb
-GEODB_PORT=5432
-GEODB_USER=postgres
-GEODB_PASSWORD="KUAa7h!G&wQEmkS3"
-GEODB_DB=postgres
-HOST_GEODB_PORT=5434
--- a/README.md
+++ b/README.md
@@ -191,7 +191,6 @@ Create an <code>.env.test</code> file with the following variables that override
     WEB_HTTP_PORT=8101
     WEB_HTTPS_PORT=8102
     HOST_POSTGRES_PORT=8103
-    MEMCACHED_PORT=11212
     QFIELDCLOUD_DEFAULT_NETWORK=qfieldcloud_test_default
     QFIELDCLOUD_SUBSCRIPTION_MODEL=subscription.Subscription
     DJANGO_DEV_PORT=8111
@@ -387,7 +386,6 @@ Based on this example
 | nginx https   | 443   | WEB_HTTPS_PORT       | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | django http   | 8011  | DJANGO_DEV_PORT      | :white_check_mark: | :x:                | :x:                |
 | postgres      | 5433  | HOST_POSTGRES_PORT   | :white_check_mark: | :white_check_mark: | :white_check_mark: |
-| memcached     | 11211 | MEMCACHED_PORT       | :white_check_mark: | :x:                | :x:                |
 | minio API     | 8009  | MINIO_API_PORT       | :white_check_mark: | :x:                | :x:                |
 | minio browser | 8010  | MINIO_BROWSER_PORT   | :white_check_mark: | :x:                | :x:                |
 | smtp web      | 8012  | SMTP4DEV_WEB_PORT    | :white_check_mark: | :x:                | :x:                |
--- a/docker-compose.override.local.yml
+++ b/docker-compose.override.local.yml
@@ -43,10 +43,6 @@ services:
       - ./docker-app/worker_wrapper:/usr/src/app/worker_wrapper
     command: python3 -m debugpy --listen 0.0.0.0:5679 manage.py dequeue
 
-  memcached:
-    ports:
-      - "${MEMCACHED_PORT}:11211"
-
   qgis:
     build:
       args:


duke-nyuki:
Task linked: [QF-6830 Add documentation for each environment variable in the .env file](https://app.clickup.com/t/2192114/QF-6830)

.env.example
```diff
@@ -1,25 +1,43 @@
...
 DEBUG=1
 
+# Main hostname of the QFieldCloud instance. Must be included in `DJANGO_ALLOWED_HOSTS`.
+# Must not have a leading `http://` or `https://` and must not have a trailing slash or port.
+# NOTE: Must be defined in the beginning of this file as it is used by other variables below.
+# DEFAULT: localhost
+QFIELDCLOUD_HOST=localhost
+
 # Environment where QFieldCloud runs
+# NOTE: the default value `test` is used to minimize changes required to run tests. Please change accordingly.
```
lukasgraf, line 17: I would challenge if it's the right decision to optimize for tests here.

99% of the `.env.example` is targeted towards development. All other settings except this one are already perfect for development and don't need to be changed, except ones that can't have generic defaults that work for everyone (like Stripe settings).

To have an `.env.test` suitable for a parallel test instance however, quite a few changes are needed. And those are [documented here](https://github.com/opengisch/QFieldCloud#:~:text=Instructions%20to%20have%20a%20test%20instance%20running%20in%20parallel%20to%20a%20dev%20instance). Including switching `ENVIRONMENT` to `test`.

So I think it would more consistent to make the `.env.example` a *development* example, and therefore change the default here to `ENVIRONMENT=development`. Otherwise this should be mentioned in the README steps for setting up a dev environment, after the instructions to `cp .env.example .env`.

_(In the long term, having dedicated `.env.dev.example`, `.env.prod.example` etc. would be nice, as discussed, but until we're there, my recommendation would be to keep the `.env.example`'s purpose focused on one environment (dev), instead of mixing several of them)._

.env.example
```diff
@@ -41,9 +59,19 @@ QFIELDCLOUD_TLS_KEY="/etc/nginx/certs/${QFIELDCLOUD_HOST}-key.pem"
...
+# DEFAULT: test@example.com
 LETSENCRYPT_EMAIL="test@example.com"
+
+# Let's Encrypt RSA key size. See https://eff-certbot.readthedocs.io/en/stable/using.html#rsa-and-ecdsa-keys.
+# VALUES: 4096
+# DEFAULT: 4096
 LETSENCRYPT_RSA_KEY_SIZE=4096
-# Set to 1 if you're testing your setup to avoid hitting request limits
+
+# Whether the Let's Encrypt certificate is used in staging or production environment. See https://eff-certbot.readthedocs.io/en/stable/using.html#revoking-certificates.
```
lukasgraf: Newline before `See ...`? 
_(If this was intentional, e.g. for line-diffability, feel free to just ignore without any response 😛 )_

.env.example
```diff
@@ -118,95 +146,158 @@ STORAGES='{
...
+#########################
+# Sentry settings
+#########################
 
 # Sentry DSN. Missing value disables Sentry logging. Can be found on https://opengisch.sentry.io/settings/projects/qfieldcloud/keys/ .
 # DEFAULT: <NO VALUE>
 SENTRY_DSN=
 
-# Sentry sample rate between 0 and 1. Read more on https://docs.sentry.io/platforms/python/configuration/sampling/ .
+# Sentry sample rate between 0 and 1. See https://docs.sentry.io/platforms/python/configuration/sampling/ .
```
lukasgraf: I would remove the `.` period at the end of this URL and the next one, to be consistent with the other `See ...` notes (and avoid unclickable URLs).

.env.example
```diff
@@ -289,75 +392,117 @@ QFIELDCLOUD_WORKER_QFIELDCLOUD_URL=http://app:8000/api/v1/
...
-QFIELDCLOUD_ACCOUNT_ADAPTER=qfieldcloud.core.adapters.AccountAdapterSignUpOpen
+# QFieldCloud QGIS image name to be used as worker by the `worker_wrapper`.
+# If empty value, a default name will be generated at build time, for example `qfieldcloud-qgis`.
+# DEFAULT: ""
+# QFIELDCLOUD_QGIS_IMAGE_NAME=""
 
-# QFieldCloud auth token expiration hours. For example 720 hours (30 days).
-# DEFAULT: 720
-QFIELDCLOUD_AUTH_TOKEN_EXPIRATION_HOURS=720
+# Directory used by QFieldCloud workerfor temporary files shared between the host, the worker_wrapper and the worker containers.
```
lukasgraf: s/workerfor/worker for/

.env.example
```diff
@@ -289,75 +392,117 @@ QFIELDCLOUD_WORKER_QFIELDCLOUD_URL=http://app:8000/api/v1/
...
 
-# Prefix used by docker compose for each of the containers, e.g. app will be `qfieldcloud_app`
-# DEFAULT: qfieldcloud
-COMPOSE_PROJECT_NAME=qfieldcloud
+# Docker image tag used for local/standalone database setup. See https://hub.docker.com/r/postgis/postgis/tags.
+# When setting up a new local instance, do not hesitate to pick a recent version.
+# For migrating a local postgis to a more recent major version (risky operation),
+# see this gist: https://gist.github.com/gounux/2c0779fcb22e512cbdc613eb78200571 .
+# Migration to a newer database version is a risky operation to your data, so prepare and test the backup of the `postgres_data` volume.
+# NOTE: Ignored if `db` and `geodb` is not used.
```
lukasgraf: After killing geodb support, I think this reference to `geodb` can also be removed.

.env.example
```diff
@@ -290,75 +392,117 @@ QFIELDCLOUD_WORKER_QFIELDCLOUD_URL=http://app:8000/api/v1/
...
+# DEFAULT: 13-3.1-alpine
+POSTGIS_IMAGE_VERSION=13-3.1-alpine
```
gounux, line 454:
```suggestion
# DEFAULT: 18-3.6-alpine
POSTGIS_IMAGE_VERSION=18-3.6-alpine
```

Wouldn't it be the opportunity to upgrade the postgis version [to this one](https://hub.docker.com/layers/postgis/postgis/18-3.6-alpine/images/sha256-b771a56a4054adbdef4ebfdcc6d421f7a592a597d61e4f1984dbef7a07629c0e) ?

.env.example
```diff
@@ -118,96 +146,158 @@ STORAGES='{
...
-# Messages are logged at the specified level and all more severe levels. The nginx default is `error`. Read more on https://nginx.org/en/docs/ngx_core_module.html#error_log.
-# OPTIONS: debug, info, notice, warn, error, crit, alert, emerg
-# DEFAULT: error
-NGINX_ERROR_LOG_LEVEL=error
 
 ###################
 # Database settings
 ###################
 
+# PostgreSQL connection user. This user must have access to create and modify the `POSTGRES_DB` database.
```
gounux:
```suggestion
# PostgreSQL connection user. This user must have access to create, modify the `POSTGRES_DB` database and add the `postgis` extension.
```

I think we faced this once, kind of pg superuser needed I believe.

.env.example
suricactus: Since we are on prod on the older DB, I would like to keep the tests running on this more ancient version. We have a backlog task that addresses this.

gounux: Alright, thanks.

.env.example
suricactus: Done

.env.example
suricactus: Done

.env.example
suricactus: Done.

.env.example
suricactus: Done.

.env.example
suricactus: We need to address this in a follow-up IMO, because we need to change also the CI, which just copy-pastes the .env.example. 

So in practice 99% of the humans change it to `development`, but 99% of the copy-pastes are resulting in `test` when done on the CI.

Very open to change, just not in this PR.

suricactus:
@lukasgraf @gounux allowed myself to force push to reduce the number of commits. @lukasgraf see my comment regarding testing/development. The purpose of this PR is to add docs, not to introduce new changes (which we need to do sooner than later).

.env.example
suricactus: See https://github.com/opengisch/QFieldCloud/pull/1398

suricactus:
Allowed myself to merge this PR, as the comments were mostly addressed.

PR #1396, opened 2025-10-21
feat: add support for virtual layers without embedded layers

suricactus:
Since the virtual and memory layers do not have filenames, we should not
treat them as regular postgres or WFS layers when checking for `Project.
has_online_vector_data`.

On the other hand, the virtual layers may embed a postgresql layer and
we should disable this workaround when a `Plan.is_external_db_supported`
is `True`.

We take a very radical approach. Instead of looking at the current
embedded layer provider, we just say - if there are any embedded layers
(which may be file based and part of the project already), we do assume
the layer has online vector data and therefore external db.

A few other fixes/improvements along the way.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_utils_project.py
+++ b/docker-app/qfieldcloud/core/tests/test_utils_project.py
@@ -11,14 +11,22 @@ logging.disable(logging.CRITICAL)
 
 
 class QfcTestCase(TestCase):
+    def test_is_virtual_layer_without_query_question_mark(self):
+        source = (
+            "layer=ogr:%2Fhome%2Fuser%2Ftestdata.gpkg%7Clayername%3Dpoints_xy:points_xy:&"
+            "query=SELECT%20*%20FROM%20points"
+        )
+
+        self.assertFalse(is_virtual_layer_with_embedded_layers(source))
+
     def test_is_virtual_layer_without_embedded_layers(self):
-        source = "query=SELECT%20*%20FROM%20points"
+        source = "?query=SELECT%20*%20FROM%20points"
 
         self.assertFalse(is_virtual_layer_with_embedded_layers(source))
 
     def test_is_virtual_layer_with_embedded_layers(self):
         source_with_embedded_layers = (
-            "layer=ogr:%2Fhome%2Fuser%2Ftestdata.gpkg%7Clayername%3Dpoints_xy:points_xy:&"
+            "?layer=ogr:%2Fhome%2Fuser%2Ftestdata.gpkg%7Clayername%3Dpoints_xy:points_xy:&"
             "query=SELECT%20*%20FROM%20points"
         )
 


docker-app/qfieldcloud/core/models.py
```diff
@@ -1564,28 +1564,12 @@ def users(self):
...
-            if layer_data.get("type_name") in (
-                "VectorLayer",
-                "Vector",
-            ) and not layer_data.get("filename", ""):
-                has_online_vector_layers = True
-                break
+        from qfieldcloud.core.utils2.project import has_online_vector_data
 
-        return has_online_vector_layers
+        return has_online_vector_data(self)
```
suricactus, line 1572: Decided to move the logic to a new function outside the models as a POC of what we discussed with some teammates. Happy to revert this decision.

NOTE the new function returns only boolean, while the `Project.has_online_vector_data` may return `None`. While returning `None` is probably more correct, I am not sure what is the additional value added here besides making the typing harder.

docker-qgis/qfc_worker/utils.py
```diff
@@ -589,21 +589,26 @@ def get_layers_data(project: QgsProject) -> dict[str, dict]:
...
         layer_source = LayerSource(layer)
         filename = layer_source.filename
-        datasource = None
+        data_provider = layer.dataProvider()
+        data_provider_source: str | None = None
+        data_provider_name: str | None = None
 
         # TODO: Move localized layer handling functionality inside libqfieldsync (ClickUp: QF-5875)
         if layer_source.is_localized_path:
-            datasource = bad_layer_handler.invalid_layer_sources_by_id.get(layer_id)
```
nirvn, line 596: Good variable rename, thanks

docker-app/qfieldcloud/core/utils2/project.py
```diff
@@ -0,0 +1,72 @@
...
+        "layers_by_id"
+    )
+
+    if layers_by_id is None:
+        return False
+
+    has_online_vector_layers = False
+
+    for layer_data in layers_by_id.values():
+        # NOTE QGIS 3.30.x returns "Vector", while previous versions return "VectorLayer"
```
nirvn, line 42: What's our minimum QGIS version supported? Do we really care about supporting versions that have reached EOL? Minimum version that's not death is 3.34 ATM right?

I know that's merely moving code around here but maybe worth to just clean that up :)

docker-app/qfieldcloud/core/utils2/project.py
```diff
@@ -0,0 +1,72 @@
...
+    Example of a virtual layer source with embedded layers:
+
+    - "?query=SELECT%20*%20FROM%20points"
+    - "?layer=ogr:%2Fhome%2Fuser%2Ftestdata.gpkg%7Clayername%3Dpoints_xy:points_xy:&layer=ogr:%2Fhome%2Fuser%2Ftestdata.gpkg%7Clayername%3Dpoints_xy:points_xy:&query=SELECT%20*%20FROM%20points"
+    """
+
+    url = urlparse(datasource)
+    query = parse_qs(url.query)
+
+    has_embedded_layers = "layer" in query and len(query["layer"]) > 0
```
nirvn, line 21: Are we planning to add a libqfieldsync project check to warn users when uploading such a virtual vector layer definition? 

Also, some people might not realize that they can use project layer names vs. layer definitions that rely on file path. Their configuration might work just fine with project layer names provided they are told it's a thing :)

suricactus: > Are we planning to add a libqfieldsync project check to warn users when uploading such a virtual vector layer definition?

Yes, we can and probably should. I leave it to you in the next months to prioritize/manage. I don't have the time right now :) .

> Also, some people might not realize that they can use project layer names vs. layer definitions that rely on file path. Their configuration might work just fine with project layer names provided they are told it's a thing :)

Not sure if I follow the "Also" part, it's pretty much the same suggestion - we should add a check to `libqfieldsync` to reduce confusion because people might use an embedded layer definition for already existing layer in the project? I think the QGIS people should actually kill the "Embedded layer" option OR hide it by default as an "Advanced setting". Right now it might trick people (including me) that adding the embedded layer is mandatory.

docker-app/qfieldcloud/core/utils2/project.py
suricactus: If you have a project you stored on QFC from 2 years ago that used QGIS 3.28 to parse your project data, the `layer_data` will still store the old way of defining the data. So we should keep the support for now OR patch all the data in the database. I prefer the former.

nirvn: @suricactus , fair enough.

nirvn (approved):

PR #1397, opened 2025-10-22
feat: remove GET param to display third-party auth providers on login page

gounux:
Not much more than the title, with this PR no need to add the GET param to display third-party providers login button.

### DIFF
--- a/docker-app/qfieldcloud/core/templates/admin/login.html
+++ b/docker-app/qfieldcloud/core/templates/admin/login.html
@@ -3,7 +3,5 @@
 {% block content %}
     {{ block.super }}
     {% comment %}TODO add a configuration based mechanism to render the social login instead of depending on query param {% endcomment %}
-    {% if request.GET.sso %}
-        {% include "socialaccount/snippets/login.html" %}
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/templates/admin/login.html
```diff
@@ -3,7 +3,5 @@
 {% block content %}
     {{ block.super }}
     {% comment %}TODO add a configuration based mechanism to render the social login instead of depending on query param {% endcomment %}
```
gounux, line 5: Regarding this former comment, maybe we could add a setting / envvar like `SOCIALACCOUNT_PROVIDERS_ENABLED`, even if an empty `SOCIALACCOUNT_PROVIDERS` envvar would basically be the same.

suricactus: Open to any mechanism. Note that this particular login page will be gone once: https://github.com/opengisch/QFieldCloud/pull/1293 is merged.

suricactus (approved):

PR #1398, opened 2025-10-22
chore: change the default ENVIRONMENT from test to development

suricactus:
Will easy a bit the getting started for new people.


gounux (approved):
Thanks !

PR #1399, opened 2025-10-22
Change nginx resolver to docker internal DNS for storage-download redirects

gounux:
Some QFieldCloud configurations may require resolving hosts not known by the 8.8.8.8 DNS servers, thus using the docker's internal DNS could help in such cases.

This initial `resolver 8.8.8.8` line in the nginx config was introduced by this commit : https://github.com/opengisch/QFieldCloud/commit/c3d978d2bb06472b35ae31f6b6a5a5abb10181f2, I don't see any reason why the internal DNS resolver could not be used, and this resolving config should maybe be handled by the host machine rather than the nginx IMHO.

### DIFF
--- a/docker-nginx/templates/default.conf.template
+++ b/docker-nginx/templates/default.conf.template
@@ -165,8 +165,8 @@ server {
     # if a Range header is provided
     set $file_range "$upstream_http_file_range";
 
-    # required DNS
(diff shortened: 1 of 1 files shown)


suricactus (approved):
Thanks!

PR #1402, opened 2025-10-22
Limit time it takes to render a project thumbnail

nirvn:
@suricactus , @lukasgraf , as discussed during the meeting today, this commit allows us to set an arbitrary max. rendering time for the project thumbnail. I've set it to 10 seconds, that felt reasonable.

FYI, if the rendering does not finish in time, the thumbnail will still contain a partial rendering (e.g. for the offending project, it has everything _but_ the labels). That's actually quite nice, most of the time that'd be better than nothing at all.


suricactus (approved):
Tested locally, thanks a lot!

nirvn:
@suricactus , nice!

PR #1411, opened 2025-11-05
Qf-6822 add restricted files update timestamp

Rakanhf:
This change will make it easier for QField to detect if the project file has been changed on the cloud and whether it needs to be re-downloaded on the mobile.

---

A new field `restricted_data_last_updated_at` is introduced, which will track date time whenever a restricted file
`(.qgs, .qgz, *_attachments.zip, .qgd, .qml)` is modified.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_project.py
+++ b/docker-app/qfieldcloud/core/tests/test_project.py
@@ -1,4 +1,7 @@
+import io
 import logging
+import time
+from io import StringIO
 from unittest.mock import patch
 
 from django.core.exceptions import ValidationError
@@ -18,7 +21,7 @@ from qfieldcloud.core.models import (
 )
 from qfieldcloud.subscription.models import Subscription
 
-from .utils import set_subscription, setup_subscription_plans
+from .utils import set_subscription, setup_subscription_plans, testdata_path
 
 logging.disable(logging.CRITICAL)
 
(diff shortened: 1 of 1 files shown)


suricactus:
Please add a test before considering merge.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -874,6 +874,7 @@ class ProjectAdmin(QFieldCloudModelAdmin):
         "created_at",
         "updated_at",
         "data_last_updated_at",
+        "admin_restricted_data_last_updated_at",
```
nirvn: Question: why should we append the "admin_" prefix to this property? I feel it'd work a lot better without it.

suricactus: This is how we called it at first. Probably way too verbose as Mathieu says. We can update code and ui and drop the word admin.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
         self.assertEqual(data[3]["owner"], "user1")
         self.assertEqual(data[3]["name"], "public_project_of_user1")
         self.assertTrue(data[3]["is_public"])
+
+    def test_restricted_data_last_updated_at_on_file_upload(self):
+        """Test that restricted_data_last_updated_at is updated when restricted files are uploaded."""
+        from io import StringIO
```
suricactus: Move the import at the start of the file. We import within functions only to prevent circular imports.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        # Create a project
+        project = Project.objects.create(
+            name="test_restricted_data", is_public=False, owner=self.user1
```
suricactus:
```suggestion
            name="test_restricted_data", owner=self.user1
```

No need to pass this explicitly, unless you want to change the default.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+
+        project.refresh_from_db()
+        # After uploading a restricted file, restricted_data_last_updated_at should be set
+        self.assertIsNotNone(project.restricted_data_last_updated_at)
+        first_update_time = project.restricted_data_last_updated_at
+
+        # Upload a non-restricted file (e.g., a geopackage)
+        response = self.client.post(
+            f"/api/v1/files/{project.id}/data.gpkg/",
+            {"file": StringIO("GPKG content")},
```
suricactus:
```suggestion
            f"/api/v1/files/{project.id}/data.csv/",
            {"file": StringIO("c1,c2\r\nv1,v2")},
```

You better not fake the data types too much.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.token1.key)
+
+        # Create a project
+        project = Project.objects.create(
+            name="test_restricted_data", is_public=False, owner=self.user1
+        )
+
+        # Initially, restricted_data_last_updated_at should be None
+        self.assertIsNone(project.restricted_data_last_updated_at)
+
```
suricactus, line 829: If you inject uploading a regular file before directly going to a restricted file would be great.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+        # Upload a non-restricted file (e.g., a geopackage)
+        response = self.client.post(
+            f"/api/v1/files/{project.id}/data.gpkg/",
+            {"file": StringIO("GPKG content")},
+            format="multipart",
+        )
+        self.assertTrue(status.is_success(response.status_code))
+
+        project.refresh_from_db()
+        # After uploading a non-restricted file, restricted_data_last_updated_at should NOT change
```
suricactus:
```suggestion
        # After uploading a non-restricted file, `restricted_data_last_updated_at` should NOT change
```

You are going to see this a lot withing the comments, they are everything that is code related is in backticks.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+        self.assertEqual(project.restricted_data_last_updated_at, first_update_time)
+
+        # Upload another restricted file (e.g., .qml file)
```
suricactus:
```suggestion
        # Upload another restricted file (e.g. `.qml` file)
```

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +804,91 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+        response = self.client.post(
+            f"/api/v1/files/{project.id}/project.qml/",
+            {"file": StringIO("QML content")},
```
suricactus:
```suggestion
            {"file": StringIO("import QtQuick")},
```

One day we might have file validation, let's be more realistic about the file contents.

docker-app/qfieldcloud/core/tests/test_project.py
suricactus: There are a few more like this in this PR, I would not comment them all.

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -339,6 +340,10 @@ def post(self, request, projectid, filename, format=None):
                     )
 
             project.data_last_updated_at = timezone.now()
+            if is_admin_restricted_file(filename, project.the_qgis_file_name):
+                update_fields.append("restricted_data_last_updated_at")
+                project.restricted_data_last_updated_at = timezone.now()
```
suricactus:
```suggestion
            now = timezone.now()
            project.data_last_updated_at = now

            if is_admin_restricted_file(filename, project.the_qgis_file_name):
                update_fields.append("restricted_data_last_updated_at")
                project.restricted_data_last_updated_at = now
```

Two comments:
- when updating multiple fields that will be assigned to `now()`, create a variable that ensures the time will not differ in the nanoseconds.
- add a bit of "air" before and after blocks, such as `if`, `for`, `with` etc.

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -152,7 +152,10 @@ def upload_project_file_version(
                         project=project, created_by=request.user
                     )
 
-            project.data_last_updated_at = timezone.now()
```
suricactus, line 155: Why delete this line? Please keep it and also use the same `now`.

suricactus (changes requested):

suricactus (approved):
Thanks!

Two comments to be discussed with the rest of the team:

- I think when we delete a file, we do not update the `data_last_updated_at`, which is probably a mistake.
- not so sure the tests belong to `core.test_project` or a new `filestorage.test_project` or other place in `filestorage`.

For me it's mergable at the current stage, but a second look from @gounux would be nice.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -804,3 +806,94 @@ def test_seamless_projects_fetching_with_public_ones(self):
...
+
+        # Initially, `restricted_data_last_updated_at` should be None
+        self.assertIsNone(project.restricted_data_last_updated_at)
+
+        # Upload a QGIS project file (restricted file)
+        response = self.client.post(
+            f"/api/v1/files/{project.id}/project.qgs/",
+            {"file": StringIO("<?xml version='1.0' encoding='utf-8'?>")},
+            format="multipart",
+        )
```
gounux, line 835: I don't know if it makes sense to have a valid QGIS project file here, probably not.

Just in case, there are some valid test projects in the repository,

https://github.com/opengisch/QFieldCloud/blob/8c19e1e9042a632135be5f32c6a5fb3023bfd603/docker-app/qfieldcloud/core/tests/test_secret_packages.py#L77-L82

gounux (approved):
Code looks good, thanks for the commented test scenarios !

> not so sure the tests belong to core.test_project or a new `filestorage.test_project` or other place in `filestorage`

I probably agree with that, if it can lighten the core module tests a little.

docker-app/qfieldcloud/core/tests/test_project.py
Rakanhf: Thank you, I will use the real files and push then squash and force push. but I think that keeping the tests in the core dir will be better, since we are mainly testing the model save behavior not the files upload capabilities itself.

suricactus: But where is the code that actually updates the restricted files timesramp? Isn't it in filestorage?

And the other argument is to make the CI slightly faster. We need to make huge refactoring of tests anyways one day.

suricactus:
@Rakanhf can you review if this is true and brind it to the attention to the dev group:

- I think when we delete a file, we do not update the data_last_updated_at, which is probably a mistake.

suricactus:
This is in a mergeable state, don't take my last comment as a blocker.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -85,22 +84,6 @@ def __init__(self, *args, **kwargs) -> None:
         for _model, model_admin in self._registry.items():
             model_admin.admin_site = self
 
-    def login(
```
lukasgraf, line 88: Why is the `login` method being removed here, and added again in a later commit? Can you squash out those no-op changes please?

Rakanhf: I have reverted these changes, since It was caused by an incorrect rebase, so I have adjusted the history as discussed

PR #1414, opened 2025-11-06
fix: do not check missing Sha256sum key when checking if file exists on legacy S3 during fs migration

gounux:
Boto3 S3 storage response may not provide a `Sha256sum` key in the metadata, when checking if a project `legacy_thumbnail_uri` exists.  
Before this PR, this was made using the `check_s3_key` function, which is looking for this `Sha256sum` key in metadata and raises a `KeyError` if not present.

Thus introducing a new `check_legacy_s3_file_exists` function to simply check if the file is there on legacy s3, without trying to get the `Sha256sum` metadata, which is not needed during the filestorage migration anyway when checking if a project's `legacy_thumbnail_uri`  file exists.

### DIFF
--- a/docker-app/qfieldcloud/core/utils.py
+++ b/docker-app/qfieldcloud/core/utils.py
@@ -329,6 +329,29 @@ def get_qgis_project_file(project_id: str) -> str | None:
     return None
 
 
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM! 👍

PR #1417, opened 2025-11-10
fix: flaky file size calculations

Rakanhf:
Sometimes QFieldCloud projects get negative values in `Project.file_storage_bytes`. This is obviously wrong After digging deep we decided to use `recompute_storage=True` when saving the `Project` model, since It won't have a big impact on speed or resources

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1797,6 +1797,7 @@ class Project(models.Model):
     def save(self, recompute_storage=False, *args, **kwargs):
         self.clean()
         logger.debug(f"Saving project {self}...")
+        additional_update_fields = set()
 
         if recompute_storage:
             # TODO Delete with QF-4963 Drop support for legacy storage
@@ -1809,15 +1810,24 @@ class Project(models.Model):
                     file_storage_bytes=Sum("versions__size", default=0)
                 )["file_storage_bytes"]
 
+            additional_update_fields.add("file_storage_bytes")
+
         # Ensure that the Project's storage_keep_versions is at least 1, and reflects the plan's default storage_keep_versions value.
         if not self.storage_keep_versions:
             self.storage_keep_versions = (
                 self.owner.useraccount.current_subscription.plan.storage_keep_versions
             )
 
+            additional_update_fields.add("storage_keep_versions")
+
         assert self.storage_keep_versions >= 1, (
             "If 0, storage_keep_versions mean that all file versions are deleted!"
         )
+
+        if kwargs.get("update_fields") is not None:
+            kwargs["update_fields"] = list(
+                set(kwargs["update_fields"]) | additional_update_fields
+            )
 
         super().save(*args, **kwargs)
 


docker-app/qfieldcloud/core/models.py
```diff
@@ -1810,15 +1810,27 @@ def save(self, recompute_storage=False, *args, **kwargs):
...
         assert self.storage_keep_versions >= 1, (
             "If 0, storage_keep_versions mean that all file versions are deleted!"
         )
 
+        if "update_fields" in kwargs and kwargs["update_fields"] is not None:
+            update_fields = set(kwargs["update_fields"])
+            if recompute_storage:
+                update_fields.add("file_storage_bytes")
+
+            if storage_keep_versions_modified:
```
```suggestion
            if not self.storage_keep_versions:
```

Instead of renaming the same thing in a new variable, can't we use the original source for the boolean flag above? Avoid parasitic renaming of things whenever possible.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1810,15 +1810,27 @@ def save(self, recompute_storage=False, *args, **kwargs):
                 )["file_storage_bytes"]
 
         # Ensure that the Project's storage_keep_versions is at least 1, and reflects the plan's default storage_keep_versions value.
+        storage_keep_versions_modified = False
         if not self.storage_keep_versions:
             self.storage_keep_versions = (
                 self.owner.useraccount.current_subscription.plan.storage_keep_versions
             )
+            storage_keep_versions_modified = True
```
```suggestion
            # If the `update_fields` is not `None`, then add these fields to be updated too 
            additional_update_fields = set()

			# Ensure that the Project's storage_keep_versions is at least 1, and reflects the plan's default storage_keep_versions value.
			if not self.storage_keep_versions:
				self.storage_keep_versions = (
				    self.owner.useraccount.current_subscription.plan.storage_keep_versions
				)

	            additional_update_fields.append("storage_keep_versions")
```

Later in the function you do:

```
if kwargs.get("update_fields") is not None:
    kwargs["update_fields"] = set(kwargs["update_fields"]) + additional_update_fields
```

The benefits:
- we keep the "update_fields" list updated right next to updating the model field.
- we can easily extend in the future if there is a new additional field to be updated.
- the if condition is only at one place, even though we do some use extra memory and CPU cycles.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1809,16 +1810,25 @@ def save(self, recompute_storage=False, *args, **kwargs):
...
                 self.owner.useraccount.current_subscription.plan.storage_keep_versions
             )
 
+            additional_update_fields.add("storage_keep_versions")
+
         assert self.storage_keep_versions >= 1, (
             "If 0, storage_keep_versions mean that all file versions are deleted!"
         )
 
+        if "update_fields" in kwargs and kwargs["update_fields"] is not None:
```
```suggestion
        if kwargs.get("update_fields") is not None:
```

Do I miss a scenario where the above suggestion would not work? Most of the existing codebase uses this pattern to check if keys exists.

No strong opinion shared here.

suricactus (approved):

PR #1422, opened 2025-11-11
chore: update .gitignore to include .env.test and .venv

Rakanhf:
Adjust `.gitignore`  file to exclude virtual environments and test env.

It makes things cleaner when developing locally with a venv in the same dir


lukasgraf (approved):
LGTM! 👍

PR #1426, opened 2025-11-18
follow up fix on Project `data_last_updated_at` on file delete 

Rakanhf:
When we delete a file in a Project, we don't update `data_last_updated_at`, we only update `restricted_data_last_updated_at`. which is not the correct behavior.

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/storage.py
+++ b/docker-app/qfieldcloud/core/utils2/storage.py
@@ -643,11 +643,14 @@ def delete_project_file_permanently(
     with transaction.atomic():
         _delete_by_key_permanently(file.latest.key)
 
-        update_fields = []
+        update_fields = ["data_last_updated_at"]
+
+        now = timezone.now()
+        project.data_last_updated_at = now
 
         if is_admin_restricted_file(filename, project.the_qgis_file_name):
             update_fields.append("restricted_data_last_updated_at")
-            project.restricted_data_last_updated_at = timezone.now()
+            project.restricted_data_last_updated_at = now
 
         if qfieldcloud.core.utils.is_the_qgis_file(filename):
             update_fields.append("the_qgis_file_name")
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-7112 Fix Project `data_last_updated_at` is not being updated on file delete](https://app.clickup.com/t/2192114/QF-7112)

docker-app/qfieldcloud/core/utils2/storage.py
```diff
@@ -643,11 +643,14 @@ def delete_project_file_permanently(
     with transaction.atomic():
         _delete_by_key_permanently(file.latest.key)
 
-        update_fields = []
+        update_fields = ["data_last_updated_at"]
+
+        now = timezone.now()
+        project.data_last_updated_at = now
```
suricactus, line 649: Do we update `data_last_updated_at` for all file uploads, or for all but the attachment uploads?

If we update only for non-attachment uploads, I guess we need to update only for  non-attachment deletes.

Rakanhf: When we upload files it's updated in two places :

1. [files_views.py Line 343](https://github.com/opengisch/QFieldCloud/blob/QF-7112-Follow-up-fix-project-data_last_updated_at/docker-app/qfieldcloud/core/views/files_views.py#L343)
2. [view_helpers.py Line 156](https://github.com/opengisch/QFieldCloud/blob/QF-7112-Follow-up-fix-project-data_last_updated_at/docker-app/qfieldcloud/filestorage/view_helpers.py#L156)

I can see that we don't update the `data_last_updated_at` according to attachments, however in `view_helpers.py` it's updated only when `file_type == File.FileType.PROJECT_FILE`

suricactus: Ok, than all clear. I had false memories. 

This PR is good to be merged IMO.

suricactus (approved):
A second look from @gounux is worth it, but IMO it's good to go.

PR #1427, opened 2025-11-18
feat: make attachments of a project optionally versionable

gounux:
This PR allows to make the attachments of the project not versioned.
By adding a new `are_attachments_versioned` field to the Project model, defaults to True.

In the admin, the box can be unchecked **only if there are no attachments file already existing for the project** :

image

- If checked, the file will be stored with the following path : `/projects/<project_id>/files/<path/to/file.extension>/v<date>_<version_uuid>`
- If unchecked -> `/projects/<project_id>/files/<path/to/file.extension>`

Thus making it easier to preview the attachments in the external storage.

Next steps :

- [x] add some tests in the `filestorage` module.
- [ ] add a setting that configures the `are_attachments_versioned` field default value for a new Project, at the instance level, to be discussed (potentially a follow-up PR).

### DIFF
--- a/docker-app/qfieldcloud/filestorage/tests/test_attachments.py
+++ b/docker-app/qfieldcloud/filestorage/tests/test_attachments.py
@@ -242,8 +242,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
 
         second_content_name = file.latest_version.content.name
 
-        # Following line fails since the implementation of `bytemark/webdav` adds a random file suffix before the extension.
-        # self.assertTrue(second_content_name.endswith("DCIM/file.name"))
+        self.assertTrue(second_content_name.endswith("DCIM/file.name"))
         self.assertTrue(second_content_name.endswith(".name"))
 
         response = self._delete_file(self.u1, p, "DCIM/file.name")


docker-app/qfieldcloud/core/admin.py
```diff
@@ -855,6 +857,24 @@ def __init__(self, *args, **kwargs):
         )
         if self.instance.has_attachments_files:
             self.fields["attachments_file_storage"].disabled = True
+            self.fields["are_attachments_versioned"].disabled = True
+
+    def clean_are_attachments_versioned(self):
+        value = self.cleaned_data["are_attachments_versioned"]
+
+        if not value:
```
```suggestion
        if value:
            return value
```

And the rest would be one level less nested.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -259,3 +259,21 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
         b64_auth = base64.b64encode(self.basic_auth.encode()).decode()
         basic_auth = f"Basic {b64_auth}"
         response["webdav_auth"] = basic_auth
+
+    def is_name_available(self, name, max_length=None):
+        # TODO: remove this method implemented in Django v5+.
+        # see https://github.com/django/django/blob/6f35c2e1fd71ff8f349598a59689514e213490e7/django/core/files/storage/base.py#L54
```
suricactus: Would you mind mentioning a concrete Django version and linking to a ClickUp task as a TODO (check the legacy storage pattern).

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -259,3 +259,21 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
...
+        return not self.exists(name) and not exceeds_max_length
+
+    def get_available_name(self, name: str, max_length: int | None = None) -> str:
+        """Returns a filename that is available on the configured webdav storage.
+
+        Arguments:
+            name: desired relative path of the file on the webdav server.
+            max_length: maximum length of the filename (not used)."""
+
+        if self.is_name_available(name, max_length):
```
suricactus, line 276: I fail to follow here.

```
if name is available:
   then a better new name
 
 otherwise return the raw name
 ```
 
 Don't we miss a `not`?

```
        if not self.is_name_available(name, max_length):
```

docker-app/qfieldcloud/filestorage/backend.py
gounux: Thanks, addressed by https://github.com/opengisch/QFieldCloud/pull/1427/commits/6363af4efabcb24b541a89fc4333293bb6007740.

docker-app/qfieldcloud/filestorage/backend.py
gounux: Well, I guess the `not` then leads to `return the raw name` ?

If the name is available, we call the base class function, since this base class function will generate a random suffix if not. Trying to be polite with Django here by calling the `super().method(...)` method.

Open to discuss another approach if this one is flaky.

suricactus: Probably I read it wrong. All good if it works.

PR #1428, opened 2025-11-18
feat: add whitelabelling

Rakanhf:
Adding a new feature that makes whitelabelling possible especially for dedicated instances

- Added `whitelabel.py` with default branding configuration and context processor
- Added `whitelabel()` context processor with translation support for user-facing strings
- Updated `base.html` to utilize whitelabel settings and added some style tweaks
- Adjusted `qfieldcloud.css` to decouple `base.html` from styles so it's easier to override
- Added whitelabel configuration dict in `settings.py`

Without Whitelabel :

Screenshot from 2025-12-02 18-20-46

---

image

---

With Whitelabel :

Screenshot from 2025-12-02 18-26-16

---

image

---

Settings : 

```
WHITELABEL = {
    "site_title": "Company ipsum dolor sit amet",
    "logo_navbar": "whitelabel/logoipsum-385.svg",
    "logo_main": "whitelabel/logoipsum-367.png",
    "logo_alt": "Company ipsum dolor sit amet Logo",
    "favicon": "whitelabel/logoipsum-386.ico",
}
```

### DIFF
--- a/docker-app/qfieldcloud/core/staticfiles/css/qfieldcloud.css
+++ b/docker-app/qfieldcloud/core/staticfiles/css/qfieldcloud.css
@@ -40,8 +40,19 @@ h1, h2, h3, h4, h5, h6 {
   margin-left: 1rem;
 }
 
+.qfc-logo-wrapper {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 15rem;
+  padding: 1rem 0;
+}
+
 .qfc-main-logo {
-  height: 15rem;
+  max-width: 100%;
+  max-height: 15rem;
+  width: auto;
+  height: auto;
 }
 
 .ml-5rem {
--- a/docker-app/qfieldcloud/core/templates/account/base.html
+++ b/docker-app/qfieldcloud/core/templates/account/base.html
@@ -8,7 +8,7 @@
   <meta name="color-scheme" content="dark light">
 
   {% sri_static 'css/vendor.css' %}
-  {% sri_static whitelabel.custom_css %}
+  {% sri_static 'css/qfieldcloud.css' %}
   <link rel="shortcut icon" type="image/x-icon" href="{% static whitelabel.favicon %}" />
 
   <title>
@@ -37,9 +37,9 @@
       <div class="flex-shrink-0 col-12">
         <div class="row">
           <div class="col-lg-6 offset-lg-3">
-            <p class="text-center">
+            <div class="qfc-logo-wrapper">
               <img src="{% static whitelabel.logo_main %}" alt="{{ whitelabel.logo_alt }}" class="qfc-main-logo">
-            </p>
+            </div>
 
             {% block content %}{% endblock content %}
 
--- a/docker-app/qfieldcloud/core/whitelabel.py
+++ b/docker-app/qfieldcloud/core/whitelabel.py
@@ -13,8 +13,6 @@ DEFAULT_WHITELABEL = {
     "logo_navbar": "logo_sidetext_white.svg",
     "logo_main": "logo_undertext.svg",
     "favicon": "favicon.ico",
-    # Styling
-    "custom_css": "css/qfieldcloud.css",
 }
 
 
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -881,5 +881,4 @@ JAZZMIN_SETTINGS = {
 #     "logo_main": "path/to/logo_main.svg",
 #     "logo_alt": "Your logo description",
 #     "favicon": "path/to/favicon.ico",
-#     "custom_css": "path/to/custom.css",
 # }


docker-app/qfieldcloud/core/templates/account/base.html
```diff
@@ -8,12 +8,12 @@
   <meta name="color-scheme" content="dark light">
 
   {% sri_static 'css/vendor.css' %}
-  {% sri_static 'css/qfieldcloud.css' %}
-  <link rel="shortcut icon" type="image/x-icon" href="{% static 'favicon.ico' %}" />
+  {% sri_static whitelabel.custom_css %}
+  <link rel="shortcut icon" type="image/x-icon" href="{% static whitelabel.favicon %}" />
```
suricactus, line 12: I am wondering instead of customizing the path to the favicon, can we hardcode the path but overwrite the default static file location. 

S8nce we can have multiple static file locations, we can have:

default_static/favicon.icon
whitelabel_static/favicon.icon

So by default we leave the whitelabel_static empty and add a file only if we want to change it.

No strong idea, just brain food.

Rakanhf: I think that the current way will enable us to have more flexibility and would be easier to understand and trace back later on.  I prefer to set my own path and folder rather than relying on the html file to decide where is should be. and as a best practice I will be the one creating a new folder called whitelabel inside our same static dir. 

TL;DR I think it's just simpler this way

docker-app/qfieldcloud/settings.py
```diff
@@ -870,3 +871,18 @@ def before_send(event, hint):
...
+#     "site_title": "Your Custom Title",
+#     "logo_navbar": "path/to/logo_navbar.svg",
+#     "logo_main": "path/to/logo_main.svg",
+#     "logo_alt": "Your logo description",
+#     "favicon": "path/to/favicon.ico",
+#     "custom_css": "path/to/custom.css",
+#     "footer_text": "Made with ❤️ by ",
+#     "footer_company_name": "Your Company",
+#     "footer_company_url": "https://example.com",
+# }
```
gounux, line 884: Wondering if it's possible and if it makes sense to add and parse the `WHITELABEL` variable in the environment file, maybe the same way it's done [with storages](https://github.com/opengisch/QFieldCloud/blob/d14ddab89fc89a8a7633be866c337a9e95dd7096/docker-app/qfieldcloud/settings_utils.py#L34) or [with social accounts providers](https://github.com/opengisch/QFieldCloud/blob/d14ddab89fc89a8a7633be866c337a9e95dd7096/docker-app/qfieldcloud/settings_utils.py#L100-L101) ?  

Thinking about when deploying, IMHO it seems easier to edit the `.env` file than editing this `settings.py` file, I don't know,

docker-app/qfieldcloud/core/staticfiles/css/qfieldcloud.css
```diff
@@ -35,8 +35,13 @@ h1, h2, h3, h4, h5, h6 {
   }
 }
 
-.qfc-header-logo {
+.header-logo {
```
suricactus: Is there a strong reason to drop the `qfc- ` prefix? I would insist on keeping the `qfc-prefix` for all the html and CSS thatwe are responsible for.

docker-app/qfieldcloud/core/templates/account/base.html
suricactus: I would be curious to hear @lukasgraf and @gounux opinion on this .

docker-app/qfieldcloud/core/templates/account/base.html
```diff
@@ -50,16 +50,14 @@
...
 
-  <!-- Footer (kept exactly like original) -->
+  <!-- Footer -->
   <footer class="footer mt-auto py-3">
     <div class="container">
       <nav class="navbar navbar-expand-md">
         <div class="collapse navbar-collapse" id="navbarText">
-          {% blocktrans %}
           <span class="text-muted navbar-text">
-            Made with 💝 by <a href="//opengis.ch" target="blank">OPENGIS.ch</a>. ©
```
suricactus, line 60: I think those two lines must stay no matter what, because we developed QFieldCloud.

docker-app/qfieldcloud/core/whitelabel.py
```diff
@@ -0,0 +1,67 @@
+import copy
+from typing import Any
+
+from django.conf import settings
+from django.http import HttpRequest
+from django.utils.translation import gettext as _
+
+# Default whitelabel configuration
+DEFAULT_WHITELABEL = {
```
suricactus, line 9: What if we add all of these as a new `constance` settings instead? I think it makes life much easier to modify such strings and paths without the need to restart QFieldCloud.

If we put the dict from the code into `constance`, it would be a bit weird with the logos, custom css and custom js, that's why I previously suggested we hardcode the paths to these files and allow users to mount a volume with overloads the default ones, something like:

```
- default
-- qfc.css
-- qfc.js
-- logo.svg
-- favicon.ico

- custom
-- custom.js
-- logo.svg
-- favicon.svg
```

In that case, our template can look like:

```html

{% static 'qfc.js' %}
{% static_if_exists 'custom.js' %}

```

and 

```
# templatetags.py

from django.contrib.staticfiles import finders, templatetags

def static_if_exists(source, *args, **kwargs):
    if not finders.find(source):
        return None

    return templatetags.static(source, *args, **kwargs)
```

The only downside I see is that introducing

docker-app/qfieldcloud/core/whitelabel.py
```diff
@@ -0,0 +1,67 @@
...
+    Filters out None values from user settings to preserve defaults.
+    """
+    # Start with a deep copy of defaults
+    whitelabel_settings = copy.deepcopy(DEFAULT_WHITELABEL)
+
+    # Get user settings, filtering out None values
+    user_settings = {
+        key: value
+        for key, value in getattr(settings, "WHITELABEL", {}).items()
+        if value is not None
```
suricactus: Please avoid list or dict comprehensions, better rewrite it as:

```
user_settings = {}
for k, v in s:
     if v is None:
          continue

    user_settings[k] = v
```

By the way, I would add this to some kind of utils.

docker-app/qfieldcloud/core/whitelabel.py
```diff
@@ -0,0 +1,67 @@
...
+    # Get merged settings
+    whitelabel_config = get_whitelabel_settings()
+
+    # Translate user-facing strings
+    whitelabel_config["site_title"] = _(whitelabel_config["site_title"])
+    whitelabel_config["logo_alt"] = _(whitelabel_config["logo_alt"])
+    whitelabel_config["footer_text"] = _(whitelabel_config["footer_text"])
+    whitelabel_config["footer_company_name"] = _(
+        whitelabel_config["footer_company_name"]
+    )
```
suricactus: This is not going to really work. We have automated string extraction that expects the actual string to be already there, so `_("Translate me")`. If you do `_(translate_me)` as a variable, then the string will not be extracted and the translation strings has to be somehow injected in the translation templates (`.po`(`t`)). While technically this can work, ain't nobody got time for that.

suricactus (commented):
Thanks for this new awesome functionality! (Even though I miss the nyancat demo ;) )

I have a few comments here on the approach. 

The approach we had until recently with the whitelabeled admin login was following a filename based paradigm withn no reasource path flexibility. I would suggest do not introduce path flexibility until required by a project, because it introduces a new set of complications.

The proposed workaround is to use the multi directory support for staticfiles already present in Django.

I think @gounux and @toebivankenoebi experience with whitelabeling client installations can be very valuable here.

docker-app/qfieldcloud/settings.py
Rakanhf: Technically we can use the env to load everything, but I think it's going to be better if it's hard coded in settings.py. and that's how we use it in things like `JAZZMIN_SETTINGS`, which it's what gets modified in dedicated.

docker-app/qfieldcloud/core/staticfiles/css/qfieldcloud.css
Rakanhf: The main reasoning behind this, is just to make it more generic. so it's easier to understand when someone is overwriting them.

suricactus: Please revert, we do all custom css to be qfc- prefixed.

docker-app/qfieldcloud/core/whitelabel.py
Rakanhf: I would probably prefer not to use `constance`. While runtime updates are indeed useful, in our case, especially in a dedicated environment, there’s a high chance that changes will also apply to the Jazzmin settings, which do not use `constance`. Because of that, it makes more sense to maintain a unified structure across both configurations. Either we stick with the current approach, or we introduce `constance` for Jazzmin as well, which in my view is something better suited as a future enhancement.

Regarding your suggestion to use “hard-coded paths,” I don’t necessarily disagree, but I believe having the flexibility to define the paths ourselves provides better control over where static files can be placed. It also avoids introducing additional code such as template tags in cases like the .js example.

Additionally, I’m not sure what the motivation is for white-labeling JavaScript files. We don’t currently use any custom JS files, and as far as I know, there are no immediate plans to add them. Even if there is a possibility of supporting custom JS in the future, that should be treated as a standalone feature rather than something we prepare for now.


Note that my approach to implementing this feature was mainly inspired by how **Jazzmin Admin** itself is structured. The intention is to keep things consistent and aligned with the existing pattern of how we customize the admin panel on dedicated environments. Maintaining this consistency is important to avoid conflicting configuration styles.

docker-app/qfieldcloud/core/whitelabel.py
Rakanhf: Great I will avoid using dict comprehensions, I believe since this file is solely for whitelabeling, I don't think that there is a need to decouple to utils file, just to make things compact.

docker-app/qfieldcloud/core/whitelabel.py
Rakanhf: So do you prefer to remove it ? Since I saw that we already use translations for the original implementation, and it's mainly there just not to break it, and have the option to add custom translations when needed.

docker-app/qfieldcloud/core/whitelabel.py
suricactus: ok, I just thought the code aboce is a good candidate for `merge_dicts_without_none` or something might be a common pattern across the code.

docker-app/qfieldcloud/core/whitelabel.py
suricactus: I would suggest we translate the defauts but the non-defaults (whitelabeled settings) could not be translatable until it is really needed to find a solution there.

docker-app/qfieldcloud/core/whitelabel.py
suricactus: > I would probably prefer not to use constance.

I am ok with that. Probably we need to expose this to an envvar then. Or I am missing how this will be configured in the end for each dedicated environment?

> Regarding your suggestion to use “hard-coded paths,” I don’t necessarily disagree, but I believe having the flexibility to define the paths ourselves provides better control over where static files can be placed. It also avoids introducing additional code such as template tags in cases like the .js example.

Note the JS example I have is well applicable for CSS as well. I dunno why I chose `.js` there. So in short, yes, we should avoid whitelabeling `.js` files, so `sed s/js/css/` and I think my suggestion above is still relevant.

> Note that my approach to implementing this feature was mainly inspired by how Jazzmin Admin itself is structured.

There is a slight difference between what Jazzmin and what we are doing.

Jazzmin targets the admin interface, which is usually configured once and for all with particular look and feel. So there are little to no variations expected  between instances of a django project using Jazzmin.

This makes it  very convenient for the developer to add custom CSS paths since they are in full control of the static files AND they have all the instances with the exact same configuration.

In our case we don't care about the look and feel of Jazzmin much, it can remain the same on all our instances (probably with few exceptions about the logo and stuff), but we want to control the look and feel (whitelabel) for each installed instance.

This is usually done by an ops person, who in theory could be with 0 programming knowledge.

So it will be much easier for a that person to just mount a few files to a predefined directory, rather than working with custom import paths IMO.

---

Probably what I am missing is how  your approach with custom paths is going to work on multiple instances. Can you provide an example how this is going to work if you have to configure a new dedicated instance.

Rakanhf: While there is indeed a difference between Jazzmin and what we are doing. I don't think that it's that much different, I don't see how whitelabeling won't be set also set once, there might be changes later on, but realistically I believe it will be rare. 

Also while we can technically use `envvar` I also don't support this approach as well, Since I want it to be as easy as it can when doing it, and I would highly suggest that you check the internal logic behind dedicated customizations, Since IMO I find it way easier to have it in this approach. (drag and drop your custom files, override `WHITELABEL` dict. and that's it).

Regarding the file paths, I'm not against your suggestion, but I see that the flexibility will be better, and also I don't think that this require programming knowledge, some one with ops experience should simply know where the files exist. otherwise we should adopt a 0 knowledge needed solution and make everything configurable in the admin panel (using `constance`) 


Let me know if you think that I'm missing something.

suricactus: Hey, I was missing how we are configuring the dedicated instances. Now I see your approach will work there and some decisions have already been made.

I am good with it as long as @toebivankenoebi and @gounux are good with it, since they are going to be the primary users of it.

I will still slightly prefer unchangeable filenames and simple(r for me) approach of new static directory that (maybe) overrides defaults,  but it is a matter of taste probably.

No more pushback from my side.

docker-app/qfieldcloud/core/templates/account/base.html
lukasgraf: I think I would side with @Rakanhf on this. I find that approach a bit more explicit and straightforward.
Layers of resource overrides are slightly more implicit ("magic"), I feel.

Another consideration is that browsers cache favicons very aggressively (for good reason). Probably a non-issue in production, but that could lead to some confusing development experiences locally, when switching work between different QFC projects and the favicon for `localhost` is cached.

lukasgraf (commented):
This looks good to me, nice work! 🎉 
Besides the things that have already been mentioned / discussed, I see nothing else that should be adressed :) 

But since I'm clearly lacking some context about how overrides happen on dedicated, and there's still some pending changes, I'm just leaving this as a comment instead of a green checkmark approval :)

toebivankenoebi:
Had an offline discussion with @Rakanhf today about it. General setup looks good to me too, flexible configuration in regards to paths and filenames works for me. CSS configuration should be simplified to be able to only change maybe base color for branding (e.g. Navbar, primary button color etc.) without having to copy all qfieldcloud related values as well. Will for now still be a WIP PR.

docker-app/qfieldcloud/settings.py
```diff
@@ -870,3 +871,14 @@ def before_send(event, hint):
...
+# Whitelabel configuration
+# Uncomment and customize to override default branding
+# See qfieldcloud/core/whitelabel.py for all available options and defaults
+# WHITELABEL = {
+#     "site_title": "Your Custom Title",
+#     "logo_navbar": "path/to/logo_navbar.svg",
+#     "logo_main": "path/to/logo_main.svg",
+#     "logo_alt": "Your logo description",
+#     "favicon": "path/to/favicon.ico",
+# }
```
gounux, line 884: I'm wondering if it makes sense to add a quick comment, about what which key will override ?

Rakanhf: If you visit `whitelabel.py` you can check the `DEFAULT_WHITELABEL` and see what the default keys, let me know if this make sense.

nirvn:
@Rakanhf , @toebivankenoebi , @suricactus , one objective for 2026Q1 is to apply this white labeling (instance title + logo).

For that, we'd need an endpoint to get the details. Something like /api/v1/characteristics/ or something along those lines. 

I was thinking that we could add a GET endpoint that would provide all customizable server settings (also bundle the /api/v1/auth/providers/ stuff) to avoid an increasingly large nb of requests to the server. Thoughts?

PR #1433, opened 2025-11-26
feat: disable `.qgs` for `shared_datasets` Project

Rakanhf:
In the past all projects on QFC required a .qgs project file. In shared_datasets  it is actually the opposite. we should never add a QGIS project there.

- Added validation in `admin` and `serializers` to prevent renaming projects to the shared datasets name if they contain a QGIS project file.
- Fix project status logic to account for shared datasets projects.
- Add warning message for QGIS project files in shared dataset (to warn pre-existing Projects that violates this Rule)
- Added checks to disallow QGIS project files in shared datasets during file uploads.

**Notes to handle pre-existing Projects that violate this Rule** (Project name "shared_datasets" and has QGIS Project files): 
* The Upload / Sync for QGIS will raise an error. but It will work just fine on other file types despite the Project violating the Rule.

---

image

---

image

---

image

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -49,6 +49,7 @@ from rest_framework.authtoken.models import TokenProxy
 
 from qfieldcloud.core import exceptions
 from qfieldcloud.core.models import (
+    SHARED_DATASETS_PROJECT_NAME,
     ApplyJob,
     ApplyJobDelta,
     Delta,
@@ -883,6 +884,26 @@ class ProjectForm(ModelForm):
             )
 
         return value
+
+    def clean(self):
+        cleaned_data = super().clean()
+        name = cleaned_data.get("name")
+
+        if name and name.lower() == SHARED_DATASETS_PROJECT_NAME:
+            if (
+                self.instance.pk
+                and self.instance.name.lower() == SHARED_DATASETS_PROJECT_NAME
+            ):
+                pass
+
+            elif self.instance.has_the_qgis_file:
+                raise ValidationError(
+                    _(
+                        "Cannot rename project to '{}' because it contains a QGIS project file."
+                    ).format(name)
+                )
+
+        return cleaned_data
 
 
 class ProjectAdmin(QFieldCloudModelAdmin):
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -11,6 +11,7 @@ from rest_framework.request import Request
 from qfieldcloud.authentication.models import AuthToken
 from qfieldcloud.core import exceptions
 from qfieldcloud.core.models import (
+    SHARED_DATASETS_PROJECT_NAME,
     ApplyJob,
     Delta,
     Job,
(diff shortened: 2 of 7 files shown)


docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -269,6 +269,11 @@ def post(self, request, projectid, filename, format=None):
 
         is_the_qgis_file = utils.is_the_qgis_file(filename)
 
+        if project.is_shared_datasets_project and is_the_qgis_file:
```
suricactus:
```suggestion
        if is_the_qgis_file and project.is_shared_datasets_project:
```

Only because all the `if`s down there first check for `is_the_qgis_file` and then for the rest.

docker-app/qfieldcloud/core/views/files_views.py
```diff
@@ -269,6 +269,11 @@ def post(self, request, projectid, filename, format=None):
...
+            raise exceptions.ValidationError(
```
suricactus: Is it really a validation error? I would add a new error here, probably `UnsupportedFileFormatError` or similar.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -856,6 +857,20 @@ def __init__(self, *args, **kwargs):
         if self.instance.has_attachments_files:
             self.fields["attachments_file_storage"].disabled = True
 
+    def clean(self):
```
suricactus, line 888: Very nice addition! Not sure if it covers everything, but myself I forgot about this. Good job!

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -124,6 +125,14 @@ def validate(self, data):
             if matching_projects != 0:
                 raise exceptions.ProjectAlreadyExistsError()
 
+            if data["name"].lower() == SHARED_DATASETS_PROJECT_NAME:
+                if self.instance and self.instance.has_the_qgis_file:
+                    raise ValidationError(
```
suricactus: Same validation error issue, see previous comment.

docker-app/qfieldcloud/core/views/files_views.py
Rakanhf: I will add a new validation class `QGISProjectFileNotAllowedError` since I don't think that there are better builtin exceptions

docker-app/qfieldcloud/core/serializers.py
Rakanhf: will use the same `QGISProjectFileNotAllowedError` exception.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -866,38 +866,45 @@ def test_restricted_data_last_updated_at_on_file_upload(self):
         self.assertIsNotNone(project.restricted_data_last_updated_at)
         self.assertGreater(project.restricted_data_last_updated_at, first_update_time)
 
-    def test_restricted_data_last_updated_at_on_file_delete(self):
```
suricactus, line 869: Is there a praticular reason why `test_restricted_data_last_updated_at_on_file_delete` is killed?

Rakanhf: Thanks for spotting that, This shouldn't be deleted at all!

I will revert this

PR #1439, opened 2025-12-08
feat: do not repeat the same log twice for libqfieldsync

suricactus:
By introducing `IS_WITHIN_QFC_WORKER` we are going to instruct `libqfieldsync` to not redirect logs to the `QgsMessageLog`, therefore avoiding repetitions such as:

```
14:22:54.842 QGSMSGLOG INFO     Layer "list_salutation" will use attribute "fid" as a primary key.
14:22:54.842 libqfieldsync INFO     Layer "list_salutation" will use attribute "fid" as a primary key.
```

We still want to pipe all the info from `QgsMessageLog` to the loggers, because there might be valuable debugging information for QFC project administrators.

See https://github.com/opengisch/libqfieldsync/pull/133/

### DIFF
--- a/docker-qgis/Dockerfile
+++ b/docker-qgis/Dockerfile
@@ -81,6 +81,10 @@ ENV XDG_RUNTIME_DIR=/root
 ENV PYTHONPATH=/libqfieldsync
 ENV PYTHONPATH=/qfieldcloud-sdk-python:${PYTHONPATH}
 
+# QFC specific variables
+# Controls whether libqfiedlsync should log also to the QGIS message log
+ENV IS_WITHIN_QFC_WORKER=1
+
 # Install debugpy for local development and debuggability
 ARG DEBUG_BUILD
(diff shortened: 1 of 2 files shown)


docker-qgis/Dockerfile
```diff
@@ -81,6 +81,10 @@ ENV XDG_RUNTIME_DIR=/root
 ENV PYTHONPATH=/libqfieldsync
 ENV PYTHONPATH=/qfieldcloud-sdk-python:${PYTHONPATH}
 
+# QFC specific variables
+# Controls whether libqfiedlsync should log also to the QGIS message log
```
gounux, line 85:
```suggestion
# QFC specific variables.
# Controls whether libqfiedlsync should log also to the QGIS message log.
```

:wink:

gounux (approved):

docker-qgis/Dockerfile
suricactus: For consistency within the file, I would suggest we don't add them here.

PR #1440, opened 2025-12-08
fix: ensure there is some kind of extent when generating a thumbnail

suricactus:
When creating a project with xlsform2qgis, it is possible that here is no project extent. In that case, the process proejct job just hangs forever.

FYI @nirvn

### DIFF
--- a/docker-qgis/qfc_worker/commands/process_projectfile.py
+++ b/docker-qgis/qfc_worker/commands/process_projectfile.py
@@ -124,6 +124,22 @@ def _generate_thumbnail(
     tmp_project = tmp_project_details["project"]
     map_settings = tmp_project_details["map_settings"]
 
+    if map_settings.extent().isEmpty():
+        logger.warning(
+            "Project has empty extent, using the full extent for the thumbnail."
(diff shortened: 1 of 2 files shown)


docker-qgis/qfc_worker/commands/process_projectfile.py
```diff
@@ -124,6 +124,17 @@ def _generate_thumbnail(
     tmp_project = tmp_project_details["project"]
     map_settings = tmp_project_details["map_settings"]
 
+    if map_settings.extent().isEmpty():
+        map_settings.setExtent(map_settings.fullExtent())
+
+    if map_settings.extent().isEmpty():
+        logger.warning("Project has empty extent, using layer extent for thumbnail.")
```
nirvn: That warning is wrong right? It should go in the if above?

suricactus: Thanks for spotting. Yes

suricactus: Addressed.

PR #1442, opened 2025-12-10
feat: add user inactivity threshold and welcome back message on login

Rakanhf:
Adding a new message that indicates if the user didn't login from a while (passed the settings threshold) 

- Adjust `AccountAdapter` to add message if the user passed the threshold without login
- Add `WEB_USER_INACTIVITY_THRESHOLD_DAYS` to `CONSTANCE_CONFIG_FIELDSETS` and `CONSTANCE_CONFIG`

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -750,6 +750,11 @@ CONSTANCE_CONFIG = {
         "Status page URL",
         str,
     ),
+    "WEB_USER_INACTIVITY_THRESHOLD_DAYS": (
+        30,
+        "User inactivity threshold in days. Set to 0 to disable the inactivity threshold.",
+        int,
+    ),
 }
 CONSTANCE_ADDITIONAL_FIELDS = {
     "textarea": [
@@ -772,7 +777,10 @@ CONSTANCE_CONFIG_FIELDSETS = {
     ),
     "Debug": ("SENTRY_REQUEST_MAX_SIZE_TO_SEND",),
     "Subscription": ("TRIAL_PERIOD_DAYS",),
-    "Web": ("STATUS_PAGE_URL",),
+    "Web": (
+        "STATUS_PAGE_URL",
+        "WEB_USER_INACTIVITY_THRESHOLD_DAYS",
+    ),
 }
 
 # Minimum number of bytes to ask a range when requesting a file part, otherwise a HTTP 416 is returned. Set to 0 to allow any number of bytes in the range.


docker-app/qfieldcloud/settings.py
```diff
@@ -489,6 +489,9 @@ def before_send(event, hint):
 # For now, the default is set to `True` to maintain backwards compatibility.
 ACCOUNT_PRESERVE_USERNAME_CASING = True
 
+# User inactivity threshold
+USER_INACTIVITY_THRESHOLD = timedelta(days=30)
```
gounux: Should this new setting also be available in the environment through `.env.example` ?

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -27,6 +31,29 @@ class AccountAdapter(DefaultAccountAdapter, BaseInvitationsAdapter):
...
+
+        response = super().login(request, user)
+
+        if previous_last_login:
+            delta = timezone.now() - previous_last_login
+            threshold = getattr(
+                settings,
+                "USER_INACTIVITY_THRESHOLD",
+                timedelta(days=30),
+            )
```
gounux: Am curious here why this is used, over `settings.USER_INACTIVITY_THRESHOLD` ?

JW, that's all.

docker-app/qfieldcloud/settings.py
Rakanhf: I don't think so, since we want to have more control on the delta function, if we convert this into an env variable we will need to define a default unit of time e.g `days` and we can control it that way. e.g 
```
# evn
USER_INACTIVITY_THRESHOLD=30

#Settings
USER_INACTIVITY_THRESHOLD = timedelta(days=int(os.environ.get("USER_INACTIVITY_THRESHOLD_DAYS", 30)))
```

I don't have a strong opinion on both, I think both are fine.

docker-app/qfieldcloud/core/adapters.py
Rakanhf: It's just a safer option to prevent `AttributeError` if the settings was missing for some reason. but it's non safe if `USER_INACTIVITY_THRESHOLD` isn't a `timedelta` object

gounux: Thanks for the explanation !

docker-app/qfieldcloud/settings.py
gounux: > # env
> USER_INACTIVITY_THRESHOLD=30

This looks great !

gounux: Maybe prefix with `QFIELDCLOUD_` to be iso with other envvar, so -> `QFIELDCLOUD_USER_INACTIVITY_THRESHOLD_DAYS` ?

Rakanhf: Yes, I totally agree, Thanks for the suggestion !

docker-app/qfieldcloud/settings.py
```diff
@@ -489,6 +489,9 @@ def before_send(event, hint):
 # For now, the default is set to `True` to maintain backwards compatibility.
 ACCOUNT_PRESERVE_USERNAME_CASING = True
 
+# User inactivity threshold
+QFIELDCLOUD_USER_INACTIVITY_THRESHOLD_DAYS = 30
+
```
suricactus: Please convert it to constance configuration.

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -27,6 +31,26 @@ class AccountAdapter(DefaultAccountAdapter, BaseInvitationsAdapter):
     to overcome this limitation by providing custom `new_user` method.
     """
 
+    def login(self, request, user):
+        # last_login here is the *previous* login time
+        previous_last_login = user.last_login
+
+        response = super().login(request, user)
+
+        threshold = getattr(config, "WEB_USER_INACTIVITY_THRESHOLD_DAYS", 0)
```
suricactus, line 40: I am more in favor to add `WEB_USER_INACTIVITY_THRESHOLD_DAYS` in constance. Now it will magically appear.

suricactus (approved):

PR #1443, opened 2025-12-11
feat: add public ID for user avatars

Rakanhf:
The avatar URL currently includes the username, which allows potential user enumeration. To prevent this, the username should be removed from the avatar path while keeping avatars publicly accessible.

- Added a `public_id` property to the `User` model for generating signed identifiers.
- Updated avatar URL generation to use `public_id` instead of `username`.
- Modified URL patterns and view logic to handle `public_id` for avatar file access.

### QField

Screenshot from 2025-12-18 17-07-28

### QField Mobile

Image 2025-12-18 at 16 58 55

### QFieldSync

image

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -307,8 +307,8 @@ class User(AbstractUser):
     def is_person(self):
         return self.type == User.Type.PERSON
 
-    @property
-    def public_id(self):
+    @cached_property
+    def public_id(self) -> str:
         return signing.dumps({"id": self.pk})
 
     @property
--- a/docker-app/qfieldcloud/filestorage/views.py
+++ b/docker-app/qfieldcloud/filestorage/views.py
@@ -4,6 +4,7 @@ from uuid import UUID
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core import signing
 from django.db.models import Q, QuerySet
+from django.http import Http404
 from django.http.response import HttpResponse, HttpResponseBase
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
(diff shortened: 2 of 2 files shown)


duke-nyuki:
Task linked: [QF-7129 Remove `username` from avatar URL](https://app.clickup.com/t/2192114/QF-7129)

docker-app/qfieldcloud/filestorage/views.py
```diff
@@ -219,7 +220,14 @@ def get(
         Returns:
             internal redirect to the Object Storage
         """
-        useraccount = get_object_or_404(UserAccount, user__username=username)
+
+        try:
+            data = signing.loads(public_id)
+            user_id = data["id"]
+        except Exception:
+            return redirect(staticfiles_storage.url("logo.svg"))
```
suricactus: Isn't more "correct" to return 404 in such case? Also, I guess we can be slightly more precise with the expected error if the signing fails rather than the generic `Exception` class.

docker-app/qfieldcloud/core/models.py
```diff
@@ -306,6 +307,10 @@ def get_absolute_url(self):
     def is_person(self):
         return self.type == User.Type.PERSON
 
+    @property
```
```suggestion
    @cached_property
```

We can cache this forever basically.

docker-app/qfieldcloud/filestorage/views.py
Rakanhf: I changed it to `raise Http404("Invalid avatar ID.")`, IMO I think that leaving the execption generic will be better, since I coud't find a solid exepction coming from django source code. but we can do something like `except (signing.BadSignature, KeyError, TypeError)` but I lean towards the simple `except Exception`

docker-app/qfieldcloud/core/models.py
Rakanhf: Great suggestion, Thank you !

docker-app/qfieldcloud/filestorage/views.py
suricactus: I think we only care about BadSignature error here, all other errors are actually something that is very wrong and we better catch it in sentry, no?

Rakanhf: Make sense, I agree

PR #1445, opened 2025-12-15
[R&D] feat: add object storage cleaner script and testing

Rakanhf:
Currently QFieldCloud keeps all project files using the non-legacy storage even if they are deleted by the user. This is due to enabled versioning on Exoscale and the lack of Exoscale mechanism to delete files after certain amount of time after they are deleted. This causes excessive file storage and therefore costs.

- Add `storage_cleaner.py` for scanning and deleting logically deleted objects in S3 compatable storages
- Add `test.py` for testing the script
- Add a `README.MD`

---
A small CLI to **analyze and clean logically deleted objects** in **versioned S3-compatible buckets**.

This script is **provider-agnostic** works with any S3-compatible service that supports versioning such as `MinIO`, etc.

It helps you to the following:
1. Detect whether a bucket has versioning configured.
2. Find keys whose **latest version is a delete marker** (logically deleted).
3. Report how much storage their historical versions consume.
4. **Deleting all versions** for those keys to free space.

#### Requirements
- boto3
- pytest (just for testing)

##### 1. No action provided or `--info`
```
python storage_cleaner.py my-versioned-bucket-name
```
```
No action specified. Use --info, --scan, or --prune.
Use --help for usage information.
Defaulting to --info.

Bucket Information:
--------------------------------------------------------------------------------
Bucket:            my-versioned-bucket-name
Versioning Status: Enabled
Region:            us-east-1
Endpoint URL:      http://localhost:8009
Prefix Filter:     No prefix specified
Deleted After:     No deleted after date specified
--------------------------------------------------------------------------------
✓ Connection successful.
✓ Bucket is accessible.

Use --scan to find logically deleted objects.
```

##### 1. Scanning for logically deleted objects
```
python storage_cleaner.py my-versioned-bucket-name --scan
```
##### 1. Deleting logically deleted objects
```
python storage_cleaner.py my-versioned-bucket-name --prune
```

#### Arguments
1. `--noinput` Skip confirmation prompt (use with --purge for automation)
2. `--deleted-after` Only process objects deleted after this date (ISO 8601 format, e.g. 2024-12-01:00:00:00) Cannot be used with `--deleted-since.`
3. `--deleted-since` Time duration ago (e.g. '3 days', '2 weeks'). Cannot be used with `--deleted-after.`
4. `--prefix` Key prefix to scan
5. `--profile` Object storage profile
6. `--debug` Enable debug logging
7. `--log-file` Write a rotating logbook file (e.g. /tmp/s3-cleaner.log)
8. `--scan` Scan only
9. `--prune` Delete found logically deleted objects (requires confirmation unless --noinput)
10. `--info` Show bucket information (default behavior)

By default it will load the `default` local S3 configuration 

### Testing 
Added multiple test cases 
1. Upload files to object storage - no deleted files (shouldn't find any logically deleted objects) -> Nothing deleted
2. Upload files to object storage - delete some files (should find logically deleted objects) -> Deleted files
3. Upload files to object storage - delete some files - restore (shouldn't find any logically deleted objects) -> Nothing deleted
4. Upload files, delete some, test deleted-after filter.
11. File created, deleted, created again, deleted again. Ensure ALL versions (history) are cleaned up.

# Demo
#### Scan
Screencast from 2025-12-25 00-35-29

#### Prune
Screencast from 2025-12-25 00-37-48

### DIFF
--- a/scripts/object_storage_cleaner/storage_cleaner.py
+++ b/scripts/object_storage_cleaner/storage_cleaner.py
@@ -58,6 +58,20 @@ class LogicallyDeletedObject:
     versions_count: int
 
 
+# Utility functions
+def format_bytes(num_bytes: int) -> str:
+    """Convert a byte count to a human-readable string."""
+    units = ["B", "KB", "MB", "GB", "TB", "PB"]
+    size = float(num_bytes)
+    for unit in units:
+        if size < 1024.0 or unit == units[-1]:
+            return f"{size:.2f} {unit}"
+
+        size /= 1024.0
+
+    return f"{size:.2f} PB"
+
+
 # Main class
 class ObjectStorageScanner:
     BATCH_SIZE = 1000
@@ -90,19 +104,6 @@ class ObjectStorageScanner:
         # Validate bucket immediately
         self._validate_bucket_versioning()
 
-    @staticmethod
-    def _format_bytes(num_bytes: int) -> str:
-        """Convert a byte count to a human-readable string."""
-        units = ["B", "KB", "MB", "GB", "TB", "PB"]
-        size = float(num_bytes)
-        for unit in units:
-            if size < 1024.0 or unit == units[-1]:
-                return f"{size:.2f} {unit}"
-
-            size /= 1024.0
-
-        return f"{size:.2f} PB"
-
     def _create_s3_client(
         self,
         region: str | None,
@@ -129,16 +130,16 @@ class ObjectStorageScanner:
         except (BotoCoreError, ClientError) as e:
             raise RuntimeError(f"Failed to initialize S3 client: {e}") from e
 
-    def _validate_bucket_versioning(self) -> bool:
+    def _validate_bucket_versioning(self) -> None:
         """Ensure the bucket has versioning enabled or suspended."""
         try:
             response = self.s3_client.get_bucket_versioning(Bucket=self.bucket)
             status = response.get("Status")
 
-            if status not in ("Enabled", "Suspended"):
+            if status not in ("Enabled"):
                 raise RuntimeError(
                     f"Bucket '{self.bucket}' does not have versioning history (Status: {status}). "
-                    "Nothing to clean."
+                    "Nothing to scan."
                 )
         except ClientError as e:
             raise RuntimeError(
@@ -157,20 +158,18 @@ class ObjectStorageScanner:
             else "No deleted after date specified",
         )
 
-        return status == "Enabled"
-
     def _print_progress(self, delta: int, label: str = "Deletable") -> None:
         self._scanned_count += delta
         if sys.stdout.isatty():
             sys.stdout.write(
                 f"Scanned {self._scanned_count} versions | Found {self._stats_keys} keys | "
-                f"{label}: {self._stats_versions} versions ({self._format_bytes(self._stats_bytes)})"
+                f"{label}: {self._stats_versions} versions ({format_bytes(self._stats_bytes)})"
                 + " " * 20
                 + "\r",
             )
             sys.stdout.flush()
 
-    def _print_filters(self):
+    def _print_filters(self) -> None:
         msg = [f"Scanning bucket: '{self.bucket}'"]
         if self.prefix:
             msg.append(f"Prefix filter: {self.prefix}")
@@ -182,15 +181,15 @@ class ObjectStorageScanner:
 
         logger.info("\n".join(msg) + "\n")
 
-    def _print_summary(self, is_clean: bool = False):
-        versions_action = "deleted" if is_clean else "to delete"
-        storage_action = "freed" if is_clean else "to free"
+    def _print_summary(self, prune: bool = False) -> None:
+        versions_action = "deleted" if prune else "to delete"
+        storage_action = "freed" if prune else "to free"
 
         logger.info("-" * 100)
         logger.info(f"  Total logically deleted keys: {self._stats_keys}")
         logger.info(f"  Total versions {versions_action}: {self._stats_versions}")
         logger.info(
-            f"  Total storage {storage_action}: {self._format_bytes(self._stats_bytes)}"
+            f"  Total storage {storage_action}: {format_bytes(self._stats_bytes)}"
         )
         logger.info("-" * 100)
 
@@ -236,7 +235,7 @@ class ObjectStorageScanner:
 
             yield from items
 
-    def _aggregate_versions(
+    def _aggregate_if_logically_deleted(
         self, key: str, versions: list[VersionRef]
     ) -> tuple[LogicallyDeletedObject, list[VersionRef]] | None:
         """Determine if a key is logically deleted and aggregate its stats."""
@@ -294,7 +293,7 @@ class ObjectStorageScanner:
             logger.error(f"Failed to delete batch: {e}")
             return 0
 
-    def get_bucket_info(self):
+    def print_bucket_info(self) -> None:
         """Display bucket and connection information."""
 
         logger.info("\nBucket Information:")
@@ -312,7 +311,7 @@ class ObjectStorageScanner:
         logger.info("✓ Bucket is accessible.")
         logger.info("\nUse --scan to find logically deleted objects.")
 
-    def _find_logically_deleted(
+    def _iter_logically_deleted(
         self,
         progress: Callable[[int], None] | None = None,
     ) -> Iterator[tuple[LogicallyDeletedObject, list[VersionRef]]]:
@@ -328,7 +327,9 @@ class ObjectStorageScanner:
 
             # Another key, process the previous key.
             if current_key is not None:
-                result = self._aggregate_versions(current_key, current_versions)
+                result = self._aggregate_if_logically_deleted(
+                    current_key, current_versions
+                )
                 if result:
                     yield result
 
@@ -338,7 +339,7 @@ class ObjectStorageScanner:
 
         # Process the final key
         if current_key is not None:
-            result = self._aggregate_versions(current_key, current_versions)
+            result = self._aggregate_if_logically_deleted(current_key, current_versions)
             if result:
                 yield result
 
@@ -347,10 +348,10 @@ class ObjectStorageScanner:
 
         self._print_filters()
 
-        def progress(delta):
+        def progress(delta: int) -> None:
             self._print_progress(delta)
 
-        for aggregate, versions in self._find_logically_deleted(progress=progress):
+        for aggregate, versions in self._iter_logically_deleted(progress=progress):
             self._stats_keys += 1
             self._stats_versions += aggregate.versions_count + 1
             self._stats_bytes += aggregate.total_size_bytes
@@ -366,22 +367,22 @@ class ObjectStorageScanner:
                         v.last_modified,
                     )
 
-        self._print_summary(is_clean=False)
+        self._print_summary(prune=False)
 
-    def clean(self) -> None:
+    def prune(self) -> None:
         """Scan and delete logically deleted objects."""
 
         self._print_filters()
 
         batch: list[dict[str, str]] = []
 
-        def progress(delta):
+        def progress(delta: int) -> None:
             self._print_progress(delta, label="Deleted")
 
         logger.debug("Starting to delete logically deleted objects...")
 
         # Iterate
-        for aggregate, versions in self._find_logically_deleted(progress=progress):
+        for aggregate, versions in self._iter_logically_deleted(progress=progress):
             self._stats_keys += 1
             self._stats_bytes += aggregate.total_size_bytes
 
@@ -396,15 +397,15 @@ class ObjectStorageScanner:
         if batch:
             self._stats_versions += self._delete_batch(batch)
 
-        self._print_summary(is_clean=True)
+        self._print_summary(prune=True)
 
 
-def signal_handler(sig, frame):
+def _handle_sigint(sig: int, frame: object) -> None:
     logger.info("\n\nScan interrupted by user.")
     sys.exit(0)
 
 
-signal.signal(signal.SIGINT, signal_handler)
+signal.signal(signal.SIGINT, _handle_sigint)
 
 
 def setup_logging(debug: bool = False, log_file: str | None = None) -> None:
@@ -444,7 +445,7 @@ def setup_logging(debug: bool = False, log_file: str | None = None) -> None:
         logger.addHandler(file_handler)
 
 
-def parse_since(value: str) -> datetime:
+def parse_since_cutoff(value: str) -> datetime:
     """
     Parse a duration like '3s', '30d', '2w', '3 days' and return a datetime object.
 
@@ -452,7 +453,7 @@ def parse_since(value: str) -> datetime:
         value: The duration string to parse.
 
     Returns:
-        A datetime object representing the time delta from now.
+        A UTC datetime equal to now minus the parsed duration.
     """
     value = value.strip().lower()
 
@@ -484,7 +485,7 @@ def parse_since(value: str) -> datetime:
 
 def main() -> int:
     parser = argparse.ArgumentParser(
-        description="Scan and clean logically deleted objects in S3.",
+        description="Scan and clean logically deleted objects in object storage like S3.",
         formatter_class=argparse.RawDescriptionHelpFormatter,
         epilog="""
         Examples:
@@ -495,10 +496,10 @@ def main() -> int:
         %(prog)s my-bucket --prefix logs/ --deleted-after 2024-12-01
 
         # Delete with confirmation prompt
-        %(prog)s my-bucket --purge
+        %(prog)s my-bucket --prune
 
         # Delete without confirmation (automation)
-        %(prog)s my-bucket --purge --noinput
+        %(prog)s my-bucket --prune --noinput
         """,
     )
 
@@ -520,7 +521,7 @@ def main() -> int:
 
     parser.add_argument(
         "--since",
-        type=parse_since,
+        type=parse_since_cutoff,
         metavar="DAYS",
         help="Number of days ago (e.g. 3, 3d, '7 days')",
     )
@@ -532,15 +533,15 @@ def main() -> int:
     parser.add_argument("--scan", action="store_true", help="Scan only")
 
     parser.add_argument(
-        "--purge",
+        "--prune",
         action="store_true",
-        help="Delete found objects (requires confirmation unless --noinput)",
+        help="Delete found logically deleted objects (requires confirmation unless --noinput)",
     )
 
     parser.add_argument(
         "--noinput",
         action="store_true",
-        help="Skip confirmation prompt (use with --purge for automation)",
+        help="Skip confirmation prompt (use with --prune for automation)",
     )
 
     parser.add_argument(
@@ -573,7 +574,7 @@ def main() -> int:
 
     # Create scanner
     try:
-        cleaner = ObjectStorageScanner(
+        scanner = ObjectStorageScanner(
             bucket=args.bucket,
             region_name=args.region,
             profile=args.profile,
@@ -586,16 +587,16 @@ def main() -> int:
         return 1
 
     if args.info:
-        cleaner.get_bucket_info()
+        scanner.print_bucket_info()
         return 0
 
     # Mode: Scan only
     elif args.scan:
-        cleaner.scan()
+        scanner.scan()
         return 0
 
-    # Mode: Purge (with or without confirmation)
-    elif args.purge:
+    # Mode: Prune (with or without confirmation)
+    elif args.prune:
         if not args.noinput:
             logger.info(f"\nTarget Bucket: {args.bucket}")
             if args.prefix:
@@ -610,14 +611,12 @@ def main() -> int:
                 logger.info("Operation cancelled.")
                 return 0
 
-        cleaner.clean()
+        scanner.prune()
 
     # Default fallback to info
     else:
-        logger.info("No action specified. Use --info, --scan, or --purge.")
-        logger.info("Use --help for usage information.")
-        logger.info("Defaulting to --info.")
-        cleaner.get_bucket_info()
+        logger.info("No action specified. Defaulting to --info.")
+        scanner.print_bucket_info()
         return 0
 
     return 0
--- a/scripts/object_storage_cleaner/storage_cleaner.py
+++ b/scripts/object_storage_cleaner/storage_cleaner.py
@@ -1,13 +1,20 @@
 import argparse
+import logging
 import signal
 import sys
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from datetime import datetime, timezone
+from logging.handlers import RotatingFileHandler
+from pathlib import Path
 
 import boto3
 from botocore.client import BaseClient
 from botocore.exceptions import BotoCoreError, ClientError
+
+logger = logging.getLogger(__name__)
+TRACE_LEVEL = 5
+logging.addLevelName(TRACE_LEVEL, "TRACE")
 
 
 # Data classes
@@ -137,42 +144,54 @@ class ObjectStorageScanner:
                 f"Could not check versioning status for '{self.bucket}': {e}"
             ) from e
 
+        logger.debug(
+            "\nBucket: '%s'\nVersioning: %s\nRegion: %s\nEndpoint: %s\nPrefix: %s \nDeleted After: %s",
+            self.bucket,
+            status,
+            self.s3_client.meta.region_name,
+            self.s3_client.meta.endpoint_url,
+            self.prefix or "No prefix specified",
+            self.deleted_after.isoformat()
+            if self.deleted_after
+            else "No deleted after date specified",
+        )
+
         return status == "Enabled"
 
     def _print_progress(self, delta: int, label: str = "Deletable") -> None:
         self._scanned_count += delta
-        print(
-            f"Scanned {self._scanned_count} versions | Found {self._stats_keys} keys | "
-            f"{label}: {self._stats_versions} versions ({self._format_bytes(self._stats_bytes)})"
-            + " "
-            * 20,
-            end="\r",
-            flush=True,
-        )
+        if sys.stdout.isatty():
+            sys.stdout.write(
+                f"Scanned {self._scanned_count} versions | Found {self._stats_keys} keys | "
+                f"{label}: {self._stats_versions} versions ({self._format_bytes(self._stats_bytes)})"
+                + " " * 20
+                + "\r",
+            )
+            sys.stdout.flush()
 
     def _print_filters(self):
-        print(f"\nScanning bucket: '{self.bucket}'")
+        msg = [f"Scanning bucket: '{self.bucket}'"]
         if self.prefix:
-            print(f"  Prefix filter: {self.prefix}")
+            msg.append(f"Prefix filter: {self.prefix}")
 
         if self.deleted_after:
-            print(
-                f"  Date filter:   Objects deleted after {self.deleted_after.isoformat()}"
+            msg.append(
+                f"Date filter: Objects deleted after {self.deleted_after.isoformat()}"
             )
 
-        print()
+        logger.info("\n".join(msg) + "\n")
 
     def _print_summary(self, is_clean: bool = False):
         versions_action = "deleted" if is_clean else "to delete"
         storage_action = "freed" if is_clean else "to free"
 
-        print("-" * 100)
-        print(f"  Total logically deleted keys: {self._stats_keys}")
-        print(f"  Total versions {versions_action}: {self._stats_versions}")
-        print(
+        logger.info("-" * 100)
+        logger.info(f"  Total logically deleted keys: {self._stats_keys}")
+        logger.info(f"  Total versions {versions_action}: {self._stats_versions}")
+        logger.info(
             f"  Total storage {storage_action}: {self._format_bytes(self._stats_bytes)}"
         )
-        print("-" * 100)
+        logger.info("-" * 100)
 
     def _iter_all_versions(
         self, progress: Callable[[int], None] | None = None
@@ -255,6 +274,15 @@ class ObjectStorageScanner:
         if not versions:
             return 0
 
+        if logger.isEnabledFor(TRACE_LEVEL):
+            for v in versions:
+                logger.log(
+                    TRACE_LEVEL,
+                    "Deleting object: Key=%s Version=%s",
+                    v["Key"],
+                    v["VersionId"],
+                )
+
         try:
             self.s3_client.delete_objects(
                 Bucket=self.bucket, Delete={"Objects": versions, "Quiet": True}
@@ -262,26 +290,26 @@ class ObjectStorageScanner:
 
             return len(versions)
         except ClientError as e:
-            print(f"Failed to delete batch: {e}")
+            logger.error(f"Failed to delete batch: {e}")
             return 0
 
     def get_bucket_info(self):
         """Display bucket and connection information."""
 
-        print("\nBucket Information:")
-        print("-" * 80)
-        print(f"Bucket:            {self.bucket}")
-        print("Versioning Status: Enabled")
-        print(f"Region:            {self.s3_client.meta.region_name}")
-        print(f"Endpoint URL:      {self.s3_client.meta.endpoint_url}")
-        print(f"Prefix Filter:     {self.prefix or 'No prefix specified'}")
-        print(
+        logger.info("\nBucket Information:")
+        logger.info("-" * 80)
+        logger.info(f"Bucket:            {self.bucket}")
+        logger.info("Versioning Status: Enabled")
+        logger.info(f"Region:            {self.s3_client.meta.region_name}")
+        logger.info(f"Endpoint URL:      {self.s3_client.meta.endpoint_url}")
+        logger.info(f"Prefix Filter:     {self.prefix or 'No prefix specified'}")
+        logger.info(
             f"Deleted After:     {self.deleted_after.isoformat() if self.deleted_after else 'No deleted after date specified'}"
         )
-        print("-" * 80)
-        print("✓ Connection successful.")
-        print("✓ Bucket is accessible.")
-        print("\nUse --scan to find logically deleted objects.")
+        logger.info("-" * 80)
+        logger.info("✓ Connection successful.")
+        logger.info("✓ Bucket is accessible.")
+        logger.info("\nUse --scan to find logically deleted objects.")
 
     def _find_logically_deleted(
         self,
@@ -321,10 +349,21 @@ class ObjectStorageScanner:
         def progress(delta):
             self._print_progress(delta)
 
-        for aggregate, _ in self._find_logically_deleted(progress=progress):
+        for aggregate, versions in self._find_logically_deleted(progress=progress):
             self._stats_keys += 1
             self._stats_versions += aggregate.versions_count + 1
             self._stats_bytes += aggregate.total_size_bytes
+
+            if logger.isEnabledFor(TRACE_LEVEL):
+                for v in versions:
+                    logger.log(
+                        TRACE_LEVEL,
+                        "Version: Key=%s VersionId=%s Size=%s LastModified=%s",
+                        v.key,
+                        v.version_id,
+                        v.size,
+                        v.last_modified,
+                    )
 
         self._print_summary(is_clean=False)
 
@@ -337,6 +376,8 @@ class ObjectStorageScanner:
 
         def progress(delta):
             self._print_progress(delta, label="Deleted")
+
+        logger.debug("Starting to delete logically deleted objects...")
 
         # Iterate
         for aggregate, versions in self._find_logically_deleted(progress=progress):
@@ -358,11 +399,48 @@ class ObjectStorageScanner:
 
 
 def signal_handler(sig, frame):
-    print("\n\nScan interrupted by user. Cleaning up...")
+    logger.info("\n\nScan interrupted by user.")
     sys.exit(0)
 
 
 signal.signal(signal.SIGINT, signal_handler)
+
+
+def setup_logging(debug: bool = False, log_file: str | None = None) -> None:
+    """Setup logging for the script."""
+    level = TRACE_LEVEL if log_file else (logging.DEBUG if debug else logging.INFO)
+    logger.setLevel(level)
+    logger.handlers.clear()
+    logger.propagate = False
+
+    # 2. Console Handler (The User UI)
+    console_handler = logging.StreamHandler(sys.stdout)
+    console_handler.setLevel(logging.DEBUG if debug else logging.INFO)
+    console_handler.setFormatter(logging.Formatter("%(message)s"))
+
+    # Filter out errors (they go to stderr)
+    console_handler.addFilter(lambda record: record.levelno <= logging.INFO)
+    logger.addHandler(console_handler)
+
+    # 3. Stderr Handler (Errors)
+    error_handler = logging.StreamHandler(sys.stderr)
+    error_handler.setLevel(logging.WARNING)
+    error_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
+    logger.addHandler(error_handler)
+
+    # 4. File Handler (The Audit Log)
+    if log_file:
+        Path(log_file).parent.mkdir(parents=True, exist_ok=True)
+        file_handler = RotatingFileHandler(
+            log_file, maxBytes=1000 * 1024 * 1024, backupCount=5, encoding="utf-8"
+        )
+        file_handler.setLevel(TRACE_LEVEL)
+        file_handler.setFormatter(
+            logging.Formatter(
+                "%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
+            )
+        )
+        logger.addHandler(file_handler)
 
 
 def main() -> int:
@@ -419,7 +497,22 @@ def main() -> int:
         help="Skip confirmation prompt (use with --purge for automation)",
     )
 
+    parser.add_argument(
+        "--debug",
+        action="store_true",
+        default=False,
+        help="Enable debug logging",
+    )
+
+    parser.add_argument(
+        "--log-file",
+        default=None,
+        help="Write a rotating logbook file (e.g. /tmp/s3-cleaner.log)",
+    )
+
     args = parser.parse_args()
+
+    setup_logging(debug=args.debug, log_file=args.log_file)
 
     if args.deleted_after:
         args.deleted_after = args.deleted_after.replace(tzinfo=timezone.utc)
@@ -435,7 +528,7 @@ def main() -> int:
             deleted_after=args.deleted_after,
         )
     except RuntimeError as e:
-        print(f"ERROR: {e}")
+        logger.error(e)
         return 1
 
     if args.info:
@@ -450,26 +543,26 @@ def main() -> int:
     # Mode: Purge (with or without confirmation)
     elif args.purge:
         if not args.noinput:
-            print(f"\nTarget Bucket: {args.bucket}")
+            logger.info(f"\nTarget Bucket: {args.bucket}")
             if args.prefix:
-                print(f"Prefix: {args.prefix}")
+                logger.info(f"Prefix: {args.prefix}")
 
-            print(
-                "WARNING: You are about to permanently delete logically deleted objects. This action cannot be undone!"
+            logger.warning(
+                "You are about to permanently delete logically deleted objects. This action cannot be undone!"
             )
             confirm = input("Type 'yes' to confirm deletion: ").strip().lower()
 
             if confirm != "yes":
-                print("Operation cancelled.")
+                logger.info("Operation cancelled.")
                 return 0
 
         cleaner.clean()
 
     # Default fallback to info
     else:
-        print("No action specified. Use --info, --scan, or --purge.")
-        print("Use --help for usage information.")
-        print("Defaulting to --info.")
+        logger.info("No action specified. Use --info, --scan, or --purge.")
+        logger.info("Use --help for usage information.")
+        logger.info("Defaulting to --info.")
         cleaner.get_bucket_info()
         return 0
 
--- a/scripts/object_storage_cleaner/storage_cleaner.py
+++ b/scripts/object_storage_cleaner/storage_cleaner.py
@@ -1,0 +1,550 @@
+import argparse
+import signal
+import sys
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Dict, Iterator, List, Optional, Tuple
+
+import boto3
+from botocore.client import BaseClient
+from botocore.exceptions import BotoCoreError, ClientError
+
+# Data classes
+
+
+@dataclass
+class VersionRef:
+    """Represents a single object version or delete marker.
+
+    Attributes:
+        key: Object key.
+        version_id: Version ID of this entry.
+        is_delete_marker: True if this entry is a delete marker.
+        size: Size in bytes for a normal version. For delete markers, this is 0.
+        last_modified: Timestamp when this version was created/modified.
+    """
+
+    key: str
+    version_id: str
+    is_delete_marker: bool
+    size: int
+    last_modified: datetime
+
+
+@dataclass
+class LogicallyDeletedObject:
+    """Aggregated info about a logically deleted object.
+
+    A logically deleted object is one whose latest version is a delete marker,
+    even though older versions still exist and consume storage.
+
+    Attributes:
+        key: Object key.
+        deleted_at: Timestamp of the delete marker (latest version).
+        total_size_bytes: Sum of sizes of all non-delete-marker versions.
+        versions_count: Number of non-delete-marker versions.
+    """
+
+    key: str
+    deleted_at: datetime
+    total_size_bytes: int
+    versions_count: int
+
+
+# Main class
+class ObjectStorageScanner:
+    BATCH_SIZE = 1000
+
+    def __init__(
+        self,
+        bucket: str,
+        region_name: Optional[str] = None,
+        profile: Optional[str] = None,
+        endpoint_url: Optional[str] = None,
+        access_key_id: Optional[str] = None,
+        secret_access_key: Optional[str] = None,
+        prefix: Optional[str] = None,
+        deleted_after: Optional[datetime] = None,
+    ):
+        self.bucket = bucket
+        self.prefix = prefix
+        self.deleted_after = deleted_after
+        self.region_name = region_name
+        self.profile = profile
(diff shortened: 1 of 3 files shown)


scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+    def __init__(
+        self,
+        bucket: str,
+        region_name: Optional[str] = None,
+        profile: Optional[str] = None,
+        endpoint_url: Optional[str] = None,
+        access_key_id: Optional[str] = None,
+        secret_access_key: Optional[str] = None,
+        prefix: Optional[str] = None,
+        deleted_after: Optional[datetime] = None,
```
gounux: Could we use python >= 3.9 typings, with `variable: str | None` ?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+    def _validate_bucket_versioning(self) -> bool:
+        """Ensure the bucket has versioning enabled or suspended."""
+        try:
+            response = self.s3_client.get_bucket_versioning(Bucket=self.bucket)
+            status = response.get("Status")
+            if status not in ("Enabled", "Suspended"):
+                raise RuntimeError(
+                    f"Bucket '{self.bucket}' does not have versioning history (Status: {status}). "
+                    "Nothing to clean."
+                )
```
gounux, line 143:
```suggestion
            response = self.s3_client.get_bucket_versioning(Bucket=self.bucket)
            status = response.get("Status")

            if status not in ("Enabled", "Suspended"):
                raise RuntimeError(
                    f"Bucket '{self.bucket}' does not have versioning history (Status: {status}). "
                    "Nothing to clean."
                )
```

Space for the :eyes: of @suricactus :)

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+        print(f"Region:            {self.s3_client.meta.region_name}")
+        print(f"Endpoint URL:      {self.s3_client.meta.endpoint_url}")
+        print(f"Prefix Filter:     {self.prefix or 'No prefix specified'}")
+        print(
+            f"Deleted After:     {self.deleted_after.isoformat() if self.deleted_after else 'No deleted after date specified'}"
+        )
+        print("-" * 80)
+        print("✓ Connection successful.")
+        print("✓ Bucket is accessible.")
+        print("\nUse --scan to find logically deleted objects.")
```
gounux: Any reason to use a print and not a logger from the `logging` module ?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+
+    def _find_logically_deleted(
+        self,
+    ) -> Iterator[Tuple[LogicallyDeletedObject, List[VersionRef]]]:
+        """Yields logically deleted objects."""
+        current_key: Optional[str] = None
+        current_versions: List[VersionRef] = []
```
gounux:
```suggestion
        current_versions: list[VersionRef] = []
```

We can also allow ourselves to use modern python typings IMHO :)

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+
+def parse_dt(value: str) -> datetime:
+    """Parse a datetime string from ISO 8601 format."""
+    if value:
+        try:
+            return datetime.fromisoformat(value).replace(tzinfo=timezone.utc)
+        except ValueError:
+            raise argparse.ArgumentTypeError(f"Invalid ISO 8601 date: '{value}'")
+    else:
+        return None
```
gounux: Seems that the code might produce some null results, and accepts null `value` argument as well, so maybe :

```python
def parse_dt(value: str | None) -> datetime | None:
```

?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+        help="Delete found objects (requires confirmation unless --noinput)",
+    )
+
+    parser.add_argument(
+        "--noinput",
+        action="store_true",
+        help="Skip confirmation prompt (use with --purge for automation)",
+    )
+
+    args = parser.parse_args()
```
gounux, line 573: Could we extract the `ArgumentParser` to another function ?

No hard opinion there, simply personal taste that I am not necessarily attached to :)

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+        """,
+    )
+
+    # Required
+    parser.add_argument("bucket", help="Target S3 bucket")
+
+    # Connection options
+    parser.add_argument("--prefix", help="Key prefix to scan")
+    parser.add_argument("--region", help="AWS region")
+    parser.add_argument("--profile", help="AWS profile")
```
gounux: I believe it is not necessarily `AWS`-related, S3 can also be provided by some other hosting services,

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+    parser.add_argument("--endpoint-url", help="Custom S3 endpoint URL")
+    parser.add_argument("--access-key-id", help="Access key ID")
+    parser.add_argument("--secret-access-key", help="Secret access key")
```
gounux: On linux I believe that it is possible to configure those S3 configuration settings in the home folder, e.g. [via the `aws` CLI](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html) or equivalent. Thus abstracting the credentials and making it _not_ necessary to give them explicitly. How would the script handle that ?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+                print(f"Prefix: {args.prefix}")
+
+            print(
+                "WARNING: You are about to permanently delete logically deleted objects. This action cannot be undone!"
+            )
+            confirm = input("Type 'yes' to confirm deletion: ").strip().lower()
+
+            if confirm != "yes":
+                print("Operation cancelled.")
+                return 0
```
gounux: This is nice :)

gounux (commented):
Thanks a lot, looks promising !

This is a first review round, did not checkout locally neither tested _yet_, mainly some typings & formalism comments.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Yes Sure

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I preferred `print` since everything is facing the user, and a bit simpler. but this can be enhanced later, or if you have something in mind.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Sure I agree

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Yes, that's actually correct, I will adjust the code

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Yes, we can extract the parser function, which mainly was used in an early version of the script 

```python
def parse_args(argv: Optional[Iterable[str]] = None) -> argparse.Namespace:
    """Parse command-line arguments."""
```

But I felt that since the file size is growing I kept it in the main function, personally I don't have a strong opinion about it, both option are fine.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Yes I agree, Thank you for spotting it !

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Mainly I kept this just as an option or even to use it with env e.g : 

```
export S3_ACCESS_KEY <>

python storage_cleaner.py --access-key-id S3_ACCESS_KEY ....... 
```

But boto3 by default looks for the default profile on `~/.aws/config` and `~/.aws/credentials` 

e.g 
```
# ~/.aws/credentials

[default]
aws_access_key_id = my_access_key
aws_secret_access_key = my_secret_access_key

# ~/.aws/config

[default]
endpoint_url = http://localhost:8009
region = us-east-1
```

Ref https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#cli-configure-files-using-profiles


TL;DR : You don't need to use it you should always use profiles, but it's there for some quick access for testing / dev for those don't have aws cli setup.

gounux: Alright, thanks for the detailed explanation !  
I don't remember if the `parser.add_argument` needs a `required=False` in such a case, leaving it up to you as well as resolving the comment if it's okay :)

scripts/object_storage_cleaner/storage_cleaner.py
gounux: In school a teacher once told me "hey if you're function is longer than 10 lines, I'll not read it"...  
Anyway, I don't want to spread my traumas here, leaving it up to you since I'm also okay if the parser definition is in the main function :)

scripts/object_storage_cleaner/storage_cleaner.py
gounux: Maybe just a `logger = logging.getLogger(__name__)`, then `logger.info("blablabla")` could do the trick ?

Also, offer the ability to use `logger.info` or `logger.debug`, etc. depending on the granularity of printed information we want,


Don't know ATM how the script is supposed to run : in a separate cronjob ? In a container ?  
Thinking that future us might be happy if we have some log timestamps, IDK,

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+        profile: Optional[str],
+        endpoint: Optional[str],
+        key_id: Optional[str],
+        secret: Optional[str],
+    ) -> BaseClient:
+        """Create the Boto3 S3 client."""
+        session_kwargs = (
+            {"profile_name": profile}
+            if profile
+            else {
```
suricactus: Don't do inline `if`s like this. Write a proper if/else

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+                "aws_access_key_id": key_id,
+                "aws_secret_access_key": secret,
+                "region_name": region,
+            }
+        )
+
+        try:
+            # Filter None values to let boto3 use defaults
+            clean_session_kwargs = {
+                k: v for k, v in session_kwargs.items() if v is not None
```
suricactus: Don't write cryptic inline `for` loops, write a proper one.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+            }
+            session = boto3.Session(**clean_session_kwargs)
+
+            client_kwargs = {"endpoint_url": endpoint} if endpoint else {}
```
suricactus: You can rewrite this either as:

```
if endpoint:
  client_kwargs = {
    "endpoint_url": endpoint
  }
else:
  client_kwargs = {}
```

or 

```
client_kwargs = {}
if endpoint:
  client_kwargs = { ... }
```

or
 
```
client_kwargs = {}
if endpoint:
  client_kwargs.update({ ... })
```

It's up to you. In case you are interested, I believe the first option is the cleanest for such a critical script.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+            return session.client("s3", **client_kwargs)
+        except (BotoCoreError, ClientError) as e:
+            raise RuntimeError(f"Failed to initialize S3 client: {e}") from e
+
+    def _validate_bucket_versioning(self) -> bool:
+        """Ensure the bucket has versioning enabled or suspended."""
+        try:
+            response = self.s3_client.get_bucket_versioning(Bucket=self.bucket)
+            status = response.get("Status")
+            if status not in ("Enabled", "Suspended"):
```
suricactus: This can be written in slightly different way with an enum check here, but it's fine like this, no worries.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Surprisingly enough it's also mentioned In the book `Clean Code` it argues that functions must be very small, often just a few lines, and that anything beyond ~20 lines is a red flag. which I totally disagree with, since I don't think that lines of code are the only measure to judge if the code is `clean` or not. I usually prefer to have logical chunks and keeping separation, I don't personally mind a 80 lines function but does exactly two or less logical processes.

This is highly debatable 😄 , since every implementation comes with it's own restrictions. 

My thinking process is "I have two other until functions in same file = adding parser into it's own function will make file full of function to click and read" which rings a bill that Oh I need to separate to multiple files to make things more organized and easy to ready. e.g utlis.py, parser.py, main.py, cli.py, etc.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+                    "Nothing to clean."
+                )
+        except ClientError as e:
+            raise RuntimeError(
+                f"Could not check versioning status for '{self.bucket}': {e}"
+            ) from e
+
+        return status == "Enabled"
+
+    def _iter_all_versions(self) -> Iterator[VersionRef]:
```
suricactus: I still don't get the purpose of this and the `VersionRef` object. 

Why do we need to "rename" the objects in any way, since they are already properly structured? 

If you need it just for typing, use https://youtype.github.io/boto3_stubs_docs/mypy_boto3_s3/service_resource/#objectversion .

I am also curious can't we abandon this whole thing and just use `Bucket.list_versions`? See https://boto3.amazonaws.com/v1/documentation/api/latest/guide/collections.html and https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/bucket/object_versions.html#filter .

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+
+    def _aggregate_versions(
+        self, key: str, versions: List[VersionRef]
+    ) -> Optional[Tuple[LogicallyDeletedObject, List[VersionRef]]]:
+        """Determine if a key is logically deleted and aggregate its stats."""
+        if not versions:
+            return None
+
+        # Sort by time descending (latest first)
+        versions.sort(key=lambda x: x.last_modified, reverse=True)
```
suricactus, line 250: Please confirm this, but there is no need to sort by specificication:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html#API_ListObjectVersions_Examples

I think this is a basic assumption that can be made for all the Object storage implementations.

You can verify this by writing something like this:

```
# given
# /path/file1 v1
# /path/file2 v1
# /path/file1 v2

bucket.object_versions(Prefix="/path/file")

# expected:
# /path/file1 v1
# /path/file1 v2
# /path/file2 v1
```

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+            return None
+
+        # Filter by deleted_after date if set
+        if self.deleted_after and latest.last_modified < self.deleted_after:
+            return None
+
+        total_size = sum(v.size for v in versions)
+        count = sum(
+            1 for v in versions if not v.is_delete_marker
+        )  # count actual data versions
```
suricactus: `actual_versions_count` is a bit more descriptive.

Anyways, please avoid the inline `if`. If you want to be fancy, you can do like this.

```
len(filter(lambda v: v.is_delete_marker, versions))
```

Regular for loop is prefered though.

scripts/object_storage_cleaner/storage_cleaner.py
gounux: Sure, no hard opinion honestly, I agree that logic should be the separation reason and not an arbitrary number of lines.  
Was just sharing an episode of my studies, that's all :laughing: 

Will be glad to have such discussions :)

> My thinking process is "I have two other until functions in same file = adding parser into it's own function will make file full of function to click and read" which rings a bill that Oh I need to separate to multiple files to make things more organized and easy to ready. e.g utlis.py, parser.py, main.py, cli.py, etc.

Okay, then if we want to keep this script in one single file, it's probably a better idea not to create _yet another utils function_, thanks for the argumentation :) Resolving this conversation.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+        """Yields logically deleted objects."""
+        current_key: Optional[str] = None
+        current_versions: List[VersionRef] = []
+
+        for item in self._iter_all_versions():
+            if item.key != current_key:
+                if current_key is not None:
+                    result = self._aggregate_versions(current_key, current_versions)
+                    if result:
+                        yield result
```
suricactus: Too much nesting, I am pretty sure this can be done with a few early returns/continue.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+            stats_keys += 1
+            stats_versions += aggregate.versions_count + 1
+            stats_bytes += aggregate.total_size_bytes
+
+            # Print row if <= 10
+            if stats_keys <= 10:
+                deleted_at_iso = aggregate.deleted_at.astimezone(
+                    timezone.utc
+                ).isoformat()
+                print(
```
suricactus:
```suggestion

                print(
```

Please separate the outputting with the calculations with a new line. Also, please replace all the `print` with a custom function that will make switching the output in easy way OR just use a proper logger using the logging module.

```
def log(*args):
  print(*args)
```

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Technically we can use something like 
```python
logging.basicConfig(
    level=logging.INFO,
    format='%(message)s'
)

# Log something
logger.info("Versioning Status: Enabled")
```


> Don't know ATM how the script is supposed to run : in a separate cronjob ? In a container ?
Thinking that future us might be happy if we have some log timestamps, IDK,

As far as I know, the way of running this script is not yest defined, can be via a cronjob, in container, locally by remote connection. nothing is clear yet. But I agree that having such a log file generate will be helpful in the future in case we wanted to take a deep dive in what happened. 

But I used `print` since most of the output are program output rather than actual logs. I don't have strong opinion, I think it's also fine to switch to logging instead of print. what do you think ?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+                    f"{self._format_bytes(aggregate.total_size_bytes):>12}  "
+                    f"{aggregate.versions_count:>8}  "
+                    f"{deleted_at_iso:<25}  "
+                    f"{aggregate.key}"
+                )
+
+            elif stats_keys == 11:
+                # Clear previous progress line
+                print(" " * 100, end="\r")
```
suricactus: Why only `/r`? What about mortals using unix?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+                # Print aligned continuation markers (matches table columns)
+                print(f"{'....':>9}  {'....':>11}  {'....':<32}  ....")
+                # Add blank vertical space
+                print()
```
suricactus: This is unreadable and not sure if necessary tbh. If really needed, you better abstract it in a function. Same for all the prints in the current function.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: It's not required to set r`equired=False`, since In argparse, arguments that start with `--` (or `-`) are automatically optional by default. thus doing `parser.add_argument("--prefix", required=False)` is redundant.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+            1 for v in versions if not v.is_delete_marker
+        )  # count actual data versions
+
+        aggregate = LogicallyDeletedObject(
+            key=key,
+            deleted_at=latest.last_modified,
+            total_size_bytes=total_size,
+            versions_count=count,
+        )
+        return aggregate, versions
```
suricactus, line 274:
```suggestion
        aggregate = LogicallyDeletedObject(
            key=key,
            deleted_at=latest.last_modified,
            total_size_bytes=total_size,
            versions_count=count,
            versions=versions,
        )

        return aggregate
```

Avoid tuple returns, they are hard to read, especially when you have an object under your conrol.

Prefer a newline before returns :)

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: IMO this function should never accept an invalid date or return `None`.


According to https://stackoverflow.com/a/51702332, you can directly:

```
parser.add_argument('-s', "--startdate",
    help="The Start Date - format YYYY-MM-DD",
    required=True,
    type=datetime.date.fromisoformat)
```

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: I would not allow these CLI arguments. We should use the native boto3 auth, if not custom envvars are acceptable. Testing with envvars is as easy as passing args.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,550 @@
...
+    parser.add_argument("--endpoint-url", help="Custom S3 endpoint URL")
+    parser.add_argument("--access-key-id", help="Access key ID")
+    parser.add_argument("--secret-access-key", help="Secret access key")
+
+    # Filter options
+    parser.add_argument(
+        "--deleted-after",
+        type=parse_dt,
+        help="Only process objects deleted after this date (ISO 8601 format)",
+    )
```
suricactus, line 533: Why is that? We don't need a fixed date, we need a rolling date. So it should be `--since-days`, otherwise how the caller of the script will calculate the proper date?

suricactus (commented):
Summary of the review:

The script needs a few iterations before it gets in a better shape. So be patient :)

There are a few things that are still not addressed since the preliminary review via chat.

In short, I think the script is more complex than it should be. Both in terms of algorithmic approach, as well as way of writing it. A bit more OOP than my taste, this is a script in the end.

Note this script will be used by people who are not necessarily experts in Python. Write the script like someone who sees python for the first time will read it.

Please ping me when the current comments are addressed for a second round. Feel free to simplify things that you recognize as complex or not obvious, but were not commented in this round.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I don't think it's needed here, since we will have extra code, no default enums provided from boto. 

It will look like this 

```python

class BucketVersioningStatus(str, Enum):
    ENABLED = "Enabled"
    SUSPENDED = "Suspended"

# If 
if status not in (
    BucketVersioningStatus.ENABLED,
    BucketVersioningStatus.SUSPENDED,
):
```
unless we have something like **[mypy-boto3-s3](https://boto3-stubs.readthedocs.io/en/stable/mypy_boto3_s3/literals/#bucketversioningstatustype)** installed we can use 
```python
from mypy_boto3_s3.literals import BucketVersioningStatusType
```

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: `bucket.object_versions` doesn’t group by key. It just loops over what S3 returns. Since S3 sends `Versions` and `DeleteMarkers` as separate lists, keys can appear again later in the same page (like file1 → file2 → file1).


I have created a small function to test this out :

```python 
def print_buckt_order():
    # Setup
    s3 = boto3.resource("s3")
    bucket = s3.create_bucket(Bucket="demo-order")
    s3.BucketVersioning("demo-order").enable()

    # Create 1st key
    key1 = "A"
    bucket.put_object(Key=key1, Body=b"data")
    bucket.Object(key1).delete()

    # Create 2nd key
    key2 = "B"
    bucket.put_object(Key=key2, Body=b"data")
    bucket.Object(key2).delete()

    print(f"{'Key':<5} Type")
    for v in bucket.object_versions.all():
        print(f"{v.key:<5} {'DeleteMarker' if v.size is None else 'Version'}")
```

Output

```
Key   Type
A     Version
B     Version
A     DeleteMarker
B     DeleteMarker
```

which confirm that it's going to be an issue if we didn't sort them out first.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: > I still don't get the purpose of this and the VersionRef object.

Versions have a Size attribute, DeleteMarkers do not. Accessing `.size` on a Delete Marker object from the raw API would raise an error.

> I am also curious can't we abandon this whole thing and just use Bucket.list_versions? See

I'm not sure I got this right, but I couldn't find anything related to `Bucket.list_versions` am I missing something ?

suricactus: You are right- there is no size attribute- but I still find

```
size = 0
if v.is_delete_marker:
  # delete markers do not have a size attribute
  size += 0
else:
  size += v["size"]
```

OR 

```
# delete markers...
size += v.get("size", 0) # might be getattr
```

This way we drop the need of two classes and a special function. Less code is more, even if we have the false perception that the classes will cahe explicit typing etc. 

Again, this is a script we must be able to debug easily by anyone, most likely people outside our team. Keep it as simple as possible. The native types from boto3 will always be better documented.

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Thanks! This means that we are pretty much doomed, because what if you create a key1 at t1, then create another 10000000 files, then delete key1 at t2 and then recreate key1 at t3. We need to keep in memory the whole list of 100000003 versions to be able to sort them.

Probably we will need to send a specific prefix, so we can have a smaller subset of files in memory to sort? Or sensing the prefix will (hopefully) sort the keys?

Rakanhf: While it's not mentioned in the s3 documentation, the response is actually sorted alphabetically by key. so in this case we can guarantee that we won't be deleting the wrong key or miss logically deleted keys as long as we sort them together correctly **(_merge sort_)** (which we are doing currently in `_iter_all_versions`).
 

here is small script I created to prove that the order is alphabetically correct for versions and deleted versions lists (I tried to make it very simple aka bad code but easy to read)

### Function summery
1. Create 100 files that start with a random char from the alphabit
2. Create 2 files with special chars, to compare python sort and s3 sort
3. Delete some files randomly 
4. Check the order using `object_versions.all()`
5. Check the order using `list_object_versions`

The function output confirm that sorting is correct regardless of page size or objects creation deletion position.

meaning that if we created FileA -> created 9999999 files (File_rndm) -> Deleted FileA -> restored FileA we will still get the following : 

```
FileA (version)
FileA (version)
... (other versions) ...
`FileA (DeleteMarker)`  <-- APPEARS LATER
```

Even though thousands of other files were created in between the events of FileA, they have different names (e.g., File_rndm...). Therefore, in the alphabetical index, they appear strictly after FileA.

### Function 

```python 
def prove_s3_alphabetical_ordering_with_deletes():
    import string
    import random
    import boto3

    # Setup
    s3 = boto3.resource("s3")
    bucket_name = "demo-alphabetical-ordering-deletes"

    # Cleanup previous run if needed
    try:
        bucket = s3.Bucket(bucket_name)
        bucket.object_versions.delete()
        bucket.delete()
    except Exception as e:
        pass

    bucket = s3.create_bucket(Bucket=bucket_name)
    s3.BucketVersioning(bucket_name).enable()

    # List of all possible letters (ascii_letters, digits)
    alphabet = list[str](string.ascii_letters + string.digits)

    # 1. Create 100  alphabetically ordered files
    keys = []
    for i in range(100):
        letter = random.choice(alphabet)
        key = f"{letter}-{i}"
        keys.append(key)
        bucket.put_object(Key=key, Body=b"data")


    # Create 2 files with punctuation to check s3 sorting
    bucket.put_object(Key="Image!.jpg", Body=b"data")
    keys.append("Image!.jpg")
    bucket.put_object(Key="Image#.jpg", Body=b"data")
    keys.append("Image#.jpg")



    # 2. Delete a random sample of them to create DeleteMarkers
    delete_keys = []
    for key in keys:
        if random.choice([True, False]):
            bucket.Object(key).delete()
            delete_keys.append(key)


    manually_alphabetically_ordered_keys = sorted(keys)
    manually_alphabetically_ordered_keys_2 = manually_alphabetically_ordered_keys.copy()

    manually_alphabetically_ordered_delete_keys = sorted(delete_keys)
    manually_alphabetically_ordered_delete_keys_2 = manually_alphabetically_ordered_delete_keys.copy()

    # iterate through all versions and delete markers and check if the ordering is correct
    # Using bucket.object_versions.all()
    for v in bucket.object_versions.all():
        # Determine type
        v_type = 'DeleteMarker' if v.size is None else 'Version'

        # Check if the sorted correctly
        if v_type == 'Version' and v.key == manually_alphabetically_ordered_keys[0]:
            manually_alphabetically_ordered_keys.pop(0)
        elif v_type == 'DeleteMarker' and v.key == manually_alphabetically_ordered_delete_keys[0]:
            manually_alphabetically_ordered_delete_keys.pop(0)
        else:
            print(f"Ordering is incorrect for {v.key}")
            break


    # Using paginator
    paginator = bucket.meta.client.get_paginator('list_object_versions')
    page_iterator = paginator.paginate(Bucket=bucket_name, MaxKeys=10000)
    for page in page_iterator:
        for v in page.get('Versions', []):
            if v['Key'] == manually_alphabetically_ordered_keys_2[0]:
                manually_alphabetically_ordered_keys_2.pop(0)
            else:
                print(f"Ordering is incorrect for {v['Key']}")
                break

        for dm in page.get('DeleteMarkers', []):
            if dm['Key'] == manually_alphabetically_ordered_delete_keys_2[0]:
                manually_alphabetically_ordered_delete_keys_2.pop(0)
            else:
                print(f"Ordering is incorrect for {dm['Key']}")
                break

    print("Ordering is correct")

    # Cleanup
    bucket.object_versions.delete()
    bucket.delete()

```

### Output 


```bash
Ordering is correct
```

### TL;DR
AWS S3 guarantees that `ListObjectVersions` returns keys in **alphabetical (lexicographical) order**. This allows us to process the bucket as a continuous stream without needing to load everything into memory. 

By manually sorting the `Versions` and `DeleteMarkers` from each response page into a single interleaved list (as `ObjectStorageScanner` does), we ensure that **all history for a specific file is grouped together**. This guarantees that the script correctly identifies the "latest" version to determine deletion status, regardless of when other files were created.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I think having the data classes makes it easier to read, at least from my perspective, Since if I removed `VersionRef` I would need to manually map out the structure, but when using the data class, I can clearly check the reference and it's structure.

This is my personal way of how I view code and anything with definition / types will make me more comfortable even if it takes a couple more lines of code.

At the begging I didn't lean into having an OOP structure at all, but due to the complexity of this script, making it OOP and have data classes made it a bit easier and clear for me.

suricactus: I don't want to overrule here if it is simpler for you. Don't take this comment as a blocker, but more as inspiration.

I will just summarize my suggestions that will make this script and it's maintenence easier based on my experience:

- avoid OOP, especially in scripts. Use imperative code, it's simpler mental model. The complexity in OOP comes mostly from state management, so even if you stick with OOP, avoid using `self` a s a default storage for stuff whenever possible. My experience showed me that passing state as readonly parameters/return values is not immediately obvious at writing time, but easier to follow at reading.
- avoid introducing new types when types already exist and are well documented. If this tiny wrapper helps you to follow, I am ok with it, I find it redundant though.
- Keep it very obvious what is going on by reducing the block nesting using early returns. Reduce the number of methods/functions to avoid jumping around the code.

Please check with @toebivankenoebi who will most likely deploy the script in the end how he feels about the OOP approach at a bit later stage when the script is ready for a semi-final review.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I remember that this was introduced in the first revision of the script, and we moved forward with a specific date time argument rather than a rolling days window. 

So instead of using  `--since-days 30` we use `--deleted-after 2025-12-01` 

The reason behind this approach, is to have more control over the date, and also easier to test locally without mocking

suricactus: I am sorry for this misunderstanding, I either didn't mean that or I just said something that does not make sense at all. Feel free to point it to me over chat so I can print it and hang it on my wall of shame.

`--since-days`/`--deleted-for-days` should be the case, as the script is always rolling forward and a strict date cannot be passed. In theory the arg can be smarter like `--deleted-for 30 days` or `--deleted-for 30 seconds`. Make sure the days and seconds are in their full name, so there is no accidental overlook.

Rakanhf: This is a great idea, while it will have some complexity, but I agree with you it's worth it. since it will make it easier for a cron with a predefined calls rather than calculating the dates again. 

Thank you for the suggestion

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: `versions` are used for the clean function where we actually delete them

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I converted it to logging after an internal discussion and a bit of research.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Resolving comment, changed to logging

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -443,6 +444,44 @@ def setup_logging(debug: bool = False, log_file: str | None = None) -> None:
         logger.addHandler(file_handler)
 
 
+def parse_since(value: str) -> datetime:
+    """
+    Parse a duration like '3s', '30d', '2w', '3 days' and return a datetime object.
```
suricactus, line 461: Do not accept the short names such as s, secs, d, w, but only the full ones. This can be disastrous for such important script, check my original comment on this highlighting the same argument.

Rakanhf: Sorry for misunderstanding, I took inspiration from other places on how it's usually handled, and faced a dead end when I realized `m` can be for both minute and month 😸, so I just decided to disable months. 

But yes I agree with you about this, later on in the future we can expand or enhance it.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+    logger.info("\n\nScan interrupted by user.")
+    sys.exit(0)
+
+
+signal.signal(signal.SIGINT, _handle_sigint)
+
+
+def setup_logging(debug: bool = False, log_file: str | None = None) -> None:
+    """Setup logging for the script."""
+    level = TRACE_LEVEL if log_file else (logging.DEBUG if debug else logging.INFO)
```
suricactus: Please do not use inline `if` in QFieldCloud code. This is unreadable by the common folks. Do not apply such magic, make as many of the configuration required and explicit. The log level, if not by default, should be passed as a cli param/envvar and no magical logic should be applied.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+        file_handler = RotatingFileHandler(
+            log_file, maxBytes=1000 * 1024 * 1024, backupCount=5, encoding="utf-8"
+        )
+        file_handler.setLevel(TRACE_LEVEL)
+        file_handler.setFormatter(
+            logging.Formatter(
+                "%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
+            )
+        )
+        logger.addHandler(file_handler)
```
suricactus, line 456: Leave this to the outer level, there are log rotators that are doing this much better than this code. Just kill this functionality.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+
+    # Filter out errors (they go to stderr)
+    console_handler.addFilter(lambda record: record.levelno <= logging.INFO)
+    logger.addHandler(console_handler)
+
+    # 3. Stderr Handler (Errors)
+    error_handler = logging.StreamHandler(sys.stderr)
+    error_handler.setLevel(logging.WARNING)
+    error_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
+    logger.addHandler(error_handler)
```
suricactus, line 442: Have a single output from the script, ideally to STDOUT.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+            self.s3_client.meta.endpoint_url,
+            self.prefix or "No prefix specified",
+            self.deleted_after.isoformat()
+            if self.deleted_after
+            else "No deleted after date specified",
+        )
+
+    def _print_progress(self, delta: int, label: str = "Deletable") -> None:
+        self._scanned_count += delta
+        if sys.stdout.isatty():
```
suricactus, line 163: Do not do any branching on printing based on `tty`-ness. Just print a regular log from the caller instead. Kill this function all together.

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Please do not use tuple returns

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+                logger.log(
+                    TRACE_LEVEL,
+                    "Deleting object: Key=%s Version=%s",
+                    v["Key"],
+                    v["VersionId"],
+                )
+
+        try:
+            self.s3_client.delete_objects(
+                Bucket=self.bucket, Delete={"Objects": versions, "Quiet": True}
```
suricactus, line 293: Can you please document what `Quiet` parameter means?

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,628 @@
...
+            f"  Total storage {storage_action}: {format_bytes(self._stats_bytes)}"
+        )
+        logger.info("-" * 100)
+
+    def _iter_all_versions(
+        self, progress: Callable[[int], None] | None = None
+    ) -> Iterator[VersionRef]:
+        """Low-level generator that yields all versions and delete markers, sorted by Key."""
+        paginator = self.s3_client.get_paginator("list_object_versions")
+        kwargs = {"Bucket": self.bucket, "MaxKeys": 10000}
```
suricactus: Are you sure the `MaxKeys: 10000` really works on different object storage providers? S3 explicitly says the page is no more than 1000 objects/versions.

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: @Rakanhf if you run your script multiple times, you are going to see that order is sometimes random and it prints "ordering incorrect" with `rustfs`.

suricactus: In a single page of versions/objects (up to 1000), we are guaranteed to have both versions and deletion markers, which is actually good. Therefore the limitation of 1000 object versions, now makes sense.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I believe that the main motivation behind using Logging over all was to improve quality and  have this flexibility, since why we didn't just simply leave it to `print` anyways from the beginning ? 

Main motivation behind the separation of handler, is when using piping or redirecting the output, we don't get the user facing output to the files or pipe. 

Also this approach will allow us to advance the logger later on, and for something like the log reporting. I think if these are invalid then `print` was just fine even if it's not the go to practice, It will get the job done and will be easy to read. 

Let me know If I'm missing something.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I'm not sure If I'm getting it quite well, I will appreciate if you can share other log rotators, so I have a reference. 

The main idea behind this, was enabling log files, so we can trace what will be deleted and what is deleted. and taking advantage of the current logging setup.

e.g : 

```python 
           for v in versions:
                logger.log(
                    TRACE_LEVEL,
                    "Deleting object: Key=%s Version=%s",
                    v["Key"],
                    v["VersionId"],
                )
```

and 

```python 
                for v in versions:
                    logger.log(
                        TRACE_LEVEL,
                        "Version: Key=%s VersionId=%s Size=%s LastModified=%s",
                        v.key,
                        v.version_id,
                        v.size,
                        v.last_modified,
                    )
```

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: I think that this function is very helpful, since it gives the feel that the script is actually running, since If you have 100k+ files, this script will take time to run a scan or a prune. 

and without any output / feedback, I think it will be a bit confusing. 

Regarding "printing based on tty-ness". this was a workaround so looks better and not keep printing new lines and also to avoid anything going to pipes or logging files.

Let me know if you think that there is a better solution for this.

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: Sure.

Just for reference.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_objects.html

*The operation supports two modes for the response: verbose and quiet. By default, the operation uses verbose mode in which the response includes the result of deletion of each key in your request. In quiet mode the response includes only keys where the delete operation encountered an error. For a successful deletion in a quiet mode, the operation does not return any information about the delete in the response body.*

scripts/object_storage_cleaner/storage_cleaner.py
Rakanhf: My bad, it's a typo, it should be `1000` only. 

Ref : https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html

Thank you !

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Flexibility here is a debt that I am not sure with the current team we can pay.

Single file with a lot of grepping is easier to manage than multiple files. It might sound ridiculous, but that is the sad truth of my experience managing multiple scripts.

The purpose of switching from `print` to `logging` has multiple benefits, including log level control, something you can never do with prints without `if`-ing all the time. Multiple output streams is something that we may benefit in the future, but nobody meant to be implemented right away.

By the way, check the https://docs.python.org/3/library/logging.config.html#dictionary-schema-details which is doing similar things, without the need to write any code.

I think I know what was the main (good) motivation behind all this, but I am pretty sure we need it as of now. I personally don't know when and how this script is going to fail, I guess nobody in the universe knows. Write the basic logs, write them in a easy to manage way. Let's first collect some experience with the real world problems before we try to solve probably non-existing ones by adding too much branching and clumsiness in the code.

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Just print to STDOUT and you don't have to solve any of these problems. 

https://github.com/opengisch/QFieldCloud/pull/1445#discussion_r2650033258

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: One run of the script should produce one log file. We can easily do that with a timestamp suffix.

If we want cleanup or something more advanced, our good friend https://linux.die.net/man/8/logrotate exists.

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: BTW, avoid adding dicts with more than 1 key without a trailing comma.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,639 @@
...
+    )
+
+    # Required
+    parser.add_argument("bucket", help="Target S3 bucket")
+
+    # Connection options
+    parser.add_argument("--prefix", help="Key prefix to scan")
+    parser.add_argument("--region", help="Region")
+    parser.add_argument("--profile", help="Profile")
+    parser.add_argument("--endpoint-url", help="Custom S3 endpoint URL")
```
suricactus, line 526: Kill all of these args

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Please still kill it :)

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,639 @@
...
+        "--deleted-since",
+        type=parse_since_cutoff,
+        metavar="DURATION",
+        help="Time duration ago (e.g. '3 days', '2 weeks'). Cannot be used with `--deleted-after.`",
+    )
+
+    # Action flags
+    parser.add_argument(
+        "--info", action="store_true", help="Show bucket information (default behavior)"
+    )
```
suricactus, line 545: Kill it

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,639 @@
...
+    parser.add_argument("--scan", action="store_true", help="Scan only")
+
+    parser.add_argument(
+        "--prune",
```
suricactus, line 549:
```suggestion
    parser.add_argument(
        "--permanently-delete-versions",
```

Be scary when there is a chance to destroy data. Should be very very explicit what it is going to do.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,639 @@
...
+    )
+
+    parser.add_argument(
+        "--noinput",
+        action="store_true",
+        help="Skip confirmation prompt (use with --prune for automation)",
+    )
+
+    parser.add_argument(
+        "--debug",
```
suricactus, line 561:
```suggestion
        "--log-level",
```

Make it use an enum (or something native, no idea) that accepts the built-in log levels.

scripts/object_storage_cleaner/storage_cleaner.py
```diff
@@ -0,0 +1,639 @@
...
+        action="store_true",
+        default=False,
+        help="Enable debug logging",
+    )
+
+    parser.add_argument(
+        "--log-file",
+        default=None,
+        help="Write a rotating logbook file (e.g. /tmp/s3-cleaner.log)",
+    )
```
suricactus, line 571: Kill

scripts/object_storage_cleaner/storage_cleaner.py
suricactus: Because I am lazy, note for myself:

```
export AWS_ENDPOINT_URL=http://172.17.0.1:8009 AWS_ACCESS_KEY_ID=rustfsadmin AWS_ACCESS_SECRET_KEY=rustfsadmin
```

```
import boto3

s3 = boto3.resource("s3")
bucket_name = "qfieldcloud-local"

bucket = s3.Bucket(bucket_name)
bucket.object_versions.delete()
bucket.delete()

bucket = s3.create_bucket(Bucket=bucket_name)
s3.BucketVersioning(bucket_name).enable()

root_key = "file"
special_key = "SPECIAL__FILE__000"
bucket.put_object(Key=special_key, Body="v1")
bucket.put_object(Key=special_key, Body="v2")

for i in range(1, 1000):
    key = f"{root_key}__{i:>03d}"
    bucket.put_object(Key=key, Body=key)


bucket.Object(special_key).delete()
bucket.Object("file__998").delete()


for idx, v in enumerate(bucket.object_versions.all()):
    print(f"{idx:>03d}", v.key, v.version_id, v.size)

```

suricactus (commented):
We are moving forward, but there are more things to be removed:

- please do not use inline ifs (binary/ternary operators) in Python code related to qfield and qfieldcloud. Python syntax for it is unreadable for people coming from other languages and it is confusing even for regular pythonists sometimes. The only reason we don't have https://github.com/afonasev/flake8-if-expr enabled is because it is slow and that one can argue they might be useful in simple situations.
- simplify logging logic: it should just log simple strings with minimal dynamic part.
- avoid flushing the buffers: there is PYTHONUNBUFFERED for people who really need it
- print everything line by line to stdout and leave the caller to do whatever they want with the buffer. It must be easy for grepping.
- do not add functionality that is very well handled by other tools, e.g. log rotation.
- avoid complex logic with logging and if statements when log appears. That's why different log levels exist. It should be very simple to follow what is going on from the log without any evaluation of if statements.
- please remove the "Connection options" params and use envvars for that purpose.
- avoid using tuple as return values, unless well argumented and documented why there is no other option.
- avoid using default variables for scripts. If you need a parameter, they should be explicit. If you don't need it explicit, then drop the parameter.
- in scripts like this do not use inline `for`s and `map`, `filter` and `reduce`. Functional code is less readable in case of emergency. 

Once addressed, please consider the following:

- remove the OOP abstraction for state management and using pure imperative code.
- avoid using the non-native container object `VersionRef`. 

The short summary is: start small and only add functionality if really required.

suricactus:
I am closing this, unless we want to keep it open for some reason?

PR #1447, opened 2025-12-17
Docs: Describe how to connect QField to local QFC dev instance.

lukasgraf:
Add some docs that describe how to connect QField to local QFC dev instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,34 @@ establish a secure connection to it. To learn more about this situation and
 how to fix it, please visit the web page mentioned above.
 ```
 
+## Connecting QField to a local QFieldCloud instance
+
+The QFieldCloud REST API is used across clients, including the QField app.
+One can test their local QFieldCloud instance by running QField on their development machine.
+[Install QField](https://qfield.org/get) for your operating system.
+
+If you need to connect to your local QFieldCloud instance from a device different from your development machine however (e.g. your mobile phone), the following prerequisites must be met:
+
+- The QFieldCloud app container's ports must be exposed on the public interface (0.0.0.0) of your dev machine.
+  To check this, run `docker compose ps app`. It should say `0.0.0.0:8011->8000/tcp`.
+- Your dev machine and your mobile phone need to be on the same network (e.g. same WiFi)
+- Find out the external IP address of the interface that the app container's HTTP port is bound to.
+  E.g. something like `ip -f inet addr show eth1` (make sure to pick the right interface if you have several NICs, like Ethernet and WiFi).
+- For easier access via hostname, you need to have an mDNS service like `avahi` running on your dev machine.
+  On Linux, check that `avahi` is running: `systemctl status avahi-daemon`
+  If not, install `avahi-daemon` and make sure it's running.
+  That should then allow you to access your dev machine as `<your-dev-machine-hostname>.local` from devices on the same network.
+  E.g. if the hostname of your dev machine is `foo`, you should be able to reach it using `ping foo.local` from another machine on the same network.
+- In QFieldCloud's settings, the external IP address and hostname need to be added to `DJANGO_ALLOWED_HOSTS`.
+
+This should then be enough to connect the QField mobile app to your local QFieldCloud instance:
+
+- Start QField
+- When asked to sign in, double tap on the Nyuki logo to enter a custom QFieldCloud server URL
+- Enter the URL using your mDNS hostname (or IP address, if mDNS doesn't work for you):
+  - `http://192.168.1.x:8011/` or
+  - `http://<your-dev-machine-hostname>.local:8011/`
+
 ## Code style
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -305,6 +305,29 @@ Now connecting with `curl` should fail with a similar error:
     establish a secure connection to it. To learn more about this situation and
     how to fix it, please visit the web page mentioned above.
 
+## Connecting QField to a local QFieldCloud instance
+
+In order to connect the QField mobile app to your local QFieldCloud development instance, the following prerequisites must be met:
+
+- The QFieldCloud app container's ports must be exposed on the public interface (0.0.0.0) of your dev machine
+  To check this, run `docker compose ps app`. It should say `0.0.0.0:8011->8000/tcp`.
+- Your dev machien and your mobile phone need to be on the same network (e.g. same WiFi)
```
gounux:
```suggestion
- Your dev machine and your mobile phone need to be on the same network (e.g. same WiFi)
```

gounux (approved):
Just a spotted typo :)  
Thanks for this guide !

README.md
```diff
@@ -305,6 +305,29 @@ Now connecting with `curl` should fail with a similar error:
...
+- Find out the external IP address of the interface that the app container's HTTP port is bound to.
+  E.g. something like `ip -f inet addr show eth1` (make sure to pick the right interface if you have several NICs, like Ethernet and WiFi)
+- For easier access via hostname, you need to have an mDNS service like avahi running on your dev machine.
```
suricactus:
```suggestion
- For easier access via hostname, you need to have an mDNS service like `avahi` running on your dev machine.
```

README.md
```diff
@@ -305,6 +305,29 @@ Now connecting with `curl` should fail with a similar error:
...
+  On Linux, check that `avahi` is running: `systemctl status avahi-daemon`
+  If not, install `avahi-daemon` and make sure it's running. That should then allow you to access your dev machine as `<your-dev-machine-hostname>.local` from devices on the same network. E.g. if the hostname of your dev machine is `foo`, you should be able to reach it using `ping foo.local` from another machine on the same network.
```
suricactus:
```suggestion
  If not, install `avahi-daemon` and make sure it's running.
  That should then allow you to access your dev machine as `<your-dev-machine-hostname>.local` from devices on the same network.
  E.g. if the hostname of your dev machine is `foo`, you should be able to reach it using `ping foo.local` from another machine on the same network.
```

We tend to add new markdown with one sentence per line for better history tracking.

README.md
```diff
@@ -305,6 +305,29 @@ Now connecting with `curl` should fail with a similar error:
     establish a secure connection to it. To learn more about this situation and
     how to fix it, please visit the web page mentioned above.
 
+## Connecting QField to a local QFieldCloud instance
+
+In order to connect the QField mobile app to your local QFieldCloud development instance, the following prerequisites must be met:
```
suricactus:
```suggestion

QFieldCloud REST API is used across clients, including the QField app.
One can test their local QFieldCloud instance by running QField on their development machine.
[Install QField](https://qfield.org/get) for your operating system.

In order to connect QField from a device different from your development machine (e.g. your mobile phone), the following prerequisites must be met:
```

Not sure what was the use case, but 95% of the time you better test the APIs using the desktop client, because you also have (easy) access to the configuration files. I guess it needs a bit of reordering, so in the injected paragraph above we can point to the bullet with instructions - start, type ip etc etc.

suricactus (commented):
Good stuff here, I guess for less GIS people is helpful.

A few comments:
 - Don't forget a symbol a the end of the line otherwise the letters will fall off.
 - Make it clear that most of the cases the easier (I would even use the preferable) way for testing the majority of cases is the desktop app. (I guess SSO and integrated positioning are the exclusions).
 - If you use ordered bullet list as it is always easier to read and ask questions later.
 - Try to keep one sentence per line when markdown allows it. Helps with git history later. 

Side note not directly related to this PR: the README is getting massive, we either need to reorganize it, or we split it. No need to do anything for now.

suricactus:
@lukasgraf I think this PR needs a final state at some point, I think it's around time to get it there.

suricactus:
@lukasgraf are you ok if we move this out of WIP and merge as it is, then we can improve the doc?

README.md
lukasgraf: These instructions were intended to document how to connect the QField app running on an actual mobile device.

For me, this was relevant when developing SSO related features, where the actual platform specifics matter a lot. For example to see and test how the app opens the web browser when kicking off the OAuth flow (in-application browser, or does it use the default platform browser). Or, when testing support for client certificates / mutual TLS, to check how the browser that is used for the OAuth flow accesses the device specific certificate trust store, whether it has the needed permissions to access that trust store, etc...

lukasgraf:
@suricactus I rebased and addressed the review feedback. IMHO this can be merged like this.

PR #1455, opened 2025-12-29
Introduce flake8-if-expr to forbid inline ifs

suricactus:
We always had trouble to explain what is the tolerance of using inline `if`s.

Now with this flake8 rule, there is no longer arguemntation if something
subjectively passes.

I am quite happy with the readability improvement. As someone that
switches languages all the time, this is a huge leap forward.


gounux:
Okay, strict approach considered here, will have to say goodbye to the functional JVM languages part of me :)

I'm okay with that though,

Relaunching CI tests since lot of them are failing.

gounux (approved):
LGTM 👌

PR #1456, opened 2025-12-29
feat: add `ModelAdminSearchParserMixin` to enhance code quality

Rakanhf:
Prevously when we want to use `search_parser` function we had to override `get_search_results` everytime, which makes the code longer and redundant.

Now we can simply use `ModelAdminSearchParserMixin` and `search_parser_config` to enable this feature without the need to override functions over and over again

- Add an admin mixin class `ModelAdminSearchParserMixin` to make the use of `search_parser` easier and cleaner.

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -155,7 +155,9 @@ class ModelAdminSearchParserMixin:
 
     search_parser_config: dict[str, dict[str, Any]] | None = None
 
-    def get_search_results(self, request, queryset, search_term):
+    def get_search_results(
+        self, request: HttpRequest, queryset: QuerySet, search_term: str
+    ) -> tuple[QuerySet, bool]:
         if self.search_parser_config:
             filters = search_parser(
                 request,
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/admin.py
```diff
@@ -148,8 +148,33 @@ class ModelAdminEstimateCountMixin:
...
 
 
+class ModelAdminSearchParserMixin:
+    """
+    Mixin to add search parser to the model admin.
+    """
+
+    search_parser_config: dict[str, dict[str, Any]] | None = None
+
+    def get_search_results(self, request, queryset, search_term):
```
suricactus: Typings

docker-app/qfieldcloud/core/admin.py
```diff
@@ -148,8 +148,33 @@ class ModelAdminEstimateCountMixin:
...
+        if self.search_parser_config:
+            filters = search_parser(
+                request,
+                queryset,
+                search_term,
+                self.search_parser_config,
+            )
+
+            if filters:
+                return queryset.filter(**filters), True
```
suricactus, line 172: please verify if it is correct to:

1) use the queryset without first applying the filters from the super method
2) always return true

I see this is how it used to work before, but I don't remember the motivation. Probably a small comment why is like this was/is missing.

suricactus (approved):
LGTM, don't forget typings on new code.

docker-app/qfieldcloud/core/admin.py
suricactus: A small comment before resolving the thread is always appreciated.

suricactus (approved):

PR #1458, opened 2026-01-06
test: add test for setting session role through `QFC_PG_EFFECTIVE_USER` Secret

gounux:
CF the title, this feature well deserves a CI test :)

The test considers a Project`s `project_details` JSON, as well as a Package Job's `feedback` output JSON.

Also, a simple QGIS `.qgs` project file is added, that contains an OSM XYZ layer + a layer with the SQL view automatically created during the test.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_pg_roles.py
+++ b/docker-app/qfieldcloud/core/tests/test_pg_roles.py
@@ -1,0 +1,174 @@
+import io
+
+from rest_framework.test import APITransactionTestCase
+
+from qfieldcloud.authentication.models import AuthToken
+from qfieldcloud.core.models import (
+    Organization,
+    OrganizationMember,
+    Person,
+    Project,
+    ProjectCollaborator,
+    Secret,
+)
+from qfieldcloud.core.tests.mixins import QfcFilesTestCaseMixin
+from qfieldcloud.core.tests.utils import (
+    get_test_postgis_connection,
+    set_subscription,
+    setup_subscription_plans,
+    testdata_path,
+    wait_for_project_ok_status,
+)
+from qfieldcloud.core.utils2.jobs import repackage
+
+
+class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
+    def setUp(self):
+        setup_subscription_plans()
+
+        self.u1 = Person.objects.create(username="u1", password="u1")
+        self.t1 = AuthToken.objects.get_or_create(user=self.u1)[0]
+
+        self.u2 = Person.objects.create(username="u2", password="u2")
+        self.t2 = AuthToken.objects.get_or_create(user=self.u2)[0]
+
+        self.o1 = Organization.objects.create(username="o1", organization_owner=self.u1)
+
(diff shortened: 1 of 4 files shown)


docker-app/qfieldcloud/core/tests/test_pg_roles.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self.p1.direct_collaborators.bulk_create(collaborators)
+
+        self.conn = get_test_postgis_connection()
+
+    def test_pg_effective_user_override(self):
+        # create database table and view.
+        cur = self.conn.cursor()
+        cur.execute(
+            "CREATE TABLE IF NOT EXISTS point (id integer primary key, geometry geometry(point, 4326))"
```
suricactus: Why if not exists? Should be just create table and fail of the db somehow is there. If the table was created beforehand, it should be cleaned. Same for all the occurrences kof the if not exists

docker-app/qfieldcloud/core/tests/test_pg_roles.py
```diff
@@ -0,0 +1,178 @@
...
+      CREATE ROLE that_user_that_overrides LOGIN PASSWORD 'the_password_of_that_user_that_overrides';
+   END IF;
+END
+$$;""")
+        cur.execute("GRANT SELECT ON point_view TO that_user_that_overrides")
+        self.conn.commit()
+
+        # create project-level pg_service secret.
+        Secret.objects.create(
+            name="PG_SERVICE_GEODB",
```
suricactus: Please avoid geodb as a name, it is have bad memories with the name. It had a real meaning soon enough to reuse it.

docker-app/qfieldcloud/core/tests/test_pg_roles.py
gounux: That was mainly to facilitate the local run of several following tests.

Do you mean a `cur.execute("DROP TABLE point")` at the end of the function ?

docker-app/qfieldcloud/core/tests/test_pg_roles.py
gounux: Thanks for spotting this, indeed, addressed by https://github.com/opengisch/QFieldCloud/pull/1458/commits/ab019270a8c6307af53aadcd1e9f146edac5e6c1.

docker-app/qfieldcloud/core/tests/test_pg_roles.py
gounux: Likely addressed by https://github.com/opengisch/QFieldCloud/pull/1458/commits/21019a470a2b15fc0107c531fd9374a5ad8db3db.

docker-app/qfieldcloud/core/tests/test_pg_roles.py
gounux: BTW, took the liberty to remove such occurrences in https://github.com/opengisch/QFieldCloud/pull/1458/commits/0869a0f10cfb8b30d4d5abf11b3a640f7b638898.

PR #1459, opened 2026-01-06
feat: expand `Person` quick link navigation 

Rakanhf:
Adding a couple of extra quick link navigation To make the navigation for jobs, subscriptions, logs, etc. easier. 

- Add `Job`, `Delta`, `FaultyDelta`, `LogEntry` quick links to `Person` change form
- Add `subscription` quick link to to `Person` change form using `is_frontend_user_editable`

---

Person change form : 

image

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -596,7 +596,13 @@ class PersonAdmin(QFieldCloudModelAdmin):
         obj.clean()
         obj.save()
 
-    def change_view(self, request, object_id, form_url="", extra_context=None):
+    def change_view(
+        self,
+        request: HttpRequest,
+        object_id: str,
+        form_url: str = "",
+        extra_context: Any | None = None,
+    ) -> HttpResponse:
         # Add the subscription model is editable flag to the extra context
         extra_context = extra_context or {}
 


docker-app/qfieldcloud/core/admin.py
```diff
@@ -595,6 +596,20 @@ def save_model(self, request, obj, form, change):
         obj.clean()
         obj.save()
 
+    def change_view(self, request, object_id, form_url="", extra_context=None):
```
gounux: What are the typings here ?

Rakanhf: Thank you for noticing this ! While there are no typings from Django itself [Ref](https://github.com/django/django/blob/stable/4.2.x/django/contrib/admin/options.py#L1888)

But we can use django-stubs typings.

docker-app/qfieldcloud/core/admin.py
```diff
@@ -595,6 +596,26 @@ def save_model(self, request, obj, form, change):
         obj.clean()
         obj.save()
 
+    def change_view(
+        self,
+        request: HttpRequest,
+        object_id: str,
+        form_url: str = "",
+        extra_context: Any | None = None,
+    ):
```
suricactus: return typing is missing.

suricactus (approved):

PR #1461, opened 2026-01-12
feat: add simple object storage cleaner script

Rakanhf:
Currently QFieldCloud keeps all project files using the non-legacy storage even if they are deleted by the user. This is due to enabled versioning on Exoscale and the lack of Exoscale mechanism to delete files after certain amount of time after they are deleted. This causes excessive file storage and therefore costs.

- Add `purge_deleted_objects.py` for scanning and deleting logically deleted objects in S3 compatable storages
- Add `test.py` for testing the script
- Add a `README.MD`

---
A small CLI to **analyze and clean logically deleted objects** in **versioned S3-compatible buckets**.

This script is **provider-agnostic** works with any S3-compatible service that supports versioning such as `MinIO`, etc.

It helps you to the following:
1. Detect whether a bucket has versioning configured.
2. Find keys whose **latest version is a delete marker** (logically deleted).
3. Report how much storage their historical versions consume.
4. **Deleting all versions** for those keys to free space.

#### Requirements
- boto3
- mypy-boto3-s3  _Optional for type hinting_

##### 1. Scanning for logically deleted objects
```
python purge_deleted_objects.py my-versioned-bucket-name --retention-period <time-interval> --dry-run
```
##### 2. Deleting logically deleted objects
```
python purge_deleted_objects.py my-versioned-bucket-name --retention-period <time-interval>
```

### Options

| Option | Description |
|--------|-------------|
| `bucket_name` | The name of the target S3 bucket (Required). |
| `--dry-run` | Only scan and report statistics without deleting. |
| `--force` | Skip confirmation prompt (useful for automation). |
| `--prefix <path>` | Limit scan to a specific prefix (folder). |
| `--retention-period <time-interval>` | Only process objects deleted BEFORE this duration (e.g., `7 days`, `30 minutes`, `1 hour`). Recent deletions are preserved. |
| `--log-level` | Log level (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). |

### DIFF
--- a/scripts/object_storage_cleaner/README.MD
+++ b/scripts/object_storage_cleaner/README.MD
@@ -1,0 +1,119 @@
+# Object Storage Cleaner
+
+A utility script to identify and clean "logically deleted" objects in S3-compatible object storage (e.g., AWS S3, MinIO) when bucket versioning is enabled.
+
+## The Problem
+
+When **Versioning** is enabled on an S3 bucket, deleting an object does not actually remove the data. Instead, S3 creates a **Delete Marker** as the latest version. The older versions (the actual data) remain in the bucket and continue to incur storage costs.
+
+A "logically deleted object" is an object where the current (latest) version is a Delete Marker. This means the object appears deleted to applications, but all its historical versions are still stored.
+
+This script helps you:
+1.  **Scan**: Find all objects that are logically deleted and calculate how much space they are wasting.
+2.  **Permanently Delete Versions**: Permanently delete all versions of these objects (including the delete marker) to free up storage.
+
+> **Note**: This script specifically targets objects that are *currently deleted*. It does **not** touch objects that are still active (i.e., the latest version is a file, not a delete marker), even if they have old versions.
+
+## Requirements
+
+- Python 3.8+
+- `boto3` library
+
+1. Create a virtual environment
+
+2. Install dependencies:
+
+```bash
+pip install boto3
+
+# Optional
+pip install mypy-boto3-s3 # for type hinting
+```
+
+## Usage
+
+Run the script directly from the command line.
+
+### Basic Syntax
+
+```bash
+python purge_deleted_objects.py <bucket_name> --retention-period "30 days" [options]
+```
+
+### Options
+
+Run the script with `--help` to see all available options.
+
+---
+By default the script uses the `default` S3 profile (default `boto3` behavior)
+check [Configuration and credential file settings in the AWS CLI](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html)
+
+### Examples
+
+**1. Scan a bucket to see how much space can be freed:**
+
+```bash
+python purge_deleted_objects.py my-production-bucket --retention-period "30 days" --dry-run
+```
+
+**2. Scan a specific folder:**
+
+```bash
+python purge_deleted_objects.py my-production-bucket --retention-period "30 days" --dry-run --prefix "logs/2023/"
+```
+
+**3. Cleanup objects deleted more than 30 days ago (Retention Policy):**
+
+- `--retention-period '30 days'`: Only selects objects that were deleted **more than 30 days ago**. Objects deleted recently (within the last 30 days) are preserved to allow for restoration.
+
+```bash
+# Clean objects deleted OLDER than 30 days
+python purge_deleted_objects.py my-bucket --retention-period '30 days'
+```
+
+**4. Permanently delete objects:**
+
+```bash
+python purge_deleted_objects.py my-bucket --retention-period "30 days"
+# You will be prompted to confirm.
+```
+
+**5. Automated cleanup for MinIO (local/custom endpoint):**
+
+```bash
+python purge_deleted_objects.py test-bucket --retention-period "30 days" --force
+```
+
+## Running Tests
+
+The repository includes a test suite (`test.py`) that uses `unittest` and requires a running S3-compatible service (like MinIO).
+
+### Test Requirements
+- `boto3`
+- S3 service (e.g., MinIO) running.
+
+### Steps to Run Tests
+
+1.  **Start MinIO** (or ensure a mock S3 is running).
+    Example using Docker:
+    ```bash
+    docker run -d -p 8009:9000 --name minio-test \
+      -e "MINIO_ACCESS_KEY=access_key_xyz" \
+      -e "MINIO_SECRET_KEY=secret_key_xyz" \
+      minio/minio server /data
+    ```
+
+2.  **Configure Environment Variables**
+    Export the following variables to configure `boto3` to talk to your local MinIO instance (matching the credentials above):
+
+    ```bash
+    export AWS_ACCESS_KEY_ID=access_key_xyz
+    export AWS_SECRET_ACCESS_KEY=secret_key_xyz
+    export AWS_ENDPOINT_URL=http://localhost:8009
+    export AWS_DEFAULT_REGION=us-east-1
+    ```
+
+3.  **Run Tests**
+    ```bash
+    python test.py
+    ```
--- a/scripts/object_storage_cleaner/purge_deleted_objects.py
+++ b/scripts/object_storage_cleaner/purge_deleted_objects.py
@@ -1,0 +1,519 @@
+"""
+Object Storage Logically Deleted Object Cleaner
+
+This script scans an object storage bucket for "logically deleted" objects (objects where the
+latest version is a Delete Marker) and optionally permanently deletes all versions
+of those objects to reclaim storage space.
+
+Usage:
+    python purge_deleted_objects.py <bucket> --retention-period "30 days" [options]
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import re
+import signal
+import sys
+from collections.abc import Iterable, Iterator
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from typing import TYPE_CHECKING, Any, Literal
+
+import boto3
+from botocore.client import BaseClient
+from botocore.exceptions import ClientError
+
+logger = logging.getLogger(__name__)
+
+# AWS S3 limit for delete objects batch up to 1000 keys (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_objects.html)
+DELETE_BATCH_SIZE = 1000
+
+# AWS S3 limit for list object versions max keys up to 1000 (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_object_versions.html)
+LIST_OBJECT_VERSIONS_MAX_KEYS = 1000
+
+if TYPE_CHECKING:
+    from mypy_boto3_s3.type_defs import (
+        DeleteObjectsOutputTypeDef,
+        ListObjectVersionsOutputTypeDef,
+        ObjectVersionTypeDef,
+    )
+
+    class QfcObjectVersionTypeDef(ObjectVersionTypeDef):
+        IsDeleteMarker: Literal[False]
+
+    class QfcDeleteMarkerEntryTypeDef(ObjectVersionTypeDef):
+        IsDeleteMarker: Literal[True]
+
+    ObjectVersionOrDeleteMarker = QfcObjectVersionTypeDef | QfcDeleteMarkerEntryTypeDef
+
+
+@dataclass
+class LogicallyDeletedObject:
+    key: str
+    deleted_at: datetime
+    total_size_bytes: int
+    versions_count: int
+    versions: list[dict[str, Any]]
+
+
+def format_bytes(num_bytes: int) -> str:
+    """Convert a byte count to a human-readable string."""
+    units = ["B", "KB", "MB", "GB", "TB", "PB"]
+
+    if num_bytes == 0:
+        return "0 B"
+
+    size = float(num_bytes)
+    for unit in units:
+        if size < 1024.0 or unit == units[-1]:
+            return f"{size:.2f} {unit}"
+
+        size /= 1024.0
+
+    return f"{size:.2f} PB"
+
+
+def iter_all_versions(
+    s3_client: BaseClient, bucket: str, prefix: str | None = None
+) -> Iterator[ObjectVersionOrDeleteMarker]:
+    """
+    Iterate over all object versions and delete markers from object storage, paginated.
+
+    Object storage's `list_object_versions` returns `Versions` and `DeleteMarkers` as separate
+    lists within each page. This function merges and sorts them by Key locally
+    to ensure that all history (versions + markers) for a specific Key is yielded
+    contiguously before the iterator moves to the next Key.
+
+    Args:
+        s3_client: The configured boto3 object storage client.
+        bucket: The bucket name.
+        prefix: Optional key prefix to limit the scan.
+
+    Yields:
+        ObjectVersionOrDeleteMarker: A dictionary representing either a Version or a Delete Marker.
+    """
+    paginator = s3_client.get_paginator("list_object_versions")
+    paginate_kwargs = {
+        "Bucket": bucket,
+        "MaxKeys": LIST_OBJECT_VERSIONS_MAX_KEYS,
+    }
+    if prefix:
+        paginate_kwargs["Prefix"] = prefix
+
+    for page in paginator.paginate(**paginate_kwargs):
+        # Type from mypy_boto3_s3.type_defs
+        page: ListObjectVersionsOutputTypeDef
+        versions: list[ObjectVersionOrDeleteMarker] = []
+
+        # Yield Versions
+        for version in page.get("Versions", []):
+            version: QfcObjectVersionTypeDef
+            version["IsDeleteMarker"] = False
+            versions.append(version)
+
+        # Yield Delete Markers
+        for delete_marker in page.get("DeleteMarkers", []):
+            delete_marker: QfcDeleteMarkerEntryTypeDef
+            delete_marker["IsDeleteMarker"] = True
+            versions.append(delete_marker)
+
+        # Sort by Key to keep versions of the same object together
+        versions.sort(key=lambda v: (v["Key"], v["LastModified"]), reverse=True)
+
+        yield from versions
+
+
+def analyze_key_history(
+    key: str,
+    versions: list[ObjectVersionOrDeleteMarker],
+    retention_cutoff_ts: datetime | None,
+) -> LogicallyDeletedObject | None:
+    """
+    Analyze a list of versions for a single key to see if it's logically deleted
+    and return a LogicallyDeletedObject if it is, otherwise None.
+
+    Args:
+        key: The key of the object.
+        versions: The list of versions.
+        retention_cutoff_ts: Optional cutoff timestamp. If provided, the object is only returned if it was deleted *before* this timestamp. Recent deletions are ignored.
+
+    Returns:
+        A LogicallyDeletedObject if the key is logically deleted, None otherwise.
+    """
+    if not versions:
+        return None
+
+    latest_version = versions[0]
+
+    # Criteria: Latest version must be a delete marker
+    if not latest_version["IsDeleteMarker"]:
+        return None
+
+    # Filter by time
+    # If the delete marker is NEWER than the retention cutoff, skip it (retain it)
+    if retention_cutoff_ts and latest_version["LastModified"] > retention_cutoff_ts:
+        return None
+
+    # Calculate stats efficiently
+    # Note: latest is size 0 (delete marker), so we can just sum all
+    total_size = 0
+
+    for v in versions:
+        if v["IsDeleteMarker"]:
+            # Delete markers have no size
+            continue
+        # Versions have a size
+        total_size += v["Size"]
+
+    return LogicallyDeletedObject(
+        key=key,
+        deleted_at=latest_version["LastModified"],
+        total_size_bytes=total_size,
+        versions_count=len(versions),
+        versions=versions,
+    )
+
+
+def delete_versions_batch(
+    s3_client: BaseClient, bucket: str, versions: list[ObjectVersionOrDeleteMarker]
+) -> None:
+    """
+    Delete a batch of versions, up to 1000 versions per batch.
+
+    Args:
+        s3_client: The configured S3 client.
+        bucket: The bucket name.
+        versions: The list of versions to delete.
+    """
+    if not versions:
+        return
+
+    assert len(versions) <= DELETE_BATCH_SIZE
+
+    # Transform to boto3 format
+    objects: list[ObjectVersionOrDeleteMarker] = []
+    for v in versions:
+        objects.append(
+            {
+                "Key": v["Key"],
+                "VersionId": v["VersionId"],
+            }
+        )
+
+    try:
+        # Quiet=False returns success details and errors for logging
+        response: DeleteObjectsOutputTypeDef = s3_client.delete_objects(
+            Bucket=bucket,
+            Delete={
+                "Objects": objects,
+                "Quiet": False,
+            },
+        )
+
+        # Log successes
+        for success in response.get("Deleted", []):
+            logger.info(
+                f"Permanently deleted: {success['Key']} (VersionId: {success.get('VersionId')})"
+            )
+
+        # Log errors
+        for error in response.get("Errors", []):
+            logger.error(
+                f"Failed to delete: {error['Key']} (VersionId: {error.get('VersionId')}) | "
+                f"Code: {error.get('Code')} | Message: {error.get('Message')}"
+            )
+
+    except ClientError as err:
+        logger.error(f"Failed to delete versions batch: {err}")
+
+
+def iter_logically_deleted(
+    version_iterator: Iterator[ObjectVersionOrDeleteMarker],
+    retention_cutoff_ts: datetime | None = None,
+) -> Iterator[LogicallyDeletedObject]:
+    """
+    Consumes an iterator of all object versions and yields ONLY the objects that are logically deleted.
+
+    This function relies on the input iterator being pre-sorted by `key`. It groups all versions
+    belonging to the same key into a list, and then passes that complete history to
+    `analyze_key_history` to determine if the object is currently in a deleted state.
+
+    Args:
+        version_iterator: An iterator of version dictionaries, MUST be sorted by `key`.
+        retention_cutoff_ts: Optional cutoff timestamp. If provided, the object is only returned if it was deleted *before* this timestamp. Recent deletions are ignored.
+
+    Yields:
+        LogicallyDeletedObject: A summary object for every key that is currently deleted.
+    """
+    # We maintain a buffer of versions for the "current" key we are processing.
+    # Since the input iterator is sorted, all versions for "Key A" will arrive sequentially
+    # before any version for "Key B" appears, so we can group them together.
+    current_key: str | None = None
+    current_versions: list[ObjectVersionOrDeleteMarker] = []
+
+    for version in version_iterator:
+        # Case 1: First item in the entire iterator
+        if current_key is None:
+            current_key = version["Key"]
+            current_versions.append(version)
+            continue
+
+        # Case 2: Same key as before -> Add to buffer
+        if version["Key"] == current_key:
+            current_versions.append(version)
+
+        # Case 3: New key detected -> Process the previous buffer
+        else:
+            # We have collected all versions for 'current_key'. Now analyze it.
+            result = analyze_key_history(
+                current_key, current_versions, retention_cutoff_ts
+            )
+            if result:
+                yield result
+
+            # Reset buffer for the new key
+            current_key = version["Key"]
+            current_versions = [version]
+
+    # Case 4: End of iterator -> Process the final buffer remaining in memory
+    if current_key and current_versions:
+        result = analyze_key_history(current_key, current_versions, retention_cutoff_ts)
+        if result:
+            yield result
+
+
+def action_scan(objects_iterator: Iterable[LogicallyDeletedObject]) -> None:
+    """
+    Consume an iterator of logically deleted objects and log the details.
+
+    Args:
+        objects_iterator: The iterator of logically deleted objects.
+    """
+    logger.info("Scanning for logically deleted objects...")
+
+    # Counters for summary logging
+    key_count = 0
+    total_size_bytes = 0
+    total_versions_count = 0
+
+    for obj in objects_iterator:
+        # Update counters
+        key_count += 1
+        total_size_bytes += obj.total_size_bytes
+        total_versions_count += obj.versions_count
+
+        # Log details for each key
+        logger.info(
+            f"Found: {obj.key} | Deleted At: {obj.deleted_at} | "
+            f"Wasted Size: {format_bytes(obj.total_size_bytes)} | {obj.versions_count} versions"
+        )
+
+    # Log summary
+    logger.info(f"Total deleted keys found: {key_count}")
+    logger.info(f"Total size wasted: {format_bytes(total_size_bytes)}")
+    logger.info(f"Total versions wasted: {total_versions_count}")
+
+
+def action_permanently_delete_versions(
+    s3_client: BaseClient,
+    bucket: str,
+    objects_iterator: Iterable[LogicallyDeletedObject],
+) -> None:
+    """
+    Consume an iterator of logically deleted objects and delete the versions in batches.
+
+    Args:
+        s3_client: The configured S3 client.
+        bucket: The bucket name.
+        objects_iterator: The iterator of logically deleted objects.
+    """
+    batch: list[ObjectVersionOrDeleteMarker] = []
+
+    for obj in objects_iterator:
+        for version in obj.versions:
+            batch.append(version)
+
+            # Flush batch if full
+            if len(batch) == DELETE_BATCH_SIZE:
+                delete_versions_batch(s3_client, bucket, batch)
+                batch.clear()
+
+    # Flush remaining
+    if batch:
+        delete_versions_batch(s3_client, bucket, batch)
+
+
+def get_s3_client(bucket: str, profile: str | None) -> BaseClient:
+    """
+    Create an object storage client and validate that the bucket has versioning enabled.
+
+    Args:
+        bucket: The name of the object storage bucket.
+        profile: Optional AWS profile name.
+
+    Returns:
+        A configured boto3 object storage client.
+    """
+
+    session_kwargs = {}
+    if profile:
+        session_kwargs["profile_name"] = profile
+
+    client = boto3.Session(**session_kwargs).client("s3")
+
+    # Validate bucket versioning is enabled
+    try:
+        response = client.get_bucket_versioning(Bucket=bucket)
+        status = response.get("Status")
+        if status != "Enabled":
+            raise RuntimeError(
+                f"Bucket '{bucket}' versioning is not Enabled (Status: {status})."
+            )
+    except ClientError as err:
+        raise RuntimeError(
+            f"Could not check versioning status for '{bucket}': {err}"
+        ) from err
+
+    return client
+
+
+def parse_retention_period(value: str) -> datetime:
+    """
+    Parse a duration string (e.g., '3 seconds', '30 days', '2 weeks') and return a datetime object.
+
+    Args:
+        value: The duration string to parse. Supports `seconds`, `minutes`, `hours`, `days`, and `weeks`.
+
+    Returns:
+        A UTC datetime equal to now minus the parsed duration.
+    """
+    # Handle None and non-string inputs
+    if value is None:
+        raise argparse.ArgumentTypeError("Duration cannot be None")
+
+    if not isinstance(value, str):
+        raise argparse.ArgumentTypeError(
+            f"Duration must be a string, got {type(value).__name__}"
+        )
+
+    value = value.strip().lower()
+
+    # Handle empty string after stripping
+    if not value:
+        raise argparse.ArgumentTypeError("Duration cannot be empty")
+
+    pattern = (
+        r"^(\d+)\s*(second|seconds|minute|minutes|hour|hours|day|days|week|weeks)$"
+    )
+    match = re.match(pattern, value)
+    if not match:
+        raise argparse.ArgumentTypeError(
+            "Invalid duration. Use: '3 seconds', '10 hours', '15 minutes', '30 days', '2 weeks'"
+        )
+
+    amount = int(match.group(1))
+    unit = match.group(2)
+
+    try:
+        if unit in ("second", "seconds"):
+            delta = timedelta(seconds=amount)
+        elif unit in ("minute", "minutes"):
+            delta = timedelta(minutes=amount)
+        elif unit in ("hour", "hours"):
+            delta = timedelta(hours=amount)
+        elif unit in ("day", "days"):
+            delta = timedelta(days=amount)
+        elif unit in ("week", "weeks"):
+            delta = timedelta(weeks=amount)
+        else:
+            raise argparse.ArgumentTypeError(f"Unsupported time unit: {unit}")
+    except OverflowError:
+        raise argparse.ArgumentTypeError("Duration is too large")
+
+    return datetime.now(timezone.utc) - delta
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Scan and clean logically deleted objects in object storage like S3."
+    )
+    parser.add_argument("bucket", help="Target Object Storage bucket")
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--retention-period",
+        type=parse_retention_period,
+        dest="retention_cutoff_ts",
+        required=True,
+        help="Only process objects deleted BEFORE this time interval (e.g. '30 days'). Recent deletions are preserved.",
+    )
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Scan for logically deleted objects without deleting them",
+    )
+    parser.add_argument(
+        "--force",
+        action="store_true",
+        help="Skip permanently delete confirmation prompt",
+    )
+    parser.add_argument(
+        "--log-level",
+        default="INFO",
+        help="Log level",
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+    )
+
+    args = parser.parse_args()
+
+    logging.basicConfig(
+        level=args.log_level,
+        format="%(asctime)s - %(levelname)s - %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        stream=sys.stdout,
+    )
+
+    try:
+        # 1. Setup Connection
+        client = get_s3_client(args.bucket, args.profile)
+
+        # 2. Build Pipeline
+        raw_iterator = iter_all_versions(client, args.bucket, args.prefix)
+
+        # Transform
+        clean_iterator = iter_logically_deleted(raw_iterator, args.retention_cutoff_ts)
+
+        # 3. Execute
+        if args.dry_run:
+            action_scan(clean_iterator)
+            return 0
+
+        else:
+            if not args.force:
+                confirmation_input = input(
+                    f"Permanently delete from '{args.bucket}'? (yes/no): "
+                ).lower()
+
+                if confirmation_input != "yes":
+                    return 0
+
+            action_permanently_delete_versions(client, args.bucket, clean_iterator)
+
+    except Exception as e:
+        logger.error(f"Error: {e}")
+
+        raise e
+
+    return 0
+
+
+def handle_sigint(sig: int, frame: object) -> None:
+    logger.info("\n\nScan interrupted by user.")
+    sys.exit(0)
+
+
+if __name__ == "__main__":
+    signal.signal(signal.SIGINT, handle_sigint)
+    raise SystemExit(main())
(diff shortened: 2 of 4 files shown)


scripts/object_storage_cleaner/README.MD
```diff
@@ -0,0 +1,117 @@
...
+## Requirements
+
+- Python 3.8+
+- `boto3` library
+
+Install dependencies:
+
+```bash
+pip install boto3
+```
```
gounux: Should we recommend here to use a virtual environment ?

Rakanhf: This is a great point, thank you for noticing that ! I will also add `mypy-boto3-s3` which is optional for type hinting

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+    Args:
+        s3_client: The configured boto3 object storage client.
+        bucket: The bucket name.
+        prefix: Optional key prefix to limit the scan.
+
+    Yields:
+        VersionRef: A unified object representing either a Version or a Delete Marker.
+    """
+    paginator = s3_client.get_paginator("list_object_versions")
+    kwargs = {
```
suricactus:
```suggestion
    paginate_kwargs = {
```

It can be kwargs for many things, avoid such names unless used for function arguments.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+    }
+    if prefix:
+        kwargs["Prefix"] = prefix
+
+    for page in paginator.paginate(**kwargs):
+        items: list[VersionRef] = []
+
+        # Yield Versions
+        for version in page.get("Versions", []):
+            version: ObjectVersionTypeDef  # Type from mypy_boto3_s3.type_defs
```
suricactus: Does this work if mypy_boto3_... is not installed? Asking out of curiosity.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+                f"Failed to delete: {error['Key']} (VersionId: {error.get('VersionId')}) | "
+                f"Code: {error.get('Code')} | Message: {error.get('Message')}"
+            )
+
+    except ClientError as e:
+        logger.error(f"Failed to delete versions batch: {e}")
+
+
+def iter_logically_deleted(
+    versions_stream: Iterator[VersionRef],
```
suricactus:
```suggestion
    version_chunks: Iterator[VersionRef],
```

It is not a stream in the classical sense, it's just iterable chunks of 1000 slices of objects, no?

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,501 @@
...
+    # Case 4: End of stream -> Process the final buffer remaining in memory
+    if current_key and current_versions:
+        result = analyze_key_history(
+            current_key, current_versions, deleted_after_threshold
+        )
+        if result:
+            yield result
+
+
+def action_scan(objects_stream: Iterable[LogicallyDeletedObject]) -> None:
```
suricactus: I failed to understand the function name. I guess we can name it better. A word like "summary" or "report" can help.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,501 @@
...
+    )
+    parser.add_argument("bucket", help="Target Object Storage bucket")
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--deleted-within",
+        type=parse_deleted_after_threshold,
+        dest="deleted_after_threshold",
+        help="Only process objects deleted in the last DURATION (e.g. '7 days'). Older deletions are IGNORED.",
+    )
```
suricactus, line 451: Hm, I think we need exactly the opposite of this, no?

We only want to scan and delete the deleted versions that are NOT within the last 30 days. This way we have the last 30 days deleted versions and we can restore them in case things go south.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--deleted-within",
+        type=parse_deleted_after_threshold,
+        dest="deleted_after_threshold",
+        help="Only process objects deleted in the last DURATION (e.g. '7 days'). Older deletions are IGNORED.",
+    )
+    parser.add_argument(
+        "--scan", action="store_true", help="Scan for logically deleted objects"
```
suricactus:
```suggestion
        "--print-report-only", action="store_true", help="Print a report for the logically deleted objects"
```

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,501 @@
...
+        # 2. Build Pipeline
+        raw_stream = iter_all_versions(client, args.bucket, args.prefix)
+
+        # Transform
+        clean_stream = iter_logically_deleted(raw_stream, args.deleted_after_threshold)
+
+        # 3. Execute
+        if args.scan:
+            action_scan(clean_stream)
+
```
suricactus:
```suggestion
            action_scan(clean_stream)
            return 0

```

explicit return/exit here would not hurt anyone

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+            action_permanently_delete_versions(client, args.bucket, clean_stream)
+
+    except Exception as e:
+        logger.error(f"Error: {e}")
+        if args.log_level == "DEBUG":
+            import traceback
+
+            traceback.print_exc()
+
+        return 1
```
suricactus:
```suggestion
    except Exception as err:
        logger.error(f"Error: {e}")
        
        raise
```

1) avoid single char variable names, especially with `e`.
2) just raise, no need to outsmart python, it will print everything we need.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+        level=args.log_level,
+        format="%(asctime)s - %(levelname)s - %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        stream=sys.stdout,
+    )
+    # Suppress noisy logging
+    logging.getLogger("boto3").setLevel(logging.CRITICAL)
+    logging.getLogger("botocore").setLevel(logging.CRITICAL)
+    logging.getLogger("s3transfer").setLevel(logging.CRITICAL)
+    logging.getLogger("urllib3").setLevel(logging.CRITICAL)
```
suricactus: Not convinced about this one. I would rather keep it OR if it is real PITA, add a --surpress-other-logs options (I am -0.5 on adding such option).

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+        and latest_version.last_modified < deleted_after_threshold
+    ):
+        return None
+
+    # Calculate stats efficiently
+    # Note: latest is size 0 (delete marker), so we can just sum all
+    total_size = 0
+
+    for v in versions:
+        total_size += v.size
```
suricactus: This line here and line 142 are the only places where `VersionRef` is used. IMO adding that abstraction is causing more damage than solving a real world problem. If you are under stress, you ask yourself if the key attribute should be `key` or `Key`, or `last_modified` or `LastModified` or `lastModified`.  Unless you have a very strong opinion why we should keep it, I would request removing that level of abstraction and rewriting it in simpler terms.

However, the `LogicallyDeletedObject` abstraction is just the right sweet spot, so you should keep it.

Assuming `instanceof(v, ObjectVersionTypeDef)`:
```suggestion
        # the delete markers do not have a `Size` attribute, assume their size to be 0 bytes.
        total_size += getattr(v, 'Size', 0)
```

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,501 @@
...
+                    key=delete_marker["Key"],
+                    version_id=delete_marker["VersionId"],
+                    is_delete_marker=True,
+                    size=0,
+                    last_modified=delete_marker["LastModified"],
+                )
+            )
+
+        # Sort by Key to keep versions of the same object together
+        items.sort(key=attrgetter("key"))
```
suricactus: Why don't we sort by key AND `last_modified` here? This way we don't need to sort later for a second time om line 138.


```suggestion
        items.sort(lambda v: (v["key"], v["last_modifed]))
```

Also the name `items` is probably better replaced with `versions`?

suricactus (commented):
Did a quick round and this is very close to the final result.

The only major thing is that the date anchor is inverted, the rest are style, readability and maintainability comments.

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: Yes that's why I use 
1. `from __future__ import annotations` --> Prevents annotations from being evaluated at runtime (avoiding `NameError`)


2. ```python
      from typing import TYPE_CHECKING
      if TYPE_CHECKING:
          ......
    ```

Prevents importing the type stub package at runtime

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: IMO I lean towards `scan` because it describes the active process of traversing the S3 bucket versions.

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: same comment here also 
> IMO I lean towards `scan` because it describes the active process of traversing the S3 bucket versions.

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: It it indeed PITA, when having the log-level to `DEBUG`, especially `botocore`. If less code is better then we can remove all lines and just keep :
```python
logging.getLogger("botocore").setLevel(logging.CRITICAL)
```

IMO, I think we should keep it like this for now, then when we feel the need to enable them / stop Suppressing them, we can enable them back.

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: Yes, I also got the same feeling when I was developing it. I will adjust this to be `retention-period` where it looks for logically deleted object after this period. 

e.g : 
```bach 
python object_storage_cleaner.py my-bucket --permanently-delete-versions --retention-period '30 days'
```

where `--retention-period '30 days'` Only selects objects that were deleted **more than 30 days ago**.

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: I missed the future import. Smart!

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: Trust you here in judgment, I hope we don't regret it in the future. 

Because if all is dumped in the logs, we can still use grep and show only the loggers we need.

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: I am afraid `scan` is very cryptic.  Feel free to discuss within the team about the name, I dont want to be pain myself.

Rakanhf: I had a discussion about it with the team, and we suggested that we change the implementation to have a `--dry-run` instead and remove `--permanently-delete-versions`

---

How it will be implemented

```bash
# should start the delete process
python object_storage_cleaner.py <my-versioned-bucket-name>  
.... open prompt (Yes) -> safety input prompt
```

```bash
# will just scan (print the summery report)
python object_storage_cleaner.py <my-versioned-bucket-name> --dry-run 
```

```bash
# for automation (replacing `--noinput`)
python object_storage_cleaner.py <my-versioned-bucket-name> --force  
```

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: Discussed this with the team, and there suggestion was that we can introduce `--quiet` or `--verbose`. 

while I think it's a good idea. I have another suggestion that might make the code simpler. which is we can remove the suppressing logic that I added above, and change the log level for :

```python
# Log details for each key
logger.debug(
    f"Found: {obj.key} | Deleted At: {obj.deleted_at} | "
    f"Wasted Size: {obj.total_size_bytes} bytes ({obj.versions_count} versions)"
)
```

```python
# Log successes
for success in response.get("Deleted", []):
    logger.debug(
        f"Permanently deleted: {success['Key']} (VersionId: {success.get('VersionId')})"
    )
```

instead of using `debug` here, we can switch it to `info`. the down side of this that if a user used this script interactively they will keep seeing the logs in terminal. personally I don't prefer it, but it's quite similar to something like `docker system prune --all` 

but now if you pipe / redirect the file output we will always get these logs without the need to even do `--log-level DEBUG`

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: I would prefer to keep "permanently delete" option to make it very clear about the consequences of running the script. I leave it up to you.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: Let's comment only hete https://github.com/opengisch/QFieldCloud/pull/1461#discussion_r2686761687

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: Yes, I overlooked those are not info. I like your suggestion or at least what I understood.

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: I believe that anyone would run this script should already know by it's name that it will delete files, and there is an input safety mechanism, that will require the user to type yes after seeing a warning message, unless they specify `--force`, I guess at this point the user should know what they are forcing. 

Maybe renaming the file name itself to a more aggressive name to prevent accidental data loss ? 
e.g `purge_deleted_objects.py` , `destroy_soft_deleted.py`, etc.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+    key: str
+    deleted_at: datetime
+    total_size_bytes: int
+    versions_count: int
+    versions: list[dict[str, Any]]
+
+
+def iter_all_versions(
+    s3_client: BaseClient, bucket: str, prefix: str | None = None
+) -> Iterator[dict[str, Any]]:
```
suricactus:
```suggestion
class QfcObjectVersionTypeDef(ObjectVersionTypeDef):
    IsDeleteMarker: Literal[False]
    
class QfcDeleteMarkerEntryTypeDef(ObjectVersionTypeDef):
    IsDeleteMarker: Literal[True]

ObjectVersionOrDeleteMarker = QfcObjectVersionTypeDef | QfcDeleteMarkerEntryTypeDef

def iter_all_versions(
    s3_client: BaseClient, bucket: str, prefix: str | None = None
) -> Iterator[ObjectVersionOrDeleteMarker]:
```

Very nice idea to just add a new property on the dict. I think we can improve the typing as suggested.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+    if prefix:
+        paginate_kwargs["Prefix"] = prefix
+
+    for page in paginator.paginate(**paginate_kwargs):
+        page: ListObjectVersionsOutputTypeDef  # Type from mypy_boto3_s3.type_defs
+        versions: list[dict[str, Any]] = []
+
+        # Yield Versions
+        for version in page.get("Versions", []):
+            version: ObjectVersionTypeDef  # Type from mypy_boto3_s3.type_defs
```
suricactus: Ideally we always add the comments on the line before. Actually the comment will not be needed if we use `QfcObjectVersionTypeDef`

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+            versions.append(version)
+
+        # Yield Delete Markers
+        for delete_marker in page.get("DeleteMarkers", []):
+            delete_marker: DeleteMarkerEntryTypeDef  # Type from mypy_boto3_s3.type_defs
+            delete_marker["IsDeleteMarker"] = True
+            versions.append(delete_marker)
+
+        # Sort by Key to keep versions of the same object together
+        versions.sort(key=lambda x: (x["Key"], x["LastModified"]), reverse=True)
```
suricactus:
```suggestion
        versions.sort(key=lambda v: (v["Key"], v["LastModified"]), reverse=True)
```

Nothing critical here, but isn't it more logical to name the variable `v` instead of `x`?

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+
+    # Transform to boto3 format
+    objects: list[dict[str, str]] = []
+    for v in versions:
+        objects.append({"Key": v["Key"], "VersionId": v["VersionId"]})
+
+    try:
+        # Quiet=False returns success details and errors for logging
+        response: DeleteObjectsOutputTypeDef = s3_client.delete_objects(
+            Bucket=bucket, Delete={"Objects": objects, "Quiet": False}
```
suricactus:
```suggestion
            Bucket=bucket,
            Delete={
                "Objects": objects,
                "Quiet": False,
            }
```

Don't be shy to add new lines, especially for dangerous lines like this one. :)

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+    # If the delete marker is NEWER than the retention cutoff, skip it (retain it)
+    if retention_cutoff and latest_version["LastModified"] > retention_cutoff:
+        return None
+
+    # Calculate stats efficiently
+    # Note: latest is size 0 (delete marker), so we can just sum all
+    total_size = 0
+
+    for v in versions:
+        total_size += v.get("Size", 0)
```
suricactus: Since you have the `IsDeleteMarker`, you can use:


```suggestion
        if v["IsDeleteMarker"]:
            total_size += 0
        else:
            total_size += v["Size"]
```

I know we did a full circle here, but we didn't have this hard type available. The benefit is that we will get a beautiful KeyError if things go south, instead of silently calculating 0 bytes. The only problem is that typing might be complaining, as `v` could be `QfcDeleteMarkerEntryTypeDef` according to typing, even though we have checked `IsDeleteMarker`. Feel free to find a solution or ignore this comment.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+        bucket: The bucket name.
+        versions: The list of versions to delete.
+    """
+    if not versions:
+        return
+
+    # Transform to boto3 format
+    objects: list[dict[str, str]] = []
+    for v in versions:
+        objects.append({"Key": v["Key"], "VersionId": v["VersionId"]})
```
suricactus:
```suggestion
    objects: list[BetterNameForObjects] = []
    for v in versions:
        objects.append({
            "Key": v["Key"],
            "VersionId": v["VersionId"],
        })
```

New lines and typing.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+            )
+
+        # Log errors
+        for error in response.get("Errors", []):
+            logger.error(
+                f"Failed to delete: {error['Key']} (VersionId: {error.get('VersionId')}) | "
+                f"Code: {error.get('Code')} | Message: {error.get('Message')}"
+            )
+
+    except ClientError as e:
```
suricactus:
```suggestion
    except ClientError as err:
```

`e` for `event`, `element` or something else? Avoid single letter `e` variables anywhere in the codebase.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,453 @@
...
+    )
+    parser.add_argument("bucket", help="Target Object Storage bucket")
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--retention-period",
+        type=parse_retention_period,
+        dest="retention_cutoff",
+        help="Only process objects deleted BEFORE this duration (e.g. '30 days'). Recent deletions are preserved.",
+    )
```
suricactus, line 451: IMO we should make this one a required argument. Keep the `--` in front, just pass `required`.

If you want to replicate the behaviour, just pass `0 seconds` and should be the same? Explicit is always better than implicit. Also will fix a few `datetime | None` types in the codebase.

scripts/object_storage_cleaner/README.MD
```diff
@@ -0,0 +1,125 @@
...
+| Option | Description |
+|--------|-------------|
+| `bucket_name` | The name of the target S3 bucket (Required). |
+| `--dry-run` | Only scan and report statistics without deleting. |
+| `--force` | Skip confirmation prompt (useful for automation). |
+| `--prefix <path>` | Limit scan to a specific prefix (folder). |
+| `--retention-period <duration>` | Only process objects deleted BEFORE this duration (e.g., `7 days`, `30 minutes`, `1 hour`). Recent deletions are preserved. |
+| `--log-level` | Log level (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). |
+---
+By default the script uses the `default` S3 profile (default boto3 behavior)
```
suricactus:
```suggestion
By default the script uses the `default` S3 profile (default `boto3` behavior)
```

scripts/object_storage_cleaner/README.MD
```diff
@@ -0,0 +1,125 @@
...
+
+| Option | Description |
+|--------|-------------|
+| `bucket_name` | The name of the target S3 bucket (Required). |
+| `--dry-run` | Only scan and report statistics without deleting. |
+| `--force` | Skip confirmation prompt (useful for automation). |
+| `--prefix <path>` | Limit scan to a specific prefix (folder). |
+| `--retention-period <duration>` | Only process objects deleted BEFORE this duration (e.g., `7 days`, `30 minutes`, `1 hour`). Recent deletions are preserved. |
+| `--log-level` | Log level (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). |
+---
```
suricactus: Instead of maintaining this, I am totally fine if drop this chapter and you just say "Run the script with `--help` to see all available options". In addition, you may want to move these descriptions to `argparse` definitions. The closer to code, the better.

scripts/object_storage_cleaner/test.py
```diff
@@ -0,0 +1,309 @@
...
+        # 2. Delete it (creates marker)
+        self.delete_file(key)
+
+        # 3. Manually delete the data version, leaving ONLY the marker
+        versions = self.list_versions(key)
+
+        # Find the data version
+        data_version = None
+        for v in versions:
+            if not v.get("IsDeleteMarker", False):
```
suricactus, line 317: How do you get `IsDeleteMarker` set? I checked `list_versions`, it does not set it.

scripts/object_storage_cleaner/test.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # 2. Run scan
+        result = self.run_script(["--dry-run", "--prefix", self.unique_prefix])
+
+        # 3. Verify output
+        self.assertIn("Total keys found: 0", result.stdout)
+        self.assertIn("Total size wasted: 0 bytes", result.stdout)
+        self.assertIn("Total versions wasted: 0", result.stdout)
+
+    def test_scan_restored_files_ignored(self):
```
suricactus: We need to run the same test, but also if the actual deletion spared the version ,similar to `test_permanently_delete_complex_version_history`.

Actually we need this for pretty much all the `--dry-run` commands, so I would suggest to write your tests:

1. conditions
2. dry run
3. expected results for dry run
4. no dry run
5. expected results for no dry run

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,453 @@
...
+    Delete a batch of versions.
+
+    Args:
+        s3_client: The configured S3 client.
+        bucket: The bucket name.
+        versions: The list of versions to delete.
+    """
+    if not versions:
+        return
+
```
suricactus, line 192:
```suggestion

    assert len(versions) <= 1000

```

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,453 @@
...
+    Returns:
+        A LogicallyDeletedObject if the key is logically deleted, None otherwise.
+    """
+    if not versions:
+        return None
+
+    latest_version = versions[0]
+
+    # Criteria: Latest version must be a delete marker
+    if not latest_version.get("IsDeleteMarker", False):
```
suricactus: I guess `IsDeleteMarker` is always set, so we better expect a KeyError rather than silently get a fallback.

Please check your code for silent `KeyError`s, they are actually your friend.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,453 @@
...
+            )
+    except ClientError as e:
+        raise RuntimeError(
+            f"Could not check versioning status for '{bucket}': {e}"
+        ) from e
+
+    return client
+
+
+def parse_retention_period(value: str) -> datetime:
```
suricactus, line 382: Please write excessive unit tests targeting this specific function.

Feel free to get a robotic friend to help you with creative test cases that can potentially break it.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,453 @@
...
+    total_size_bytes: int
+    versions_count: int
+    versions: list[dict[str, Any]]
+
+
+def iter_all_versions(
+    s3_client: BaseClient, bucket: str, prefix: str | None = None
+) -> Iterator[dict[str, Any]]:
+    """
+    Stream all object versions and delete markers from object storage, paginated.
```
suricactus: Stick to one name - stream, chunks, iterator. Now they are mixed throughout the code and comments.

IMO this is not a stream in the classical sense.

scripts/object_storage_cleaner/object_storage_cleaner.py
```diff
@@ -0,0 +1,453 @@
...
+    parser = argparse.ArgumentParser(
+        description="Scan and clean logically deleted objects in object storage like S3."
+    )
+    parser.add_argument("bucket", help="Target Object Storage bucket")
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--retention-period",
+        type=parse_retention_period,
+        dest="retention_cutoff",
```
suricactus:
```suggestion
        "--retention-period",
        type=parse_retention_period,
        dest="retention_cutoff_ts",
```

If you don't mind, the `_ts` suffix will make it clear that the variable contains a timestamp, and no longer a relative peirod.

suricactus (commented):
Finally managed to have a look on tests. They are good, IMO they can become even better, check the comments

Pretty sure this is the last round of review from my side, we are almost there. Once addressed the review, take it as approved from my side.

The biggest requested change is to make the `--retention-period` a required parameter. The rest are typing and "don't shoot my foot" suggestion.

Since this is a critical piece of software, please request a review from some of the other maintainers of QFieldCloud.

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: `purge_deleted_objects` or `cleanup_deleted_objects` sounds scarier than the innocent `object_storage_cleaner` :)

toebivankenoebi: I like `purge_deleted_objects` - very clear. The rest we discussed, like the `--dry-run` and `--force`

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: I don't think that this is necessary, since IMO I believe it should be implicit. I think that adding more logic to handle `0 seconds` will just introduce more code.

scripts/object_storage_cleaner/test.py
Rakanhf: It's default from boto response

scripts/object_storage_cleaner/test.py
Rakanhf: I'm not sure If I get it but what would be the propose of this refactor ? Since I don't see any extra coverage / cases covered by having this approach, since in the background the same logic runs for both scan and deletion, and 2 of 5 `--dry-run` testing won't have anything to delete in the first place

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: IMO, I don't prefer to raise an error in this case, since it should simply just be skipped instead. same goes for the comment above for `assert`

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: Same comment as bellow

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: IMO, I think that they should all be streams, which makes more sense to me. and I believe this chunk / stream mix up can just simply be `stream`.

Maybe to make it more technical and easier, we can use  `iterator` e.g 

raw_stream -> raw_iterator

What do you think ?

scripts/object_storage_cleaner/object_storage_cleaner.py
Rakanhf: let me know what do you think about this :

```python
for v in versions:
    if v["IsDeleteMarker"]:
        # Delete markers have no size
        continue
    # Versions have a size
    total_size += v["Size"]
```

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: Why do you add zero seconds? It is just relative to a timestamp that is 0 seconds before the timestamp.

This script shall not be run with 0 seconds in real life.

Please make it required parameter.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: Why would you prefer a silent error instead of explicit one? If the key is mandatory, we should be very vocal if something breaks the protocol.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: Iterator is good.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: We have the assumption we have no more than 1000 results. If we suddenly get 10000000, the memory usage of the script will blow up. The 1000 limit is set by s3 api specification, so it shall never raise. But what if we use a funky Object storage implementation that doesn't?

Rakanhf: I guess that might not be a problem that we will have in the future, since I don't think we will be using any funky object storage implementation.

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: I guess we can simply use `latest_version["IsDeleteMarker"]` instead ?

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: What about if someone wan't to run the script without a `--retention-period` ? 

let's assume that someone have an old bucket dates to 1y+ and they just want to clean it, they will need to explicitly use   `--retention-period 52 weeks`, while if they run it without specifying `--retention-period` they will be good to go within the first run.

suricactus: This will most probably never happen. Let's dont overthink the purpose of the script.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: We are using 4 different implementations of Object storage and they all have their subtle quirks.

Please add the assert, it does not hurt anyone and there is no need to guess the future.

scripts/object_storage_cleaner/purge_deleted_objects.py
suricactus: Exactly, this will throw a KeyError if it is missing.

scripts/object_storage_cleaner/object_storage_cleaner.py
suricactus: Great! I love early returns

scripts/object_storage_cleaner/test.py
suricactus: Then no need for the .get, you can access it with square brackets.

scripts/object_storage_cleaner/test.py
suricactus: Great, then we test that your explanation is valid.

suricactus: Finally managed to have a laptop review of the code again.

Yes, you should ensure that the `--dry-run` did not accidentally deleted anything.

Also please validate that the file was actually created, because now it's a good hope it did.

I think you can add such asserts `create_file` and `delete_file`

scripts/object_storage_cleaner/test.py
Rakanhf: Since in this test we are using the `list_versions` function which gets the versions from boto3 `"list_object_versions"`, and it returns both versions and deleted markers, we will need to safety check for `IsDeleteMarker`.  It's not related to the `IsDeleteMarker` logic coming from the script objects e.g `QfcObjectVersionTypeDef`

suricactus: Ok within the test.

scripts/object_storage_cleaner/test.py
```diff
@@ -0,0 +1,744 @@
...
+        """
+        test that all versions (history) are permanently deleted when the key is deleted with a large amount of data in an unordered way.
+        """
+        # Create key1
+        key1 = f"{self.unique_prefix}_SHOULD_BE_DELETED"
+        self.create_file(key1, "a" * 100)
+
+        # Create 100 files
+        for i in range(100):
+            self.create_file(f"{self.unique_prefix}file_{i}.txt", "a" * 100)
```
suricactus:
```suggestion
            self.create_file(f"{self.unique_prefix}file_{i}.txt", str(i).ljust(100, "a"))
```

Just to be super safe when checking the right file is created in the assert that checks the file contents, make sure the file contents are always different. There are a few for loops in the tests doing this.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+                "Objects": objects,
+                "Quiet": False,
+            },
+        )
+
+        # Log successes
+        for success in response.get("Deleted", []):
+            logger.info(
+                f"Permanently deleted: {success['Key']} (VersionId: {success.get('VersionId')})"
+            )
```
gounux, line 219: Wondering if it might create a lot of log lines here, but it's alright we can grep,

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+        # Update counters
+        key_count += 1
+        total_size_bytes += obj.total_size_bytes
+        total_versions_count += obj.versions_count
+
+        # Log details for each key
+        logger.info(
+            f"Found: {obj.key} | Deleted At: {obj.deleted_at} | "
+            f"Wasted Size: {obj.total_size_bytes} bytes ({obj.versions_count} versions)"
+        )
```
gounux, line 311: During this `action_scan`, also wondering if this log should be a debug one.  
No hard opinion, am also okay with its current state.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+
+    # Log summary
+    logger.info(f"Total keys found: {key_count}")
```
gounux:
```suggestion
    logger.info(f"Total deleted keys found: {key_count}")
```

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+    Args:
+        s3_client: The configured S3 client.
+        bucket: The bucket name.
+        objects_iterator: The iterator of logically deleted objects.
+    """
+    batch: list[ObjectVersionOrDeleteMarker] = []
+
+    for obj in objects_iterator:
+        for v in obj.versions:
+            batch.append(v)
```
gounux:
```suggestion
        for version in obj.versions:
            batch.append(version)
```

I believe verbose variable names help keeping in mind why the variable is here for.

Just a suggestion, feel free to ignore since this one is only used here, I think.

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+    )
+    parser.add_argument("bucket", help="Target Object Storage bucket")
+    parser.add_argument("--prefix", help="Filter by prefix")
+    parser.add_argument("--profile", help="Optional AWS Profile")
+    parser.add_argument(
+        "--retention-period",
+        type=parse_retention_period,
+        dest="retention_cutoff_ts",
+        required=True,
+        help="Only process objects deleted BEFORE this duration (e.g. '30 days'). Recent deletions are preserved.",
```
gounux:
```suggestion
        help="Only process objects deleted BEFORE this time interval (e.g. '30 days'). Recent deletions are preserved.",
```

At first glance, I found this sentence difficult to understand. Re-reading it again, it's alright. Maybe this suggestion can help, IDK,

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+        bucket: The bucket name.
+        prefix: Optional key prefix to limit the scan.
+
+    Yields:
+        ObjectVersionOrDeleteMarker: A dictionary representing either a Version or a Delete Marker.
+    """
+    paginator = s3_client.get_paginator("list_object_versions")
+    paginate_kwargs = {
+        "Bucket": bucket,
+        "MaxKeys": 1000,
```
gounux:
```suggestion
        "MaxKeys": DELETE_BATCH_SIZE,
```

I believe this is also the meaning of the `DELETE_BATCH_SIZE` constant declared on top of the file ?

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+
+    Args:
+        s3_client: The configured S3 client.
+        bucket: The bucket name.
+        versions: The list of versions to delete.
+    """
+    if not versions:
+        return
+
+    assert len(versions) <= 1000
```
gounux:
```suggestion
    assert len(versions) <= DELETE_BATCH_SIZE
```

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
...
+    for obj in objects_iterator:
+        # Update counters
+        key_count += 1
+        total_size_bytes += obj.total_size_bytes
+        total_versions_count += obj.versions_count
+
+        # Log details for each key
+        logger.info(
+            f"Found: {obj.key} | Deleted At: {obj.deleted_at} | "
+            f"Wasted Size: {obj.total_size_bytes} bytes ({obj.versions_count} versions)"
```
gounux: Regarding the `Wasted Size: {obj.total_size_bytes}`, what about printing it in KB, MB, GB, etc ? Feels more natural to me,

scripts/object_storage_cleaner/README.MD
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## Usage
+
+Run the script directly from the command line.
+
+### Basic Syntax
+
+```bash
+python purge_deleted_objects.py <bucket_name> [options]
```
gounux:
```suggestion
python purge_deleted_objects.py <bucket_name> --retention-period "2 days" [options]
```

I'd add the retention period in the example, since it's required,

scripts/object_storage_cleaner/purge_deleted_objects.py
```diff
@@ -0,0 +1,499 @@
+"""
+Object Storage Logically Deleted Object Cleaner
+
+This script scans an object storage bucket for "logically deleted" objects (objects where the
+latest version is a Delete Marker) and optionally permanently deletes all versions
+of those objects to reclaim storage space.
+
+Usage:
+    python purge_deleted_objects.py <bucket> [options]
```
gounux:
```suggestion
    python purge_deleted_objects.py <bucket> --retention-period "2 days" [options]
```

Same here, no hard opinion though but since it's required, having it in the example makes sense to me,

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: Yes, this is intended to log every deleted version, e.g If you deleted one file with 2 versions it will log 2 times

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: Having it set to `INFO` will enhance the user experience if this script is ever run in a non-automated way.

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: While they have the same value, they are technically different, Since boto3 documentation have `DELETE_BATCH_SIZE` for deleted objects limit and the `MaxKeys` for pagination limit that can be 1,000 Max. 

What I suggest is that maybe we can have a new constant e.g `MAX_PAGINATION_KEYS`

what do you think about this ?

scripts/object_storage_cleaner/purge_deleted_objects.py
Rakanhf: This is a great idea, It was implemented in earlier implementations, but removed to simplify the code. but we can add it, I don't have a strong opinion on this, since it's a simple function and a couple of things to change in the tests.

scripts/object_storage_cleaner/purge_deleted_objects.py
gounux: > What I suggest is that maybe we can have a new constant e.g MAX_PAGINATION_KEYS

Ok, yes, leaving it up to you to assess if that value would change,

scripts/object_storage_cleaner/purge_deleted_objects.py
gounux: Fair enough, thanks :)  
Resolving this,

scripts/object_storage_cleaner/purge_deleted_objects.py
gounux: That makes sense, thanks. Indeed having the deleted version logged is quite valuable,

scripts/object_storage_cleaner/purge_deleted_objects.py
toebivankenoebi: Discussed offline today and agreed on putting it back in.

scripts/object_storage_cleaner/test.py
```diff
@@ -0,0 +1,748 @@
...
+                "--prefix",
+                self.unique_prefix,
+                "--retention-period",
+                "1 second",
+            ]
+        )
+
+        # 3. Verify output
+        self.assertIn("Total deleted keys found: 0", result.stdout)
+        self.assertIn("Total size wasted: 0.00 B", result.stdout)
```
suricactus: 0.00 bytes? I will be surprised to see 0.5 bytes. Can we just print 0 in these cases?

Rakanhf: Total missed that, probably I was thinking about quantum computer.

Rakanhf:
I believe that now V1 of this script is ready @suricactus @gounux 

Before I merge I will clean up the comments so it merges cleanly

PR #1471, opened 2026-02-03
feat: add global search functionality in admin panel

Rakanhf:
Currently searching for a given username or email needs switching between multiple Django admin pages. So we wanted to add a new search page to search accross mutliple models at the same time.

- Add a new top menu link for global search page
- Add a new page `global_search.html` to handle display of `Organization`, `Project`, `subscriptions`>
- Add a new admin function to handle searching, paginating model results.

Screencast from 2026-02-03 21-24-03

### DIFF
--- a/docker-app/qfieldcloud/core/templates/admin/global_search.html
+++ b/docker-app/qfieldcloud/core/templates/admin/global_search.html
@@ -9,24 +9,23 @@
 {% endblock %}
 
 {% block content %}
-<div id="content-main" style="width: 75%; margin: 0 auto;">
+<div id="content-main" class="w-75 mx-auto">
     <h1>{% trans 'Global Search' %}</h1>
 
     {# Search Form #}
-    <div class="module" style="margin-bottom: 10px;">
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/templates/admin/global_search.html
```diff
@@ -0,0 +1,211 @@
...
+{% endblock %}
+
+{% block content %}
+<div id="content-main" style="width: 75%; margin: 0 auto;">
+    <h1>{% trans 'Global Search' %}</h1>
+
+    {# Search Form #}
+    <div class="module" style="margin-bottom: 10px;">
+        <form method="get" action="{% url 'admin:global_search' %}" style="padding: 20px;">
+            <div style="display: flex; gap: 10px; align-items: center;">
```
suricactus:
```suggestion
            <div class="...">
```

Please move all of these to a CSS class, preferable something native from django admin. If it does not exist and it is not completely ugly, then better not style it. The reason is to keep the compatibility with admin themes in the future.

suricactus (approved):

PR #1475, opened 2026-02-10
chore: update django storage import from `s3boto3.S3Boto3Storage` to `s3.S3Storage`

gounux:
Since [v1.14](https://github.com/jschneier/django-storages/blob/master/CHANGELOG.rst#114-2023-09-04) of `django-storage` introduces a dependency import change :

```python
from storages.backends.s3boto3 import S3Boto3Storage
```

becomes

```python
from storages.backends.s3 import S3Storage
```

### DIFF
--- a/docker-app/qfieldcloud/core/management/commands/extracts3data.py
+++ b/docker-app/qfieldcloud/core/management/commands/extracts3data.py
@@ -7,7 +7,7 @@ import mypy_boto3_s3
 from django.conf import settings
 from django.core.files.storage import storages
 from django.core.management.base import BaseCommand, CommandParser
-from storages.backends.s3boto3 import S3Boto3Storage
+from storages.backends.s3 import S3Storage
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-7331 Django-storages namespace changed to s3.S3Storage](https://app.clickup.com/t/2192114/QF-7331)

lukasgraf (approved):
LGTM 👍 

The changelog also mentions that the install extra is now `s3` (so that would be `django-storages[s3]`). However, I noticed that we're not even using that extra, we just [depend on `django-storages==1.14.6`](https://github.com/opengisch/QFieldCloud/blob/f5ef59f0ee15fb40dfb45733119af0fa53671ee4/docker-app/requirements/requirements.in#L29).

Is there a particular reason for that? It's not a big deal, since that extra does nothing more than [pull in`boto3>=1.4.4`](https://github.com/jschneier/django-storages/blob/ca89a94a7462a2423df460e7bfd5f847457042ca/pyproject.toml#L61-L63), which we also pull in directly, but it's still usually good practice to declare dependencies like that.

Just something I noticed though, totally fine to merge without that extra IMHO.

PR #1476, opened 2026-02-10
feat: add upload_thumbnail ViewSet action

Rakanhf:
Currently, the thumbnails are assumed to be accessed via a shared volume from the worker_wrapper after they are produced by the worker (qgis) container, which is not convenient.

This endpoint is intended to help us stop using shared volumes

- Added a new endpoint `/projects/{PROJECT_ID}/thumbnail/` to upload new thumbnail to projects
- Added `ProjectThumbnailSerializer` for handling request body
- Added `MaxFileSizeValidator` and `MaxImageDimensionValidator` validation functions


suricactus:
I would suggest we add a few limitations on the file being uploaded:

- smaller than 5MB
- png or jpg, but no other format
- if possible validate it is one of these formats, so noone can upload `virus.bat`
- probably resolution limitation of 10000x10000

I think the first two should be very straightforward to add.

Each of these would need a test case to ensure they fail.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -194,6 +195,51 @@ class Meta:
...
+        1. File size (max 5MB)
+        2. Actual image format/content
+        """
+        if not value:
+            return value
+
+        # 1. Validate file size (max 5MB)
+        max_size_mb = 5
+        max_size_bytes = max_size_mb * 1024 * 1024  # 5MB in bytes
+        if value.size > max_size_bytes:
```
```suggestion
        if value.size > settings.QFIELDCLOUD_PROJECT_THUMBNAIL_MAX_BYTES:
```

Let's not throw magic numbers everywhere.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -194,6 +195,51 @@ class Meta:
...
+        # 2. Validate image format/content and dimensions using PIL
+        try:
+            # If the file is not an image, PIL will raise a ValidationError
+            image = Image.open(value)
+
+            # Validate dimensions
+            width, height = image.size
+            max_dimension = 10000
+
+            if width > max_dimension or height > max_dimension:
```
```suggestion
            # Validate dimensions
            if width > settings.QFIELDCLOUD_PROJECT_THUMBNAIL_MAX_DIMENSION or height > settings.QFIELDCLOUD_PROJECT_THUMBNAIL_MAX_DIMENSION:
```

docker-app/qfieldcloud/core/serializers.py
suricactus: Don't forget to udpate the function doc too.

suricactus:
> Let me know what do you think, since I lean towards either making the validation just for this serializer or implement a more general function if there is a need in the feature to use it elsewhere.

> Mainly I didn't consider adding the validation to the models, since this is the only point we expose thumbnail upload, it'a mainly handled in internal Jobs / background tasks. and I didn't want to have any unexpected behavior from the Jobs.

Yes, but what if we add django admin tomorrow (we should btw)? Or add user interface so users can change the thumbnail themselves? Then we need to reimplement the same validation in these other endpoints. That's the reason I tend to prefer validation as close as possible to the database. It also helps a lot with the mental model - you know all thumbnails, regardless of the source, are matching the same requirements in your database.

In this particular case, I am okayish to add the validation on serializer level too, as the validation function is almost identical for models and serializers and we if needed we can adapt it. But since I already mentioned we should add the thumbnail upload also on Django admin, I think it makes more sense to be on the model already.

> And since the serializer is tightly coupled with the view itself and validations won't be used anywhere else , I decided not to add any constances in the settings, since it's not global.

Most of the times assume that such validation etc are going to be used in more than one place. Through the life of this project it proved to be more preferable to have stuff as a libs rather than in the implementation details of any particular model/serializer. I am aware that this is not always applied in the current state of the code, but this is something I learned through the time and prefer in any new contribution.
 
> If you think that these functions should be more vague or global to be used for avatars, etc. I think we should also re write the validation functions towards that, and probably name the contance something like `QFIELDCLOUD_UPLOADED_IMAGE_MAX_SIZE` and 'QFIELDCLOUD_UPLOADED_IMAGE_MAX_DIMENSION' instead. Since I believe that both avatar and thumbnail will realistically share the same limits (in our case). which will be better than having multiple contances in the settings, and If we spot the need for it, we can simply implement it in a later PR.

I tend to agree that both avatars and thumbnails should have the same validation in the future. But it's not the future yet, and we never know if the future we are predicting is going to be the one that we think, so I would suggest to keep the naming that clearly states it is related to the thumbnails for now. Use the names from above as a template, ok to change them to your taste. If we need the same validations for the avatar in the future, then we can easily rename them in `settings.py` without any problem (actually it happens quite often). Please add them to `settings.py` instead of constance, or I am missing a reason why Johnny should be able to change this configuration?

docker-app/qfieldcloud/settings.py
```diff
@@ -349,6 +349,10 @@
         "Attachments can be unversioned only when using an attachment storage of type WebDAV (qfieldcloud.filestorage.backend.QfcWebDavStorage)"
     )
 
+QFIELDCLOUD_PROJECT_THUMBNAIL_MAX_DIMENSION = 10000  # 10000px per dimension
+
+QFIELDCLOUD_PROJECT_THUMBNAIL_MAX_BYTES = 5 * 1024 * 1024  # 5MB
+
```
suricactus: Please avoid inline comments, they are reserved for the tooling we use (ruff, mypy, coverage etc). Feel free to either drop, add an inline comment before the line, or add a docstring below the variable.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -74,6 +80,11 @@ def has_permission(self, request, view) -> bool:
         description="""Create a new project owned by the specified
         user or organization"""
     ),
+    upload_thumbnail=extend_schema(
+        description="Update the project thumbnail",
+        responses={200: None},
```
suricactus: 200 or 201? I think 201 probably suits better the endpoint, no?

Rakanhf: I think that status `200 ok` is a bit better than `201 created`, since we are not uploading it once, it's also used for updating the thumbnail

suricactus: Actually after second read, I guess it has to be 204, as there is no content returned, right?

gounux (commented):
Thanks for the PR, LGTM !

One thing : apologies for coming with this after @suricactus's review & approval, but we faced some support cases in the past, during which there were important amounts of projects thumbnails created - the ones generated and pushed to S3 during the `PROCESS_PROJECTFILE` typed Jobs.

We realized that each time a new project thumbnail is generated / created / pushed on S3, it creates a new version there, thus sometimes the number of thumbnail file versions was reaching the limit.

So the fix at that moment was to introduce this `purge_previous_thumbnails_versions` utils functions, declared here :

https://github.com/opengisch/QFieldCloud/blob/3d4a9cb74ae22ded721526e1a83446e60a31fe8f/docker-app/qfieldcloud/core/utils2/storage.py#L448-L450

And called at the end of a `PROCESS_PROJECTFILE` Job run, so that there is basically only one version of the thumbnail kept, and the previous ones versions of this thumbnail are deleted :

https://github.com/opengisch/QFieldCloud/blob/3d4a9cb74ae22ded721526e1a83446e60a31fe8f/docker-app/worker_wrapper/wrapper.py#L656-L658

In my opinion this topic and limitation is related to this PR, and we wouldn't want to get in a situation where there are more than the limit of possible file versions for a project thumbnail,

Leaving it up to you to assess if such a purge method should be used here, or if there is maybe a better way to address this file versions limit, in a follow-up PR or so.

docker-app/qfieldcloud/core/views/projects_views.py
Rakanhf: Yes, I guess it will be better than `201`, and we don't return the updated object, we can indeed change it to 204 no content

PR #1478, opened 2026-02-11
chore: drop support of legacy storage

gounux:
The title says it : drop support of the historic and legacy QFieldCloud storage.  

After this potential long-life PR gets merged, QFieldCloud will not support any S3 storage flagged as `"QFC_IS_LEGACY": true` in the .env's `STORAGES` var anymore.

The work has partly consisted of grepping for any `QF-4963` / `legacy` occurrences.  

Note : the pre-commit tools were quite useful regarding deletions that were sometimes too abrupt.


duke-nyuki:

docker-app/worker_wrapper/wrapper.py
gounux: Wondering if it makes sense to keep it like this, i.e. if `project__the_qgis_file_name` might be empty on default storage only,

.github/workflows/test.yml
```diff
@@ -102,7 +88,7 @@ jobs:
           }'
           EOF
 
-          echo "STORAGES_PROJECT_DEFAULT_STORAGE=${{ matrix.storage }}" >> .env
+          echo "STORAGES_PROJECT_DEFAULT_STORAGE=default" >> .env
```
suricactus: You can drop this line, it already assumes empty value means default.

docker-app/worker_wrapper/wrapper.py
suricactus: Then throw errors whenever possible, so we can fix them :)

.github/workflows/test.yml
gounux: Addressed in https://github.com/opengisch/QFieldCloud/pull/1478/changes/7ab6a8ca1a92fddf1772667603a4d80b45eb03c3

docker-app/qfieldcloud/filestorage/management/commands/migrateavatars.py
suricactus, line 1: I would suggest we remove this, nobody isi going to use it anymore, right?

docker-app/qfieldcloud/filestorage/serializers.py
```diff
@@ -46,7 +46,6 @@ class Meta:
             "uploaded_at",
             "display",
             "is_latest",
-            # TODO delete the fields below, they are deprecated and exists only as a compatibility layer for the legacy storage
```
suricactus, line 49: I would keep this comments. We really need to kill them and make our life easier without the `sha256`. None of our clients should use them.

suricactus (approved):
Nice to see 3.5k lines deleted <3 .

Few small comments.

docker-app/qfieldcloud/filestorage/serializers.py
gounux: Sure, see https://github.com/opengisch/QFieldCloud/pull/1478/commits/2b3bfa07b903f85de67875ac84694086946c23c9

docker-app/qfieldcloud/filestorage/management/commands/migrateavatars.py
gounux: Deleted in https://github.com/opengisch/QFieldCloud/pull/1478/commits/aef902a848b3f99d2c74bdcb92168b8410e368ad

suricactus (approved):
Please do NOT squash before merging.

suricactus:
@gounux let me know when you are ready for the final review. I see the checks are green, but there is still one conflict.

gounux:
> @gounux let me know when you are ready for the final review. I see the checks are green, but there is still one conflict.

Thanks @suricactus, rebased on latest master and conflicts resolved,  
I also removed the `do-not-merge` label, a final review round would be appreciated :)

docker-app/qfieldcloud/core/utils.py
```diff
@@ -398,153 +248,6 @@ def get_deltafile_schema_validator() -> jsonschema.Draft7Validator:
...
-                md5sum=f.e_tag.replace('"', ""),
-            )
-        )
-
-    files.sort(key=lambda f: f.name)
-
-    return files
-
-
 def list_versions(
```
suricactus: Kill? I don't see who uses this function?

docker-app/qfieldcloud/core/utils.py
suricactus: This is used only in one migration: https://github.com/opengisch/QFieldCloud/blob/414d7715cad18ec581a1daa527c1d5750296da01/docker-app/qfieldcloud/core/migrations/0041_auto_20210705_0837.py#L49

I think we can safely remove the usage from the migration, as there will be no files for someone who is just getting started with QFieldCloud.

suricactus: This can be deleted, completely useless usage here: https://github.com/opengisch/QFieldCloud/blob/414d7715cad18ec581a1daa527c1d5750296da01/docker-app/qfieldcloud/core/tests/test_delta.py#L285

suricactus: This is not used anywhere, which will allow `get_legacy_s3_credentials` to go away too.

docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -282,9 +281,7 @@ def test_push_apply_delta_file_invalid_json_schema(self):
         self.assertEqual(FaultyDeltaFile.objects.count(), 1)
 
         faulty_deltafile = FaultyDeltaFile.objects.first()
-        prefix = utils.safe_join(f"projects/{project.id}/deltafiles/")
 
-        self.assertTrue(faulty_deltafile.deltafile.name.startswith(prefix))
```
suricactus, line 287: You took it a bit too literal. You can restore it as:

```
        prefix = f"projects/{project.id}/deltafiles/"

        self.assertTrue(faulty_deltafile.deltafile.name.startswith(prefix))
```

docker-app/qfieldcloud/core/migrations/0041_auto_20210705_0837.py
```diff
@@ -141,11 +132,6 @@ def forwards_find_project_filename_func(apps, schema_editor):
             model_name="processprojectfilejob",
             name="feedback_old",
         ),
-        # / add feedback field for all jobs
-        # / add feedback field for all jobs
-        migrations.RunPython(
-            forwards_find_project_filename_func, migrations.RunPython.noop
-        ),
```
suricactus, line 148: The first comment is to stay, it is closing the opening at line 125 (116).

Here we can also add a NOTE:

> NOTE: The `Project.project_filename` field will be polulated once the first `.qgs`/`.qgz` file is uploaded after this migration.

docker-app/qfieldcloud/core/tests/test_delta.py
gounux: Thanks indeed...

Restored in https://github.com/opengisch/QFieldCloud/pull/1478/changes/7490fdfe9f47ca9767ddabef0e0e71bfcf5a4b31

suricactus (approved):
Looks good! I am happy to give it a try!

gounux:
:tada:

PR #1479, opened 2026-02-11
fix: add missing env vars in docker compose file

gounux:
Somehow those 2 recently added env variables were not present in the `app`'s environment declaration, so until this PR they didn't have any effect.


Rakanhf (approved):
LGTM

PR #1481, opened 2026-02-11
fix: update signup_closed template to use blocktrans with about_url variable

Rakanhf:
Href for "join the waiting list" was not being rendered correctly. 

I also suggest that we can either remove it or create a new contance for it.


gounux:
Approval + such a "little" and not scary change, merging this,

PR #1484, opened 2026-02-18
fix: allow social account linking only to useraccounts of type `PERSON`

gounux:
This PR intends to allow the creation of allauth's `SocialAccount` only to useraccount having the `PERSON` type. Thus disabling Social Accounts creation and SSO login for email addresses related to Organization.

It has been observed that if an Organization and a User have the same email address, login via SSO using this email address may result in being connected as the Organization, which is not alright. The meaning of this PR is also to avoid that.

## Steps to reproduce :

1. in Django Admin create an Organization `o` with email address `A`
2. in Django Admin create a People `p` with same email address `A`
3. login using a 3rd party IDP with this same email address `A`

-> without the changes brought by this PR, you are logged in as the Organization`o`, created at step 1.  
-> with the changes, you are logged in as the People `p`, created at step 2.

### Login with 3rd party IDP with only the Organization having the `A` email address -> blocked

1. delete any `Social Account` related to the email address `A` mentioned above, in Django Admin
2. change the `p` People's email address
-> there is only the Organization `o` having the `A` email address
3. login using a 3rd party IDP with this same email address `A`
-> it fails, since SSO login for Organization is not allowed, with an error message displayed : `Can login via SSO only to user accounts, not organizations.`

### Login with 3rd party IDP with no `A` email address anywhere  -> new Social Account and new QFieldCloud user created

1. change the `o` Organization's email address
-> there is no more Organization neither People having this `A` email address
2. login using a 3rd party IDP with this same email address `A`
-> you're logged in as a new QFieldCloud user, that was created during the login and to which a new Social Account is binded (for this you need `QFIELDCLOUD_ACCOUNT_ADAPTER=qfieldcloud.core.adapters.AccountAdapterSignUpOpen` is the `.env` environment file)

### DIFF
--- a/docker-app/qfieldcloud/core/adapters.py
+++ b/docker-app/qfieldcloud/core/adapters.py
@@ -7,21 +7,18 @@ from typing import Literal
 from allauth.account import app_settings
 from allauth.account.adapter import DefaultAccountAdapter
 from allauth.account.models import EmailConfirmationHMAC
-from allauth.core.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
-from allauth.socialaccount.models import SocialLogin
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 from constance import config
 from django.contrib import messages
 from django.contrib.auth.models import AbstractUser
 from django.core.exceptions import ValidationError
 from django.http import HttpRequest
-from django.shortcuts import redirect
 from django.utils import timezone
 from invitations.adapters import BaseInvitationsAdapter
 
 from qfieldcloud.authentication.sso.provider_styles import SSOProviderStyles
-from qfieldcloud.core.models import Person, User
+from qfieldcloud.core.models import Person
 
 logger = logging.getLogger(__name__)
 
@@ -196,47 +193,6 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
 
     Logs stack trace and error details on 3rd party authentication errors.
     """
-
-    def pre_social_login(self, request: HttpRequest, sociallogin: SocialLogin) -> None:
-        """
-        Invoked just after a user successfully authenticates via a
-        social provider, but before the login is actually processed
-        (and before the pre_social_login signal is emitted).
-
-        This is a good hook to check that the user is of type `PERSON`,
-        and not of type `ORGANIZATION` or `TEAM`,
-        as SSO login should only be allowed for user accounts.
-        """
-
-        # check if allauth can find a useraccount with this email,
-        # meaning that the accounts would be linked and a social account created.
-        if sociallogin.is_existing:
-            # here we check that the user trying to login is of type `PERSON`.
-            user = sociallogin.user
-
-            if user.type != User.Type.PERSON:
-                # if the user is not of type `PERSON`,
-                # we try to find a user with the same email that is of type `PERSON`,
-                # and link the social account to that user instead.
-                # Otherwise we block the social account link to an Organization.
-
-                email = sociallogin.account.extra_data.get("email", "")
-
-                if not email:
-                    messages.error(
-                        request, "No email returned by the Identity Provider."
-                    )
-                    raise ImmediateHttpResponse(redirect("account_login"))
-
-                try:
-                    person_user = User.objects.get(email=email, type=User.Type.PERSON)
-                    sociallogin.user = person_user
-                except User.DoesNotExist:
-                    messages.error(
-                        request,
-                        "Can login via SSO only to user accounts, not organizations.",
-                    )
-                    raise ImmediateHttpResponse(redirect("account_login"))
 
     def on_authentication_error(
         self,
--- a/docker-app/qfieldcloud/core/adapters.py
+++ b/docker-app/qfieldcloud/core/adapters.py
@@ -7,18 +7,21 @@ from typing import Literal
 from allauth.account import app_settings
 from allauth.account.adapter import DefaultAccountAdapter
 from allauth.account.models import EmailConfirmationHMAC
+from allauth.core.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
+from allauth.socialaccount.models import SocialLogin
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 from constance import config
 from django.contrib import messages
 from django.contrib.auth.models import AbstractUser
 from django.core.exceptions import ValidationError
 from django.http import HttpRequest
+from django.shortcuts import redirect
 from django.utils import timezone
 from invitations.adapters import BaseInvitationsAdapter
 
 from qfieldcloud.authentication.sso.provider_styles import SSOProviderStyles
-from qfieldcloud.core.models import Person
+from qfieldcloud.core.models import Person, User
 
 logger = logging.getLogger(__name__)
 
@@ -193,6 +196,47 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
 
     Logs stack trace and error details on 3rd party authentication errors.
     """
+
+    def pre_social_login(self, request: HttpRequest, sociallogin: SocialLogin) -> None:
+        """
+        Invoked just after a user successfully authenticates via a
+        social provider, but before the login is actually processed
+        (and before the pre_social_login signal is emitted).
+
+        This is a good hook to check that the user is of type `PERSON`,
+        and not of type `ORGANIZATION` or `TEAM`,
+        as SSO login should only be allowed for user accounts.
+        """
+
+        # check if allauth can find a useraccount with this email,
+        # meaning that the accounts would be linked and a social account created.
+        if sociallogin.is_existing:
+            # here we check that the user trying to login is of type `PERSON`.
+            user = sociallogin.user
+
+            if user.type != User.Type.PERSON:
+                # if the user is not of type `PERSON`,
+                # we try to find a user with the same email that is of type `PERSON`,
+                # and link the social account to that user instead.
+                # Otherwise we block the social account link to an Organization.
+
+                email = sociallogin.account.extra_data.get("email", "")
+
+                if not email:
+                    messages.error(
+                        request, "No email returned by the Identity Provider."
+                    )
+                    raise ImmediateHttpResponse(redirect("account_login"))
+
+                try:
+                    person_user = User.objects.get(email=email, type=User.Type.PERSON)
+                    sociallogin.user = person_user
+                except User.DoesNotExist:
+                    messages.error(
+                        request,
+                        "Can login via SSO only to user accounts, not organizations.",
+                    )
+                    raise ImmediateHttpResponse(redirect("account_login"))
 
     def on_authentication_error(
         self,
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/core/adapters.py
```diff
@@ -194,6 +197,47 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
...
+                email = sociallogin.account.extra_data.get("email", "")
+
+                if not email:
+                    messages.error(
+                        request, "No email returned by the Identity Provider."
+                    )
+                    raise ImmediateHttpResponse(redirect("account_login"))
+
+                try:
+                    person_user = User.objects.get(email=email, type=User.Type.PERSON)
```
lukasgraf: I would try to emulate allauth's [`SocialLogin._lookup_by_email`](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/models.py#L386-L395) as much as possible here. And just where it gets the results from [`filter_users_by_email`](https://github.com/pennersr/django-allauth/blob/main/allauth/account/utils.py#L243-L282) then filter that list of users to only keep the ones that match `type == User.Type.PERSON`.

Because with this implementation, we're missing several parts of the functionality that allauth does during that lookup:
- Making sure to prefer verified email addresses
- Lower casing emails to make lookups case insensitive
- Making unicode-aware comparisons for the email addresses
- Properly setting `sociallogin._did_authenticate_by_email = email`

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -194,6 +197,47 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
...
+                    sociallogin.user = person_user
+                except User.DoesNotExist:
+                    messages.error(
+                        request,
+                        "Can login via SSO only to user accounts, not organizations.",
+                    )
+                    raise ImmediateHttpResponse(redirect("account_login"))
```
lukasgraf: I'm not sure if the exception handling here (and above, for missing email) is ideal.

For one, this code path will also be taken when a user signs up using QField or the QFieldSync plugin. In that case, `messages.error()` will have zero effect, and the user never sees that message.

The other thing is that this would lead to the following situation: When an `Organization` exists with a particular email address, but no `Person`, then a user with that email in his IDP account would not be able to sign up, because we prevent it here.

I think we should instead modify the original behavior of django-allauth *as little as possible*, and only solve the issue of not allowing organizations / teams to be picked during lookup. And this would mean an implementation like this:
- Check if the account is anything else than a `Person`
- If yes:
  - Restore the `sociallogin` object to exactly the state it was before allauth did the lookup that found the origanization via the email lookup. That means undoing [these changes](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/models.py#L392-L394):
    - Reset `sociallogin.user`
    - Reset `sociallogin._did_authenticate_by_email`
  - Then, we perform another email lookup ourselves. Emulating as much as possible from the original `SocialLogin._lookup_by_email`, but: Instead of directly using the results of `filter_users_by_email(email, prefer_verified=True)` [here](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/models.py#L391C21-L391C71), first filter them to throw out anything that is not a `Person`. Then continue as normal.

For the case described above (org exists with email, but no person), this should then result in `sociallogin.is_existing` being `False`, because we reset it, and didn't find an existing user, and allauth proceeding normally, which means it will create the account and link it.

lukasgraf (changes requested):

docker-app/qfieldcloud/core/adapters.py
gounux: Okay I see, thanks ! Regarding the exception handling, should we just throw an Exception ?

> The other thing is that this would lead to the following situation: When an Organization exists with a particular email address, but no Person, then a user with that email in his IDP account would not be able to sign up, because we prevent it here.

Yes, that's the tricky point. And indeed it makes sense to allow such a case, rather than throwing and blocking it.

Though, after having tried to implement the logic, resetting the `sociallogin.user` to `None` creates a problem later on, since [the `is_finalized` property is called again](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/internal/flows/login.py#L75), and the `user` property is `None` so [this `if self.user.pk is None` call fails](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/models.py#L327-L333).

> For the case described above (org exists with email, but no person), this should then result in sociallogin.is_existing being False, because we reset it, and didn't find an existing user, and allauth proceeding normally, which means it will create the account and link it.

Wondering what the "Reset sociallogin.user" means here and how we could get in the same state before the lookup, need to investigate allauth deeper,

Also, I tried to create a new Person and assign it to the `sociallogin`, somehow the flow then lands into the allauth's raw Social Signup form, saying `You are about to use your <IDP> account to login to QFieldCloud. As a final step, please complete the following form: ` + Email and Username pre-filled. With a `An account already exists with this email address. Please sign in to that account first, then connect your Google account.` after validating.  
IDK if this is a good approach to dig further,

lukasgraf: > Though, after having tried to implement the logic, resetting the sociallogin.user to None creates a problem later on, since the is_finalized property is called again, and the user property is None so this if self.user.pk is None call fails.

Ahh, right :-/
django-allauth does this weird thing during signup / login where it first creates a "fake" / empty user, and assigns it to the sociallogin object. The further it progresses during the flow and gets more pieces information _(is the useracount already linked to a socialaccount? Does the user with that email exist? What profile metadata did the IDP provide? ...)_ it then continously enriches / populates that empty user with more data. In the case of new user creation, it then saves it, and otherwise (I think, if I remember correctly) it replaces it with a real user that is re-fetched from the DB.

It's all a bit convoluted how allauth does this, and the exact order in which those things happen make it difficult to customize the behavior.

So I think I agree with you here - not sure if this approach is worth following.

Monkey patching the `filter_users_by_email` would be much simpler and cleaner, I think. It should be as simple as patching it like this, more or less (untested):
```python
from django.allauth.account import utils as allauth_utils

orig_filter_users_by_email = allauth_utils.filter_users_by_email


def patched_filter_users_by_email(
    email: str, is_active: Optional[bool] = None, prefer_verified: bool = False
) -> List:

    users = orig_filter_users_by_email()
    return list(filter(lambda user: user.type == User.Type.PERSON, users))

allauth_utils.filter_users_by_email = patched_filter_users_by_email
```

The main challenge with this approach is running this code that does the patching in the right module / at the right time. It needs to run before any modules have already imported the `filter_users_by_email` function, otherwise they will not be affected by the patch.

Let's maybe discuss this tomorrow.

gounux: Thanks a lot @lukasgraf :)

So this commit (https://github.com/opengisch/QFieldCloud/pull/1484/changes/51bef4e51269702636fad7c7b75e5aa9823fe694) intends to patch the several places where allauth is using the `filter_users_by_email`, identified with a grep.

Current feeling is that it's quite hacky, also, what if some new versions of allauth change the way this utils filter function is used, or at new code places ? Will it be necessary to update this patch ?

Anyway, LF to review / discussion regarding this,

Also, with this patch applied I tested those different cases :

- Only org has the email address -> a new QFieldCloud user is created, social login does not result in linking the Org account so this case is okay I believe :white_check_mark: 
- When there is already an existing user with the email address -> raw allauth sign up form is displayed, `email` pre-filled but `username` not pre-filled, and `An account already exists with this email address. Please sign in to that account first, then connect your <IDP> account.`, so this case is KO I guess :x:

lukasgraf: As discussed:
- I would wrap up the patching into its own function and own module, and just make that a one line call from `apps.py`, to keep it self contained
- It should only be necessary to patch the function in the module where it is defined, but not in the places where it's imported (IF our patch is early enough -> test)
- The second case that you describe that fails is surprising to me. d.allauth should just link the account in that case. I have a feeling that this is caused by something else, and not the patch

I'd also like to hear what @suricactus thinks about the monkey patch approach.

suricactus: I will be able to have a look at this next week, needs a bit more thinking than what I can afford now :)

lukasgraf: @suricactus so just to summarize where things are at regarding this:

At the moment we feel neither of these approaches (`pre_social_login` hook, or monkey patching `filter_users_by_email`) is a solution that we're comfortable shipping.

Actually fixing this requires an upstream change in `django-allauth` IMHO. For example, introducing a `lookup()` hook on the `SocialAccountAdapter` that allows you to customize how the lookup is done. And I mean completely customize, by entirely replacing the original implementation.

Because that's the issue with [how the `pre_social_login` hook is implemented](https://github.com/pennersr/django-allauth/blob/d729357926074a1af64a4b881afac45db305aa16/allauth/socialaccount/internal/flows/login.py#L35): Unlike almost all of the [other adapter hooks](https://github.com/search?q=repo%3Apennersr%2Fdjango-allauth+get_adapter%28%29.&type=code), it doesn't _replace_ the original implementation, but it just gets executed _in addition_ to it, giving you an option to intervene. So that means we can't avoid the call to the original `sociallogin.lookup()`, and we therefore would have to undo whatever it did, which is rather tricky and brittle.

suricactus: I think it's pretty clear that ideally we want to have this upstreamed to the `allauth` library.

>> When there is already an existing user with the email address -> raw allauth sign up form is displayed, email pre-filled but username not pre-filled, and An account already exists with this email address. Please sign in to that account first, then connect your <IDP> account., so this case is KO I guess ❌
> The second case that you describe that fails is surprising to me. d.allauth should just link the account in that case. I have a feeling that this is caused by something else, and not the patch

I have the same feeling. Does it fail also without your patch? @gounux 

> At the moment we feel neither of these approaches (pre_social_login hook, or monkey patching filter_users_by_email) is a solution that we're comfortable shipping.

I think the monkey patching is ok for now. The only problem I see is if we move forward with the allauth version. We have two options (not necessarily mutually exclusive):
 - pin the `allauth` version in the `requirements.in` so we are sure it works with this one and only update when the upstreamed patch is available.
 - make some kind of `checksum` of the patched function/module and don't start django if they mismatch with our expectation. This way we protect ourselves from accidental bump OR allows us not to pin the version.

As for the patch itself, it's looks good, makes sense. Have not tried it myself.

On the best way to apply it, I tend to agree with @lukasgraf feeling here:

> It should only be necessary to patch the function in the module where it is defined, but not in the places where it's imported (IF our patch is early enough -> test)

The only problem with this is the order of imports AFAIK. But I believe it would be possible.

---

I would suggest we merge this PR as soon as we have opened PR upstream. Probably a topic for the next sprint.

suricactus:
@gounux did you open your upstream PR?

gounux:
> @gounux did you open your upstream PR?

Indeed, you can find it here -> https://codeberg.org/allauth/django-allauth/pulls/4728

docker-app/qfieldcloud/core/adapters.py
suricactus: I am closing this thread, as we went for an upstream PR directly.

suricactus:
Allowe myself to convert to draft.

gounux:
Closing this, superseded by #1623 after the upstream change in `django-allauth` that make the QFieldCloud PR & changes less _patchy_.

PR #1486, opened 2026-02-19
feat: implement thumbnail upload functionality from sdk

Rakanhf:
Creating a thumbnail should not block successfully using the QGIS project on QField. Instead of worker wrapper directly looking for a thumbnail.png  file via a shared volume we better send the file using the SDK within the worker.

- Removed thumbnail saving from `after_docker_run`
- Add `upload_thumbnail` step at `ProcessProjectfileCommand`
- Add `upload_project_thumbnail` function to upload thumbnail using the sdk

In order for these changes to function properly, we need to resolve these PRs first 
* [PR-1476](https://github.com/opengisch/QFieldCloud/pull/1476) ✅
* [PR-94](https://github.com/opengisch/qfieldcloud-sdk-python/pull/94) ✅

~~Then we should update `qfieldcloud-sdk==0.11.0` to the latest version when PR-94 is released~~

### DIFF
--- a/docker-qgis/qfc_worker/commands/process_projectfile.py
+++ b/docker-qgis/qfc_worker/commands/process_projectfile.py
@@ -25,6 +25,7 @@ from qfc_worker.utils import (
     open_qgis_project_temporarily,
     start_app,
     stop_app,
+    upload_project_thumbnail,
 )
 from qfc_worker.workflow import (
     Step,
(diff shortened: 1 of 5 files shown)


docker-qgis/qfc_worker/commands/process_projectfile.py
```diff
@@ -231,9 +234,21 @@ def get_workflow(self, project_id: UUID, project_file: str) -> Workflow:  # type
                     name="Generate Thumbnail Image",
                     arguments={
                         "the_qgis_file_name": WorkDirPathAsStr("files", project_file),
-                        "thumbnail_filename": Path("/io/thumbnail.png"),
+                        "thumbnail_filename": WorkDirPath("thumbnail.png"),
```
suricactus, line 237: Just make sure we do not upload the `thumbnail.png` by accident as part of the project files.

If I remember correctly, it should not be the case, becase the project files are in `<workdirpath>/files/`, but double checking should be easy.

Rakanhf: No they are not uploaded with the project files, I double checked also on object storage level, and it's not.

PR #1488, opened 2026-02-23
Update libqfieldsync revision to catch mbtiles shared datasets fix

nirvn:
Changes between the last revision and this one: https://github.com/opengisch/libqfieldsync/compare/6eeff164d28b899b997b0dcee14977781b6a0e7e..6221ad54fda3d44c81c12ada441d6c73f888291c


nirvn:
@lukasgraf , correct. It's deployed as we speak on qt5 QGIS installations across the globe

PR #1489, opened 2026-02-23
feat: enhance social login ui and improve provider display

Rakanhf:
- Updated login template

---

Before: 

image

After : 

image

### DIFF
--- a/docker-app/qfieldcloud/core/templates/account/login.html
+++ b/docker-app/qfieldcloud/core/templates/account/login.html
@@ -1,6 +1,7 @@
 {% extends 'account/base.html' %}
 {% load i18n %}
 {% load bootstrap4 %}
+{% load socialaccount %}
 
 {% block title_contents %}{% trans 'Sign-In' %}{% endblock %}
 
@@ -38,6 +39,19 @@
   </div>
 </div>
 
-{% include "socialaccount/snippets/login.html" %}
+{% get_providers as socialaccount_providers %}
+
+{% if socialaccount_providers %}
+    {% if not SOCIALACCOUNT_ONLY %}
+      <div class="d-flex align-items-center my-3">
+        <hr class="flex-grow-1 m-0">
+        <span class="px-3 text-muted">{% trans 'Or' %}</span>
+        <hr class="flex-grow-1 m-0">
+      </div>
+    {% endif %}
+  {% include "socialaccount/snippets/provider_list.html" with process="login" %}
+  {% include "socialaccount/snippets/login_extra.html" %}
+{% endif %}
+
 
 {% endblock content %}


docker-app/qfieldcloud/core/templates/allauth/elements/provider.html
```diff
@@ -1,11 +1,11 @@
 <li class="provider-button-parent">
 
-    <a title="{{ attrs.name }}" href="{{ attrs.href }}">
+    <a title="{{ attrs.styles.name|default:attrs.name }}" href="{{ attrs.href }}">
```
lukasgraf: I would drop this part of the customization.

The provider name (`Microsoft`, `Google`, etc..) that is defined as part of `SOCIALACCOUNT_PROVIDERS` gets used in many places. For example also in QField and QFieldSync, exposed in the [`ListProvidersView`](https://github.com/opengisch/QFieldCloud/blob/c95d37a0d800ec191f786d10ff85aa98e6fcb57d/docker-app/qfieldcloud/authentication/views.py#L181). The only way to keep user interfaces around provider lists for these three application (QFC, QF, QFS) somewhat consistent is if we use the same source of information in all of them, and keep the data model around provider infos clear and unambiguous.

docker-app/qfieldcloud/core/templates/account/login.html
```diff
@@ -11,6 +12,22 @@
...
+{% get_providers as socialaccount_providers %}
+
+{% if socialaccount_providers %}
+  <div class="mt-3">
+    {% include "socialaccount/snippets/provider_list.html" with process="login" %}
+    {% if not SOCIALACCOUNT_ONLY %}
+      <div class="d-flex align-items-center my-3">
+        <hr class="flex-grow-1 m-0">
+        <span class="px-3 text-muted">{% trans 'Or' %}</span>
+        <hr class="flex-grow-1 m-0">
```
lukasgraf: This essentially inlines allauth's `socialaccount/snippets/login.html` here - I like it, that seems a lot more stable and easier to customize than overriding `provider_list.html` 👍 

---

One thing I would consider for the future:
Maybe we want to still insert a placeholder here for a message or a short text that goes _before_ the provider list.

For example, a `Continue with ...` could go there. But also a message like `"If you are inside the internal <company> network, you can use your <company> account listed below to login."` that would be highly customer specific.

I would simply provide an empty string for the English translation as the default, and this would allow us to customize that preamble by simply overriding the translation for a particular deployment.


Maybe something along these lines (untested).

```
{% load i18n %}

{% trans "msg_before_provider_list" as before_message %}

{% if before_message %}
  <div>
    {{ before_message|safe }}
  </div>
{% endif %}
```

```
msgid "msg_before_provider_list"
msgstr ""
```

docker-app/qfieldcloud/core/templates/account/login.html
```diff
@@ -38,6 +55,5 @@ <h5 class="card-title">{% trans 'Sign-In' %}</h5>
   </div>
 </div>
 
-{% include "socialaccount/snippets/login.html" %}
```
lukasgraf, line 41: As discussed: Switch back the order of username/password login and providers: Password at the top, providers at the bottom.

docker-app/qfieldcloud/core/templates/account/login.html
suricactus: > One thing I would consider for the future:

I like the direction of the thought, but I would argue this is out of scope of the current PR, as we still do not have the usecase? It's going to be trivial to add such functionality later if really needed.

Rakanhf: This will be determined in an internal discussion and then open a flow up PR

PR #1490, opened 2026-02-26
UPDATE Readme: Give context to pre-commit

t-book:
## Expand pre-commit documentation in README

The existing "Code style" section only listed the install commands without much context. This update expands it to be a little  more helpful for new contributors like me :)

No code changes — documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ Now connecting with `curl` should fail with a similar error:
 
 ## Code style
 
-Code style done with [`pre-commit`](https://pre-commit.com):
+This project uses [`pre-commit`](https://pre-commit.com) to automatically run linters and formatters before each commit.
 
     pip install pre-commit
     # install pre-commit hook


README.md
```diff
@@ -307,12 +307,15 @@ Now connecting with `curl` should fail with a similar error:
...
-Code style done with [`pre-commit`](https://pre-commit.com):
+This project uses [`pre-commit`](https://pre-commit.com) to automatically run linters and formatters before each commit, so style issues never make it into the repository.
 
     pip install pre-commit
     # install pre-commit hook
     pre-commit install
 
+This registers pre-commit as a Git hook for this repository. From now on, every git commit will automatically check your staged changes.
+
+All hooks are defined in .pre-commit-config.yaml at the repository root. Check this file if you want to understand which checks are being run.
```
suricactus:
```suggestion
All hooks are defined in `.pre-commit-config.yaml` file at the repository root.
Check this file if you want to understand which checks are being run.
```

Please quote "programmer" identifiers such adds variables, functions, filenames etc with backticks.

Also we try to follow one sentence per line policy in the markdown files, as it makes git diff tracking easier.

There is one more sentence above that needs a separate line in the above paragraph.

README.md
```diff
@@ -307,12 +307,15 @@ Now connecting with `curl` should fail with a similar error:
 
 ## Code style
 
-Code style done with [`pre-commit`](https://pre-commit.com):
+This project uses [`pre-commit`](https://pre-commit.com) to automatically run linters and formatters before each commit, so style issues never make it into the repository.
```
suricactus:
```suggestion
This project uses [`pre-commit`](https://pre-commit.com) to automatically run linters and formatters using a pre-commit hook, so style issues can never be commited.
Check the configured checks in `.pre-commit-config.yaml` file.
```

I think this should be enough and the rest can be dropped. We should not explain every single piece of software we use, especially when we link to the docs ;)

t-book:
Thanks @suricactus for your feedback.

I agree that keeping the README concise and pointing users to the pre-commit documentation makes sense.
I removed the last two paragraphs and squashed the commits while keeping you credited in the commit message and hope that works for you.

The only change left is explicitly mentioning that pre-commit runs linters and formatters automatically before each commit. I think this makes the section slightly more visible and helps developers better understand its purpose at a glance, but I am equally fine leaving it as it was if you prefer (:

PR #1491, opened 2026-02-27
Reduce default QGIS network timeout

vaahtokarkki:
This PR changes the default QGIS network timeout from 60 seconds to 5 seconds.

The current default of 60 seconds feels a bit overkill when processing a project file that may contain invalid or unreachable layers (e.g. service down, network outage, DNS issues, etc.). This is especially problematic for layers that have QFieldSync `cloud_action` set to `remove` or `no_action`. In those situations, each failing layer can block processing unnecessary for up to a minute, which adds up quickly.

We initially ran into this issue in our self-hosted setup, where outbound network access outside the private network is blocked. Because of that, any external layer in the project would wait the full 60 seconds before failing, which in turn caused QFC jobs to time out. While this affected our environment, I think the upstream project could also benefit from this change as well.

### DIFF
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -33,6 +33,7 @@ from qgis.core import (
     QgsProject,
     QgsProjectArchive,
     QgsProviderRegistry,
+    QgsSettings,
     QgsZipUtils,
 )
 from qgis.PyQt import QtCore, QtGui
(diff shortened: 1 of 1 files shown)


docker-qgis/qfc_worker/utils.py
```diff
@@ -153,6 +175,9 @@ def start_app() -> str:
         QtCore.qInstallMessageHandler(_qt_message_handler)
         QgsApplication.messageLog().messageReceived.connect(_write_log_message)
 
+        settings = QgsSettings()
+        settings.setValue("network/network-timeout", 5000)
```
suricactus:
```suggestion
        settings.setValue("network/network-timeout", NETWORK_TIMEOUT_MS)
```

Would you mind switching the magic number into a constant in the start of the file. Would nice if you also add the important points as documentation for like the one you provided in the PR.

vaahtokarkki: Sure! I’ve also added the `QGIS_` prefix to the constant name for clarity.

suricactus (approved):

docker-qgis/qfc_worker/utils.py
suricactus: Lovely!

PR #1492, opened 2026-02-27
Bump various deps

suricactus:
Check individual commits for more info.


suricactus:
@Rakanhf I have the feeling, but not the confindence that we should be ok without explicitly setting this to `True`. What do you think?

PR #1487, opened 2026-02-20
Manage CORS configuration via environment variables using django-cors-headers

t-book:
Hi there!

Previously, allowing cross-origin access to the API required manually adding `add_header` directives in the nginx configuration. This made it cumbersome for developers building custom frontends or apps against the QFieldCloud API.

This PR introduces django-cors-headers so that allowed origins can be managed entirely through environment variables:

```env
CORS_ALLOWED_ORIGINS=https://app.example.com,http://localhost:5173
CORS_ALLOW_CREDENTIALS=1
```

**Changes:**

- Added `django-cors-headers` as a dependency
- CORS middleware and settings configured in settings.py, driven by env vars
- CORS restricted to /api/ endpoints via CORS_URLS_REGEX
- Removed hardcoded `Access-Control-Allow-Origin` header from nginx (`/swagger.yaml`). https://docs.qfield.org should go into .env
- Passed `CORS_ALLOWED_ORIGINS` and `CORS_ALLOW_CREDENTIALS` through docker-compose.yml
- Updated .env.example with documentation

This allows developers to build their own frontends against the API by simply adding their origin to `CORS_ALLOWED_ORIGINS` — no nginx changes or container rebuilds required.

> Feature request https://ideas.qfield.org/qfieldcloud-feature-requests/p/enable-configurable-cors-via-environment-variables

PR #1493, opened 2026-02-28
chore: edit the way the boolean `CORS_ALLOW_CREDENTIALS` envvar is loaded

gounux:
Follow-up of #1487 :)

If somehow the new CORS-related env var is not there or commented, the default value would not be applied and the app would not boot, with this log :

```
CORS_ALLOW_CREDENTIALS = bool(int(os.environ.get("CORS_ALLOW_CREDENTIALS", 0)))
app-1  | ValueError: invalid literal for int() with base 10: ''
```

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -1000,4 +1000,4 @@ CORS_URLS_REGEX = r"^/api/.*$"
 
 # Whether to include credentials (cookies, authorization headers) in
 # cross-origin requests. Required when clients send auth tokens.
-CORS_ALLOW_CREDENTIALS = bool(int(os.environ.get("CORS_ALLOW_CREDENTIALS", 0)))
+CORS_ALLOW_CREDENTIALS = bool(int(os.environ.get("CORS_ALLOW_CREDENTIALS") or 0))


docker-app/qfieldcloud/settings.py
```diff
@@ -1000,4 +1000,4 @@ def before_send(event, hint):
 
 # Whether to include credentials (cookies, authorization headers) in
 # cross-origin requests. Required when clients send auth tokens.
-CORS_ALLOW_CREDENTIALS = bool(int(os.environ.get("CORS_ALLOW_CREDENTIALS", 0)))
+CORS_ALLOW_CREDENTIALS = bool(int(os.environ.get("CORS_ALLOW_CREDENTIALS") or 0))
```
suricactus, line 1003:
```suggestion
CORS_ALLOW_CREDENTIALS = bool(int(os.environ["CORS_ALLOW_CREDENTIALS"]))
```

> the app would not boot

This is a feature, not a bug. If we have a required envvar, we better not boot. See my favored approach in https://github.com/opengisch/QFieldCloud/pull/1494

gounux: Ok I see, LGTC then 🚀

PR #1494, opened 2026-03-01
feat: be stricter when expecting and reading envvars into `settings.py`

suricactus:
Currently a lot of variables are treated as optional while they are not, therefore they are accessed with `os.environ.get`. This gives false expectation that the system can properly run without setting those envvars in `.env.example` and does not throw if these variables are missing.

Now the code switches to direct dict lookups for these variables, so we get a `KeyError` as soon as such variable is not set.

Further validation can be applied later, which probably mimics what `django-configurations` or similar tools are doing, but for now it's enough to throw a `KeyError` when the `.env` file is not updated.

Note the failing linting CI has been fixed on master and it is no longer an issue.

### DIFF
--- a/docker-app/qfieldcloud/settings_utils.py
+++ b/docker-app/qfieldcloud/settings_utils.py
@@ -121,7 +121,15 @@ def get_socialaccount_providers_config() -> dict:
 def parse_string_to_list(input_str: str, *, delimiter: str) -> list[str]:
     assert delimiter
 
-    return [item.strip() for item in input_str.split(delimiter) if item.strip()]
+    items = []
+
+    for item in input_str.split(delimiter):
+        stripped_item = item.strip()
+
+        if stripped_item:
+            items.append(stripped_item)
+
+    return items
 
 
 def parse_string_to_bool(input_str: str) -> bool:


gounux:
Thanks @suricactus, I agree on that.  
Also, I suppose we should not specify some default values in the `.env.example` then, that could be misleading.

docker-app/qfieldcloud/settings.py
```diff
@@ -513,19 +512,15 @@ def before_send(event, hint):
 SOCIALACCOUNT_PROVIDERS = get_socialaccount_providers_config()
 
 # Third-party auth header names configuration
-QFIELDCLOUD_IDP_ID_HEADER_NAME = os.environ.get(
-    "QFIELDCLOUD_IDP_ID_HEADER_NAME", "X-QFC-IDP-ID"
-)
-QFIELDCLOUD_ID_TOKEN_HEADER_NAME = os.environ.get(
-    "QFIELDCLOUD_ID_TOKEN_HEADER_NAME", "X-QFC-ID-Token"
-)
+QFIELDCLOUD_IDP_ID_HEADER_NAME = "X-QFC-IDP-ID"
```
Rakanhf, line 515: Love it 💯 It won't change anyway, so why bother.

docker-app/qfieldcloud/settings.py
```diff
@@ -338,8 +341,8 @@
         f"Missing {STORAGES_PROJECT_DEFAULT_ATTACHMENTS_STORAGE=} from the `STORAGES` configuration, available storages: {STORAGES.keys()}"
     )
 
-STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED = bool(
-    int(os.environ.get("STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED") or 1)
+STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED = parse_string_to_bool(
+    os.environ.get("STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED") or "1"
```
Rakanhf, line 345: Why does this have a default value ?

I think it either should follow the new file updates. Or other variables should also have a default value if they are not critical.

docker-app/qfieldcloud/settings_utils.py
```diff
@@ -118,5 +118,16 @@ def get_socialaccount_providers_config() -> dict:
     return providers
 
 
-def parse_string_to_list(input_str: str) -> list[str]:
-    return [item.strip() for item in input_str.split(",") if item.strip()]
+def parse_string_to_list(input_str: str, *, delimiter: str) -> list[str]:
+    assert delimiter
+
+    return [item.strip() for item in input_str.split(delimiter) if item.strip()]
```
Rakanhf: Doesn't follow code style "inline for"

docker-app/qfieldcloud/settings.py
suricactus: It's because `STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED` is not required in the `.env` file.

docker-app/qfieldcloud/settings_utils.py
suricactus: I hoped it can go under the radar :satellite:

docker-app/qfieldcloud/settings.py
Rakanhf: Got it, I double checked it, and I remember that both `STORAGES_PROJECT_DEFAULT_ATTACHMENTS_STORAGE` and `STORAGE_PROJECT_DEFAULT_ATTACHMENTS_VERSIONED` are optional

PR #1498, opened 2026-03-03
feat: Add server info endpoint

suricactus:
For now we return information about the whitelabeling, but in the future we can add more and more, as `auth/providers` is already a good candidate to join.

```
$ curl https://localhost:8002/api/v1/server/info/ | jq
{
    "whitelabel": {
        "site_title": "QFieldCloud",
        "logo_navbar": "https://localhost/staticfiles/logo_sidetext_white.svg",
        "logo_main": "https://localhost/staticfiles/logo_undertext.svg",
        "favicon": "https://localhost/staticfiles/favicon.ico"
    }
}
```

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from typing import Any
 
+from django.templatetags.static import static
 from django.urls import reverse
 from django.utils.translation import gettext as _
 from django_stubs_ext import StrOrPromise
@@ -672,3 +673,25 @@ class TeamMemberSerializer(serializers.ModelSerializer):
     class Meta:
         model = TeamMember
         fields = ("member",)
+
+
+class WhitelabelSerializer(serializers.Serializer):
+    site_title = serializers.CharField(help_text="The title of the site.")
+    logo_navbar = serializers.URLField(help_text="The URL of the navigation bar logo.")
+    logo_main = serializers.URLField(help_text="The URL of the main logo.")
+    favicon = serializers.URLField(help_text="The URL of the favicon.")
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        request = self.context.get("request")
+        if request:
+            for key in ("logo_navbar", "logo_main", "favicon"):
+                value = data.get(key)
+                if value:
+                    data[key] = request.build_absolute_uri(static(value))
+
+        return data
+
+
+class ServerInfoSerializer(serializers.Serializer):
+    whitelabel = WhitelabelSerializer()
--- a/docker-app/qfieldcloud/core/tests/test_server_views.py
+++ b/docker-app/qfieldcloud/core/tests/test_server_views.py
@@ -1,0 +1,64 @@
+from django.core.cache import cache
+from django.templatetags.static import static
+from django.test import override_settings
+from rest_framework.test import APITransactionTestCase
+
+from qfieldcloud.core.whitelabel import get_whitelabel_settings
+
+
+class QfcTestCase(APITransactionTestCase):
+    def setUp(self):
+        # Clear the cache before each test to prevent cached responses
+        cache.clear()
+        self.whitelabel_settings = get_whitelabel_settings()
+        super().setUp()
+
+    def test_server_info_default_settings(self):
+        """Test that the endpoint returns the correct default whitelabel settings"""
+        response = self.client.get("/api/v1/server/info/")
+        data = response.json()
+
+        self.assertTrue("whitelabel" in data)
+
+        system_info = data["whitelabel"]
+
+        # Check title
+        self.assertEqual(
+            system_info["site_title"], self.whitelabel_settings["site_title"]
+        )
+
+        # Check that the URLs have been transformed to absolute URLs properly
+        self.assertTrue(system_info["logo_navbar"].startswith("http"))
+        self.assertTrue(
+            system_info["logo_navbar"].endswith(
+                static(self.whitelabel_settings["logo_navbar"])
+            )
+        )
+
+        self.assertTrue(system_info["logo_main"].startswith("http"))
+        self.assertTrue(
+            system_info["logo_main"].endswith(
+                static(self.whitelabel_settings["logo_main"])
+            )
+        )
+
+        self.assertTrue(system_info["favicon"].startswith("http"))
+        self.assertTrue(
+            system_info["favicon"].endswith(static(self.whitelabel_settings["favicon"]))
+        )
+
+    @override_settings(
+        WHITELABEL={
+            "site_title": "My Custom Title",
+        }
+    )
+    def test_server_info_custom_whitelabel_settings(self):
+        """Test that the endpoint correctly reflects custom WHITELABEL settings from settings.py"""
+        response = self.client.get("/api/v1/server/info/")
+        data = response.json()
+
+        self.assertTrue("whitelabel" in data)
+
+        system_info = data["whitelabel"]
+
+        self.assertEqual(system_info["site_title"], "My Custom Title")
--- a/docker-app/qfieldcloud/core/views/server_views.py
+++ b/docker-app/qfieldcloud/core/views/server_views.py
@@ -1,0 +1,28 @@
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_page
+from drf_spectacular.utils import extend_schema, extend_schema_view
+from qfieldcloud.core.serializers import ServerInfoSerializer
+from qfieldcloud.core.whitelabel import get_whitelabel_settings
+from rest_framework import status, views
+from rest_framework.permissions import AllowAny
+from rest_framework.request import Request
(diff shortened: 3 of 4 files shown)


docker-app/qfieldcloud/core/tests/test_server_views.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertEqual(system_info["site_title"], DEFAULT_WHITELABEL["site_title"])
+
+        # Check that the URLs have been transformed to absolute URLs properly
+        self.assertTrue(system_info["logo_navbar"].startswith("http"))
+        self.assertIn(
+            static(DEFAULT_WHITELABEL["logo_navbar"]), system_info["logo_navbar"]
+        )
+
+        self.assertTrue(system_info["logo_main"].startswith("http"))
+        self.assertIn(static(DEFAULT_WHITELABEL["logo_main"]), system_info["logo_main"])
```
suricactus:
```suggestion
        self.assertTrue(system_info["logo_main"].endswith(static(DEFAULT_WHITELABEL["logo_main"])))
```
I think this should world, no? You may need a trailing slash.

suricactus: Second comment is the `DEFAULT_WHITELABEL`. What if we use `settings.py` that overwrites those? I would use the "official" API for getting those values with `get_whitelabel_settings`.

docker-app/qfieldcloud/core/tests/test_server_views.py
```diff
@@ -0,0 +1,47 @@
...
+    def setUp(self):
+        # Clear the cache before each test to prevent cached responses
+        cache.clear()
+        super().setUp()
+
+    def test_server_info_default_settings(self):
+        """Test that the endpoint returns the correct default whitelabel settings"""
+        response = self.client.get("/api/v1/server/info/")
+        data = response.json()
+        system_info = data["system"]
```
suricactus: Why do we need the parasitic `system` level here? I don't see any additional value on nesting here, the flatter, the better, as we don't wonder "what if that key does not exist" in clients.

docker-app/qfieldcloud/core/tests/test_server_views.py
Rakanhf: Sure nice catch, and about the `.endswith(static(DEFAULT_WHITELABEL["logo_main"]))` it will technically work  but I'm not sure if the tests will be stable since sometimes when running locally I got errors because  static files added some prefixes at the end of the file,  I'm not sure why it's happening but it might be a miss-configuration from my initial tests iteration

docker-app/qfieldcloud/core/tests/test_server_views.py
Rakanhf: Here since we may add other things in the future I decided to change the `whitelabel` dict to be `system` instead. to make it more general, as we can add more to it in the future and should not be just linked to whitelable. 

and also just a way to organize the response structure a bit more. you can also check [PR-1511](https://github.com/opengisch/QFieldCloud/pull/1511)

lukasgraf: I would generally agree with "not unnecessarily nested", but one level of namespacing I think we need. I wouldn't cram everything in the top level namespace. That will eventually lead to naming conflicts from different providers that want to use the same name, and having to make long, ugly key names just to make them unique and group them together logically. And when you then start prefixing keys with things like `auth_`, congratulations, you've just invented namespaces.

However, I would really prefer naming things just after what they are, instead of using meaningless, generic terms. If it's whitelabel settings, name them `whitelabel`, instead of `system`.

If I were to implement this, I would have one `ServerInfoSerializer` that manages the top level namespace. And it essentially does nothing else than collect the other serializers, and place them in their own namespace. E.g. there's a `WhitelabelSerializer` that goes into the `whitelabel` namespace, an `AuthSerializer` that goes into `auth`, etc...

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -672,3 +672,13 @@ def to_internal_value(self, data: dict[str, Any]) -> dict[str, Any]:
     class Meta:
         model = TeamMember
         fields = ("member",)
+
+
+class ServerInfoSerializer(serializers.Serializer):
+    class SystemSerializer(serializers.Serializer):
```
lukasgraf: As discussed this morning, I wouldn't nest this serializer. It should be importable separately, and then it can also be tested completely in isolation by a trivial unit test, that doesn't even need to do network requests using a client.

docker-app/qfieldcloud/core/tests/test_server_views.py
suricactus: > However, I would really prefer naming things just after what they are, instead of using meaningless, generic terms. If it's whitelabel settings, name them whitelabel, instead of system.

I thought we put all of these in a `whitelabel` namespace already. I mistakenly read it is `system.whitelabel.logo`.

Rakanhf: I renamed it to be "whitelabel"

docker-app/qfieldcloud/core/tests/test_server_views.py
```diff
@@ -0,0 +1,58 @@
...
+        # Clear the cache before each test to prevent cached responses
+        cache.clear()
+        self.whitelabel_settings = get_whitelabel_settings()
+        super().setUp()
+
+    def test_server_info_default_settings(self):
+        """Test that the endpoint returns the correct default whitelabel settings"""
+        response = self.client.get("/api/v1/server/info/")
+        data = response.json()
+        system_info = data["whitelabel"]
```
suricactus, line 23: Don't assume `whitelabel` is present in the answer IMO.

docker-app/qfieldcloud/core/views/server_views.py
```diff
@@ -0,0 +1,34 @@
...
+    serializer_class = ServerInfoSerializer
+
+    @method_decorator(cache_page(60))
+    def get(self, request: Request) -> Response:
+        whitelabel_settings = get_whitelabel_settings()
+
+        for key in ["logo_navbar", "logo_main", "favicon"]:
+            value = whitelabel_settings.get(key)
+            if value:
+                whitelabel_settings[key] = request.build_absolute_uri(static(value))
```
suricactus: If we have a serializer, wouldn't it be more natural to move this to the serializer class? Ideally we always want the absolute URL for any response.

docker-app/qfieldcloud/core/tests/test_server_views.py
Rakanhf: But its always in the response, can you let me know why you think that it shouldn't be there ?

suricactus: Because it's a test, and we test if the whitelabel is present :)

docker-app/qfieldcloud/core/tests/test_server_views.py
suricactus: Happy to look at this together.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -672,3 +673,25 @@ def to_internal_value(self, data: dict[str, Any]) -> dict[str, Any]:
     class Meta:
         model = TeamMember
         fields = ("member",)
+
+
+class WhitelabelSerializer(serializers.Serializer):
+    site_title = serializers.CharField(help_text="The title of the site.")
+    logo_navbar = serializers.URLField(help_text="The URL of the navbar logo.")
```
nirvn: Random Q: do we have a preference on whether to use diminutives like navbar for navigation bar (and below favicon for favorite icon) in the help_text?

suricactus: I would prefer the full name, the few saved bytes is something we can afford.

nirvn: I _personally_ prefer the full name for help text too.

docker-app/qfieldcloud/core/tests/test_server_views.py
suricactus:
```suggestion
		self.assertTrue("whitelabel" in data)

        system_info = data["whitelabel"]
```

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -677,9 +677,9 @@ class Meta:
 
 class WhitelabelSerializer(serializers.Serializer):
     site_title = serializers.CharField(help_text="The title of the site.")
-    logo_navbar = serializers.URLField(help_text="The URL of the navbar logo.")
+    logo_navbar = serializers.URLField(help_text="The URL of the navigation bar logo.")
     logo_main = serializers.URLField(help_text="The URL of the main logo.")
-    favicon = serializers.URLField(help_text="The URL of the favicon.")
+    favicon = serializers.URLField(help_text="The URL of the favorite icon.")
```
suricactus: I think you overdid it a bit, the term is favicon :)

PR #1501, opened 2026-03-09
chore: block unstyled allauth pages

Rakanhf:
There are a couple of pages that any qfc user can access which is the default pages related to allauth. These urls needs to be restricted, since it's un-styled and also redundant and can be found in account settings

### DIFF
--- a/docker-app/qfieldcloud/urls.py
+++ b/docker-app/qfieldcloud/urls.py
@@ -33,6 +33,7 @@ from rest_framework import permissions
 
 from qfieldcloud.authentication import views as auth_views
 from qfieldcloud.core.admin import qfc_admin_site
+from qfieldcloud.core.utils2.view_utils import blocked_view
 from qfieldcloud.core.views.redirect_views import redirect_to_admin_project_view
 from qfieldcloud.filestorage.views import (
     compatibility_file_crud_view,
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/urls.py
```diff
@@ -59,6 +60,10 @@ def wrapper(request, *args, **kwargs):
     return wrapper
 
 
+def blocked_view(request, *args, **kwargs):
+    raise Http404
```
suricactus: Would you mind moving this to `core.utils2.view_utils`?

Rakanhf: I'm not sure if i get `utils2.view_utils` correctly since I don't think there is a file related to this. IMO keeping it in `urls.py` will be more contained, But I can move it to `utils.py`

suricactus: Please do create the new file. I see `blocked_view` used in other `urls.py` files, therefore it's more of a utils function/class.

docker-app/qfieldcloud/core/utils.py
```diff
@@ -614,3 +615,7 @@ def get_file_storage_choices() -> list[tuple[str, str]]:
     """
     storages = list(settings.STORAGES.keys())[:-1]
     return [(storage, storage) for storage in storages]
+
+
+def blocked_view(request, *args, **kwargs):
+    raise Http404
```
suricactus: Would you mind creating a new file `core.utils2.view_utils` as originally requested, because we want to eventually retire this whole file, it's a conglomerate of unrelated things that should eventually go to their own file and this file should be deleted.

PR #1504, opened 2026-03-11
fix: edit unique constraints when there are nullable fields

gounux:
This PR intends to :

- make use of [`nulls_distinct`](https://docs.djangoproject.com/en/5.2/ref/models/constraints/#nulls-distinct) on the `UniqueConstraint`s applied to `File` and `Secret` models, available with Django 5+ on PostgreSQL 15+ backends,
- replace `unique_together` with `UniqueConstraint`,


docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
```diff
@@ -0,0 +1,30 @@
...
+            "0096_remove_secret_secret_project_type_name_assigned_to_uniq_and_more",
+        ),
+        ("filestorage", "0005_fileversion_file_storage"),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name="file",
+            unique_together=set(),
+        ),
```
suricactus, line 19: What is this? Never seen it before? Shall we have empty `set()`?

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -40,8 +40,13 @@ def with_type_project(self) -> FileQueryset:
 
 class File(models.Model):
     class Meta:
-        unique_together = [
-            ("project", "name", "file_type", "package_job"),
+        constraints = [
+            models.UniqueConstraint(
+                fields=["project", "name", "file_type", "package_job"],
+                name="unique_file_per_project_name_type_packagejob",
+                deferrable=models.Deferrable.DEFERRED,
```
suricactus: Is it required to make it deferred? A comment might help.

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
gounux: Before this PR, the "constraint" was defined like this :

```python
unique_together = [
            ("project", "name", "file_type", "package_job"),
]
```

Which is now redundant with the introduced `UniqueConstraint`,  
The migration part you mention is the automatic django's `makemigrations` result, maybe it's worth removing this part of the migration,

docker-app/qfieldcloud/filestorage/models.py
gounux: > Is it required to make it deferred?

Maybe not, [doc](https://docs.djangoproject.com/en/5.2/ref/models/constraints/#deferrable) says "A deferred constraint will not be enforced until the end of the transaction. An immediate constraint will be enforced immediately after every command.", so, maybe enforcing it earlier might be a good option.

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
suricactus: Ah, I guess it's something from Django migration internals. If it works, it works :)

docker-app/qfieldcloud/filestorage/models.py
suricactus: What was your motivation to make it deferred?

gounux: > What was your motivation to make it deferred?

A kind recommendation expressed by a senior of mine :-)

Explicitly set to `IMMEDIATE` in https://github.com/opengisch/QFieldCloud/pull/1504/changes/8475d5cec5ec428f1e21710234274f6e88aecb6e

suricactus: But is it required to explicitly set it though?

gounux: That's right, explicitly defining the default value is maybe too verbose,  
Addressed in https://github.com/opengisch/QFieldCloud/pull/1504/changes/ce4619459fb9ad56bea3322fa1aa37d15ef53078

PR #1506, opened 2026-03-11
feat: Add externally accessible API for user creation POST /users

spwoodcock:
### Why did I make this PR?

- I have been developing an application that uses QFieldCloud underneath to manage projects, then the QField mobile app for mapping.
- As part of the project creation process, I want to create a new user account that I can give to the user of my app.
- This way they don't need to be given QFC admin credentials linked to my app, or pre-configured accounts - we can do it programmatically.
- Once the user has an account, they can log in and manage the project as needed (adding new users to it mainly).

> [!NOTE]
> I only realised very recently that the UI at app.qfield.cloud isn't part of the OSS self-host solution (which is perfectly fine), but that does mean this API is mostly useful for creators of external apps, that wish to use the official app.qfield.cloud instance (so they can log in with the newly created user and manage).
>
> As a result, I will create a tiny wrapper around qfieldcloud-sdk-python as a backend to manage users, with a small frontend to manage this. The user logs in and can add new users to the project - that's about it.

### What's included

- Adds a simple POST request option to the existing `/users` route, that requires `is_staff=true` (instance admin) for the user.
- Standard users probably shouldn't be allowed to create new users (at least for now / until discussed how).

Request body:

```json
 {
    "username": "newmapper",
    "password": "securepass123",
    "email": "newmapper@example.com"
  }
```

Success response (201 Created):

```json
  {
    "username": "newmapper",
    "type": 1, # a 'person'
    "full_name": "",
    "avatar_url": null,
    "username_display": "newmapper"
  }
```

Error responses:

  - 400 Bad Request: invalid payload (e.g. bad username format / too short / user already exists)
  - 401 Unauthorized: no auth token
  - 403 Forbidden: authenticated but not staff

### How did I verify

- Added a bunch of extra tests.
- Confirmed all tests pass `docker compose run --rm -e ENVIRONMENT=test app python manage.py test --keepdb -v2 --exclude-tag="flaky" qfieldcloud.core.tests.test_user.CreatePersonAPITestCase`
- Ran all pre-commit hooks.
- Non-breaking change - additive only).
- No db migration needed.

Disclaimer: I used Opus 4.6 to better target where changes should be made & for assisting with test creation.

Also, I was writing this code to solve **my** problems and totally understand if it's not in the roadmap to be accepted. I thought I may as well PR it, as it was written anyway!

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -283,6 +283,14 @@ class CreateUserSerializer(serializers.ModelSerializer):
             "last_name": {"required": False, "default": ""},
         }
 
+    def validate_email(self, value):
+        if Person.objects.filter(email__iexact=value).exists():
+            raise serializers.ValidationError(
+                _("This email is already taken by another user!")
+            )
+
+        return value
+
     def create(self, validated_data):
         return Person.objects.create_user(
             has_accepted_tos=False,
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -19,6 +19,7 @@ from qfieldcloud.core.models import (
     Organization,
     OrganizationMember,
     PackageJob,
+    Person,
     ProcessProjectfileJob,
     Project,
     ProjectCollaborator,
@@ -266,6 +267,36 @@ class CompleteUserSerializer(serializers.ModelSerializer):
             "last_name",
         )
         read_only_fields = ("full_name", "avatar_url")
+
+
+class CreateUserSerializer(serializers.Serializer):
+    """Serializer for programmatic Person account creation.
+
+    Intentionally does *not* validate username uniqueness here so that the view
+    can return HTTP 409 Conflict (rather than 400) when the username is taken.
+    """
+
+    username = serializers.RegexField(
+        r"^[-a-zA-Z0-9_]+$",
+        max_length=150,
+        min_length=3,
+        error_messages={
+            "invalid": _(
+                "Enter a valid username. This value may contain only letters, "
+                "numbers, and -/_ characters."
+            )
+        },
+    )
+    password = serializers.CharField(write_only=True, style={"input_type": "password"})
+    email = serializers.EmailField(required=True)
+
+    def create(self, validated_data):
+        return Person.objects.create_user(
+            username=validated_data["username"],
+            password=validated_data["password"],
+            email=validated_data["email"],
+            has_accepted_tos=True,
+        )
 
 
 class PublicInfoUserSerializer(serializers.ModelSerializer):
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -276,25 +276,22 @@ class CreateUserSerializer(serializers.Serializer):
     can return HTTP 409 Conflict (rather than 400) when the username is taken.
     """
 
-    username = serializers.RegexField(
-        r"^[-a-zA-Z0-9_]+$",
+    username = serializers.CharField(
         max_length=150,
-        min_length=3,
-        error_messages={
-            "invalid": _(
-                "Enter a valid username. This value may contain only letters, "
-                "numbers, and -/_ characters."
-            )
-        },
+        validators=User._meta.get_field("username").validators,
     )
     password = serializers.CharField(write_only=True, style={"input_type": "password"})
     email = serializers.EmailField(required=True)
+    first_name = serializers.CharField(max_length=150, required=False, default="")
+    last_name = serializers.CharField(max_length=150, required=False, default="")
 
     def create(self, validated_data):
         return Person.objects.create_user(
             username=validated_data["username"],
             password=validated_data["password"],
             email=validated_data["email"],
+            first_name=validated_data["first_name"],
+            last_name=validated_data["last_name"],
             has_accepted_tos=False,
         )
 
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -295,7 +295,7 @@ class CreateUserSerializer(serializers.Serializer):
             username=validated_data["username"],
             password=validated_data["password"],
             email=validated_data["email"],
-            has_accepted_tos=True,
+            has_accepted_tos=False,
         )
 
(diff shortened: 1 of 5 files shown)


docker-app/qfieldcloud/core/serializers.py
```diff
@@ -220,6 +221,36 @@ class Meta:
...
+    )
+    password = serializers.CharField(write_only=True, style={"input_type": "password"})
+    email = serializers.EmailField(required=True)
+
+    def create(self, validated_data):
+        return Person.objects.create_user(
+            username=validated_data["username"],
+            password=validated_data["password"],
+            email=validated_data["email"],
+            has_accepted_tos=True,
```
gounux: Seems complicated to make a new user automatically created accept the ToS, waht about putting this new value  to `False`, and then find a mechanism / popup to make the new user aware of the ToS during first connexion on QFieldCloud ?

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -220,6 +221,36 @@ class Meta:
...
+        r"^[-a-zA-Z0-9_]+$",
+        max_length=150,
+        min_length=3,
+        error_messages={
+            "invalid": _(
+                "Enter a valid username. This value may contain only letters, "
+                "numbers, and -/_ characters."
+            )
+        },
+    )
```
gounux: Here, maybe we could stick with the model definition ?

Can be found here : https://github.com/opengisch/QFieldCloud/blob/ae3934ada6e695df1f8ffa5b360069cb46018635/docker-app/qfieldcloud/core/models.py#L265-L280

So if possible IMO it could be a good idea to keep the same errors messages and same validation regexp(s), maybe it's possible to not redeclare the regexps and get them directly from the model's field definition ? ping @Rakanhf ;-)

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -220,6 +221,36 @@ class Meta:
...
+    password = serializers.CharField(write_only=True, style={"input_type": "password"})
+    email = serializers.EmailField(required=True)
```
gounux: Should we also allow the possibility to specify first name / last name ?

Maybe nullable, no hard opinion, just wondering if it might help in some case,

gounux (commented):
Thanks a lot @spwoodcock !

Here is a first review with some remarks and questions :-)

docker-app/qfieldcloud/core/serializers.py
suricactus: IMO we should be able to define all the fields that one defined on sign up.

docker-app/qfieldcloud/core/serializers.py
spwoodcock: Thanks for flagging!

Does this work `validators=User._meta.get_field("username").validators`, to keep things nicely in sync?

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -220,6 +221,33 @@ class Meta:
         read_only_fields = ("full_name", "avatar_url")
 
 
+class CreateUserSerializer(serializers.Serializer):
```
suricactus: Is there a specific reason why you didn't use `ModelSerializer`?

spwoodcock: The docstring just below on 227 explains it:

```
Intentionally does *not* validate username uniqueness here so that the view
can return HTTP 409 Conflict (rather than 400) when the username is taken.
```

I thought a 409 would be more 'correct' as a HTTP response, but reviewing a bit more I do see that 400 responses are used for other duplicate resources (e.g. ProjectAlreadyExistsError), so perhaps we should keep the 400 for better consistency with the rest?

suricactus: You get the automated validation as defined on the model directly for free.

The consistency is also there, as you stated.

So please do change to model serializer.

spwoodcock: Sure! I'll need no override .create anyway to ensure the password is hashed (I think), but should be all good - will come back to this in the evening 👍

docker-app/qfieldcloud/core/views/users_views.py
```diff
@@ -5,26 +5,31 @@
...
+from rest_framework import generics, permissions, status
 from rest_framework.response import Response
 
 User = get_user_model()
 
 
-class ListUsersViewPermissions(permissions.BasePermission):
+class ListCreateUsersViewPermissions(permissions.BasePermission):
     def has_permission(self, request, view):
+        if request.method == "POST":
```
suricactus, line 20:
```suggestion
        if view.action == "create":
```

In newer code we want to be a bit more precise about the `action` not only the method as it might not uniquely identify what we are trying to do.

suricactus (commented):
Code looks good, thanks for the tests too. Single comment here.

suricactus:
@spwoodcock I saw your commit directly on GH, did you try locally? If you confirm yes, then this and the SDK PR can go ahead. :)

docker-app/qfieldcloud/core/views/users_views.py
suricactus: I lied to you, there is no `action` in this base class.

spwoodcock: Oh damn, sorry! I just went ahead and changed, but haven't had a chance to setup local again for testing. I'll be back on Monday to test this out! 🙏

spwoodcock:
Sorry I didn't actually test the suggestion and just commited 😅 

Reverted back to previous =) hopefully all set

suricactus:
@spwoodcock one last obstacle remaining in front of us:

```
FAIL: test_duplicate_email_returns_400 (qfieldcloud.core.tests.test_user.CreatePersonAPITestCase)
----------------------------------------------------------------------
  File "/usr/src/app/qfieldcloud/core/tests/test_user.py", line 490, in test_duplicate_email_returns_400
    self.assertIn("email", response.data)
AssertionError: 'email' not found in {'code': 'api_error', 'message': 'API Error'}

----------------------------------------------------------------------```

spwoodcock:
Fixing now!

spwoodcock:
Done =)

suricactus (approved):

suricactus:
Thank you @spwoodcock  !

PR #1507, opened 2026-03-12
ARM64 Architecture Compatibility for Docker

manylon:
When the host machine uses ARM64 architecture, Docker will build images in the same architecture by default. However QGIS is not available for ARM64 in the Debian repository and we need to specify the platform in the QGIS Dockerfile.

Other images that do not provide ARM64 builds are:
- postgis/postgis
- vishnunair/docker-mkcert
- bytemark/webdav:2.4

if forcing the platform is not suitable, an alternative is to document this requirement in the README for developers using ARM64 machines.

### DIFF
--- a/docker-compose.override.standalone.yml
+++ b/docker-compose.override.standalone.yml
@@ -2,6 +2,7 @@ services:
 
   db:
     image: postgis/postgis:${POSTGIS_IMAGE_VERSION}
+    platform: linux/amd64
     restart: unless-stopped
     environment:
       POSTGRES_DB: ${POSTGRES_DB}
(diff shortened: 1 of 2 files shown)


spwoodcock:
It's pretty trivial to make the QFC images ARM-compatible in the workflow here: https://github.com/opengisch/QFieldCloud/blob/master/.github/workflows/build_and_push.yml

I'm happy to make a PR for that if helpful 👍 

But as you say, the additional components are a problem to make the full stack for ARM. I wouldn't count on the postgis image being updated anytime soon. I built PostGIS images myself instead & there are loads of alternative sources out there (CloudNativePG, etc).

PR #1508, opened 2026-03-16
feat: add project seed and xls file to project creation api

Rakanhf:
Currently we have the xlsform creation only via the web UI. We should be able to do the same via API, so remote systems can easily create XLSForms based projects easily.

Added the ability to create a project from a custom seed / xls form file, using the existing `/projects` endpoint
Which will be easy to maintain and also be compatible with future changes, where it can now handle both `Json` and `multipart/form-data` requests 
Depending on the use case

e.g to create a new `Project` with a custom seed & xls form file we can simply use a `multipart/form-data` request
or we can create a new `Project` with only custom seed **without** xls file using Json as follow :

```json
# POST /api/v1/projects/
{
    "name": "my-demo-seed-project",
    "seed": {
        "basemap_provider": "osm",
        "basemap_style": "standard",
        "extent": "800000,5900000,1100000,6100000"
    }
}
```

- Add `seed` and `xlsform_file` fields to the `ProjectSerializer`.
- Implement validation to ensure that an XLSForm file cannot be uploaded without seed data.
- Add a new utility `utils2.seed.py` for building seed data and handling basemap configurations.
- Update `perform_create` method in `ProjectViewSet` to handle seed data and create associated `ProjectSeed` and `Job` entries.
- Add tests for various project creation scenarios involving seed data and XLSForm uploads.
- Drop unnecessary  `create_project` job exposure to `/jobs` endpoint and it's `CreateProjectJobSerializer`

related PR : [PR-837](https://github.com/opengisch/QFieldCloud-Nyuki/pull/837)

### DIFF
--- a/docker-app/qfieldcloud/core/serializers.py
+++ b/docker-app/qfieldcloud/core/serializers.py
@@ -168,8 +168,10 @@ class ProjectSerializer(serializers.ModelSerializer):
         return data
 
     def create(self, validated_data):
-        self._seed_data = validated_data.pop("seed", None)
-        self._xlsform_file = validated_data.pop("xlsform_file", None)
+        # remove seed and xlsform_file from validated_data to avoid errors
+        # since these are not part of the model and will be set in the view
+        validated_data.pop("seed", None)
+        validated_data.pop("xlsform_file", None)
         return super().create(validated_data)
 
     def get_needs_repackaging(self, obj: Project) -> bool:
--- a/docker-app/qfieldcloud/core/views/projects_views.py
+++ b/docker-app/qfieldcloud/core/views/projects_views.py
@@ -172,11 +172,11 @@ class ProjectViewSet(viewsets.ModelViewSet):
     def perform_create(self, serializer: ProjectSerializer) -> None:
         super().perform_create(serializer)
 
-        seed_data = getattr(serializer, "_seed_data", None)
+        seed_data = serializer.validated_data.get("seed", None)
         if not seed_data:
             return
 
-        xlsform_file = getattr(serializer, "_xlsform_file", None)
+        xlsform_file = serializer.validated_data.get("xlsform_file", None)
         project = serializer.instance
 
         basemaps, extent, xlsform_config = project_seed.build_seed_data(
@@ -203,6 +203,7 @@ class ProjectViewSet(viewsets.ModelViewSet):
             type=Job.Type.CREATE_PROJECT,
             created_by=self.request.user,
         )
+        print("Job created")
 
     @transaction.atomic
     def perform_update(self, serializer: ProjectSerializer) -> None:


docker-app/qfieldcloud/core/utils2/project_seed.py
suricactus, line 1: I would rename this to `project_seed`. It's way to generic otherwise.

docker-app/qfieldcloud/core/views/jobs_views.py
```diff
@@ -60,8 +60,6 @@ def get_serializer_by_job_type(self, job_type, *args, **kwargs):
             return serializers.PackageJobSerializer(*args, **kwargs)
         elif job_type == Job.Type.PROCESS_PROJECTFILE:
             return serializers.ProcessProjectfileJobSerializer(*args, **kwargs)
-        elif job_type == Job.Type.CREATE_PROJECT:
-            return serializers.CreateProjectJobSerializer(*args, **kwargs)
```
suricactus, line 104: Is it a problem if we keep the possibility to trigger the job? No strong opinion, just asking.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -241,7 +242,7 @@ def test_create_project_from_xlsform(self):
 
         ProjectSeed.objects.create(
             project=self.p1,
-            extent=Polygon.from_bbox((-20000000, 20000000, -20000000, 20000000)),
+            extent=Polygon.from_bbox(seed.DEFAULT_EXTENT),
```
suricactus:
```suggestion
            extent=Polygon.from_bbox(seed.DEFAULT_PROJECT_EXTENT),
```

It's a bit more readable name IMO.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -125,6 +127,42 @@ def get_queryset(self):
 
         return projects
 
+    @transaction.atomic
+    def perform_create(self, serializer: ProjectSerializer) -> None:
+        super().perform_create(serializer)
+
+        seed_data = getattr(serializer, "_seed_data", None)
```
suricactus: hm, I feel a bit weird using `getattr` and underscored property. Can we use `serializer.validated_data["seed"]` instead?

I see the `_seed_data` is something under our control, but I don't see where it is used besides here?

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -571,12 +600,6 @@ class Meta(JobMixin.Meta):
         allow_parallel_jobs = True
 
 
-class CreateProjectJobSerializer(JobMixin, serializers.ModelSerializer):
-    class Meta(JobMixin.Meta):
-        model = Job
-        allow_parallel_jobs = True
-
-
```
suricactus, line 587: I think we still need to keep it, as if you call `jobs/<project-id>`, we need to seriealize the result. See `JobViewSet`.

docker-app/qfieldcloud/core/views/jobs_views.py
Rakanhf: IMO I don't think it's necessary, since we won't use the endpoint anyway, and if we want to use it later, we can add it, but for now, it's dead code

docker-app/qfieldcloud/core/serializers.py
Rakanhf: Since we removed the `elif job_type == Job.Type.CREATE_PROJECT:` from the comment above, this code basically becomes dead and it's not used anywhere, so it's not needed for now, just to make things cleaner.

docker-app/qfieldcloud/core/views/projects_views.py
Rakanhf: Here since we don't want to include the `seed_data` and `xlsform_file` in the serializer creation we pop it into a private variable so `return super().create(validated_data)`  doesn't raise errors, since they are not a `Project` model fields, and also serves as a private data store so we can use it in step 2 in the view which is creating the `ProjectSeed`

docker-app/qfieldcloud/core/views/jobs_views.py
suricactus: We do use the endpoint in the SDK. 

Also since it is exposed, people may use it.

Rakanhf: Yes they technically can use it, but what will be the benefit of them using it ?

docker-app/qfieldcloud/core/views/projects_views.py
suricactus: I am not convinced this is the right way to do it, let's discuss it next week. 

It just feels weird to have private property (prefixed with `_`) and use from the outside.

docker-app/qfieldcloud/core/views/jobs_views.py
suricactus: Please do keep it. It is used for serialization of the response. If you list the jobs, you probably also want to see what seed was given for example.

Just a few lines below we have: `            raise NotImplementedError(f'Unknown job type "{job_type}"')`, I guess we don't want to break the API for anyone that was unlucky enough to create a project with a seed.

docker-app/qfieldcloud/core/serializers.py
suricactus: Closing, see https://github.com/opengisch/QFieldCloud/pull/1508#discussion_r2968376394

docker-app/qfieldcloud/core/views/jobs_views.py
Rakanhf: Reverted the changes as discussed internally

docker-app/qfieldcloud/core/views/projects_views.py
suricactus:
```suggestion
        seed_data = serializer.validated_data.get("seed_data", None)
```

This is supposed to work, I missed initially that the `seed_data` is optional.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -142,8 +144,33 @@ def validate(self, data):
...
+                        ]
+                    }
+                )
+
         return data
 
+    def create(self, validated_data):
+        self._seed_data = validated_data.pop("seed", None)
+        self._xlsform_file = validated_data.pop("xlsform_file", None)
+        return super().create(validated_data)
```
suricactus, line 175: The good thing is that `validated_data` is hard top-level copy of `self.validated_data`, which means that we can remove keys from the `validated_data` without affecting the `self.validated_data`. See https://github.com/encode/django-rest-framework/blob/main/rest_framework/serializers.py#L202-L213

PR #1509, opened 2026-03-17
DRY up docker-compose env vars

manylon:
Found two TODOs in docker-compose.yml and addressed both of them.

The volumes `static_volume`, `media_volume` and the shared environment variables are extracted into YAML anchors so they are defined once and reused across app and worker_wrapper.

Edit: shared volumes part is moved to #1516 as suggested by @suricactus

NOTE: I am not a YAML expert! Happy to get feedback on the anchor approach if there's a cleaner way.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -171,6 +171,7 @@ services:
       - transformation_grids:/transformation_grids
       - /var/run/docker.sock:/var/run/docker.sock
       - custom_ca_certificates:/etc/ssl/certs/:ro
+      - ${TMP_DIRECTORY}:/tmp
     environment:
       <<: *django-env
       TMP_DIRECTORY: ${TMP_DIRECTORY}


docker-compose.yml
```diff
@@ -5,6 +5,66 @@ x-logging: &default-logging
     max-file: "10"
     tag: "{{.ImageName}}|{{.Name}}|{{.ImageFullID}}|{{.FullID}}"
 
+x-volume-static: &volume-static static_volume:/usr/src/app/staticfiles
+x-volume-media: &volume-media media_volume:/usr/src/app/mediafiles
+
+x-django-env: &django-env
```
gounux: What about `x-qfieldcloud-env: &qfieldcloud-env` here ?  
Since the "django" one would also be used by the worker wrapper, no hard opinion though.

docker-compose.yml
```diff
@@ -164,10 +170,11 @@ services:
       target: worker_wrapper_runtime
     command: python manage.py dequeue
     user: root # TODO change me to least privileged docker-capable user on the host (/!\ docker users!=hosts users, use UID rather than username)
+    environment:
+      <<: [*django-env, *worker-wrapper-env]
```
gounux: Wondering if here we should maybe find a way to make the worker wrapper receive only the env vars it needs ?

Could something like the following make sense and work ?

```yaml
x-worker-wrapper-env: &worker-wrapper-env
  TMP_DIRECTORY: ${TMP_DIRECTORY}
  QFIELDCLOUD_HOST: ${QFIELDCLOUD_HOST}
  QFIELDCLOUD_ADMIN_URI: ${QFIELDCLOUD_ADMIN_URI}
  QFIELDCLOUD_WORKER_QFIELDCLOUD_URL: ${QFIELDCLOUD_WORKER_QFIELDCLOUD_URL}
  ...

x-qfieldcloud-env: &qfieldcloud-env
  <<: *worker-wrapper-env
  DJANGO_ALLOWED_HOSTS: ${DJANGO_ALLOWED_HOSTS}
  DJANGO_USE_X_FORWARDED_HOST: ${DJANGO_USE_X_FORWARDED_HOST}
  DJANGO_SETTINGS_MODULE: ${DJANGO_SETTINGS_MODULE}
  ...
```

YAML expert needed ;-)

docker-compose.yml
suricactus: TBH I slightly prefer the `django-env` or `common-django-env` for now, as the common denominator between the two services is Django.

docker-compose.yml
```diff
@@ -5,6 +5,66 @@ x-logging: &default-logging
     max-file: "10"
     tag: "{{.ImageName}}|{{.Name}}|{{.ImageFullID}}|{{.FullID}}"
 
+x-volume-static: &volume-static static_volume:/usr/src/app/staticfiles
+x-volume-media: &volume-media media_volume:/usr/src/app/mediafiles
```
suricactus: @manylon would you mind if we handle the volumes in a follow-up PR? Let's focus only on the envvars in this PR.

docker-compose.yml
suricactus: I wonder why we would even define `worker-wrapper-env`? Can't we define here:

```
environment:
  <<: [*django-env]
  VAR: VAL
  ...
```

docker-compose.yml
manylon: No strong opinion here. I followed the existing name from `app: &default-django`

docker-compose.yml
manylon: I prefere to define env variable anchors at the top of the file. This keeps the service definitions cleaner and easier to manage the variables without scanning through service configurations.

I also considered splitting the shared variables into three anchors: core-env, worker-wrapper-env and django-env, because app and worker_wrapper share the same settings.py file, but have different needs.
```yaml
x-core-env: &core-env
  ...

x-worker-wrapper-env: &worker-wrapper-env
  <<: *core-env
  ...

x-django-env: &django-env
  <<: *core-env
  ...
```

Some vars are read unconditionally via os.environ["VAR"], so LOGGER_SOURCE env var could help conditionally load only what each service requires.
```python
IS_WORKER_SERVICE = LOGGER_SOURCE == "worker_wrapper"

if IS_WORKER_SERVICE:
    # Name of the qgis docker image used as a worker by worker_wrapper
    QFIELDCLOUD_QGIS_IMAGE_NAME = os.environ["QFIELDCLOUD_QGIS_IMAGE_NAME"]
```

Thanks for the feedback @gounux and @suricactus! If this direction doesn't align with the project's goals, I'm happy to revert.

suricactus: > @gounux  Wondering if here we should maybe find a way to make the worker wrapper receive only the env vars it needs ?
> @manylon Some vars are read unconditionally via os.environ["VAR"], so LOGGER_SOURCE env var could help conditionally load only what each service requires.

Let us not do this refactoring in `settings.py` for now. There are questions we need to answer that we didn't even asked ourselves yet.

> I prefere to define env variable anchors at the top of the file. This keeps the service definitions cleaner and easier to manage the variables without scanning through service configurations.

I am not a big fan of moving things away from where they are being used. Would it be possible to kill `x-worker-wrapper-env` and use the proposed:

```
environment:
  <<: [*django-env]
  VAR: VAL
  ...
```

PR #1516, opened 2026-03-21
Refactor: Extract shared volumes into YAML anchors

manylon:
Followup to #1509 as suggested by @suricactus.

The volumes static_volume, media_volume are extracted into YAML anchors so they are defined once and reused across app and worker_wrapper.

Shared volumes were verified with:
```shell
docker compose exec app touch /usr/src/app/staticfiles/test-shared.txt
docker compose exec worker_wrapper ls /usr/src/app/staticfiles/test-shared.txt
docker compose exec app rm /usr/src/app/staticfiles/test-shared.txt
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,9 +5,6 @@ x-logging: &default-logging
     max-file: "10"
     tag: "{{.ImageName}}|{{.Name}}|{{.ImageFullID}}|{{.FullID}}"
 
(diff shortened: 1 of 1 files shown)


suricactus:
I think for now this makes it less readable. Not happy with the current state, but the suggestion here is not much better. Thanks for the contribution @manylon .

PR #1527, opened 2026-03-29
The storage_used_bytes should return an int, not a float

nirvn:
@Rakanhf , @suricactus , spotted this while experimenting with the endpoint on staging. Do we need to make a migration if this hasn't hit mainstream yet?


suricactus:
Thanks for spotting. For a few weeks we will serve this as float and in the next release it will be integer.

PR #1529, opened 2026-03-31
feat: allow configurable http in nginx

gounux:
This PR intends to allow the nginx reverse proxy to be available using HTTP, through a new `NGINX_ALLOW_HTTP` environment variable that can be configured to do so. By default this env var is set to false (HTTPS only).

The considered approach is to use `include` directives, in order to factorize the nginx config headers and locations between HTTP and HTTPS.

Note that only docker-internal request can make use of such configured and allowed HTTP, this is intended for workers running on the same machine.


.env.example
toebivankenoebi, line 1: We'll need to add this into the release notes.

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -45,6 +45,9 @@
 
 logger = logging.getLogger(__name__)
 
+NGINX_INTERNAL_TOKEN = "static-nginx-internal-token"
```
suricactus: Does it make sense to keep this static string like this? I mean this is quite sensitive and we must be sure that nobody can pretend that the request came from nginx. Just bringing a warning here to verify nobody can do that.

gounux: Initially that was an env var, which could be generated with a random key generator that could make it not guessable. That said this would create yet another environment variable, though. Happy to discuss that.

gounux: Should we then add a new env var / setting with this internal "token" ? What do you think ?  
@lukasgraf do you have some thoughts as well ?

The idea being here to make the download file API endpoint have a way to know that a request comes from the nginx reverse-proxy, so that nginx handles the redirect to the storage and streaming the file contents back to the client. Maybe there is a better way to achieve so, would be happy to hear :)

docker-nginx/templates/includes/qfieldcloud.conf.template
```diff
@@ -0,0 +1,115 @@
...
+    error_page 502 503 504 =503 /pages/loading.html;
+
+    # Redirect 500-level errors to a dedicated handler location
+    error_page 500 501 505 = @handle_500_error;
+
+    proxy_intercept_errors on;
+    proxy_pass http://django;
+}
+
+location @handle_500_error {
```
gounux, line 44: Note: take into consideration #1536 once it's merged.

docker-app/qfieldcloud/filestorage/view_helpers.py
suricactus: I would rather replace `NGINX_ALLOW_INTERNAL_HTTP` with `NGINX_INTERNAL_HTTP_TOKEN`. If present, http is allowed and the value is passed, if not, then https only.

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -254,8 +257,13 @@ def download_field_file(
 
     range = get_range(request, field_file.size)
 
-    # Assume that if the request is secure, we are behind a nginx proxy and we can use `X-Accel-Redirect`
-    if request.is_secure() and not settings.IN_TEST_SUITE:
+    # Assume that if the request has the internal token in the headers,
+    # we are behind a nginx proxy and we can use `X-Accel-Redirect`.
+    if (
+        NGINX_INTERNAL_TOKEN_HEADER_KEY in request.headers
+        and request.headers[NGINX_INTERNAL_TOKEN_HEADER_KEY] == NGINX_INTERNAL_TOKEN
```
gounux: Here it could be possible to check if the `X-Forwarded-For` header is there.

docker-app/qfieldcloud/filestorage/view_helpers.py
suricactus: The need for the token was dropped, see https://github.com/opengisch/QFieldCloud/pull/1529#discussion_r3056419484

docker-app/qfieldcloud/filestorage/view_helpers.py
gounux: Addressed in https://github.com/opengisch/QFieldCloud/pull/1529/commits/2d5b68ae3be764a5217179a53c1b031482c44375

docker-nginx/templates/default.conf.template
```diff
@@ -59,10 +70,23 @@ server {
...
+  set $force_https 0;
+  if ($allow_internal_http != "1") {
+    set $force_https 1;
+  }
+  if ($is_docker_internal != "1") {
+    set $force_https 1;
+  }
+  if ($force_https = "1") {
     return 302 https://$host:${WEB_HTTPS_PORT}$request_uri;
   }
```
suricactus, line 79:
```suggestion
  # if HTTP is not allowed AND request is not from docker internal, redirect to HTTPS.
  if ($allow_internal_http != "1") {
    return 302 https://$host:${WEB_HTTPS_PORT}$request_uri;
  }
  if ($is_docker_internal != "1") {
    return 302 https://$host:${WEB_HTTPS_PORT}$request_uri;
  }
```

Let's try to reduce the imperative code within a configuration file as much as possible.

gounux:
Thanks for the review and simplification @suricactus,

Let's wait that the CI passes before merging this, if everybody is okay ?

PR #1530, opened 2026-04-02
Fix migration to target only project files, instead of breaking existing packages

suricactus:
If we run the migration without the introduced filter by package type, we will potentially break the packages of projects in such way, that we have to repackagee every single project for every single user that already had a package.

The reason is because we can easily have a file within a given project, name and file_type (used in the group by), but part of different package jobs. In this case we should not try to re-parent all the versions.

E.g. we have project p1 with the following project files:

- p1.qgz
- p1.gpkg

Then user u1 and u2 create their respective jobs.

In our database we will store:

| id | project | name       | file_type    | package_job_id |
|----|---------|------------|--------------|----------------|
| 1  | p1      | p1.qgz     | project_file | <null>         |
| 2  | p1      | p1.gpkg    | project_file | <null>         |
| 3  | p1      | p1.qgz     | package_file | u1_package_id  |
| 4  | p1      | data.gpkg  | package_file | u1_package_id  |
| 5  | p1      | p1.qgz     | package_file | u2_package_id  |
| 6  | p1      | data.gpkg  | package_file | u2_package_id  |

The group by project, name and type will should that file id 3 and 5, as well as 4 and 6 need to be merge, which is false, since they are part of different packages.

Then all versions of file 5 and 6 will be transferred to 3 and 4, which is wrong on two levels - it should not happen and package files should not have more than one version. What is even worse, the files 5 and 6 will be deleted, so when u2 tries to synchronize, their QField will receive an empty list of files and the project will not work at all.

Repackaging will be blocked if it is a gpkg only project and there if there is no data change, packaging is not triggered.

---

In theory the problem with duplicate files might happen also for package files, but since we are ensured by other levels of the application logic that we will never have two exactly the same names (or object storage prefixes, or versions) for package files, therefore we just filter them out from the query.

--- 

This was caught due to the "confirmation" query:
```select project_id, name, file_type, package_job_id, count(*) from filestorage_file where true group by 1, 2, 3, 4 having count(*) > 1 order by 1, 5;```

### DIFF
--- a/docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
+++ b/docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
@@ -65,7 +65,10 @@ def fix_duplicating_filenames_as_versions(apps, schema_editor):
     file_summaries_count = (
         File.objects.values("project_id", "name", "file_type")
         .annotate(count=models.Count("*"))
-        .filter(count__gt=1)
+        .filter(
+            file_type=FILE_TYPE_PROJECT_FILE,
+            count__gt=1,
+        )
         .count()
     )
 

### CODE REVIEW

File: docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py [hunk 1]
```diff
@@ -2,6 +2,8 @@
 
 from django.db import migrations, models
 
+FILE_TYPE_PROJECT_FILE = 1
```
gounux, line 5: Is there any reason not to use `File.FileType.PROJECT_FILE` here ? No big technical difference I presume, It seems more readable to me though, no hard opinion.

suricactus: We don't have access to it. In migrations the model we have access to is not the real class. That's why we don't import from `from qfc.filestorage.models import FileVersion` but we do `FileVersion = apps.get_model("filestorage", "FileVersion")`.

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
```diff
@@ -10,7 +12,10 @@ def fix_duplicating_filenames_as_versions(apps, schema_editor):
     file_summaries_qs = (
         File.objects.values("project_id", "name", "file_type")
         .annotate(count=models.Count("*"))
-        .filter(count__gt=1)
+        .filter(
+            file_type=FILE_TYPE_PROJECT_FILE,
+            count__gt=1,
```
Rakanhf, line 17: I believe that this `count__gt=1` is redundant,

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
```diff
@@ -10,7 +12,10 @@ def fix_duplicating_filenames_as_versions(apps, schema_editor):
     file_summaries_qs = (
```
Rakanhf, line 12: I think we can merge all filters within one statement instead, just to make it cleaner

```suggestion
    file_summaries_qs = (
        File.objects.values("project_id", "name", "file_type")
        .annotate(count=models.Count("*"))
        .filter(
            count__gt=1,
            file_type=FILE_TYPE_PROJECT_FILE
        )
    )
```

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
Rakanhf, line 66: Shouldn't we add the same here ?

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
suricactus: I am not sure if I get this comment. It's exactly the same as before, except:
- the order of the kwargs in the `filter` method are switched. The order does not have effect.
- A trailing comma was removed of the last kwarg, which will lead to linting error.

Actually if I apply the patch as it is suggested right now, it's not even syntatically correct, the result would be:
```
    file_summaries_qs = (
        File.objects.values("project_id", "name", "file_type")
        .annotate(count=models.Count("*"))
        .filter(
            count__gt=1,
            file_type=FILE_TYPE_PROJECT_FILE
        )
    )
        File.objects.values("project_id", "name", "file_type")  # this is syntatically incorrect line onwards
        .annotate(count=models.Count("*"))
        .filter(
            file_type=FILE_TYPE_PROJECT_FILE,
            count__gt=1,
        )
    )
```

I supposed this was intended to be a multiline patch suggestion, but it only references line 12 as of now.

@Rakanhf would you mind elaborating.

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
suricactus: Why it is redundant? The `count(*) > 1` is actually the line that filters only the problematic `File` instances, those that have a `count(*) = 1` are already good. We even have an assert a few lines below: `assert file_summary["count"] >= 2`

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
Rakanhf: I think that there was something wrong with my browser / my eyes, since I saw a double `.filter` statements added instead :

```python
        .filter(
            count__gt=1,
        )
        .filter(
            count__gt=1,
            file_type=FILE_TYPE_PROJECT_FILE
        )
```

this also I guess clarify the other comment above for the redundant `count__gt`

and basically my suggestion was what you actually did.

Sorry for the inconvenience

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
Rakanhf: clarified in the comments bellow

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
suricactus: Very much true, thanks for spotting both. Pushed a new commit.

docker-app/qfieldcloud/filestorage/migrations/0006_alter_file_unique_together_and_more.py
suricactus: I guess you were confused with the red color, which states the lines being deleted. Anyways, thanks!

PR #1531, opened 2026-04-02
Rephrase the error message so we don't mention the nonexisting "premium" word

suricactus:
Looking for neutral language that makes it clear what is the problem.

### DIFF
--- a/docker-app/qfieldcloud/core/permissions_utils.py
+++ b/docker-app/qfieldcloud/core/permissions_utils.py
@@ -699,7 +699,7 @@ def check_can_become_collaborator(user: QfcUser, project: Project) -> bool:
             if not user.useraccount.current_subscription.plan.is_premium:
                 raise ExpectedPremiumUserError(
                     _(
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -699,7 +699,7 @@ def check_can_become_collaborator(user: QfcUser, project: Project) -> bool:
             if not user.useraccount.current_subscription.plan.is_premium:
                 raise ExpectedPremiumUserError(
                     _(
-                        "Only premium users can be added as collaborators on private projects."
+                        "Only users who upgraded from free plan can be added as collaborators on private projects."
```
gounux, line 702: Could "community plan" be accurate here ?

suricactus: What if we rename it from "community" to "ninja"  ? That's why I like to keep if vague, this is burried far away from billing and we will never remember it.

PR #1532, opened 2026-04-03
fix upload project thumbnail permission

Rakanhf:
Currently `process_projectfile`  was failing. since the step for uploading a new thumbnail was limited to (ADMIN, MANAGER) roles. 

Changed the permission for thumbnail creation exclude READER role and limit job triggering permissions, so `READER` role can only trigger `PACKAGE` job, while Other write capable roles can trigger other jobs (PROCESS_PROJECTFILE, DELTA_APPLY)

- Change `upload_thumbnail` upload_permission to `can_create_files`
- Add test to cover thumbnail permissions

> [!WARNING]
> This PR #1535 should be merged first, or jobs will fail on the thumbnail step if the job was triggered by a reader user

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_project.py
+++ b/docker-app/qfieldcloud/core/tests/test_project.py
@@ -1032,15 +1032,12 @@ class QfcTestCase(APITransactionTestCase):
             name="perm_test_project", owner=self.user1, is_public=False
         )
 
-        # Equivalent to `can_create_files` permissions
-        roles = [
-            ProjectCollaborator.Roles.ADMIN,
-            ProjectCollaborator.Roles.MANAGER,
-            ProjectCollaborator.Roles.EDITOR,
-            ProjectCollaborator.Roles.REPORTER,
+        # Un authorized roles that are forbidden to upload a thumbnail
+        unauthorized_roles = [
+            ProjectCollaborator.Roles.READER,
         ]
 
-        for role in roles:
+        for role in ProjectCollaborator.Roles:
             collaborator_user = Person.objects.create_user(
                 username=f"user_{role}", password="abc123"
             )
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/views/jobs_views.py
```diff
@@ -33,10 +35,27 @@ def has_permission(self, request, view):
             job_id = permissions_utils.get_param_from_request(request, "job_id")
 
             try:
-                project = Job.objects.get(id=job_id).project
+                job = Job.objects.get(id=job_id)
```
suricactus:
```suggestion
                job = Job.objects.get(id=job_id).select_related("project")
```

Just check if we don't do this automatically in jobs.

docker-app/qfieldcloud/core/views/jobs_views.py
```diff
@@ -33,10 +35,27 @@ def has_permission(self, request, view):
...
+                project = job.project
+                job_type = job.type
```
suricactus: Please avoid such renamings whenever possible. `job.type` and `job.project` are perfectly fine and clearly state the source of the value being compared.

Rakanhf: It's to avoid name errors. since if we used `job.project` e.g `can_read_create_jobs(request.user, job.project)` in case of `should_expect_project_id` it will cause a name error

suricactus: I see, thanks!

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -1024,3 +1024,69 @@ def test_upload_project_thumbnail_invalid_file(self):
...
+            name="perm_test_project", owner=self.user1, is_public=False
+        )
+
+        # Equivalent to `can_create_files` permissions
+        roles = [
+            ProjectCollaborator.Roles.ADMIN,
+            ProjectCollaborator.Roles.MANAGER,
+            ProjectCollaborator.Roles.EDITOR,
+            ProjectCollaborator.Roles.REPORTER,
+        ]
```
suricactus: Can we instead of explicitly listing what is allowed, we list what is not allowed?

This way if we add a new role in the future, we will be made aware about this test.

PR #1546, opened 2026-04-09
Add project cloning to create job and expose it via api

Rakanhf:
Cloning projects will allow people to have some kind of templates that they reuse multiple times.

Now Users will be able to use the API to clone Projects easily
---

- Changed `ProjectSeed` model field to `clone_from_project`
- Adjust `clone_from_project` field to be null-able 
- Adjusted `ProjectViewSetPermissions` to handle cloning action
- Added `clone` action to enable `/projects/{PROJECT_UUID}/clone/`  
- Added `ProjectCloneSerializer` with necessary validations and creation logic
- Added a new exception `NotCloneableProjectError` 
- Fixed `get_param_from_request` to enhance error handling, since we accept both `JSONParser`, `MultiPartParser`
- Split `project_creation` into two streamlined functions `prepare_project_files`, `configure_qgis_project`
- Added `the_qgis_file_name` field to `ProjectSerializer`
---

#### Example : 

```
curl --location 'https://localhost/api/v1/projects/' \
--header 'Content-Type: application/json' \
--header 'Authorization: {MY_TOKEN}' \
--data '{
    "name" : "clone-me-pubic",
    "is_public": true,
    "description": "Hello from the cloned public project",
    "owner": "{USERNAME}",
    "seed": {
        "extent": "-180, -90, 180, 90"
    },
    "clone_from_project": "{PROJECT_UUID}"
}'
```

which will return the Project as a response.
---


duke-nyuki:
Task linked: [QF-6717 Allow duplicating/cloning a project and all it's files](https://app.clickup.com/t/2192114/QF-6717)

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -75,6 +77,22 @@ def has_permission(self, request, view) -> bool:
             return permissions_utils.can_update_project(user, project)
         elif view.action == "upload_thumbnail":
             return permissions_utils.can_create_files(user, project)
+        elif view.action == "clone":
+            # check if the owner is passed in the request data
+            clone_owner_id = None
+            if isinstance(request.data, dict):
```
suricactus, line 86: In what situation the `request.data` is not a `dict`? I think it is safe to assume it is always a `dict`?

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -75,6 +77,22 @@ def has_permission(self, request, view) -> bool:
...
+            if clone_owner_id:
+                try:
+                    clone_owner = User.objects.get(pk=clone_owner_id)
+                except User.DoesNotExist:
+                    return False
+
+            else:
+                clone_owner = user
+
+            return permissions_utils.can_clone_project(user, project, clone_owner)
```
suricactus: I would suggest that we avoid the third parameter here. We want to know if a user can clone the project. If the new owner is assignable, well that's another story and should be checked at a later stage.

reasons:

- the `permission_utils` should ideally always have only two parameters only. This will make it much easier in the future when we refactor the permission utils.
- if you have to dig insite the actual request body in the permissions check class, then something is not designed in a very optimal way.
- we do care if a user can close specific project. If any of the other payload is invalid - name, owner, etc, etc, then we are going to throw at a later stage.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -272,6 +289,46 @@ def seed_xlsform(self, request: Request, projectid: UUID) -> StreamingHttpRespon
...
+        request=ProjectCloneSerializer,
+    )
+    @action(detail=True, methods=["post"], url_path="clone")
+    def clone(self, request: Request, projectid: UUID) -> Response:
+        source_project = self.get_object()
+
+        if source_project.is_shared_datasets_project:
+            return Response(
+                {
+                    "detail": "This project is a special shared datasets project and cannot be cloned."
```
suricactus:
```suggestion
                    "detail": 'The "shared_ datasets" project cannot be cloned."
```

More consise.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -272,6 +289,46 @@ def seed_xlsform(self, request: Request, projectid: UUID) -> StreamingHttpRespon
...
+    @extend_schema(
+        description="Clone an existing project",
+        request=ProjectCloneSerializer,
+    )
+    @action(detail=True, methods=["post"], url_path="clone")
+    def clone(self, request: Request, projectid: UUID) -> Response:
+        source_project = self.get_object()
+
+        if source_project.is_shared_datasets_project:
+            return Response(
```
suricactus: Instead of adding return of custom responses here and there, it's better if we `raise NotClonableProjectError("message")` as we do everywhere else.

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -162,9 +164,46 @@ def _create_project_from_xlsform(
...
+    project.write(str(project_filename))
+
+    return str(project_filename)
+
+
 def create_project_from_seed(
     project_seed: ProjectSeed, tmp_project_dir: str, xlsform_filename: str
 ) -> str:
+    if project_seed.copy_from_project:
+        return clone_project_from_seed(project_seed, tmp_project_dir)
```
suricactus: Note that the new function `clone_project_from_seed` and the parent function are suspiciosly similar (and their overlap will only grow in the future).


I would suggest that we split the `create_project_from_seed` function in two parts (therefore Job steps):


1) Prepare project files.

INPUT: seed

-- if cloning, download projects
-- if not
--- if a xlsform form file, run xlsform converter
--- otherwise just create an empty project
--- Then create basemaps etc (can be part of the latter step too)

OUTPUT: QGIS filename

2) Configure QGIS project.

INPUT: seed, QGIS filename

-- open project
-- Validate CRS and set CRS (can be in 1) instead, not sure)
-- Set project extent
-- set project title
....


This way the code duplication will be greatly minimized.

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -162,9 +164,46 @@ def _create_project_from_xlsform(
...
+        )
+        if not project_filename.exists():
+            raise FileNotFoundError(
+                f"QGIS project file '{project_seed.source_qgis_file_name}' "
+                f"not found in source project {source_id}"
+            )
+    else:
+        raise FileNotFoundError(
+            f"No QGIS project file name known for source project {source_id}"
+        )
```
suricactus, line 187: No need for this new field `source_qgis_file_name`. You can always call `GET /projects/<source_project>` and get the project filename from there. Yes, it is an extra call, but worker->API is cheap and very fast. Bonus points we keep the `GET /seed` endpoint minimal.

suricactus: - avoid using overloaded operators in QFieldCloud code. The lead developer has very hard feelings against them in dynamic languages (don't use `/` to join path).
- even without those hard feelings, a few lines below the code was already written with `joinpath` so keep it for consistency.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -244,9 +246,79 @@ class Meta:  # type: ignore[reportIncompatibleVariableOverride]
     settings = serializers.JSONField()
     # TODO @suricactus: QF-7258 Adding a project extent field on project creation, see https://app.clickup.com/t/2192114/QF-7258
     crs = serializers.CharField(default="EPSG:3857")
+    source_qgis_file_name = serializers.CharField(
+        source="copy_from_project.the_qgis_file_name",
+        read_only=True,
+        allow_null=True,
+    )
```
suricactus, line 253: Please drop this property. It's redundant, you can obtain this information by calling `GET /projects/<source_project_uuid>`. In the future we can optimize this by expanding the whole `clone_from_project` with `ProjectSerializer`. But no need to do this optimization right now.

suricactus (commented):
I know this is WIP, there are a few things to be adjusted beyound what we have already discussed via chat.

docker-app/qfieldcloud/core/views/projects_views.py
Rakanhf: It's a defensive check against if the user supplied a json list instead, since we accept both `JSONParser`, `MultiPartParser`

docker-app/qfieldcloud/core/views/projects_views.py
Rakanhf: To simplify things I will drop `can_clone_project` and use `can_access_project` and `can_create_project` in the if statement instead.

docker-app/qfieldcloud/core/views/projects_views.py
suricactus: Can you provide a sample `curl` that will result in `request.data` being non-dict at this line of code?

Rakanhf: Example:

```
curl --location 'https://localhost/api/v1/projects/{PROJECT_UUID}/clone/' \
--header 'Content-Type: application/json' \
--header 'Authorization: token {MY_TOKEN}' \
--data '["1","2","3"]'
```

I will get : 
"'list' object has no attribute 'get'" STATUS `500`

But it should be : 
"Invalid data. Expected a dictionary, but got list." STATUS `400`

Anyway after looking into it twice, I see that the same issue will happen for

https://github.com/opengisch/QFieldCloud/blob/b69cbc41eb0a7efc9990e1255227f48669965905/docker-app/qfieldcloud/core/permissions_utils.py#L171

which makes my safeguard check useless, since the function `get_param_from_request` will cause the error first. But now I also noticed that we don't even need to check for `clone_owner_id` since the `owner_obj` is exactly the same. So I will remove this redundant code and move the defensive check to `get_param_from_request`. so now the code is cleaner and more robust.

docker-qgis/qfc_worker/commands/create_project.py
Rakanhf: > No need for this new field `source_qgis_file_name`. You can always call `GET /projects/<source_project>` 

I had to add  `the_qgis_file_name` to the `ProjectSerializer`, sine it wasn't there

docker-app/qfieldcloud/core/migrations/0100_alter_projectseed_extent.py
suricactus, line 1: Would you mind putting all migrations in one migration per commit whenever possible?

docker-app/qfieldcloud/core/models.py
```diff
@@ -1994,9 +1996,10 @@ class ProjectSeed(models.Model):
     )
     """The project to copy from, if any. It is mutually exclusive with `xlsform_file`."""
 
+    # TODO: revert this to not null and blank False once we create a new extent field in the project model
```
suricactus, line 1999: Hey, it's not something that I enforced yet, but it's always nicer if we add TODO with it's author and a link to the actual TODO task.

So something like:

```
TODO @Rakanhf: make the `extent` field not nullable once we add `Project.extent` field.
```

docker-app/qfieldcloud/core/models.py
```diff
@@ -2008,19 +2011,25 @@ class ProjectSeed(models.Model):
...
-        if self.xlsform_file and self.copy_from_project:
+        if self.xlsform_file and self.clone_from_project:
             raise ValidationError(
                 _(
-                    "Both `xlsform_file` or `copy_from_project` cannot be set at the same time."
+                    "Both `xlsform_file` or `clone_from_project` cannot be set at the same time."
                 )
             )
 
+        # TODO: revert this to not null and blank False once we create a new extent field in the project model
```
suricactus, line 2027: Same comment

docker-app/qfieldcloud/core/models.py
```diff
@@ -1994,9 +1996,10 @@ class ProjectSeed(models.Model):
     )
     """The project to copy from, if any. It is mutually exclusive with `xlsform_file`."""
 
+    # TODO: revert this to not null and blank False once we create a new extent field in the project model
     extent = models.PolygonField(
-        null=False,
-        blank=False,
+        null=True,
+        blank=True,
```
suricactus, line 2002: Why do we need to make the field "nullable"? Both `clone` and `create from template` can require the extent field?

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -186,17 +215,38 @@ def create_project_from_seed(
...
-    if project_seed.settings.basemaps:
+    project.setTitle(project_seed.name)
+
+    crs = QgsCoordinateReferenceSystem(project_seed.crs)
+    if crs.isValid():
+        project.setCrs(crs)
+    else:
+        logger.warning("Project CRS parameter is invalid, skipping CRS configuration.")
+
+    # TODO: set the extent of the project
```
suricactus, line 247: Same TODO

suricactus (commented):
Nice 👍

docker-app/qfieldcloud/core/models.py
Rakanhf: Yes, since when we clone a project we don't need to supply the extent for now.

docker-app/qfieldcloud/core/tests/test_project.py
```diff
@@ -1252,3 +1255,154 @@ def test_upload_project_thumbnail_permissions(self):
             )
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_clone_project(self):
```
suricactus, line 1289: Split these to smaller targeted tests, instead of a single mega function. Hard to maintains in the long term.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -145,33 +171,60 @@ def validate(self, data):
...
+                    raise ValidationError(
+                        {"seed": "Only extent field is allowed in seed."}
+                    )
+
+            target_owner = data.get("owner") or self.context["request"].user
+            if (
+                source_project.file_storage_bytes
+                > target_owner.useraccount.storage_free_bytes
+            ):
+                raise ValidationError({"storage_quota": "Insufficient storage quota."})
```
suricactus: What is `storage_quota` here? I thought it is a name of a field, but there is no such field? Also, I would suggest we use the error that we already have prepared for such cases, namely `subscriptions.QuotaError`.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -221,9 +278,20 @@ class Meta:
...
         )
+        _clone_exclude = {
+            "id",
+            "name",
+            "owner",
+            "seed",
+            "xlsform_file",
+            "clone_from_project",
+        }
+        clonable_fields = set(fields) - set(read_only_fields) - _clone_exclude
```
suricactus: We discussed over a chat this must be a whitelist instead of blacklist of fields.

docker-app/qfieldcloud/core/serializers.py
```diff
@@ -145,33 +171,60 @@ def validate(self, data):
...
                 )
 
+            if seed:
+                provider = seed.get("basemap_provider", "none")
+                url = seed.get("basemap_url", "")
+                if provider == "custom" and not url:
+                    raise ValidationError(
+                        {
+                            "seed": [
+                                "basemap_url is required when basemap_provider is 'custom'."
```
```suggestion
                                'When configuring basemaps in the project's seed, `basemap_url` is required when `basemap_provider` is set to "custom".'
```

In error messages, we try to be clear. Identifiers we quote with backticks, values with double quotes.

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -174,38 +189,62 @@ def get_queryset(self):
...
+                extent = None
+
+            ProjectSeed.objects.create(
+                project=project,
+                extent=extent,
+                clone_from_project=source_project,
+                settings={
+                    "schemaId": ProjectSeed.SETTINGS_SCHEMA_ID,
+                    "basemaps": [],
+                    "xlsform": None,
```
suricactus: Do we require those keys? If not, we can just kill them?

suricactus (approved):
Thank you! Exciting new feature!

docker-app/qfieldcloud/core/views/projects_views.py
```diff
@@ -174,38 +189,62 @@ def get_queryset(self):
     def perform_create(self, serializer: ProjectSerializer) -> None:
         super().perform_create(serializer)
 
-        seed_data = serializer.validated_data.get("seed", None)
-        if not seed_data:
-            return
-
-        xlsform_file = serializer.validated_data.get("xlsform_file", None)
         project = serializer.instance
+        source_project = serializer.validated_data.get("clone_from_project", None)
```
suricactus:
```suggestion
        clone_from_project = serializer.validated_data.get("clone_from_project", None)
```

docker-app/qfieldcloud/core/views/projects_views.py
Rakanhf: It is required, but I adjusted the `ProjectSeedSettings` dataclass so we don't need to pass them anymore.

suricactus (approved):
Thanks!

I think we can make a tiny refactoring of `perform_create` view class, but it can be in a follow-up, now it's functional and clear enough.

Good job!

PR #1548, opened 2026-04-10
fix:`PROCESS_PROJECTFILE` being triggered on shared_datasets projects

Rakanhf:
When `the_qgis_file_name` is set to be an empty string (blank), and a gpgk file is uploaded, then a `PROCESS_PROJECTFILE` job is triggered which will cause it to fail. since it was resolving to True

- Use `has_the_qgis_file` property to correctly check if th project has qgis file


duke-nyuki:
Task linked: [QF-7417 Hide status color dot of shared_datasets projects](https://app.clickup.com/t/2192114/QF-7417)

suricactus (approved):
Nice one!

PR #1549, opened 2026-04-13
Make `parse_extent_coords` to expect coords in EPSG:4326 instead of 3857

suricactus:
Not much to say.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_project.py
+++ b/docker-app/qfieldcloud/core/tests/test_project.py
@@ -89,7 +89,7 @@ class QfcTestCase(APITransactionTestCase):
             {
                 "basemap_provider": "osm",
                 "basemap_style": "standard",
-                "extent": "800000,5900000,1100000,6100000",
+                "extent": "22,41,29,45.67",
             }
         )
 
--- a/docker-app/qfieldcloud/core/utils2/project_seed.py
+++ b/docker-app/qfieldcloud/core/utils2/project_seed.py
@@ -9,7 +9,7 @@ from qfieldcloud.core.models import BasemapProvider, BasemapStyle
 
 OSM_URL = "https://tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png"
 
-DEFAULT_PROJECT_EXTENT = (-20000000, -20000000, 20000000, 20000000)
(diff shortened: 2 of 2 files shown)


duke-nyuki:
Task linked: [QF-8111 Use EPSG:4326 in user facing extent selection](https://app.clickup.com/t/2192114/QF-8111)

Rakanhf:
I think these should also be changed

1. https://github.com/opengisch/QFieldCloud/blob/cffe2a9fc448442a1fce10c74f65e8e849b0160a/docker-qgis/qfc_worker/commands/create_project.py#L261
2. https://github.com/opengisch/QFieldCloud/blob/cffe2a9fc448442a1fce10c74f65e8e849b0160a/docker-app/qfieldcloud/core/serializers.py#L246 (and also the comment above can be removed now)
3. https://github.com/opengisch/QFieldCloud/blob/cffe2a9fc448442a1fce10c74f65e8e849b0160a/docker-app/qfieldcloud/core/models.py#L2000
4. https://github.com/opengisch/QFieldCloud/blob/cffe2a9fc448442a1fce10c74f65e8e849b0160a/docker-app/qfieldcloud/core/tests/test_jobs.py#L353 (which I think the test is failing because of it)


For the change 3, there will be a migration file related to this, but I'm not sure which data format should we save the data in db. If `3857` will be saved then I think we should also change the `build_seed_data` to transform it and also do that accordingly with the api, Let me know what do you think

suricactus:
Thanks for the comments. I forgot we had an API already exposed for creating projects, so I adjusted that.

Sorry I force pushed accidentally, it's a short PR and I hope you don't bother much.

---

I guess I should have been more descriptive in the PR and I assumed GIS/QGIS knowledge.

We just changed what is the input coordianate system on app.qfield.cloud form (and API) for the project extent. However, we do not need change the project CRS, the layer CRS, the database field or anything else.

1) Not needed, QGIS projects can happily live with layers in different projections, they will be reprojected on the fly.
2) No, we still accept  and assume EPSG:3857 as an input CRS.
3) No, we still store EPSG:3857, we reproject the input coordinates from EPSG:4326.
4) No, because 1), 2) and 3).

Rakanhf (approved):
LGTM

PR #1550, opened 2026-04-14
chore: remove unnecessary webdav storage's `is_name_available` method

gounux:
Since Django >= 5.1, the `is_name_available` method is present in the base class, so we don't need it anymore.

See https://github.com/django/django/blob/84d09a547fe35e10018ab242602ca76c29ca91a1/django/core/files/storage/base.py#L54-L56


duke-nyuki:
Task linked: [QF-7176 Get rid of Storage's `is_name_available` with Django >= 5.1](https://app.clickup.com/t/2192114/QF-7176)

suricactus (approved):

PR #1551, opened 2026-04-16
chore: add default value for nginx's internal HTTP in docker compose

gounux:
With this default value in nginx docker compose service, setups without the env var set would thus be anyway configured.


duke-nyuki:
Task linked: [QF-8071 Allow HTTP in nginx](https://app.clickup.com/t/2192114/QF-8071)

suricactus (approved):

PR #1552, opened 2026-04-16
fix(nginx): Improve error handling and logging

gounux:
Never cache 500 error page.
Caching lead to nginx immediately serving a 304 not modified when error page was cached, which is confusing.

Remove "=" sign, preserving the original 500 status code and not overwriting it.

Supersed #1536  
Closes #1536


gounux:
Allowing myself to merge, since those changes were approved in #1536

suricactus (commented):
LGTM

PR #1553, opened 2026-04-16
Hardening the WebDAV storage backend

t-book:
Implements the proposal from https://ideas.qfield.org/qfieldcloud-feature-requests/p/improve-resilience-of-qfcwebdavstorage-against-real-world-webdav-servers-without

### Context

`QfcWebDavStorage` works against the `bytemark/webdav:2.4` container in the dev stack, but breaks against modern WebDAV servers. Most visibly against Nextcloud/SabreDAV, where a file upload crashes with a 500 during `make_collection`: a HEAD-based existence check misreads SabreDAV's `405 Method Not Allowed` as "collection absent" and then crashes on the redundant MKCOL.

While fixing that, a few other robustness gaps became visible: no retries on transient `5xx`, no request timeouts, and `exists()` / `delete()` swallowing all HTTP errors — survivable on S3 thanks to bucket versioning, but risky on WebDAV where there is no such safety net.

This PR addresses all five issues, scoped to `QfcWebDavStorage`. No signature changes, no new dependencies, no changes to S3, migrations, CI matrix, or documentation. **The project's stance that S3 is the only production-supported backend remains unchanged; this only makes the WebDAV option more robust for self-hosters.**

### What changes

- **`make_collection`** sends `MKCOL` unconditionally and accepts `201/405/409` as success. Fixes the Nextcloud 500 crash, saves one round-trip per directory level, race-condition-free.
- **`get_requests_session`** configures a `urllib3.Retry` adapter for `429/502/503/504` with `MKCOL` and `PROPFIND` in `allowed_methods` (urllib3 defaults do not include them).
- **`perform_webdav_request`** applies method-aware timeouts via `_timeout_for()` — longer read timeout for `PUT` to accommodate server-side processing (virus scan, version snapshot).
- **`exists`** distinguishes "absent" (404/410) from "unreachable" (401/403/5xx, network errors), which now raise `IOError` instead of silently returning `False` and risking data overwrite in `Storage.save()`.
- **`delete`** is idempotent for 404/410 and raises all other HTTP errors.

### Tests

Three new integration tests added to the existing `QfcTestCase`, covering the versioning lifecycle (old-version download after new upload, targeted version delete, delete-then-reupload). They reuse the WebDAV service the existing integration tests already depend.

Four new mock-based `SimpleTestCase` classes add 21 unit tests for the resilience changes. They require no WebDAV service and no DB.

### Verification

Tests run locally against both backends by swapping the `"webdav"` entry in `STORAGES`: `bytemark/webdav:2.4` (current dev stack) and hosted Nextcloud (SabreDAV). Both pass.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -7,7 +7,9 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import ContentFile
 from django.core.files.storage import Storage
 from django.http import HttpResponse
+from requests.adapters import HTTPAdapter
 from storages.backends.s3 import S3Storage
+from urllib3.util.retry import Retry
 
 
 class QfcBackendStorageMixin(ABC):
@@ -63,6 +65,16 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
     Copyright (c) 2020, Mikhail Porokhovnichenko
     """
 
+    # Should this go into .env?
+    DEFAULT_TIMEOUT = (10, 60)
+    UPLOAD_TIMEOUT = (10, 600)
+    RETRY_TOTAL = 5
+    RETRY_BACKOFF_FACTOR = 1
+    RETRY_STATUS_FORCELIST = (429, 502, 503, 504)
+    RETRY_ALLOWED_METHODS = frozenset(
+        ["HEAD", "GET", "PUT", "DELETE", "MKCOL", "PROPFIND"]
+    )
+
     def __init__(self, **options):
         self.requests = self.get_requests_session(**options)
 
@@ -91,11 +103,30 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
 
         return True
 
-    def get_requests_session(self, **kwargs) -> requests.Session:
+    def get_requests_session(self, **_kwargs) -> requests.Session:
         """
-        Creates a HTTP session for requesting webdav later.
+        Creates an HTTP session with retry and connection pooling.
         """
-        return requests.Session()
+        session = requests.Session()
+        retries = Retry(
+            total=self.RETRY_TOTAL,
+            backoff_factor=self.RETRY_BACKOFF_FACTOR,
+            status_forcelist=self.RETRY_STATUS_FORCELIST,
+            allowed_methods=self.RETRY_ALLOWED_METHODS,
+            respect_retry_after_header=True,
+            raise_on_status=False,
+        )
+        adapter = HTTPAdapter(max_retries=retries, pool_connections=10, pool_maxsize=20)
+        session.mount("https://", adapter)
+        session.mount("http://", adapter)
+        return session
+
+    def _timeout_for(self, method: str) -> tuple[int, int]:
+        """Returns the (connect, read) timeout to use for a given HTTP method."""
+        if method.lower() == "put":
+            return self.UPLOAD_TIMEOUT
+
+        return self.DEFAULT_TIMEOUT
 
     def perform_webdav_request(
         self, method: str, name: str, *args, **kwargs
@@ -111,6 +142,7 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         """
         url = self.get_webdav_url(name)
         method = method.lower()
+        kwargs.setdefault("timeout", self._timeout_for(method))
 
         response = getattr(self.requests, method)(url, *args, **kwargs)
         response.raise_for_status()
@@ -139,7 +171,7 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         """
         return self.public_url.rstrip("/") + "/" + name.lstrip("/")
 
-    def _open(self, name: str, mode: str = "rb") -> ContentFile:
+    def _open(self, name: str, mode: str = "rb") -> ContentFile:  # pylint: disable=unused-argument
         """Reads the content of a file from the configured webdav storage.
 
         Arguments:
@@ -149,8 +181,11 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         Returns:
             Content of the requested file.
         """
-        content = self.perform_webdav_request("GET", name).content
-        return ContentFile(content, name)
+        response = self.perform_webdav_request("GET", name, stream=True)
+        return ContentFile(
+            b"".join(response.iter_content(chunk_size=8 * 1024 * 1024)),
+            name,
+        )
 
     def _save(self, name: str, content: ContentFile) -> str:
         """Saves a file on the configured webdav storage.
@@ -174,8 +209,11 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         return name
 
     def make_collection(self, name: str) -> None:
-        """Creates a so-called collection on the configured webdav storage for a file.
-        Typically creates parent folders if not existing.
+        """Creates parent collections (folders) for a file on the webdav server.
+
+        Sends MKCOL for each parent. Treats 405 (URL already mapped) and 409
+        as success, so the loop is idempotent and tolerant of servers that
+        use 409 loosely for "already exists".
 
         Arguments:
             name: relative path of the file on the webdav server.
@@ -184,10 +222,11 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
 
         for directory in name.split("/")[:-1]:
             col = os.path.join(coll_path, directory, "")
-            resp = self.requests.head(col)
+            resp = self.requests.request(
+                "MKCOL", col, timeout=self._timeout_for("MKCOL")
+            )
 
-            if not resp.ok:
-                resp = self.requests.request("MKCOL", col)
+            if resp.status_code not in (201, 405, 409):
                 resp.raise_for_status()
 
             coll_path = os.path.join(coll_path, directory)
@@ -195,13 +234,18 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
     def delete(self, name: str) -> None:
         """Deletes a file from a configured webdav storage.
 
+        Idempotent for 404/410, raises for other errors.
+
         Arguments:
             name: relative path of the file on the webdav server.
         """
         try:
             self.perform_webdav_request("DELETE", name)
-        except requests.HTTPError:
-            pass
+        except requests.HTTPError as exc:
+            if exc.response is not None and exc.response.status_code in (404, 410):
+                return
+
+            raise
 
     def exists(self, name: str) -> bool:
         """Checks if a file exists on a configured webdav storage.
@@ -209,15 +253,28 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         Arguments:
             name: relative path of the file on the webdav server.
 
+        Raises:
+            IOError: when the server is unreachable or returns a status that
+                is neither a clear hit (2xx) nor a clear miss (404/410).
+
         Returns:
-            True if exists, False if not.
+            True if the file exists, False if not.
         """
+        url = self.get_webdav_url(name)
         try:
-            self.perform_webdav_request("HEAD", name)
-        except requests.exceptions.HTTPError:
+            resp = self.requests.head(url, timeout=self._timeout_for("HEAD"))
+        except requests.exceptions.RequestException as exc:
+            raise IOError(f"WebDAV exists() network error for {name!r}: {exc}") from exc
+
+        if 200 <= resp.status_code < 300:
+            return True
+
+        if resp.status_code in (404, 410):
             return False
 
-        return True
+        raise IOError(
+            f"WebDAV exists() got unexpected status {resp.status_code} for {name!r}"
+        )
 
     def size(self, name: str) -> int:
         """Returns the size of a file on a configured webdav storage.
@@ -235,10 +292,10 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
             return int(
                 self.perform_webdav_request("HEAD", name).headers["content-length"]
             )
-        except (ValueError, requests.exceptions.HTTPError):
-            raise IOError("Unable get size for %s" % name)
+        except (ValueError, requests.exceptions.HTTPError) as exc:
+            raise IOError(f"Unable to get size for {name}") from exc
 
-    def url(self, name: str, **kwargs) -> str:  # type: ignore
+    def url(self, name: str, **_kwargs) -> str:  # type: ignore
         """Returns the URL of a file from the configured webdav storage.
 
         Arguments:
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -4,10 +4,28 @@ from abc import ABC
 
 import requests
 from django.core.exceptions import ImproperlyConfigured
-from django.core.files.base import ContentFile
+from django.core.files.base import ContentFile, File
 from django.core.files.storage import Storage
 from django.http import HttpResponse
+from requests.adapters import HTTPAdapter
+from rest_framework import status
 from storages.backends.s3 import S3Storage
+from urllib3.util.retry import Retry
+
+# WebDAV request timeouts (connect, read) in seconds
+DEFAULT_TIMEOUT = (10, 60)
+UPLOAD_TIMEOUT = (10, 600)
+
+# Retry configuration
+RETRY_TOTAL = 5
+RETRY_BACKOFF_FACTOR = 1
+RETRY_STATUS_FORCELIST = (
+    status.HTTP_429_TOO_MANY_REQUESTS,
+    status.HTTP_502_BAD_GATEWAY,
+    status.HTTP_503_SERVICE_UNAVAILABLE,
+    status.HTTP_504_GATEWAY_TIMEOUT,
+)
+RETRY_ALLOWED_METHODS = frozenset(["HEAD", "GET", "PUT", "DELETE", "MKCOL", "PROPFIND"])
 
 
 class QfcBackendStorageMixin(ABC):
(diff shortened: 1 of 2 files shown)


suricactus:
Hello @t-book . Thank you for the PR. We are a bit short of resources at this immediate moment and we did not account for this change coming (yet). We will do our best effort to review this in the few next weeks.

The ideas and the community portal are for sharing ideas that we review on regular, but not daily basis. They are intended for non-techny users to share their wishes.

Before opening a PR it's always nice to first have an issue here on GitHub, a tool that we use and check every day, and allows immediate conversation with the developers before even coding.

It makes the process easier for both sides:

- The contributor knows in advance if such PR is going to be accepted and if there are the other requirements to consider.
- The reviewer knows that such PR is coming, has shared their expectations, and makes the review process faster.

---

So if we start from scratch, feel free to start from the ideas platform, but also open an issue here on GitHub. After a few back and forth with the maintainers as comments, the expectations on both sides will be cleaner and the actual code will be straighforward implementation.

Actually the process was aleady described in the README.md file: https://github.com/opengisch/QFieldCloud#collaboration

As we already have the PR here, no need to do something else besides waiting for a review.

I am very happy with your contributions recently, I don't want to discourage you in any way to continue improving QFieldCloud.

t-book:
Hi @suricactus Thank you very much for your reply — totally understandable that the review will take some time.

Regarding GitHub issues: this is actually how I usually work as well and what I’m used to. In the QFieldCloud project, however, I ran into the problem that the GitHub issue templates redirect feature issues to the ideas platform, and since this isn’t a bug, I wasn’t sure what the correct way would be to open an issue in this case.

Would it be appropriate to open a feature request using the bug template, or am I maybe overlooking another option?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -63,6 +65,16 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
     Copyright (c) 2020, Mikhail Porokhovnichenko
     """
 
+    # Should this go into .env?
```
gounux: Could this maybe go to the django [`settings.py`](https://github.com/opengisch/QFieldCloud/blob/master/docker-app/qfieldcloud/settings.py) ?

Also, it could be nice to have those constants prefixed with `WEBDAV_STORAGE_*`, e.g. :

```python
WEBDAV_STORAGE_DEFAULT_TIMEOUT = (10, 60)
WEBDAV_STORAGE_UPLOAD_TIMEOUT = (10, 600)
```

Then it's possible to access them via `settings.WEBDAV_STORAGE_DEFAULT_TIMEOUT`,

Also, a little comment describing the meaning of each constant would be helpful :)

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -63,6 +65,16 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
...
+    DEFAULT_TIMEOUT = (10, 60)
+    UPLOAD_TIMEOUT = (10, 600)
+    RETRY_TOTAL = 5
+    RETRY_BACKOFF_FACTOR = 1
+    RETRY_STATUS_FORCELIST = (429, 502, 503, 504)
```
gounux: Rather than using magic numbers for HTTP return codes, could we use DRF's constants ?

That would look like :

```python
from rest_framework import status
status.HTTP_429_TOO_MANY_REQUESTS
status.HTTP_502_BAD_GATEWAY
...
```

You can find such an exemple here in the code base : https://github.com/opengisch/QFieldCloud/blob/605e9da987816f61162aa1107d2f6545114908cb/docker-app/qfieldcloud/core/exceptions.py#L118

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -91,11 +103,30 @@ def check_status(self) -> bool:
...
+        session = requests.Session()
+        retries = Retry(
+            total=self.RETRY_TOTAL,
+            backoff_factor=self.RETRY_BACKOFF_FACTOR,
+            status_forcelist=self.RETRY_STATUS_FORCELIST,
+            allowed_methods=self.RETRY_ALLOWED_METHODS,
+            respect_retry_after_header=True,
+            raise_on_status=False,
+        )
+        adapter = HTTPAdapter(max_retries=retries, pool_connections=10, pool_maxsize=20)
```
gounux, line 125: Could those `pool_connections` and `pool_maxsize` also go to a specific constant in `settings.py` ?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -91,11 +103,30 @@ def check_status(self) -> bool:
...
         """
-        return requests.Session()
+        session = requests.Session()
+        retries = Retry(
+            total=self.RETRY_TOTAL,
+            backoff_factor=self.RETRY_BACKOFF_FACTOR,
+            status_forcelist=self.RETRY_STATUS_FORCELIST,
+            allowed_methods=self.RETRY_ALLOWED_METHODS,
+            respect_retry_after_header=True,
+            raise_on_status=False,
```
gounux, line 123: It seems that here we would silently ignore 429, 502, 503 and 504 errors returned by the WebDAV backend, is this what we want ?

See [urllib's doc](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html) for this `raise_on_status=False` key,

My opinion is that we should raise if our future WebDAV requests would not pass, what do you think ?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -184,40 +222,59 @@ def make_collection(self, name: str) -> None:
...
         for directory in name.split("/")[:-1]:
             col = os.path.join(coll_path, directory, "")
-            resp = self.requests.head(col)
+            resp = self.requests.request(
+                "MKCOL", col, timeout=self._timeout_for("MKCOL")
+            )
 
-            if not resp.ok:
-                resp = self.requests.request("MKCOL", col)
+            if resp.status_code not in (201, 405, 409):
```
gounux, line 238: Wondering if the `raise_for_status` would raise if the HTTP return code is 201 ?

Based on [its definition](https://github.com/psf/requests/blob/6ec76b4a36fba4a32fb72f5f9ba04e632635e477/src/requests/models.py#L1006), it seems that 2XX and 3XX return codes pass the function,

Also, wondering if we should silently ignore [HTTP 405 Method not allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405) here,

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -149,8 +181,11 @@ def _open(self, name: str, mode: str = "rb") -> ContentFile:
         Returns:
             Content of the requested file.
         """
-        content = self.perform_webdav_request("GET", name).content
-        return ContentFile(content, name)
+        response = self.perform_webdav_request("GET", name, stream=True)
```
gounux, line 194: Could you explain why adding a `stream` here ?

My bet would be to support servers don't sending the `Content-length` header and answering chunks, though I'm curious !

Anyway, a little comment might help :)

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -174,8 +209,11 @@ def _save(self, name: str, content: ContentFile) -> str:
         return name
 
     def make_collection(self, name: str) -> None:
-        """Creates a so-called collection on the configured webdav storage for a file.
-        Typically creates parent folders if not existing.
+        """Creates parent collections (folders) for a file on the webdav server.
```
gounux, line 221:
```suggestion
        """Creates parent collections (directories) for a file on the webdav server.
```

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -174,8 +209,11 @@ def _save(self, name: str, content: ContentFile) -> str:
...
+
+        Sends MKCOL for each parent. Treats 405 (URL already mapped) and 409
```
gounux, line 223: Are we sure about the `405 (URL already mapped)` ?

[Specs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405) seem to say `Method not allowed`,

docker-app/qfieldcloud/filestorage/tests/test_webdav.py
```diff
@@ -65,3 +69,249 @@ def test_upload_then_delete_file_succeeds(self):
...
+    def test_session_has_retry_adapter_with_webdav_methods(self):
+        storage = _make_storage()
+        adapter = storage.requests.get_adapter("https://webdav.example/")
+        retry = adapter.max_retries
+
+        self.assertEqual(retry.total, storage.RETRY_TOTAL)
+        for method in ("PUT", "DELETE", "MKCOL", "PROPFIND"):
+            self.assertIn(method, retry.allowed_methods)
+
+        for status_code in (429, 502, 503, 504):
```
gounux: Also here, can we use DRF's constants ?

e.g.

```python
from rest_framework import status

for status_code in (status.HTTP_429_TOO_MANY_REQUESTS, ...):
```

gounux (commented):
Thanks for the contribution @t-book ! It's appreciated that we make this `QfcWebDavStorage` more resilient :)

Here is a first review round, I'll checkout locally and test it more deeply, will likely come with feedback, questions and reviews,

> QfcWebDavStorage works against the bytemark/webdav:2.4 container in the dev stack, but breaks against modern WebDAV servers. Most visibly against Nextcloud/SabreDAV, where a file upload crashes with a 500 during make_collection: a HEAD-based existence check misreads SabreDAV's 405 Method Not Allowed as "collection absent" and then crashes on the redundant MKCOL.

Well at the moment we have it plugged and working with several NextCloud instances, which makes me curious how I could be able to reproduce those crashes you mention.  
Would you have some NextCloud logs // reproduction steps in order to be able to face the mentioned issues ?

docker-app/qfieldcloud/filestorage/backend.py
suricactus: Unless required, I would slightly prefer to keep them as constants in the file itself.

Also, if it is possible to avoid non-scalar constants will make our life easier.

t-book:
@gounux 


**Thanks for your review — really appreciate it!**

> Well at the moment we have it plugged and working with several NextCloud instances, which makes me curious how I could be able to reproduce those crashes you mention.


I tested this against two Nextcloud instances (tab.digital, client installation behind vpn) and was able to reproduce the behaviour there, but I can also try to set up a minimal reproduction case to confirm it more systematically.

One clarification: my earlier wording was probably a bit too strong. The behaviour we’re seeing with 405 on HEAD for collection paths is actually consistent with how SabreDAV-based WebDAV servers can behave — HEAD is not reliably supported for directories and may return 405 Method Not Allowed depending on the server configuration and version.

At the same time, it seems likely that this is not fully uniform across all Nextcloud deployments. Differences in WebDAV handling (SabreDAV version, settings, or even reverse proxies that normalize HEAD/GET) could explain why it works in your setup but fails in ours, especially on a fresh base path where the MKCOL flow is actually triggered.

I’ll try to narrow down a clean reproduction and can share logs/stack traces once available — that should help confirm whether this is purely environment-specific or a broader WebDAV quirk.

docker-app/qfieldcloud/filestorage/backend.py
t-book: Makes sense, But I would tend to follow @suricactus's preference and keep them in the file. If you both agree they should live in settings.py with a `WEBDAV_STORAGE_*` prefix instead, happy to move them

docker-app/qfieldcloud/filestorage/backend.py
t-book: Good suggestion, will do! From what I see urllib3.Retry(status_forcelist=...) expects integers ([urllib3 docs](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry)), and DRF status constants are [plain ints](https://github.com/encode/django-rest-framework/blob/main/rest_framework/status.py), so they should work directly:
```
from rest_framework import status

RETRY_STATUS_FORCELIST = (
    status.HTTP_429_TOO_MANY_REQUESTS,
    status.HTTP_502_BAD_GATEWAY,
    status.HTTP_503_SERVICE_UNAVAILABLE,
    status.HTTP_504_GATEWAY_TIMEOUT,
)
```

docker-app/qfieldcloud/filestorage/backend.py
t-book: Same answer as above — happy to follow whichever direction you agree on. These are common defaults for a `requests.Session` transport adapter, so keeping them as class attributes felt reasonable. But if there's a project convention to centralize connection pool config in settings.py, I'll move them there.

docker-app/qfieldcloud/filestorage/backend.py
t-book: That’s a fair concern! I agree the naming is a bit misleading.

In my understanding, `raise_on_status=False` doesn't silently ignore errors—it just ensures that when retries are exhausted for the codes in our forcelist (429, 502, etc.), we get a Response object back instead of a low-level urllib3 exception.

**I went this route because:**

1. **Exception Compatibility:** It ensures we consistently raise a standard `requests.HTTPError` (via `raise_for_status()`), which the existing except blocks are already set up to catch. If set to `True`, a persistent failure surfaces as a `MaxRetryError`, which might bypass those blocks?
3. **Access to Error Details:** Getting the Response object back even after a failed retry allows us to inspect the headers or the error body (like a 'Quota Exceeded' message) which would be lost if an exception is raised by the retry engine itself?

Does that make sense to you and the codebase?

docker-app/qfieldcloud/filestorage/backend.py
t-book: Good question. Looking at that definition, `raise_for_status()` only raises for 4xx/5xx — 2xx passes through silently.

**On 201:** Listing 201 in the guard is technically redundant, but I included it mainly for readability: these are the documented outcomes for MKCOL, and it saves the reader from having to remember how `raise_for_status()` classifies 2xx? Happy to drop 201 from the check if you prefer — the behavior would be identical I'd say.

**On 405/409:** Per [RFC 4918](https://www.rfc-editor.org/rfc/rfc4918#section-9.3.1) §9.3.1, MKCOL is intended for unmapped URLs (405), and 409 handles intermediate collection conflicts. From what I've observed, some WebDAV servers return these codes if the folder already exists. By accepting them, we ensure the operation is idempotent rather than failing because a directory is already in place?

docker-app/qfieldcloud/filestorage/backend.py
t-book: absolutely (:

docker-app/qfieldcloud/filestorage/tests/test_webdav.py
t-book: Sure, good point! Makes sense to keep tests and production code consistent!

docker-app/qfieldcloud/filestorage/backend.py
t-book: I am happy to share my thoughts ~
My reasoning was about managing peak memory usage during the file transfer.
Without `stream=True`, requests immediately buffers the entire file in its internal memory before we can interact with it. By using stream=True, we can iterate over chunks as they arrive instead.
To verify this approach, I tested with a real-world GeoPackage file and peak measuring:

**Without streaming: 856.48 MB peak memory
With streaming (processing chunks without joining): 16.03 MB peak memory**

However, I realize my current PR code uses` b"".join(response.iter_content(...))` which would still load all chunks into memory. So if my test is right, it demonstrates the potential, but my implementation doesn't fully achieve it yet.

```python
(e) tonischonbuchner@Tonis-MacBook-Pro /tmp/p $% cat test.py 
import tracemalloc
import requests

url = "https://naciscdn.org/naturalearth/packages/natural_earth_vector.gpkg.zip"

tracemalloc.start()
response = requests.get(url, stream=False)
data = response.content
_, peak_without = tracemalloc.get_traced_memory()
tracemalloc.stop()

tracemalloc.start()
response = requests.get(url, stream=True)
data = b"".join(response.iter_content(chunk_size=8*1024*1024))
_, peak_with_join = tracemalloc.get_traced_memory()
tracemalloc.stop()

tracemalloc.start()
response = requests.get(url, stream=True)
for chunk in response.iter_content(chunk_size=8*1024*1024):
    len(chunk)
_, peak_proper = tracemalloc.get_traced_memory()
tracemalloc.stop()

print(f"Without stream=True:      {peak_without / 1024**2:.2f} MB")
print(f"With stream + join():     {peak_with_join / 1024**2:.2f} MB")
print(f"With stream + chunks:     {peak_proper / 1024**2:.2f} MB")

(e) tonischonbuchner@Tonis-MacBook-Pro /tmp/p $% python test.py
Without stream=True:      856.48 MB
With stream + join():     851.38 MB
With stream + chunks:     16.03 MB

```

If you think the streaming idea is worth pursuing, I could refactor to process chunks directly without joining them, which would deliver the full memory benefit shown in the test.

docker-app/qfieldcloud/filestorage/backend.py
t-book: You're right that the general HTTP spec defines 405 as "Method Not Allowed". The parenthetical "(URL already mapped)" is specifically what 405 means in the context of MKCOL — [RFC 4918 §9.3.1](https://www.rfc-editor.org/rfc/rfc4918#section-9.3.1) says:

"405 (Method Not Allowed) - MKCOL can only be executed on an unmapped URL."

I agree my comment could be clearer about that distinction. I'd rephrase it something like?
```
# 405: "MKCOL can only be executed on an unmapped URL" (RFC 4918 §9.3.1),
# which in practice means the collection already exists.
```

docker-app/qfieldcloud/filestorage/backend.py
gounux: Thank you for the test / benchmark, that says it all !

Then I guess `stream + chunk` seems like the way to go.

docker-app/qfieldcloud/filestorage/backend.py
gounux: Okay :) then let's forget the settings, could those go on top of the file just after the imports ?

And I don't think that the `WEBDAV` prefix would be relevant anymore, since they live in this specific file,

docker-app/qfieldcloud/filestorage/backend.py
gounux: Let's forget the settings.py then :)

gounux:
Thanks @t-book, really looking forward making this WebDAV storage stronger and usable on more backends !

Would you mind rebasing to latest master ?

#1478 just got merged, which gets rid of almost 4k code lines regarding the old legacy storage,  
Would be nice to relaunch the CI tests :)

t-book:
> https://github.com/opengisch/QFieldCloud/pull/1478 just got merged, which gets rid of almost 4k code lines regarding the old legacy storage,

Wow wow wow this is great! 

> Would you mind rebasing to latest master ?

Good idea! I would also update my PR based on your comments above. Today is a bit busy so this will need some time. 

Thanks for your review @gounux !

docker-app/qfieldcloud/filestorage/backend.py
t-book: @gounux I worked on this and will also test it in live nextcloud instances.

t-book:
Good morning @gounux I've rebased and updated the code based on your comments. Lets see what tests do. Also I will test latest changes against some live nextcloud / webdav instances.

t-book:
Could someone please restart the test? Looks Like a build error not even reaching the Code. Locally they passed.

t-book:
Hi @gounux and @suricactus ,

First of all, thanks again for the thorough and thoughtful review.

However, after going through the failing tests, I realize things are becoming a bit convoluted. Rather than continuing down this path and taking up more of your time, I think it's better to take a step back and restart with a cleaner, more structured approach. 

Instead of one large PR, I'd like to open separate issues first for each improvement so we can discuss and refine each point properly. Here's what I'm planning:

| **Issue** | **Goal** |
|---|---|
| Streaming Memory Optimization | Reduce peak memory usage for large files | 
| Timeout Management | Better handling of slow uploads/downloads |
| Retry & Resilience | Auto-recovery from transient errors | 
| Error Handling | Clearer error messages for debugging |
| Race Condition Safety | Handle concurrent operations safely |
| Code Maintainability | DRF constants, visible configuration |

This should make everything clearer and less time-consuming for everyone.

Thank you again for your patience and generosity with your time!

Best regards,
Toni

suricactus:
@t-book splitting your work in smaller chunks can only be appreciated. Thank you for your understanding and collaboration. Happy to see your upcoming issues and following PRs!

PR #1556, opened 2026-04-21
Package cloud project using QGZ to save bandwidth

nirvn:
I was debugging why a ~600kb project turned into a 1.4mb download when packaged by QFieldCloud. Turns out my source project was using the .qgz format while QFieldCloud packages using the .qgs format.

I'm proposing we move to .qgz, for two reasons.

- First, the file size of a .qgs is _exponentially_ larger. In this project I was looking at, the source project file fieldtm.qgz weights 44,215 bytes. The exported fieldtm_qfield.qgs (and its _attachments.zip sidecate) size is 884,950 bytes. That's a 20-fold increase. In my part of the world, it amounts extra seconds of waiting.

- Second, the .qgs always comes with the _attachments.zip sidecar, which forces _two file downloads_ instead of a single .qgz.

Members of the jury, I rest my case.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_packages.py
+++ b/docker-app/qfieldcloud/core/tests/test_packages.py
@@ -4,7 +4,6 @@ import logging
 import os
 import tempfile
 import time
-import zipfile
 
 from django.http import FileResponse
 from django.test import tag
@@ -28,6 +27,7 @@ from qfieldcloud.core.models import (
 from qfieldcloud.core.tests.mixins import QfcFilesTestCaseMixin
 from qfieldcloud.core.tests.utils import (
     get_test_postgis_connection,
+    open_qgis_file,
     setup_subscription_plans,
     testdata_path,
     wait_for_project_ok_status,
@@ -258,12 +258,11 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         )
 
         local_file = os.path.join(tempdir, "project_qfield.qgz")
-        with zipfile.ZipFile(local_file, "r") as qgz:
-            with qgz.open("project_qfield.qgs") as qgs:
-                self.assertEqual(
-                    qgs.readline().strip(),
-                    b"<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>",
-                )
+        with open_qgis_file(local_file) as f:
+            self.assertEqual(
+                f.readline().strip(),
+                b"<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>",
+            )
 
     def test_list_files_for_qfield_broken_file(self):
         self.upload_files(
@@ -347,11 +346,10 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         )
 
         local_file = os.path.join(tempdir, "project_qfield.qgz")
-        with zipfile.ZipFile(local_file, "r") as qgz:
-            with qgz.open("project_qfield.qgs") as qgs:
-                for line in qgs:
-                    if 'name="theMapCanvas"' in str(line):
-                        return
+        with open_qgis_file(local_file) as f:
+            for line in f:
+                if 'name="theMapCanvas"' in str(line):
+                    return
 
     def test_download_project_with_broken_layer_datasources(self):
         self.upload_files_and_check_package(
--- a/docker-app/qfieldcloud/core/tests/test_packages.py
+++ b/docker-app/qfieldcloud/core/tests/test_packages.py
@@ -4,6 +4,7 @@ import logging
 import os
 import tempfile
 import time
+import zipfile
 
 from django.http import FileResponse
 from django.test import tag
@@ -207,8 +208,7 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
             ],
             expected_files=[
                 "data.gpkg",
-                "project_qfield.qgs",
-                "project_qfield_attachments.zip",
+                "project_qfield.qgz",
             ],
         )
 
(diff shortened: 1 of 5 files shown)


suricactus:
I will be able to review only next week.

nirvn:
@suricactus , no worries, there is rush per say. I did however want to open the PR now as I don't want to forget this. I feel it's a pretty worthy thing to unlock here.

t-book:
@nirvn great find!

docker-qgis/requirements_internal.txt
```diff
@@ -1,3 +1,3 @@
 # NOTE dependency versions should be defined in the `*.tar.gz` format, not `git+https://` to make `wheel` happy
-libqfieldsync @ git+https://github.com/opengisch/libqfieldsync@6eeff164d28b899b997b0dcee14977781b6a0e7e
-xlsform2qgis @ git+https://github.com/opengisch/xlsform2qgis@ab8f8014ea4465a454f709f733fcd173dd5bdd24
+libqfieldsync @ git+https://github.com/opengisch/libqfieldsync@4b2fd8eb2435433e4e480e976b90775c6da5d823
+xlsform2qgis @ git+https://github.com/opengisch/xlsform2qgis@973479e6e1ecffc566063f3b65fadec7ab150119
```
suricactus: This is unrelated change, can you please add it in a separate PR?

nirvn: Arg, my bad! I meant to update libqfieldsync once again to match a revision on master/main. Fixing.

nirvn: @suricactus , fixed.

docker-app/qfieldcloud/core/tests/test_packages.py
```diff
@@ -252,18 +252,18 @@ def test_download_file_for_qfield(self):
...
-                f.readline().strip(),
-                "<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>",
-            )
+        local_file = os.path.join(tempdir, "project_qfield.qgz")
+        with zipfile.ZipFile(local_file, "r") as qgz:
+            with qgz.open("project_qfield.qgs") as qgs:
+                self.assertEqual(
+                    qgs.readline().strip(),
+                    b"<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>",
+                )
```
suricactus: Ideally we want this to be a function of it's own. It is going to be used a lot. It's already two times here, there are a bunch of nice-to-have tests yet to be written that will need it too.

`def open_qgis_file(path: str | Path) -> <file-descriptor>`

So we can write it as simple as:


```suggestion
        local_file = os.path.join(tempdir, "project_qfield.qgz")
        with open_qgis_file(local_file, "r") as f:
            self.assertEqual(
                qgs.readline().strip(),
                b"<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>",
            )
```

Such function should live in `core/tests/utils.py`.

nirvn: Good idea, we can then actually handle both .qgz and .qgs.

Is that working for you?

```python
def open_qgis_file(path: str | Path) -> TextIO:
    if isinstance(path, str):
        path = Path(path)
    
    suffix = path.name.lower()[-4:]
    if suffix == ".qgz":
        with zipfile.ZipFile(path, 'r') as qgz:
            return qgz.open(f"{path.name[:-4]}.qgs")
    elif suffix == ".qgs":
        return open(path)

    return None
```

nirvn: (@suricactus , forgot to include user handle)

suricactus: Pathlib's Path has a more robust way of getting the extension (suffix). 

Also no need for the if statement on the first line, just pass it `path = Path(path)`.

Not sure if the `with` will work properly when you return None. You better throw.

The usage from my initial comment should be like this, the implementation is more flexible.

nirvn: @suricactus , updated

```python
def open_qgis_file(path: str | Path) -> TextIO:
    path = Path(path)
    
    if path.suffix.lower() == ".qgz":
        with zipfile.ZipFile(path, 'r') as qgz:
            return qgz.open(f"{path.name[:-4]}.qgs")
    elif path.suffix.lower() == ".qgs":
        return open(path)

    raise Exception("The QGIS project file could not be opened")
```

docker-app/qfieldcloud/core/tests/utils.py
```diff
@@ -207,3 +209,15 @@ def get_test_postgis_connection() -> psycopg2.extensions.connection:
...
 
     return conn
+
+
+def open_qgis_file(path: str | Path) -> TextIO:
+    path = Path(path)
+
+    if path.suffix.lower() == ".qgz":
+        with zipfile.ZipFile(path, "r") as qgz:
+            return qgz.open(f"{path.name[:-4]}.qgs")
```
suricactus:
```suggestion
            return qgz.open(f"{path.stem}.qgs")
```

docker-app/qfieldcloud/core/tests/utils.py
```diff
@@ -207,3 +209,15 @@ def get_test_postgis_connection() -> psycopg2.extensions.connection:
     cursor.close()
 
     return conn
+
+
+def open_qgis_file(path: str | Path) -> TextIO:
```
suricactus:
```suggestion
def open_qgis_file(filename: str | Path) -> TextIO:
```

Most of the source code already prefers `filename` as variable name when it is certainly a file name, and `path` when we don't know if it is a file or dir path.

docker-app/qfieldcloud/core/tests/utils.py
nirvn: @suricactus , does stem provide the full name prior to the last .extension? I.e. if my filename "project.house.two.qgz", will stem give us "project.house.two"? Let me try :)

nirvn: Answer: yes.

```
>> p = Path("/tmp/project.house.one.qgz")
>> p.stem
'project.house.one'
```

suricactus (approved):
Looks good to me now. Please squash all into one PR and let's hope the tests pass.

Thanks for the contribution!

PR #1557, opened 2026-04-21
fix: change mounted repository for custom ca bundles

gounux:
Since `/etc/ssl/certs` is used for storing CAs trusted by the system, we should use an empty repository for custom CA bundles.

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -1006,5 +1006,5 @@ QFIELDCLOUD_CUSTOM_CA_VOLUME_NAME = (
 )
 
 # Filename where optional custom CA certificate are stored.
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8143 Use an empty repository for mounting custom ca bundle](https://app.clickup.com/t/2192114/QF-8143)

suricactus (approved):
Thanks for the fix. I am just curious how did we miss this during testing?

gounux:
Thanks @suricactus !

>  I am just curious how did we miss this during testing?

I guess working with a local-only setup did not help spotting this, e.g. local rustfs, smtp4dev, etc.

We spotted this when a QFieldCloud instance was trying to access a real mail server using https.

PR #1558, opened 2026-04-22
Use geographical coordinates to store project seed extent

suricactus:
The reason is that most of the CRS are based on 4326 and the conversion between the CRS-es will result in more correct reprojections.

I got the wrong feeling that using a geographical CRS will cause issues.

It's more flexible. Might be problematic for projects for celestial bodies :moon:  :)


duke-nyuki:
Task linked: [QF-8147 Store the project seed extent in EPSG:4326 instead of EPSG:3857](https://app.clickup.com/t/2192114/QF-8147)

Rakanhf (approved):
LGTM 🚀

PR #1561, opened 2026-04-22
Actually respect the renderer timeout and return None if no thumbnail can be generated

suricactus:
For unknown reason when we have `QgsMapRendererCustomPainterJob` and we wait for it to finish with `job.waitForFinishedWithEventLoop()`, that call hangs forever without actually producing rendering the thumbnail. Actually the reason is known - the job never gets active, or in other words `job.isActive()` is always `False`. But we don't know why.

Even though we have a timeout `QTimer` that is supposed to control how long we wait for the renderer to finish, the
`timer.timeout.connect(lambda: job.cancel())` did not actually cancel the job, because `job.cancel()` will do nothing on a job that never been active.

That's why we add a check right after calling `job.start()` if the `job.isActive()` and only then `job.waitForFinishedWithEventLoop()`.

If the job is not active, then we just don't produce a new thumbnail.

---

Alternative solution is to call `job.renderSynchronously()`, but then we no longer have the timeout to control the rendering within reasonable time interval.

---

The most probable cause for failing renderer is when the `QgsProject` was created via `pyqgis` API. Most probably the error is caused by wrong `QgsMapSettings`, ~but it has not been what is wrong in `QgsMapSettings` yet~. EDIT:  https://github.com/opengisch/QFieldCloud/pull/1560 is fixing the problem.

This PR is still very much relevant, because we don't know what QGIS project people will upload to QFieldCloud.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_jobs.py
+++ b/docker-app/qfieldcloud/core/tests/test_jobs.py
@@ -422,3 +422,24 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
         self.assertEqual(fields[20]["type"], "String")
         self.assertEqual(fields[21]["name"], "comment")
         self.assertEqual(fields[21]["type"], "String")
+
+    def test_thumbnail_generation_with_wrong_extent_does_not_hang(self):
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-8148 Ensure the thumbnail generation does not hang forever](https://app.clickup.com/t/2192114/QF-8148)

gounux (approved):

PR #1562, opened 2026-04-23
Fix `createbuckets` by passing the arguments as list of args instead of a str

suricactus:
Let's pray :pray:

### DIFF
--- a/docker-createbuckets/createbuckets.py
+++ b/docker-createbuckets/createbuckets.py
@@ -37,25 +37,39 @@ for storage_name, storage_config in STORAGES.items():
     bucket_name = storage_config["OPTIONS"]["bucket_name"]
 
     subprocess.run(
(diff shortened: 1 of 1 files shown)


docker-createbuckets/createbuckets.py
```diff
@@ -37,25 +37,39 @@ class DjangoStorages(TypedDict):
...
+        ],
         shell=True,
         check=True,
     )
 
     subprocess.run(
-        f"/usr/bin/mc mb --ignore-existing {storage_name}/{bucket_name}",
+        ["/usr/bin/mc", "mb", "--ignore-existing", f"{storage_name}/{bucket_name}"],
         shell=True,
     )
```
gounux, line 56: Should we maybe add `check=True` here ?

It seems that the bucket creation could fail silently without this check, which is there in the other steps using `/usr/bin/mc`,

suricactus:
Fixed by https://github.com/opengisch/QFieldCloud/pull/1563

PR #1563, opened 2026-04-23
chore: replace `mc` with `rc` in the createbuckets local service

gounux:
Since local QFieldCloud is making use of rustfs and not minio anymore, it might be more logic to create local buckets with rustfs cli with the `rc` command tool.

### DIFF
--- a/docker-createbuckets/Dockerfile
+++ b/docker-createbuckets/Dockerfile
@@ -6,8 +6,9 @@ RUN apt-get update \
     python3 \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8149 Replace `mc` with `rc` in createbuckets](https://app.clickup.com/t/2192114/QF-8149)

suricactus (approved):

t-book:
@gounux with this PR merged 
https://github.com/opengisch/QFieldCloud/blob/master/docker-compose.override.minio.yml
will not work anymore, right? If so, it might be helpful to leave a comment somewhere for users trying the minio compose file (or still need to use it for standalone)?

suricactus:
No, we will remove https://github.com/opengisch/QFieldCloud/blob/master/docker-compose.override.minio.yml instead :)

PR #1567, opened 2026-04-28
feat: add storage threshold fields to Plan model

Rakanhf:
Now the clients are hardcoding the value when to show warning  and critical  messages. However this is something that should be configured at Plan  creation by the marketing team.

- Added `storage_threshold_warning_bytes` and `storage_threshold_critical_bytes` fields to `Plan` model
- Added validation logic to ensure thresholds are greater than zero and correctly ordered relative to each other and the total storage
- Updated the admin interface to display the new fields
- Added a migration to set default values based on existing `storage_mb` values
- Add validation to reject 0 values for `storage_mb`
- Refactored model validation to use django's model validation methods
- Added storage threshold fields to current subscription endpoint
- Added tests


duke-nyuki:
Task linked: [QF-8094 Add new fields to the Subscription.Plan when the warning and critical messages should be shown and return them at /subscriptions/<u>/current API](https://app.clickup.com/t/2192114/QF-8094)

docker-app/qfieldcloud/subscription/serializers.py
```diff
@@ -26,6 +32,9 @@ class Meta:
             "storage_used_bytes",
             # how many bytes of storage we have in the current subscription
             "active_storage_total_bytes",
+            # remaining bytes thresholds for client-side storage warnings
+            "storage_threshold_warning_bytes",
+            "storage_threshold_critical_bytes",
```
suricactus:
```suggestion
            "plan_storage_threshold_warning_bytes",
            "plan_storage_threshold_critical_bytes",
```

docker-app/qfieldcloud/subscription/tests/test_plans.py
```diff
@@ -0,0 +1,32 @@
...
+    def setUp(self):
+        setup_subscription_plans()
+
+    def test_plan_storage_threshold_validation(self):
+        plan = Plan.objects.first()
+
+        # update storage_threshold_warning_bytes to 0
+        plan.storage_threshold_warning_bytes = 0
+        plan.storage_threshold_critical_bytes = 0
+        with self.assertRaises(ValidationError):
```
suricactus:
```suggestion
        with self.assertRaisesRegex(ValidationError, "must be greater than 0"):
```

Avoid checks only for the error class.

suricactus (approved):
Nice work! 

Two tiny comments. Also please fix the PR title ;)

Rakanhf:
@suricactus Adjusted some tests, since they were failing, appreciate it if you give it a second look.

docker-app/qfieldcloud/subscription/migrations/0010_plan_storage_threshold_critical_bytes_and_more.py
```diff
@@ -0,0 +1,48 @@
...
+    dependencies = [
+        ("subscription", "0009_subscription_purchased_seats"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="plan",
+            name="storage_threshold_critical_bytes",
+            field=models.PositiveBigIntegerField(
+                default=1000000, help_text="Remaining storage threshold critical bytes."
```
suricactus: Stick to (lack of) underscore in the same file ;)

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
```diff
@@ -64,9 +66,30 @@ def setUp(self):
...
+        plan.storage_threshold_critical_bytes = 100_000
         plan.save()
 
+        # Simulate over-quota while keeping plan valid (storage_mb must be > 0).
+        self.more_bytes_than_plan = plan.storage_mb * 1000 * 1000 + 1
+
+        # TODO Delete with QF-4963 Drop support for legacy storage
+        if self.project.uses_legacy_storage:
+            self.project.file_storage_bytes = self.more_bytes_than_plan
+            self.project.save(update_fields=["file_storage_bytes"])
```
suricactus: Usually we avoid "faking" it like this, we would upload a real file. Since we are very close to merging #1478 , no need to change. Actually it's a race whose PR will be ready earlier :) .

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -133,6 +133,16 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
     # it's a simple fixed amount for the entire plan.
     storage_mb = models.PositiveIntegerField(default=10)
 
+    # Remaining bytes thresholds for storage usage warnings.
+    storage_threshold_warning_bytes = models.PositiveBigIntegerField(
+        default=2_000_000,
```
```suggestion
        default=0,
```

TBH having default of `0` makes more sense here, than an arbitrary number. What if you want to setup a Plan with just 100MB storage, then you have to edit this.

Setting to 0 will enforce the support to ensure they fill in the data :)

suricactus (commented):
A few small comments:

- legacy storage patch - not great, not terrible since we are almost there in removing it
- default value for thresholds - I would skip it and set the default to 0, so the user can see immediate feedback if something is wrong.

Anyways, good job! And sorry you had to fight with the legacy storage, but now you know more about it ;) .

docker-app/qfieldcloud/subscription/models.py
Rakanhf: But this will conflict with the required validations, however it will remove a lot of changes in this PR, since a lot of them are driven by non 0 value, which also related to the `storage_mb` being 0. I'm not sure if I should revert the whole changes and make things simpler ? or should we keep the validation now as is ?

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
Rakanhf: Yes, that's why I really didn't want to go fancy about the patch since it will be removed anyway very soon.

Rakanhf:
@suricactus Sorry for the annoying force push, all comments are addressed and commit history is cleaned.

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
```diff
@@ -64,9 +66,30 @@ def setUp(self):
...
+        plan.storage_mb = 1
+        plan.storage_threshold_warning_bytes = 200_000
+        plan.storage_threshold_critical_bytes = 100_000
         plan.save()
 
+        # Simulate over-quota while keeping plan valid (storage_mb must be > 0).
+        self.more_bytes_than_plan = plan.storage_mb * 1000 * 1000 + 1
+
+        # TODO Delete with QF-4963 Drop support for legacy storage
+        if self.project.uses_legacy_storage:
```
suricactus: No more!

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
```diff
@@ -93,6 +116,19 @@ def setUp(self):
     def tearDown(self):
         self.conn.close()
 
+    def _re_simulate_legacy_over_quota(self):
```
suricactus: Kill this

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -228,12 +244,58 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
...
+        if self.storage_threshold_warning_bytes >= self.storage_bytes:
+            errors.setdefault("storage_threshold_warning_bytes", []).append(
+                _("must be less than storage_mb.")
+            )
+
+        if errors:
+            raise ValidationError(errors)
+
+    def clean(self):
+        # storage_mb must be greater than 0
```
suricactus: Do we really need this kind of comments?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -228,12 +244,58 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
...
+        if self.storage_mb == 0:
+            raise ValidationError({"storage_mb": _("must be greater than 0.")})
+
+        # user_type must be PERSON or ORGANIZATION
```
suricactus: And this?

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -228,12 +244,58 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
     def storage_bytes(self) -> int:
         return self.storage_mb * 1000 * 1000
 
-    def save(self, *args, **kwargs):
+    def clean_storage_threshold_bytes(self):
+        errors = {}
+
+        if self.storage_threshold_warning_bytes == 0:
+            errors.setdefault("storage_threshold_warning_bytes", []).append(
+                _("must be greater than 0.")
```
suricactus: Is there a reason to use lower-case? Feels a bit weird to have a sentence with the anyways mandatory dot in the end, but starting with lower case.


```suggestion
                _("Must be greater than 0.")
```

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
```diff
@@ -64,9 +66,25 @@ def setUp(self):
...
 
         plan = subscription.plan
-        plan.storage_mb = 0
+        plan.storage_mb = 1
+        plan.storage_threshold_warning_bytes = 200_000
+        plan.storage_threshold_critical_bytes = 100_000
         plan.save()
 
+        # Simulate over-quota while keeping plan valid (storage_mb must be > 0).
+        self.more_bytes_than_plan = plan.storage_mb * 1000 * 1000 + 1
```
suricactus: Why add to `self`? Avoid adding stuff to self unless neccessary.

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
```diff
@@ -64,9 +66,25 @@ def setUp(self):
...
+        file_version = FileVersion.objects.add_version(
+            project=self.project,
+            filename="bigfile.name",
+            content=ContentFile(b"x", "dummy.name"),
+            file_type=File.FileType.PROJECT_FILE,
+            uploaded_by=self.user,
+        )
+        FileVersion.objects.filter(pk=file_version.pk).update(
+            size=self.more_bytes_than_plan
+        )
```
suricactus: Instead of manipulating the metadata, just upload the "large" file. If the "big file" is very, very big, then decrease the plan storage. I think in this case 1mb is very far from considered "big", so you can easilu simulate uploading such file.

suricactus (commented):
Last round of feeddback I guess.

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -228,12 +244,58 @@ def get_plans_for_user(cls, user: User, user_type: User.Type) -> QuerySet["Plan"
...
+                _("must be less than storage_mb.")
+            )
+
+        if errors:
+            raise ValidationError(errors)
+
+    def clean(self):
+        # storage_mb must be greater than 0
+        if self.storage_mb == 0:
+            raise ValidationError({"storage_mb": _("must be greater than 0.")})
```
suricactus: Please check if there are no plans configured with 0 MB on app.qfield.cloud with a simple query :)

Rakanhf: Ran a query and there are no plans with 0 storage.

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
Rakanhf: Great catch, this is a leftover from the legacy storage workaround

docker-app/qfieldcloud/core/tests/test_file_upload_policy.py
Rakanhf: Yes, I was a bit concerned that it will take more time, now I testes both ways, didn't find any difference.

suricactus (approved):
Just one note here.

We have change previously available behaviour - having a plan with `storage_mb = 0` and controlling the storage of plan entirely by `StoragePackage`-s.

We do not use it anywhere, but if we need to enable it in the future, we should make the new threshold fields AND the storage_mb equal to 0 exempt from exceptions.

PR #1571, opened 2026-04-29
Bump dependencies within the qgis worker container

suricactus:
See individual commits for more info.


gounux (approved):
Thanks for the changelogs & diff links, no breaking changes spotted,

Let's wait for the CI tests to finish successfully I presume,

PR #1578, opened 2026-05-05
Insure that the QFieldCloud logo provided through /server/info/ is dark theme friendly

nirvn:
While we're adding whitelabeling support in QField, we stumbled on this issue whereas the default QFC logo is not so friendly in the context of a dark theme.

In QField, we historically added a white buffer to Nyuki, which looks like this:

image

When switching to the QFC server provided logo, we end up with:

image

IMHO, that's a visual regression :) this PR fixes this by adding a white buffer on the QFC server hosted logo.


nirvn:
@suricactus , as discussed.

@Rakanhf , something spotted while implementing the endpoint you coded :)

suricactus (approved):

suricactus (approved):
Not sure how this looks in QFieldSync light theme, but I think it's better anyways.

PR #1582, opened 2026-05-06
fix: remove `parser_classes` from `ProjectViewSet`

Rakanhf:
The CLI/SDK was failing to create projects since the `create_project` function in sdk see [sdk.py L510](https://github.com/opengisch/qfieldcloud-sdk-python/blob/cbad4a966007e5a447eaccee916576fe5185c8c0/qfieldcloud_sdk/sdk.py#L510) is sending a request with the content type `application/x-www-form-urlencoded`.

To fix this issue I removed the `parser_classes` since the default will include all three `JSONParser`, `MultiPartParser` and `FormParser` see [rest settings](https://github.com/encode/django-rest-framework/blob/22e231cf2f77b4cfe929de875d958b93916b1a8b/rest_framework/settings.py#L35)

Reproduced Using cli e.g:

```
qfieldcloud-cli create-project 'TestProject'
```

### DIFF
--- a/docker-app/qfieldcloud/core/views/projects_views.py
+++ b/docker-app/qfieldcloud/core/views/projects_views.py
@@ -25,7 +25,7 @@ from qfieldcloud.subscription.exceptions import QuotaError
 from rest_framework import filters as drf_filters
 from rest_framework import generics, permissions, status, viewsets
 from rest_framework.decorators import action
-from rest_framework.parsers import JSONParser, MultiPartParser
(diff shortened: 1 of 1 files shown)


suricactus (approved):
Thanks!

"Test Project" will always fail as we don't allow spaces in names ;)

Rakanhf:
> Thanks!
> 
> "Test Project" will always fail as we don't allow spaces in names ;)

Nice catch, Locally I was testing with random "demo5376", I will adjust the pr description.

PR #1585, opened 2026-05-08
Automatically render foreign keys in Django admin

suricactus:
Add a new `ModelAdminListDisplayAutoFkeyFields` that list of fields to
be automatically converted to links to their corresponding foreign
model. Useful for fields such as `created_by`. It is tedious and error prone to define them one by one.

When set to `True`, all foreign key fields in `list_display` will be automatically converted to links.
When set to `False`, no fields will be automatically converted to links.
When set to an iterable of field names, only those fields will be converted to links.

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -1000,9 +1000,9 @@ class SecretAdmin(QFieldCloudModelAdmin):
         "name",
         "type",
         "assigned_to",
-        "project__name",
+        "project",
         "organization",
-        "created_by__link",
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8213 Automatically render foreign keys in Django admin](https://app.clickup.com/t/2192114/QF-8213)

Rakanhf (approved):
LGTM ! But i wonder what's the motivation behind this PR ?

suricactus:
> Useful for fields such as created_by.

It is tedious to define them one by one.

PR #1586, opened 2026-05-08
chore: Add more debug options within the app container

suricactus:
`memray` and `py-spy` proved useful and it should be easy for us to enable them whenever needed. Adding them to the local dev container in debug mode will be handy to monitor how the Django app behaves.

Sample use locally:

```
py-spy top --pid $(pgrep -f ppid)
py-spy attach --pid $(pgrep -f ppid)

```


duke-nyuki:
Task linked: [QF-8212 Add more ready-to-use debug tools in the debug build](https://app.clickup.com/t/2192114/QF-8212)

toebivankenoebi:
Makes a lot of sense, looks good to me

toebivankenoebi (approved):

PR #1542, opened 2026-04-08
Make both QGIS `3.44.10` to `4.0.2` version available

suricactus:
Unfortunately QGIS 4 is not built for ubuntu 24.04 noble and we have to push the base image too. This needs extensive testing before pushing forward.

Fix #1533 

Waits on https://github.com/opengisch/QFieldCloud/pull/1631

PR #1587, opened 2026-05-08
feat: Bump Python base image from 3.10 boookworm to 3.14 trixie

suricactus:
Besides a few deprecations/removals and a ton of performance
optimizations and syntax sugar, the new image works without any
problems.


duke-nyuki:

docker-app/Dockerfile
```diff
@@ -3,7 +3,7 @@
 ##########################
 
 # pull a builder image, the same as the base
-FROM python:3.10-slim-bookworm AS build
+FROM python:3.14-slim-trixie AS build
```
gounux, line 6: I believe we should make sure to target python >= 3.14.5, seems that the new incremental Garbage Collector _had a number of reports of significant memory pressure in production environments._

-> https://discuss.python.org/t/reverting-the-incremental-gc-in-python-3-14-and-3-15/107014

The GC was reverted in 3.14.5

-> https://blog.python.org/2026/05/python-3145rc1/

suricactus: I think by the time we merge this, 3.14.5 or newer will be the 3.14 tagged image.

duke-nyuki:
Task linked: [QF-8234 Drop deprecated unittest methods](https://app.clickup.com/t/2192114/QF-8234)

docker-app/Dockerfile
suricactus: Not yet merged https://hub.docker.com/_/python/tags?name=3.14.5

suricactus: Merged and resolved.

docker-app/Dockerfile
```diff
@@ -109,11 +113,22 @@ USER app
...
+COPY ./requirements/requirements_test.txt requirements/requirements_test.txt
+RUN echo "GDAL_VERSION=$(gdalinfo --version | cut -d, -f1 | cut -d' ' -f2)"
+RUN export GDAL_VERSION=$(gdalinfo --version | cut -d, -f1 | cut -d' ' -f2); echo "GDAL_VERSION=$GDAL_VERSION" && pip3 install $(grep ^fiona requirements/requirements_test.txt)
+
 # a separate stage for webserver test environment
 FROM base AS webserver_test
 ENV LOGGER_SOURCE=app
+
+# Copy fiona only
+COPY --from=build_webserver_test /usr/local/lib/python3.14/site-packages/fiona /usr/local/lib/python3.14/site-packages/fiona/
```
Rakanhf, line 127: I assume that this is a workaround since `Fiona` is not supported for python 3.14 ? 

I checked https://github.com/Toblerity/Fiona & https://pypi.org/project/fiona/ and don't see that it would support 3.14

Rakanhf (commented):
Everything looks good, and tried it locally, Two small things to change 

https://github.com/opengisch/QFieldCloud/blob/1eff0f3275044f23b980e45aef78379033f6c33f/docker-app/qfieldcloud/core/logging/formatters.py#L28

and 

https://github.com/opengisch/QFieldCloud/blob/1eff0f3275044f23b980e45aef78379033f6c33f/docker-app/qfieldcloud/core/admin.py#L598


Since the use of `utcnow()` was deprecated from 3.12 see : https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow


Also I have one question :
1- Shouldn't we also update the workers / demgen, etc ?

docker-app/Dockerfile
suricactus: Yup, fiona seems to be abandoned (https://github.com/Toblerity/Fiona/issues/1504). Since it still builds with Python 3.14 just fine AND is only used for tests, I kept it to reduce the number of changes.

After we get this merged, a easy to do oboarding task is to remove fiona as a dependency and use something else: geopandas, Pyogrio, or something else. I am thinking even of raw gdal, it's for sure going to be maintained.

suricactus:
> Since the use of `utcnow()` was deprecated from 3.12 see : [docs.python.org/3/library/datetime.html#datetime.datetime.utcnow](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)

Thanks for these, will push a commit.

> Also I have one question : 1- Shouldn't we also update the workers / demgen, etc ?

Yes, we will. One by one.

As for the QGIS workers, there is already a PR doing so: https://github.com/opengisch/QFieldCloud/pull/1542

The worker wrapper will be updated as part of this PR, as the worker wrapper is the same as the app image.

I am not sure what do you mean by demgen, there is no such service in this repository.

Rakanhf (approved):
LGTM

PR #1589, opened 2026-05-08
feat: Make the File and FileVersion Admin usable, meaningful, but only readonly

suricactus:
Preivously the File and FileVersion admin were the bare minimum. Viewing/editing is practically unusable.

Now add a few more columns, sorting and search of the fields. The records display readonly. If one needs to upload a file, a dedicated view calling `FileVersion.add_version` must be developed.

Since each `File` instance is having the `latest_file_versions` field, then Django Admin is rendering a dropdown list of all available `FileVersion` records, which is obviously giving great strain on the server.

| before | after |
|-|-|
| image | image |
| image | image |
| image | image |
|  image | image |

The clickable links on the left were useless, as they pointing to the File(Version) model, not to the foreign key.

This will be restored automtically as functionality once #1585 is merged.

### DIFF
--- a/docker-app/qfieldcloud/filestorage/admin.py
+++ b/docker-app/qfieldcloud/filestorage/admin.py
@@ -1,11 +1,25 @@
+from typing import Any
+
 from django.contrib import admin
 
-from qfieldcloud.core.admin import qfc_admin_site
+from qfieldcloud.core.admin import QFieldCloudModelAdmin, qfc_admin_site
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8214 Make the File and FileVersion admin usable, meaningful, and only readonly](https://app.clickup.com/t/2192114/QF-8214)

gounux (approved):
Thanks ! Looks good based on the provided screenshots,

PR #1590, opened 2026-05-08
Remove is_attachment_download_on_demand from the read-only fields within the ProjectSerializer

nirvn:
Discovered while trying to add a [x] on-demand attachments download checkbox within QFieldSync's cloud project properties dialog.


suricactus (approved):
Thanks!

PR #1592, opened 2026-05-09
chore: Add ruff rule BLE001 to checks for except clauses that catch all exceptions

### LINKED ISSUE #1575: Possible code Improvements
### What is the enhancement proposal? What problem are you trying to solve?

While exploring the codebase to learn how things work, I noticed a few patterns that might benefit from small improvements. I'm still new to the project, so please let me know if any of these are intentional or if I'm missing context.

## 1. Mutable default argument in `permission_check`

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

**File:** `docker-app/qfieldcloud/core/permission_check.py`, line 9

The `check_args` parameter uses a mutable list `[]` as its default value. While it doesn't seem to be mutated in the current code, this is a common Python footgun — if anyone ever adds an in-place modification down the line, the shared list would cause hard-to-trace bugs across decorated views.

**Current code:**

```python
def permission_check(perm: str, check_args: list[str | Callable] = []) -> Callable:
```

**Suggested fix:**

```python
def permission_check(perm: str, check_args: list[str | Callable] | None = None) -> Callable:
    if check_args is None:
        check_args = []
```

---

## 2. Bare `except Exception` hides unexpected errors

**File:** `docker-qgis/qfc_worker/utils.py`, around line 88

The code checks whether `Qgis.NoLevel` exists and falls back to setting it manually. The intent is clearly to catch `AttributeError`, but `except Exception` would also swallow unrelated errors and make debugging harder.

**Current code:**

```python
try:
    Qgis.NoLevel
except Exception:
    Qgis.NoLevel = 4
```

**Suggested fix:**

```python
try:
    Qgis.NoLevel
except AttributeError:
    Qgis.NoLevel = 4
```

---

## 3. Implicit `None` return in `wait_for_postgres`

**File:** `docker-app/wait_for_services.py`

If the `while` loop exits without a successful connection and without hitting the `except` branch's timeout path, the function falls through without a return statement, implicitly returning `None` instead of `False`.

**Current code (simplified):**

```python
def wait_for_postgres():
    start_time = time()
    while time() - start_time < TIMEOUT:
        try:
            conn = psycopg2.connect(**config)
            conn.close()
            return True
        except psycopg2.OperationalError as error:
            if time() - start_time < TIMEOUT:
                sleep(INTERVAL)
            else:
                logger.error(...)
                # no return here

    logger.error(f"Could not connect to Postgres within {TIMEOUT} seconds.")
    # falls through — returns None
```

**Suggested fix:**

```python
def wait_for_postgres() -> bool:
    start_time = time()
    while time() - start_time < TIMEOUT:
        try:
            conn = psycopg2.connect(**config)
            conn.close()
            return True
        except psycopg2.OperationalError as error:
            if time() - start_time < TIMEOUT:
                sleep(INTERVAL)
            else:
                logger.error(...)
                return False

    logger.error(f"Could not connect to Postgres within {TIMEOUT} seconds.")
    return False
```

---

## 4. `bool()` on query string gives wrong result for `"false"`

**File:** `docker-app/qfieldcloud/core/views/files_views.py`, lines 110–114

The `skip_metadata` query parameter is parsed with `bool(skip_metadata_param)`, but in Python `bool("false")` evaluates to `True` (any non-empty string is truthy). So a client sending `?skip_metadata=false` would actually *enable* metadata skipping.

**Current code:**

```python
skip_metadata_param = request.GET.get("skip_metadata", "0")
if skip_metadata_param == "0":
    skip_metadata = False
else:
    skip_metadata = bool(skip_metadata_param)  # bool("false") == True
```

**Suggested fix:**

```python
skip_metadata_param = request.GET.get("skip_metadata", "0")
try:
    skip_metadata = bool(int(skip_metadata_param))
except (ValueError, TypeError):
    skip_metadata = False
```

Or, if the API should be more forgiving with string values, a small helper could handle `"true"/"false"/"yes"/"no"/"1"/"0"` consistently.

---

### Agreements

- [x] I commit to keeping my branch up to date with the latest `master` branch until my pull request is merged.
- [x] I commit to writing tests for my enhancement proposal, if applicable.
- [x] I am aware that my enhancement proposal and related pull request may not be accepted.

### Additional context

_No response_

suricactus:
Overly broad except clauses can lead to unexpected behavior, such as catching KeyboardInterrupt or SystemExit exceptions that prevent the user from exiting the program. 
It can also hide problems e.g. AttributeError  or KeyError  that are caused by developers.
Instead of catching all exceptions, catch only those that are expected to be raised in the try block.

Fixes #1575 .

Check individual commits.

See https://docs.astral.sh/ruff/rules/blind-except/.

### DIFF
--- a/docker-qgis/qfc_worker/commands/apply_deltas.py
+++ b/docker-qgis/qfc_worker/commands/apply_deltas.py
@@ -652,7 +652,7 @@ def rollback_deltas(
         # NOTE pathlib operations will raise in case of error
         try:
             layer_backup_path.rename(layer_path)
-        except Exception as err:
+        except (FileExistsError, FileNotFoundError, PermissionError, OSError) as err:
             # TODO nothing better to do here?
             is_success = False
             logger.warning(
@@ -684,9 +684,8 @@ def cleanup_backups(layer_paths: set[str]) -> bool:
         layer_path = Path(layer_path_str)
         layer_backup_path = get_backup_path(layer_path)
         try:
-            if layer_backup_path.exists():
-                layer_backup_path.unlink()
-        except Exception as err:
+            layer_backup_path.unlink(missing_ok=True)
+        except (IsADirectoryError, PermissionError, OSError) as err:
             is_success = False
             logger.warning(
                 f"Unable to remove backup: {layer_backup_path}. Reason: {err}"
--- a/docker-qgis/qfc_worker/commands/apply_deltas.py
+++ b/docker-qgis/qfc_worker/commands/apply_deltas.py
@@ -652,7 +652,13 @@ def rollback_deltas(
         # NOTE pathlib operations will raise in case of error
         try:
             layer_backup_path.rename(layer_path)
-        except (FileExistsError, FileNotFoundError, PermissionError, OSError) as err:
+        except (
+            FileExistsError,
+            FileNotFoundError,
+            IsADirectoryError,
+            PermissionError,
+            OSError,
+        ) as err:
             # TODO nothing better to do here?
             is_success = False
             logger.warning(
--- a/docker-qgis/qfc_worker/commands/package.py
+++ b/docker-qgis/qfc_worker/commands/package.py
@@ -72,15 +72,9 @@ def call_libqfieldsync_packager(
         if vl_extent.isNull() or not vl_extent.isFinite():
             logger.info("Failed to obtain the project extent from project layers.")
 
-            try:
-                vl_extent = QgsRectangle.fromWkt(
-                    qfc_worker.utils.extract_project_details(project)["extent"]
-                )
-            except Exception as err:
-                logger.error(
-                    "Failed to get the project extent from the current map canvas.",
-                    exc_info=err,
-                )
+            vl_extent = QgsRectangle.fromWkt(
+                qfc_worker.utils.extract_project_details(project)["extent"]
+            )
 
         if vl_extent.isNull() or not vl_extent.isFinite():
             raise Exception("Failed to obtain the project extent.")
--- a/docker-qgis/qfc_worker/commands/package.py
+++ b/docker-qgis/qfc_worker/commands/package.py
@@ -9,7 +9,7 @@ from libqfieldsync.offline_converter import ExportType, OfflineConverter
 from libqfieldsync.offliners import OfflinerType, PythonMiniOffliner, QgisCoreOffliner
 from libqfieldsync.project import ProjectConfiguration
 from libqfieldsync.utils.file_utils import get_project_in_folder
-from qgis.core import QgsCoordinateTransform, QgsRectangle
+from qgis.core import QgsCoordinateTransform, QgsCsException, QgsRectangle
 
 import qfc_worker.utils
 from qfc_worker.commands_base import QfcBaseCommand
@@ -56,12 +56,13 @@ def call_libqfieldsync_packager(
             if layer_extent.isNull() or not layer_extent.isFinite():
                 continue
 
+            transform = QgsCoordinateTransform(layer.crs(), project.crs(), project)
+
             try:
-                transform = QgsCoordinateTransform(layer.crs(), project.crs(), project)
                 vl_extent.combineExtentWith(
                     transform.transformBoundingBox(layer_extent)
                 )
-            except Exception as err:
+            except QgsCsException as err:
                 logger.error(
                     "Failed to transform the bbox for layer {} from {} to {} CRS.".format(
                         layer.name(), layer.crs(), project.crs()
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -547,7 +547,8 @@ def is_localhost(hostname: str, port: int | None = None) -> bool:
                     return True
 
         return False
-    except Exception:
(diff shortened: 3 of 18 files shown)


duke-nyuki:
Task linked: [QF-8220 Reduce the number of except clauses that catch all exceptions and add a ruff rule about it](https://app.clickup.com/t/2192114/QF-8220)

duke-nyuki:

docker-qgis/qfc_worker/commands/package.py
```diff
@@ -72,15 +73,9 @@ def call_libqfieldsync_packager(
...
-                    qfc_worker.utils.extract_project_details(project)["extent"]
-                )
-            except Exception as err:
-                logger.error(
-                    "Failed to get the project extent from the current map canvas.",
-                    exc_info=err,
-                )
+            vl_extent = QgsRectangle.fromWkt(
+                qfc_worker.utils.extract_project_details(project)["extent"]
+            )
```
gounux, line 78: What was the former `except` trying to catch here ? `KeyError` maybe ?

Looking at the `extract_project_details` code, here, 

https://github.com/opengisch/QFieldCloud/blob/145bedd46b8bb9652b2e546d212b7c28bfb5c36a/docker-qgis/qfc_worker/utils.py#L579

Wondering if it's safe to remove this `except` here,

docker-qgis/qfc_worker/utils.py
```diff
@@ -91,7 +91,7 @@ def _write_log_message(message, tag, level):
     # in 3.16 it was Qgis.None, but since None is a reserved keyword, it was inaccessible
     try:
         Qgis.MessageLevel.NoLevel
-    except Exception:
+    except AttributeError:
         Qgis.MessageLevel.NoLevel = 4
```
gounux: Well, do we need to catch something here, since QGIS 3.16 is not really supported anymore ?

docker-qgis/qfc_worker/workflow.py
```diff
@@ -209,7 +209,10 @@ def json_default(obj):
 
     try:
         obj_str += f" {str(obj)}"
-    except Exception:
+    # Tipically, we should only expect `TypeError` if `__str__` or `__repr__` are not present.
```
gounux:
```suggestion
    # Typically, we should only expect `TypeError` if `__str__` or `__repr__` are not present.
```

docker-qgis/qfc_worker/commands/apply_deltas.py
```diff
@@ -684,9 +684,8 @@ def cleanup_backups(layer_paths: set[str]) -> bool:
         layer_path = Path(layer_path_str)
         layer_backup_path = get_backup_path(layer_path)
         try:
-            if layer_backup_path.exists():
-                layer_backup_path.unlink()
-        except Exception as err:
+            layer_backup_path.unlink(missing_ok=True)
+        except (IsADirectoryError, PermissionError, OSError) as err:
```
gounux, line 694:
```suggestion
        except (FileNotFoundError, IsADirectoryError, PermissionError, OSError) as err:
```

Maybe ?

gounux: Oh no sorry, seems that the `missing_ok` key avoids this potential `FileNotFoundError`.

gounux (commented):
Thanks @suricactus, so far I reviewed the majority of the changes,

docker-qgis/qfc_worker/commands/package.py
suricactus: if we are missing `extent` in `project_details` nowadays, we have a bigger problem, therefore no need to check for `KeyError` anymore.

docker-qgis/qfc_worker/utils.py
suricactus: I guess not.

suricactus:
@gounux sorry for the force push, but the comment typo was in the first commit...

docker-qgis/qfc_worker/workflow.py
suricactus: thanks!

docker-qgis/qfc_worker/utils.py
suricactus: Removed.

docker-qgis/qfc_worker/commands/package.py
gounux: Okay thanks, let's trust _the QGIS_ then :)

docker-qgis/qfc_worker/commands/apply_deltas.py
```diff
@@ -652,7 +652,7 @@ def rollback_deltas(
         # NOTE pathlib operations will raise in case of error
         try:
             layer_backup_path.rename(layer_path)
-        except Exception as err:
+        except (FileExistsError, FileNotFoundError, PermissionError, OSError) as err:
```
gounux:
```suggestion
        except (FileExistsError, FileNotFoundError, IsADirectoryError, PermissionError, OSError) as err:
```

Extra safety maybe ?

gounux (approved):
Apart from the CI getting red due to pre-commit check failing, am okay to go :)

PR #1593, opened 2026-05-09
chore: Drop all references to `skip_metadata` query param on the `GET /files` endpoint

suricactus:
This is a legacy storage artifact that was used to choose whether to send an expensive additional per file request to the Object storage to get the SHA256 sum of the file.

Now we calculate the value by default and there is no such need.


duke-nyuki:

gounux (approved):
Thanks for catching and dropping this legacy artifact :)

PR #1597, opened 2026-05-10
tests: Add a new CI step running template validity and django checks

suricactus:
See https://django-extensions.readthedocs.io/en/latest/validate_templates.html .

Ideally `VALIDATE_TEMPLATES_IGNORES` should be used, but it uses only the basename of the ignored files, see https://github.com/django-extensions/django-extensions/blob/4.1/django_extensions/management/commands/validate_templates.py#L64-L69 .

Basenames are just way too generic for our usecase, e.g. `create_form.html`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,3 +150,42 @@ jobs:
       #   timeout-minutes: 30
       #   with:
       #     limit-access-to-actor: true
+
+  django_checks:
+    name: Run Django checks
+    runs-on: ubuntu-24.04
+    continue-on-error: true
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Export the env variables file
+        run: |
+          cp .env.example .env
+          sed -ri 's/^ENVIRONMENT=(.*)/ENVIRONMENT=test/g' .env
+          sed -ri 's/^COMPOSE_FILE=(.*)/COMPOSE_FILE=\1:docker-compose.override.test.yml/g' .env
+          eval $(egrep "^[^#;]" .env | xargs -d'\n' -n1 | sed -E 's/(\w+)=(.*)/export \1='"'"'\2'"'"'/g')
+
+      - name: Pull docker containers
+        run: docker compose pull
+
+      - name: Build and run docker containers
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8228 Add a new CI step checking for template validity](https://app.clickup.com/t/2192114/QF-8228)

.github/workflows/test.yml
```diff
@@ -150,3 +150,46 @@ jobs:
...
       #   with:
       #     limit-access-to-actor: true
+
+  django_checks:
+    name: Run Django checks
+    runs-on: ubuntu-24.04
+    continue-on-error: true
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
```
gounux:
```suggestion
        uses: actions/checkout@v6
```

CF https://github.com/actions/checkout/releases

Shall we setup dependabot on `github-actions` ecosystem so that we get automatic updates of such GH actions ?

Was done on QFieldSync here : https://github.com/opengisch/qfieldsync/pull/772

.github/workflows/test.yml
```diff
@@ -150,3 +150,46 @@ jobs:
...
+          docker compose run app python manage.py migrate
+          docker compose run app python manage.py collectstatic
+
+      - name: Django check
+        run: |
+          docker compose run app python manage.py check --deploy --fail-level=ERROR
+
+      - name: Validate templates
+        run: |
+          docker compose run --rm app python3 manage.py validate_templates
```
gounux: Shall we harmonize the docker / python commands ?

```suggestion
      - name: Django check
        run: |
          docker compose run --rm app python manage.py check --deploy --fail-level=ERROR

      - name: Validate templates
        run: |
          docker compose run --rm app python manage.py validate_templates
```

Or maybe I'm missing something and a specific version is needed ?

gounux (approved):
Thanks @suricactus, happy to have those CI django checks in the future :)

.github/workflows/test.yml
suricactus: Done! Yes, we should!

.github/workflows/test.yml
suricactus: Done.

PR #1598, opened 2026-05-10
fix: Fix django-axes warning for not using an `ip_address` to block attempts

suricactus:
We should not "normalize" warnings, and one of them that is there for quite some times is:
```
WARNINGS:
?: (axes.W006) AXES_LOCKOUT_PARAMETERS does not contain 'ip_address'. This configuration allows attackers to bypass rate limits by rotating User-Agents or Cookies.
	HINT: Add 'ip_address' to AXES_LOCKOUT_PARAMETERS.
```
The main reason why it has not been tackled yet was because we didn't want to lock out a room of users behind the same IP.

Turned out we can configure lockout params to be a combo:
```
AXES_LOCKOUT_PARAMETERS = [["username", "ip_address"]]
```

See https://django-axes.readthedocs.io/en/latest/5_customization.html#customizing-lockout-parameters


duke-nyuki:
Task linked: [QF-8229 Fix django-axes warning for not using an `ip_address` to block attempts](https://app.clickup.com/t/2192114/QF-8229)

gounux (approved):

PR #1599, opened 2026-05-10
fix: Avoid dumping potentially very long field values which can affect performance

suricactus:
An incident during week 19 2026 caused QFieldCloud to use exponentially more RAM than usual. The culpriit? Adding the vector layer fields to `feedback.json`.

The `fields` are only used to print within the job logs for now and can be skipped from saving in project details. But in the future they will have a central role around project creation.

For now, we just delete it in a very ugly way once they are used.


duke-nyuki:
Task linked: [QF-8232 Avoid dumping potentially very long field values which can affect performance](https://app.clickup.com/t/2192114/QF-8232)

lukasgraf (approved):
LGTM! 👍

PR #1601, opened 2026-05-11
refactor: Change the deprecated in 5.1 `CheckConstraint.check` to `CheckConstraint.condition`

suricactus:
> The check keyword argument of CheckConstraint is deprecated in favor of condition.

See https://docs.djangoproject.com/en/5.1/releases/5.1/#id2 (three from the bottom)

See https://github.com/django/django/pull/17876

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -2691,7 +2691,7 @@ class Secret(models.Model):
             ),
             models.CheckConstraint(
                 # we want either the project to be set or the organization, but never both set or unset, therefore ^ (XOR)
-                check=Q(project__isnull=True) ^ Q(organization__isnull=True),
+                condition=Q(project__isnull=True) ^ Q(organization__isnull=True),
                 name="secret_assigned_to_organization_or_user",
             ),
         ]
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8233 Change the deprecated in 5.1 `CheckConstraint.check` to `CheckConstraint.condition`](https://app.clickup.com/t/2192114/QF-8233)

docker-app/qfieldcloud/core/models.py
```diff
@@ -2691,7 +2691,7 @@ class Meta:
             ),
             models.CheckConstraint(
                 # we want either the project to be set or the organization, but never both set or unset, therefore ^ (XOR)
-                check=Q(project__isnull=True) ^ Q(organization__isnull=True),
+                condition=Q(project__isnull=True) ^ Q(organization__isnull=True),
```
gounux, line 2694: Isn't there a migration produced ?

Or shall we edit the `0085` migration ?

https://github.com/opengisch/QFieldCloud/blob/2cee60c14d0bbab7cdb2de86d9c71df90d3deaba/docker-app/qfieldcloud/core/migrations/0085_alter_secret_options_and_more.py#L100-L110

suricactus: No migration, but you are right we have to fix the old migration (now or when we bump Django 6, better now).

gounux (approved):
Thanks !

PR #1602, opened 2026-05-11
tests: Remove the usage of `unittest.assert((Not)Equals` that was dropped in 3.12

suricactus:
See https://bugs.python.org/issue45162


duke-nyuki:
Task linked: [QF-8234 Drop deprecated unittest methods](https://app.clickup.com/t/2192114/QF-8234)

gounux (approved):
Thanks ! Code and CI looks good,

PR #1603, opened 2026-05-11
tests: Bump test dependencies to latest

suricactus:
We might not change requirements_test.in , but the depdencies of dependencies move forward. That's why we compile the in  file in txt  file, or namely requirements_test.txt . It is healthy to push to a newer version from time to time.

Just run pip-compile .

No need for individual commits here, we are talking about tests and they come from pip-compile itself.


duke-nyuki:
Task linked: [QF-8235 Bump test dependencies in `requirements_test.txt` to their latest versions](https://app.clickup.com/t/2192114/QF-8235)

docker-app/requirements/requirements_test.txt
```diff
@@ -1,22 +1,22 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.14
```
gounux: Was pip-compile ran with python 3.14 ?

Wondering if we would have some surprises, since on master it's not python 3.14 yet.

suricactus: Thanks for spotting. It was a backport from #1587

gounux (approved):

PR #1605, opened 2026-05-11
Rephrase help text message so we don't mention the nonexisting "premium" word

suricactus:
"Premium" is term that only exists in Django admin and not in the rest of the application, so rephrase to simple language.

Continuation of the work in b8fda6a2d992a39887965522df22275d9dee36d4


duke-nyuki:
Task linked: [QF-8075 Drop "premium plan" from error names so people are not confused where this wording comes from](https://app.clickup.com/t/2192114/QF-8075)

gounux (approved):

PR #1610, opened 2026-05-13
feat: add `signup_url` to the server info endpoint

gounux:
Depending on if the configured QFieldCloud adapter is open to sign up or not, it will respectively return the absolute URL for sign up or a null value.

FYI @nirvn


duke-nyuki:
Task linked: [QF-8217 Add the registration URL to server info response](https://app.clickup.com/t/2192114/QF-8217)

suricactus (approved):
Nice one!

PR #1612, opened 2026-05-13
chore: drop support of legacy storage in the codebase

gounux:
CF the title, dropping legacy storage code episode _564827_,

See individual commit messages for more information,

Maybe this one would need careful reviews, since it's based on existing _TODO_ comments : https://github.com/opengisch/QFieldCloud/commit/d0e0f36df0db54238ebba6cbe3b96a77ace60248

Also, maybe other commits to come next,


duke-nyuki:

docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -345,11 +335,7 @@ class Meta:
     )
 
     # Timestamp when the `FileVersion` record was inserted in the database.
-    # TODO We do not `auto_now_add=True` to be able to set this when migrating files from legacy to the regular storage. Switch to `auto_now_add=True` when the legacy storage is no longer supported.
-    created_at = models.DateTimeField(default=timezone.now, editable=False)
-
-    # The version id from the legacy object storage. The version id format is provider dependent, e.g. on S3 it is a random string, on others it is a UUID.
-    legacy_id = models.TextField(max_length=255, editable=False, null=True)
```
suricactus, line 352: Keep it for now. It's too early to get excited. If we lose this data, then we are doomed if we have to restore something.

suricactus:
Except the deletion of `FileVersion.legacy_id`, everything else is mergeable.

docker-app/qfieldcloud/filestorage/models.py
gounux: Ok, that makes sense thanks, reintroduced in https://github.com/opengisch/QFieldCloud/pull/1612/changes/df2db18ec580baa8b940fdd5f677290dd196bd17

gounux:
> Except the deletion of `FileVersion.legacy_id`, everything else is mergeable.

Ok thanks, removing the Draft status then.

suricactus (approved):

PR #1621, opened 2026-05-19
Add `get_file_name_from_header` to get the file name from `Content-Disposition`

suricactus:
Since #1587 is going to take a while, we don't have `cgi` in the future, and we can such a nice function, let's do that in advance.

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/http_utils.py
+++ b/docker-app/qfieldcloud/core/utils2/http_utils.py
@@ -1,0 +1,11 @@
+from email.message import Message
+from typing import cast
+
+
+def get_file_name_from_header(content_disposition: str) -> str | None:
+    """Extracts the file name from the `Content-Disposition` header."""
+    message = Message()
+    message["Content-Disposition"] = content_disposition
+    file_name = cast(str | None, message.get_filename())
+
+    return file_name


duke-nyuki:

docker-app/qfieldcloud/core/utils2/http_utils.py
```diff
@@ -0,0 +1,11 @@
+from email.message import Message
+from typing import cast
+
+
+def get_file_name_from_header(content_disposition: str) -> str | None:
+    """Extracts the file name from the `Content-Disposition` header."""
+    message = Message()
+    message["Content-Disposition"] = content_disposition
+    file_name = cast(str | None, message.get_filename())
```
lukasgraf, line 9: Obsolete, see follow up comment.

---

~This fails to decode headers with non-ASCII characters, which are usually RFC2047 encoded like this:~
```
Content-Disposition: inline; filename="=?UTF-8?Q?Research=5F2025=5FSample=5FAut=C3=B6r.pdf?='
```

I don't know the full context in which this function is used - if it exclusively operates on filenames that were produced by us, and are guaranteed to not contain any non-ASCII characters or whitespace, then it may be fine. But otherwise it's too simplistic.

`email.header.decode_header` still needs to be applied to results of `get_filename()` - the `Message` class doesn't do that for you.

E.g. apply something like this iteratively to all chunks and their encoding, if the chunks are bytestrings, and then concatenate the decoded chunks.

```python
>>> from email.message import Message
>>> from email.header import decode_header
>>>
>>> message = Message()
>>> message["Content-Disposition"] = 'Content-Disposition: inline; filename="=?UTF-8?Q?Research=5F2025=5FSample=5FAut=C3=B6r.pdf?="'
>>> message.get_filename()
'=?UTF-8?Q?Research=5F2025=5FSample=5FAut=C3=B6r.pdf?='
>>>
>>> (chunk, encoding)
(b'Research_2025_Sample_Aut\xc3\xb6r.pdf', 'utf-8')
>>> chunk, encoding = decode_header(message.get_filename())[0]
>>> chunk.decode(encoding)
'Research_2025_Sample_Autör.pdf'
```

lukasgraf (changes requested):

docker-app/qfieldcloud/core/utils2/http_utils.py
lukasgraf: Never mind - I was in the email headspace / mindset when thinking of RFC 2047.

For HTTP, RFC2047 encoding is not officially supported by standards, and we should only encounter RFC 5987 encoding - which `email.message.Message` _does_ indeed handle automatically:

```python
>>> message = Message()
>>> message["Content-Disposition"] = 'Content-Disposition: attachment; filename*=UTF-8\'\'file%20n%C3%A4me.jpg'
>>> message.get_filename()
'file näme.jpg'
```

lukasgraf (approved):
LGTM 👍

docker-app/qfieldcloud/core/utils2/http_utils.py
suricactus: Wow, learned a few things on the way! Thanks for digging deep here!

PR #1623, opened 2026-05-20
feat: block 3rd party auth for Organizations

gounux:
This PR intends to allow Social login only to `PERSON` type Users.  
If a Person and an Organization share the same email address, it makes no sense to log in as the Org, with a 3rd party IDP.

Since [this commit](https://codeberg.org/allauth/django-allauth/commit/2c7c4cbb3b66922544b23dc20c46871723a81d0a), `django-allauth` provides a new `authenticate_by_email` hook, that allows the custom implementations of `DefaultSocialAccountAdapter` to have a specific behavior regarding user <-> email matching.  
QFieldCloud's adapter behavior is filtering Users and keep only the `PERSON` type ones. See provided test scenario.

**TODO**: bump the `django-allauth` dependency to a future release version in the requirements, when the commit will be part of a release.

See #1484 for more infos and background.

### DIFF
--- a/docker-app/qfieldcloud/core/adapters.py
+++ b/docker-app/qfieldcloud/core/adapters.py
@@ -6,7 +6,7 @@ from typing import Literal
 
 from allauth.account import app_settings
 from allauth.account.adapter import DefaultAccountAdapter
-from allauth.account.models import EmailConfirmationHMAC
+from allauth.account.models import EmailAddress, EmailConfirmationHMAC
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 from allauth.socialaccount.models import SocialLogin
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
@@ -19,7 +19,7 @@ from django.utils import timezone
 from invitations.adapters import BaseInvitationsAdapter
 
 from qfieldcloud.authentication.sso.provider_styles import SSOProviderStyles
-from qfieldcloud.core.models import Person
+from qfieldcloud.core.models import Person, User
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-7958 SSO: fix getting account linked with organization](https://app.clickup.com/t/2192114/QF-7958)

gounux:
This one can go on review, since `django-allauth` [released a new version yesterday](https://codeberg.org/allauth/django-allauth/src/branch/main/ChangeLog.rst#section),

suricactus:
@lukasgraf is the man to review this :)

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -281,3 +281,44 @@ def pre_social_login(self, request, sociallogin):
...
+        ).values_list("user_id", flat=True)
+
+        matching_users = User.objects.filter(
+            type=User.Type.PERSON,
+            email__in=email_addresses,
+            is_active=True,
+            pk__in=verified_users_ids,
+        )
+
+        if matching_users.exists():
```
lukasgraf: I think we should still keep the check for `self.can_authenticate_by_email(sociallogin, email)` from the original implementation. Otherwise we would make the `SOCIALACCOUNT_EMAIL_AUTHENTICATION` ineffective.

docker-app/qfieldcloud/core/adapters.py
```diff
@@ -281,3 +281,44 @@ def pre_social_login(self, request, sociallogin):
...
+
+        verified_users_ids = EmailAddress.objects.filter(
+            email__in=email_addresses,
+            primary=True,
+            verified=True,
+        ).values_list("user_id", flat=True)
+
+        matching_users = User.objects.filter(
+            type=User.Type.PERSON,
+            email__in=email_addresses,
```
lukasgraf: This will do a case sensitive match for the email addresses. Unfortunately, Django does not come with a case insensitive `IN` operator by default, we would need to add that [ourselves](https://forum.djangoproject.com/t/elegant-way-in-insensitive-case/26217/7), or use something like `iexact` and a loop.

This could lead to very rare, confusing failures where login doesn't work for only a few users.

Since the list of email addresses associated with a social login should be very short (small handful), I would probably use a loop and an `iexact` match here.

lukasgraf (changes requested):
Looks good to me overall, just some small remarks!

docker-app/qfieldcloud/core/adapters.py
gounux: Thanks ! A for loop is brought via this commit : https://github.com/opengisch/QFieldCloud/pull/1623/commits/edfd87ff1fb3ab06283a4c194d2a7af5ddd123b6

Also a test case to ensure a match is found with insensitive case email.

docker-app/qfieldcloud/core/adapters.py
gounux: See this commit : https://github.com/opengisch/QFieldCloud/pull/1623/commits/cce4e12b14fc208f8a53f4cc471faedd8fef37fb

Also added some tests case to ensure the `can_authenticate_by_email` would behave correctly.

lukasgraf (approved):
Looks good to me! 👍 🎉 

And I love the tests - very readable, and helpful in understanding what our customization of the adapter actually does!

PR #1624, opened 2026-05-21
Add a tiny JS utility to improve navigation and the looks long textual fields

### LINKED ISSUE #1618: JSON view in Django Backend
### What is the enhancement proposal? What problem are you trying to solve?

It would be nice to have a rich-text JSON viewer instead of an indented but lengthy JSON dump.

### Description

I’ve tested https://github.com/jmrivas86/django-json-widget to avoid reinventing the wheel. Unfortunately, it does not work for read-only fields.

However, it should be fairly straightforward to add a JavaScript-based viewer to the Django admin, for example using https://github.com/andypf/json-viewer, which is MIT licensed and looks solid.

## A quick test for the job Admin 

```python
    def feedback__pre(self, instance):
        data = json.dumps(instance.feedback or {})
        from django.utils.safestring import mark_safe

        return mark_safe(f"""
            <div id="json-viewer-{instance.pk}"></div>

            <script src="https://pfau-software.de/json-viewer/dist/iife/index.js"></script>
            <script>
                (function() {{
                    const el = document.getElementById("json-viewer-{instance.pk}");
                    if (!el) return;
                    const viewer = document.createElement("andypf-json-viewer");
                    viewer.data = {data};
                    viewer.indent = 2;
                    viewer.theme = "default-light";
                    viewer.showDataTypes = true;
                    viewer.showToolbar = true;
                    viewer.expandIconType = "arrow";
                    viewer.showCopy = true;
                    viewer.showSize = true;
                    el.appendChild(viewer);
                }})();
            </script>
    """)
```
**In a PR I would not use `mark_save` but include the script as static file.**

## This gives us:

Image

where objects can nicely be toggled:

Image

### Agreements

- [x] I commit to keeping my branch up to date with the latest `master` branch until my pull request is merged.
- [x] I commit to writing tests for my enhancement proposal, if applicable.
- [x] I am aware that my enhancement proposal and related pull request may not be accepted.

### Additional context

_No response_

suricactus:
We have readonly fields such as Job.output and Job.feedback, which are
very hard to read when they are too long. We already made them monospaced,
but we can also enable syntax highlighting for further improvement.

The JSON formatter goes an extra step and makes multiline text more readable.

```
{
  "key": "multi/nline",
  "num": 5
}
```

Will be rendered as:

```
{
  "key": "
multi
line
",
  "num": 5
}
```

This makes the JSON invalid, but that's why the `copy` button was introduced.

In addition, there is also the `wrap` button, that toggles wrapping.

---

All the logic was developed in JS, because we will remove the long
textual fields one day from the database and make the downloadable
attachments. In that bright future, the admin logic will be handled on
the client side.

I think it also fixes #1618 . I know that the delta JSON is still editable in Django admin, but this is supposed to change and become readonly too.

| before | after |
|-|-|
|2026-05-21 14 28 14 localhost 7e63064a0223 | 2026-05-21 14 27 15 localhost b585de90489b |


duke-nyuki:
Task linked: [QF-8296 Add some syntax highlighting in the JSON output](https://app.clickup.com/t/2192114/QF-8296)

gounux (approved):
Could checkout have a look locally, looks great thanks !

PR #1630, opened 2026-05-21
feat: make the jobs `error_stack` a multiline string instead array of multiline strings

suricactus:
What is more, if we have a chain `raise Exception from err`, it will also print the error chain, instead of the last error only.

We long struggled with the readability of the `job.feedback.error_stack` field. Because JSON representation make it very ugly to read multiline strings `"line\nsecond"`, back then I made it an array, for slightly better UX.

But this still makes it hard to consume, so let's convert the tb to a string. Nobody expects anything about this field, it's pure human-read field.

---
Tested every single try/except block with a 1/0 exception.

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_delta.py
+++ b/docker-app/qfieldcloud/core/tests/test_delta.py
@@ -118,11 +118,8 @@ class QfcTestCase(APITransactionTestCase):
 
             if job.feedback:
                 if "error_stack" in job.feedback:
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-8297 Convert the `error_stack` in `Job.feedback` from a `list[str]` to just `str` to improve readability](https://app.clickup.com/t/2192114/QF-8297)

gounux (approved):
LGTM 👍

PR #1631, opened 2026-05-23
feat: Store the QGIS project file version in `Project.qgis_version`

suricactus:
This is leading the path to know which QGIS worker to use.

TODOs:
- [x] remove the duplicated function `qfieldcloud.tests.utils.open_qgis_file`
- [x] add tests that the version is set on file upload
- [x] add tests that the version is reset on file deletion
- [x] add tests that the version is set via `process_projectfile` if missing


duke-nyuki:
Task linked: [QF-8303 Store the QGIS project file version in `Project.qgis_version`](https://app.clickup.com/t/2192114/QF-8303)

docker-app/qfieldcloud/filestorage/view_helpers.py
```diff
@@ -406,16 +425,29 @@ def delete_project_file_version(
...
+
+            if file_qs.exists():
+                assert file_qs.count() == 1
+
+                # Recalculate the QGIS version based on the remaining file.
+                # Let's say we have `p.qgs` which has file version 1 (QGIS 4.0) and file version 2 (QGIS 4.2), and we delete file version 2,
+                # then we should recalculate the QGIS version based on file version 1 (QGIS 4.0),
+                # rather than keeping the QGIS version from the deleted file version 2 (QGIS 4.2).
+                qgis_version = get_qgis_version_from_project_file(
+                    filename, file_qs.first().latest_version.content.open()
```
lukasgraf: Can we use a `with` statement context manager here instead of the direct `open()` call?
Otherwise we might leave open file handles around, or rely on garbage collection to close them.

docker-qgis/qfc_worker/utils.py
```diff
@@ -1076,3 +1078,52 @@ def process_project_write(document: QDomDocument) -> None:
 
     project.writeProject.connect(process_project_write)
     project.write(project.fileName())
+
+
+@contextmanager
+def open_qgis_file(filename: str | Path) -> Iterator[TextIO]:
```
lukasgraf, line 1084: Since these two functions are largely duplicated in `app` and `qfc_worker`, could we at some sort of a signpost comment that they should be kept in sync?

lukasgraf (approved):
LGTM! 👍 🎉 
Just some minor comments, nothing that should block a merge IMHO.

docker-app/qfieldcloud/filestorage/view_helpers.py
suricactus: Done in https://github.com/opengisch/QFieldCloud/pull/1631/commits/4df6b21801ad0c3d319f40a224caf56a2df41173

docker-qgis/qfc_worker/utils.py
suricactus: Done in https://github.com/opengisch/QFieldCloud/pull/1631/commits/44ec00a77f858e32e010be8f01cc65102742e610

PR #1633, opened 2026-05-26
fix: also install `tzdata-legacy` to support legacy timezones

suricactus:
Apparently there are users still using `US/Pacific` or similar long deprecated timezones.

If the package is not installed, errors such as:
```
["Invalid timezone 'US/Pacific'"]
```

appear, when the `core_useraccount.timezone` is set to a legacy timezone.

See:
https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=1051973


duke-nyuki:

PR #1636, opened 2026-05-27
refactor: Add `ruff` check for deprecations and fix them

suricactus:
Literally the title.

### DIFF
--- a/docker-app/qfieldcloud/authentication/authentication.py
+++ b/docker-app/qfieldcloud/authentication/authentication.py
@@ -1,5 +1,3 @@
-from typing import Type
-
 from django.http.request import HttpRequest
 from django.utils import timezone
 from django.utils.translation import gettext as _
@@ -21,7 +19,7 @@ def invalidate_all_tokens(user: User) -> int:
 
 
 def create_token(
-    token_model: Type[AuthToken],
+    token_model: type[AuthToken],
     user: User,
     _serializer=None,
     request: HttpRequest | None = None,
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -3,11 +3,11 @@ import json
 import time
 import uuid
 from collections import namedtuple
-from collections.abc import Iterable
+from collections.abc import Generator, Iterable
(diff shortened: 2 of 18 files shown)


lukasgraf (approved):
LGTM! 👍 

One question though: Since we don't have a `pyproject.toml` with a declared range of supported Python versions, I believe `ruff` will use [`target-version` from its own settings](https://github.com/opengisch/QFieldCloud/blob/master/ruff.toml#L20) instead - which only apply to ruff and nothing else. That's currently at `py310` - is that still what we're currently targeting as a "soft" minimal supported Python version?

(Not suggesting to bump it in this PR - once we have these deprecation checks in place, it should be even easier to bump it in the future).

suricactus:
Nice question! I believe `ruff.toml#target-version` is taken into account indeed. However, I am not 100% sure how it works since we have a Python version that is different between the containers.

I did a few force pushes, intiially thinking we should change the `target-version=py314`,  since we merged #1587 anyways. But then I remembered, that QGIS3 image (ubuntu 24.04 noble) comes with Python 3.12. Once we settle with the QGIS situation, we can revisit the target version.

The long term plan is that we gradually add more and more rules here until we get to the stage of the other QField related projects, e.g. libqfieldsync, qfieldsync, convert2qgis, xlsformconverter, they all use this ruff config: https://github.com/opengisch/libqfieldsync/blob/master/pyproject.toml#L24-L25

This will take a while for a large project like QFieldCloud, but nothing stops us from dreaming :)

lukasgraf:
Yep, it makes sense to align the target-version I think. 👍

(Maybe just to the lowest common denominator, if it turns out that the difference in Python versions between contains causes some checks to fail). 

For the longer term however I would consider creating a `pyproject.toml` and declaring the required Python versions in there. There's quite a semantic difference between declaring "this code requires one of these Python versions" and "the code is linted with this version in mind".

suricactus:
The main problem we have is that our repo here is not exactly a tipical pyproject. We have at least 2 different projects under one repo - docker-app and docker-qgis.

Happy to discuss this one day, but I think we are quite far from hitting a wall with the current approach of having ruff file for now.

PR #1638, opened 2026-05-28
nginx: allow certbot acme challenge in http

gounux:
_Let's Encrypt_'s acme challenge location is over HTTP, which this PR intends to allow again, after #1529 added a redirect from all the nginx `location`s to HTTPS.

### DIFF
--- a/docker-nginx/templates/default.conf.template
+++ b/docker-nginx/templates/default.conf.template
@@ -56,15 +56,16 @@ geo $is_docker_internal {
     172.16.0.0/12  1;
 }
 
-map $uri $is_acme {
-    ~^/.well-known/acme-challenge/  1;
-    default                         0;
+# Determine if we should redirect to https
+map "$allow_internal_http:$is_docker_internal" $base_redirect_to_https {
+    "1:1"   0; # If we explicitly allow plain http and the request is coming from docker internal network, don't redirect to https.
+    default 1; # Default is to redirect to https.
 }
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8324 Fix certbot acme challenge on http](https://app.clickup.com/t/2192114/QF-8324)

docker-nginx/templates/default.conf.template
```diff
@@ -56,6 +56,17 @@ geo $is_docker_internal {
...
+map $uri $is_acme {
+    ~^/.well-known/acme-challenge/  1;
+    default                         0;
+}
+
+map "$is_acme$allow_internal_http$is_docker_internal" $allow_http {
+    ~^1     1; # ACME challenge: always allowed
+    "011"   1; # docker internal with HTTP allowed
+    default 0;
+}
```
suricactus: I know I am not in workmood right now, but this is completely unreadable what is going on. Can we add a comment explaining what is going on here?

Btw, can't we just have `location ~^/.well-known/acme-challenge/ {}` in the `server` part deleted below?

toebivankenoebi: I changed the logic to simplify the mapping to 2 variables and added some comments. Should be much clearer now.

The `location /.well-known/acme-challenge/ {}` statement is still needed.

Rakanhf (approved):
Had a quick discussion and everything LGTM !

PR #1639, opened 2026-06-01
Trials: Add support for creating trials for any plan.

lukasgraf:
This makes the `create_subscription` accept an optional `regular_plan` keyword argument, allowing it to create trial subscriptions for any plan, not just the default plan for a given user type.

If no particular `regular_plan` is specified, we choose the default plan for the given user type (same as before).

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -946,6 +946,7 @@ class AbstractSubscription(models.Model):
         plan: Plan,
         created_by: Person,
         active_since: datetime | None = None,
+        regular_plan: Plan | None = None,
     ) -> tuple[type["AbstractSubscription"] | None, "AbstractSubscription"]:
         """Creates a subscription for a given account to a given plan. If the plan is a trial, create the default subscription in the end of the period.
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8311 Fix existing bugs for trial subscriptions](https://app.clickup.com/t/2192114/QF-8311)

suricactus (approved):
LGTM!

Ideally the `regular_plan` should not be optional, but I guess this will required a few extra changes in the codebase.

I would suggest we leave this for the bright day when we refactor the logic how trial/regular plans work.

lukasgraf:
@suricactus I don't quite understand what you mean by "not optional".
In the case where `create_subscription()` is used to create a normal subscription (not a trial one), would you when expect the caller to pass in the identical plan to both the `plan` and the `regular_plan` arguments? That would be an extraordinarily weird API, IMHO.

(Because the function worked so far, `plan` holds the trial plan for trial subscriptions, and the normal plan for non-trial subscriptions)

suricactus:
Sorry, I didn't explain myself very well, and you noted it.

I meant that regular plan should be mandatory (therefore always passed) and optionally we pass the trial plan, not the other way around as it is right now: we pass a plan, trial or regular, and then we pass the regular only if the plan is trial, but also if it is not passed, use the default.

lukasgraf:
Yep, I just realized now what you meant :)

With the changes as of now, the more natural signature would be `(plan, trial_plan=None)` (and invert the meaning of `plan`).

However, as you already indicated, this is a slightly more invasive change than I feel comfortable right now. Not least because the change spans two repos, and there's also some tests that proved to be quite tricky to rewrite, because they rely a lot on the default behavior, and how exactly `create_default_plan_subscription()` invokes `create_subscription()`. So I'll leave it for now, but make a note for the bright day ;-)

PR #1640, opened 2026-06-01
fix: preserve package files for non-member users on public projects

Rakanhf:
`latest_package_jobs()` was filtering the cleanup (`delete_obsolete_packages`) protection set to only include the project owner and direct collaborators Package jobs triggered by public users were excluded, causing their files to be deleted immediately after the next cleanup run and resulting in a 400 `invalid_job` error

Added a conditional check for public projects to bypass the collaborator-only filter so the latest job per distinct `triggered_by` user is preserved regardless of membership status

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1388,6 +1388,13 @@ class Project(models.Model):
         Returns:
             QuerySet of all the last package jobs.
         """
+        if self.is_public:
+            return (
+                PackageJob.objects.filter(project_id=self.id)
+                .order_by("triggered_by", "-created_at")
+                .distinct("triggered_by")
+            )
+
         if self.owner.is_organization:
             # all the users including the organization owner
             triggered_by_qs = Person.objects.for_organization(self.owner)
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1388,12 +1388,16 @@ class Project(models.Model):
         Returns:
             QuerySet of all the last package jobs.
         """
-        if self.is_public:
-            return (
-                PackageJob.objects.filter(project_id=self.id)
-                .order_by("triggered_by", "-created_at")
-                .distinct("triggered_by")
+        jobs_qs = (
+            PackageJob.objects.filter(
+                project_id=self.id,
             )
+            .order_by("triggered_by", "-created_at")
+            .distinct("triggered_by")
+        )
+
+        if self.is_public:
+            return jobs_qs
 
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8315 Fix missing package files causing 403 on /packages/ endpoint](https://app.clickup.com/t/2192114/QF-8315)

suricactus:
I would suggest that we keep the old logic, but wrap it with `if not project.is_public`.

Rakanhf:
> I would suggest that we keep the old logic, but wrap it with `if not project.is_public`.

I added that, sorry for the force push, it will just be cleaner.

docker-app/qfieldcloud/core/models.py
```diff
@@ -1388,6 +1388,13 @@ def latest_package_jobs(self) -> PackageJobQuerySet:
         Returns:
             QuerySet of all the last package jobs.
         """
+        if self.is_public:
+            return (
+                PackageJob.objects.filter(project_id=self.id)
+                .order_by("triggered_by", "-created_at")
+                .distinct("triggered_by")
+            )
```
suricactus: Preventing ourselves from having different querysets.

```suggestion
        jobs_qs = (
                PackageJob.objects.filter(project_id=self.id)
                .order_by("triggered_by", "-created_at")
                .distinct("triggered_by")
        )

        if self.is_public:
            return jobs_qs
            
        # ...
        
        jobs_qs = jobs_qs.filter(
            triggered_by__in=triggered_by_ids_qs,
        )
```

suricactus (approved):
Acceptable to merge like this if a blocked for the release, but if you deduplicate the code I would be happy.

gounux (approved):
Thanks for this, and for the tests as well, am also okay to merge as-is,

Though I'm wondering if public projects could then keep an important amount of Jobs - e.g. db records, storage files, etc., since those are not purged anymore.

Maybe there could be some ways to limit the `latest_package_job_for_user` to max one per user, this would need more brain juice than I'm capable at the moment, and I believe that this can be part of a follow-up.

Rakanhf:
> Thanks for this, and for the tests as well, am also okay to merge as-is,
> 
> Though I'm wondering if public projects could then keep an important amount of Jobs - e.g. db records, storage files, etc., since those are not purged anymore.
> 
> Maybe there could be some ways to limit the `latest_package_job_for_user` to max one per user, this would need more brain juice than I'm capable at the moment, and I believe that this can be part of a follow-up.

We discussed that, and it will be indeed a follow up. since we might keep it as is for now and just create a simple cron to do the cleaning.

PR #1648, opened 2026-06-03
Force linux/amd64 Docker images for local development

manylon:
When starting the development environment on an arm64 machine, Docker fails with a platform mismatch. The docker-createbuckets Dockerfile uses ubuntu:noble (pulled as linux/arm64 on Apple Silicon) but installs a hardcoded linux/amd64 binary for the rustfs, causing an architecture conflict.

### DIFF
--- a/docker-qgis/Dockerfile
+++ b/docker-qgis/Dockerfile
@@ -1,5 +1,5 @@
 # NOTE if the ubuntu version is changed, also change it in `Suites:` in apt sources, and in `QGIS_VERSION`
-FROM --platform=linux/amd64 ubuntu:noble
+FROM  ubuntu:noble
 
 # Create the expected /io directory that is usually created via the `worker_wrapper`
 RUN mkdir -p /io
--- a/docker-qgis/Dockerfile
+++ b/docker-qgis/Dockerfile
@@ -1,5 +1,5 @@
 # NOTE if the ubuntu version is changed, also change it in `Suites:` in apt sources, and in `QGIS_VERSION`
-FROM  ubuntu:noble
+FROM ubuntu:noble
 
 # Create the expected /io directory that is usually created via the `worker_wrapper`
 RUN mkdir -p /io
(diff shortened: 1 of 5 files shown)


duke-nyuki:
Task linked: [QF-8369 Force linux/amd64 Docker images for local development](https://app.clickup.com/t/2192114/QF-8369)

suricactus:
May I ask both @lukasgraf and @gounux to have a look on this PR before merging?

gounux (approved):
Thanks, "light" LGTM from my side since I don't have an `arm64` machine to test,  
Though I added the new `DOCKER_DEFAULT_PLATFORM=linux/amd64` env var locally, and everything seems alright!

docker-qgis/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # NOTE if the ubuntu version is changed, also change it in `Suites:` in apt sources, and in `QGIS_VERSION`
-FROM --platform=linux/amd64 ubuntu:noble
+FROM  ubuntu:noble
```
gounux:
```suggestion
FROM ubuntu:noble
```

Maybe ? Sorry..

lukasgraf (approved):
LGTM! 👍 🎉 ❤️ 

I faced the same issue ages ago, and just worked around it by adding `export DOCKER_DEFAULT_PLATFORM=linux/amd64` to my local `.bash_profile` 😉 

But since it seems I'm not the only QFC dev on macOS any more, it now makes sense to address it here!

Tested it locally, removed the default platform setting from my profile, works like a charm 👍

suricactus (approved):
Thanks all! I also tested this PR, and I tested by accident that this PR works fine on Linux even without the new envvar. It also works with it. 

Only the tiny whitespace issue needs to be addressed that @gounux highlighted before merging.

manylon:
Thank you all for the reviews! Addressed the issue and after passed test will merge.

PR #1650, opened 2026-06-05
Include team members in the collaborator count when displaying project collaborators in overview

manylon:
Number of people involved in a project counted only direct collaborators. 

Now it counts both team members and direct collaborators, deduplicated.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1825,26 +1825,33 @@ class Project(models.Model):
         )
 
     @property
-    def collaborators_count(self) -> int:
+    def total_collaborators(self) -> PersonQueryset:
         if self.owner.is_organization:
             exclude_pks = [self.owner.organization.organization_owner_id]
         else:
             exclude_pks = [self.owner_id]
 
-        team_collab_ids = (
+        team_collaborators_ids = (
             self.collaborators.skip_incognito()  # type: ignore[attr-defined]
             .filter(collaborator__type=User.Type.TEAM)
             .values_list("collaborator_id", flat=True)
         )
-        return (
-            self.direct_collaborators.values_list("collaborator_id", flat=True)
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8362 Adding teams for collaboration in projects doesn't count the members of the team involved](https://app.clickup.com/t/2192114/QF-8362)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1824,6 +1824,28 @@ def direct_collaborators(self) -> ProjectCollaboratorQueryset:
...
         )
 
+    @property
+    def collaborators_count(self) -> int:
+        if self.owner.is_organization:
+            exclude_pks = [self.owner.organization.organization_owner_id]
+        else:
+            exclude_pks = [self.owner_id]
+
+        team_collab_ids = (
```
```suggestion
        team_collaborators_ids = (
```

No need to save the few bytes, we can afford them :)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1824,6 +1824,28 @@ def direct_collaborators(self) -> ProjectCollaboratorQueryset:
             )
         )
 
+    @property
+    def collaborators_count(self) -> int:
```
```suggestion
    def total_collaborators_count(self) -> int:
```

Just to distiguish betweeh `collaborators` that we already have as a FK from `ProjectCollaborator.project` and this one, that includes the members too.

lukasgraf (approved):
LGTM! 👍 🎉 

Gave it a quick test locally, does what it says on the tin! 😄 

Just a small nitpick on the pull request title:
I would try to phrase PR titles in the positive, if possible. So, describe what they improve or fix, not what the original problem was.

So maybe something like `Also count team members when displaying project collaborators` for this one.

suricactus (approved):

manylon:
Thanks for the reviews @suricactus @lukasgraf !

PR #1654, opened 2026-06-10
feat: configure max number of parameters via constance

gounux:
This PR intends to add the ability to dynamically configure the `DATA_UPLOAD_MAX_NUMBER_FIELDS` Django setting, via a new Constance config.

This value is set via a middleware that overrides this setting for each request.

Reason behind this is that in Django Admin, editing Models with a lot of related objects (e.g. Organizations with many Projects / many Members) may send a lot of GET/POST parameters, thus reaching the Django `DATA_UPLOAD_MAX_NUMBER_FIELDS` setting limit (1000 is the default).

### DIFF
--- a/docker-app/qfieldcloud/core/middleware/fields_limit.py
+++ b/docker-app/qfieldcloud/core/middleware/fields_limit.py
@@ -1,0 +1,17 @@
+from collections.abc import Callable
+
+from constance import config
+from django.http import HttpRequest, HttpResponse
+from django.test.utils import override_settings
+
+
+class DynamicMaxNumberFieldsLimitMiddleware:
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]):
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        """Dynamically set the DATA_UPLOAD_MAX_NUMBER_FIELDS setting limit based on the Constance value."""
+        with override_settings(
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-8373 Increase possible number of GET/POST parameters in django admin](https://app.clickup.com/t/2192114/QF-8373)

suricactus (commented):
Hm, I would actually set this as a Django setting instead? I would say we can safely increase to 3-4-5k, whatever is needed whenever we hit the limit. If needed later we can expose to an envvar. But this is not something a "manager" should configure.

t-book:
Great addition. I lean the same way as suricactus though — feels like it belongs in Django settings / an env var, managed by whoever runs the instance, rather than a Constance value. Just my two cents.

lukasgraf:
After discussing this in today's meeting, I also agree with what @suricactus is suggesting.

I'm partly to blame here, because I suggested making the config a Constance setting instead of an ENV var (to avoid yet another env var that will stay the same for 95% of deployments). When in reality we probably should not even make it configurable.

I think just changing the Django setting to something like 5000 is fine:
- It's an upper bound. It will only have an effect on requests and deployments with lots of data that actually make use of that many params.
- Eventually, there could be situations where (reverse) proxies like nginx, or WAFs on on-premise setups, kill the requests because they have their own limits for number of params, or just the request body length in general. But even in situations like those, the outcome would still be the same: Request is blocked. In one situation because of a proxy, in the other because of this Django limit.

gounux:
Ok thanks all for the feedbacks, closing this since it'll be addressed in #1656 via a simple setting value.

PR #1655, opened 2026-06-10
fix: include organization owner in user filtering for active users

Rakanhf:
Adjusted the filter to include the organization owner for `active_users`

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -631,6 +631,7 @@ class OrganizationManager(UserManager):
 
 class Organization(User):
     members: models.QuerySet["OrganizationMember"]
+    organization_owner_id: int
 
     class Meta(User.Meta):
         verbose_name = "organization"
@@ -698,7 +699,7 @@ class Organization(User):
         ).filter(
             Q(id__in=users_with_delta)
             | Q(id__in=users_with_jobs)
-            | Q(id=self.organization_owner_id)  # type: ignore[attr-defined]
+            | Q(id=self.organization_owner_id)
         )
 
     def save(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8021 Verify that org admin is shown in active user count](https://app.clickup.com/t/2192114/QF-8021)

docker-app/qfieldcloud/core/models.py
```diff
@@ -695,7 +695,11 @@ def active_users(self, period_since: datetime, period_until: datetime):
 
         return Person.objects.filter(
             is_staff=False,
-        ).filter(Q(id__in=users_with_delta) | Q(id__in=users_with_jobs))
+        ).filter(
+            Q(id__in=users_with_delta)
+            | Q(id__in=users_with_jobs)
+            | Q(id=self.organization_owner_id)  # type: ignore[attr-defined]
```
suricactus: Instead of the type exception, can you add the typing for `organization_owner_id`? You can see in the models there are quite a few of these added:

```
class Organization:
   organization_owner_id: int
```

Rakanhf: To be honest I don't really like the class here, since It personally confuses me, maybe looking into stubs, etc. will be better. I highly prefer the comment.

suricactus: I am very open to discuss a better solution for the future. As of time being, most of the time the problem is solved with the explicit setting of the new properties on the model class, as initially requested. Let's merge this PR with the already existing solution, and then discuss a new one.

suricactus (changes requested):
Let's solve the typing issue not with ignoring, but with explicit property declaration.

docker-app/qfieldcloud/core/models.py
Rakanhf: I added the typing, I got a bit confused about something else.

suricactus (approved):
Thanks! Ready to get merged!

PR #1656, opened 2026-06-11
feat: increase max number of GET/POST parameters

gounux:
Follow-up of #1654 : increases the max number of GET/POST parameters via a setting configuration, from the 1000 default to 5000.

Reason behind this is that in Django Admin, editing Models with a lot of related objects (e.g. Organizations with many Projects / many Members) may send a lot of GET/POST parameters, sometimes more than the 1000 default.


duke-nyuki:
Task linked: [QF-8373 Increase possible number of GET/POST parameters in django admin](https://app.clickup.com/t/2192114/QF-8373)

lukasgraf (approved):
LGTM! 👍 🎉

suricactus (approved):

PR #1657, opened 2026-06-12
Add ratelimit for password resets

manylon:
Default rate limiting for password reset in django-allauth is 20/m/ip, 5/m/key (20 attempts per minute per IP, 5 per minute per key). Lowered the IP limit to 5/m/ip, 5/m/key.

Also added a custom page for 429 (Too Many Requests) errors

### DIFF
--- a/docker-nginx/templates/includes/qfieldcloud.conf.template
+++ b/docker-nginx/templates/includes/qfieldcloud.conf.template
@@ -31,6 +31,9 @@ location @proxy_to_app {
     error_page 403 =403 /pages/403.html;
     error_page 404 =404 /pages/404.html;
 
+    # Rate limit exceeded, shown when app returns 429 (e.g. too many password reset attempts)
+    error_page 429 =429 /pages/429.html;
+
     # Initial loading page for upstream issues (502, 503, 504 for timeouts)
     error_page 502 503 504 =503 /pages/loading.html;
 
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-6892 Add ratelimit for password resets](https://app.clickup.com/t/2192114/QF-6892)

docker-nginx/templates/includes/qfieldcloud.conf.template
```diff
@@ -31,6 +31,9 @@ location @proxy_to_app {
     error_page 403 =403 /pages/403.html;
     error_page 404 =404 /pages/404.html;
 
+    # Rate limit exceeded, shown when app returns 429 (e.g. too many password reset attempts)
+    error_page 429 =429 /pages/429.html;
```
suricactus, line 35: This will be problematic once we expose this as a JSON-based API. Nothing to worry about right now.

suricactus (commented):
What will happen if we have a workshop with 50 participants and 20 of them forgot their passwords?

How did we come up with the number 5?

Finally, would you provide a command for resetting the ratelimit via Django command or ideally a button from the admin interface?

suricactus:
@manylon ping

manylon:
> What will happen if we have a workshop with 50 participants and 20 of them forgot their passwords?
> 

We will hit both limits in this case.

> How did we come up with the number 5?
> 

No hard opinions on the number 5, the default is 20 per IP for every minute and 5 per key for every minute. This means that after one minute the limit resets.

> Finally, would you provide a command for resetting the ratelimit via Django command or ideally a button from the admin interface?

Yes, I can provide a management command, and an admin for resetting the rate limit. `django-allauth` stores rate limit counters in Django's cache backend. The cache key format (from allauth/core/ratelimit.py) is:
```
allauth:rl:<action>:<per>:<value>
```
So resetting the limit is by deleting the specific cache key, or clearing the entire cache of this part if a more aggressive reset is acceptable.

manylon:
Based on current needs and discussions, this pull request will be closed.

PR #1658, opened 2026-06-12
fix: detect S3 Content-Type from original filename in versioned paths

00-Python:
## Summary

Fixes incorrect S3 Content-Type metadata on all uploaded files.

### The Bug

QFieldCloud stores project files with versioned S3 keys:

    projects/<uuid>/files/DCIM/photo.jpg/v20260610-a1b2c3d4

django-storages calls mimetypes.guess_type() on the full key — the
trailing version UUID has no extension, so it always returns None.
The fallback then uses content.content_type from the HTTP request
(typically application/json), causing **every uploaded file** to be
stored as Content-Type: application/json.

This breaks S3 web UIs (AWS Console, MinIO Console, rustfs) — images
do not preview, downloads get wrong MIME types.

### The Fix

Override _get_write_parameters() in QfcS3Boto3Storage to extract
the original filename from the versioned path and use it for MIME
detection. Example:

    photo.jpg/v20260610-a1b2c3d4  →  photo.jpg  →  image/jpeg
    notes.gpkg/v20260610-a1b2c3d4 →  notes.gpkg →  application/geopackage+sqlite3

Non-versioned paths (thumbnails, etc.) are unaffected — they fall
through to the original behaviour.

### Changes

- docker-app/qfieldcloud/filestorage/backend.py — +24 lines
- Adds import mimetypes (stdlib)
- Adds _get_write_parameters() override to QfcS3Boto3Storage

### Testing

- [x] Deployed on a live QFieldCloud instance with rustfs S3 backend
- [x] Uploaded JPEG, PNG, GeoPackage, QGIS project — all now get correct Content-Type
- [x] Storage health check passes ({"storage":"ok"})
- [x] Backward compatible — existing objects unaffected

### DIFF
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -9,7 +9,6 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import ContentFile
 from django.core.files.storage import Storage
 from django.http import HttpResponse
-
 from storages.backends.s3 import S3Storage
 
 
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -1,5 +1,7 @@
 import base64
+import mimetypes
 import os
+import re
 from abc import ABC
 
 import requests
@@ -7,7 +9,6 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import ContentFile
 from django.core.files.storage import Storage
 from django.http import HttpResponse
-import mimetypes
 
 from storages.backends.s3 import S3Storage
 
@@ -58,21 +59,30 @@ class QfcS3Boto3Storage(QfcBackendStorageMixin, S3Storage):
         """
         pass
 
-    def _get_write_parameters(self, name, content=None):
+    def _get_write_parameters(self, name: str, content=None) -> dict:
         params = super()._get_write_parameters(name, content)
 
-        # Detect Content-Type from the original filename embedded in the
-        # versioned S3 path.  Versioned paths look like:
-        #   projects/<uuid>/files/DCIM/photo.jpg/v20260610-a1b2c3d4
-        # The trailing version UUID has no file extension, so the parent
-        # class's mimetypes.guess_type(name) always returns None.
+        # Detect content type from the original filename embedded in the
+        # versioned S3 path, so it can be properly set in the object storage
+        # metadata and later returned as a `Content-Type` header when
+        # downloading the object.
+        # Versioned paths look like:
+        #   `projects/<uuid>/files/DCIM/photo.jpg/v20260317162354-512bd29b`
+        # Note the last part of the versioned name has the version timestamp
+        # and a random UUID fragment, which prevents `mimetypes.guess_type`
+        # from guessing a proper content type, as it uses the file extension
+        # to do so.  Without this fix, the parent class's
+        # `_get_write_parameters` would set `ContentType` to `None`.
         parts = name.rsplit("/", 1)
-        if len(parts) == 2 and parts[1].startswith("v2"):
-            original_name = parts[0].rsplit("/", 1)[-1]
-        else:
-            original_name = parts[-1]
+        base_name = parts[-1]
 
-        mime_type, encoding = mimetypes.guess_type(original_name)
+        if (
+            len(parts) == 2
+            and re.fullmatch(r"v20[0-9]{12}-[a-f0-9]{8}$", parts[-1])
+        ):
+            base_name = parts[0].rsplit("/", 1)[-1]
+
+        mime_type, encoding = mimetypes.guess_type(base_name)
         if mime_type:
             params["ContentType"] = mime_type
             if encoding:
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -3,6 +3,7 @@ import mimetypes
 import os
 import re
 from abc import ABC
+from typing import Any
 
 import requests
 from django.core.exceptions import ImproperlyConfigured
@@ -58,7 +59,7 @@ class QfcS3Boto3Storage(QfcBackendStorageMixin, S3Storage):
         """
         pass
 
-    def _get_write_parameters(self, name: str, content: ContentFile | None = None) -> dict:
+    def _get_write_parameters(self, name: str, content: ContentFile | None = None) -> dict[str, Any]:
         params = super()._get_write_parameters(name, content)
 
         # Detect content type from the original filename embedded in the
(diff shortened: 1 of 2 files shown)


suricactus:
Thank you @00-Python . Is there any chance that this is AI assisted change?

00-Python:
AI assisted with the PR write-up and documentation only. I discovered this issue while setting up QField Cloud for the company I work for, then investigated the root cause, implemented the fix, and verified the solution myself.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +58,28 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
         """
         pass
 
+    def _get_write_parameters(self, name, content=None):
```
suricactus: Can you add some typing?

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +58,28 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
...
+        params = super()._get_write_parameters(name, content)
+
+        # Detect Content-Type from the original filename embedded in the
+        # versioned S3 path.  Versioned paths look like:
+        #   projects/<uuid>/files/DCIM/photo.jpg/v20260610-a1b2c3d4
+        # The trailing version UUID has no file extension, so the parent
+        # class's mimetypes.guess_type(name) always returns None.
```
suricactus:
```suggestion
        # Detect content type from the original filename embedded in the
        # versioned S3 path, so it can be properly set to the Object storage
        # metadata and later returned as a `Content-Type` header when 
        # downloading the object.
        # Versioned paths look like:
        # `projects/<uuid>/files/DCIM/photo.jpg/v20260317162354-512bd29b`
        # Note the last part of the versioned name has the version timestamp and
        # a random UUID fragment, which prevents the `mimetypes.guess_type`
        # to guess a proper content type, as it uses the file extension to do so.
        # If the patch is not applied, then `S3Storage._get_write_parameters` 
        # method would set the `ContentType` param to `None`.
```

Slightly elaborate the comment. I guess we need to fix my English.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +58,28 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
...
+        parts = name.rsplit("/", 1)
+        if len(parts) == 2 and parts[1].startswith("v2"):
+            original_name = parts[0].rsplit("/", 1)[-1]
+        else:
+            original_name = parts[-1]
+
+        mime_type, encoding = mimetypes.guess_type(original_name)
```
suricactus:
```suggestion
       parts = name.rsplit("/", 1)
       base_name = parts[-1]

       if (
           len(parts) == 2
           # 
           and re.fullmatch(r"v20[0-9]{12}-[a-f0-9]{8}$", parts[-1])
       ):
           base_name = parts[0].rsplit("/", 1)[-1]

       mime_type, encoding = mimetypes.guess_type(base_name)
```

I think checking for `startswith` and `v2` was a bit too simplistic. The only problem with this check is that we rely that the object storage key format will remain the same. I think having a few tests will ensure that when we change https://github.com/opengisch/QFieldCloud/blob/641cb34789e0ffaaa03bbef3a1c444bb577cffbc/docker-app/qfieldcloud/filestorage/models.py#L265 , then we don't accidentally break this code.

suricactus (changes requested):
Thanks for the contribution! A few comments added.

One general comment - we need written tests, as this is very sensitive part of the codebase.

00-Python:
All review feedback addressed:

- Added type hints on _get_write_parameters
- Elaborated the comment explaining why Content-Type matters for S3 metadata and download headers
- Replaced the startswith("v2") check with re.fullmatch() against the exact version format from FileVersion.display + get_file_version_upload_to
- Added 11 tests, including contract tests that build version suffixes using the same format strings as models.py (line 265 + 345) — if the key format ever changes, those tests break and surface the mismatch

Apologies, the new commits were pushed from my work account (Joseph Webster - jwebstr-b4rn / b4rn.org.uk) rather than this GitHub account (00-Python). Same person, just pushed from the wrong machine.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +59,37 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
...
+        # and a random UUID fragment, which prevents `mimetypes.guess_type`
+        # from guessing a proper content type, as it uses the file extension
+        # to do so.  Without this fix, the parent class's
+        # `_get_write_parameters` would set `ContentType` to `None`.
+        parts = name.rsplit("/", 1)
+        base_name = parts[-1]
+
+        if (
+            len(parts) == 2
+            and re.fullmatch(r"v20[0-9]{12}-[a-f0-9]{8}$", parts[-1])
```
suricactus: NOTE to myself:

Move this regex a bit closer to the actual format definition after merging this:  (https://github.com/opengisch/QFieldCloud/blob/641cb34789e0ffaaa03bbef3a1c444bb577cffbc/docker-app/qfieldcloud/filestorage/models.py#L265)

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +59,37 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
         """
         pass
 
+    def _get_write_parameters(self, name: str, content: ContentFile | None = None) -> dict:
```
suricactus:
```suggestion
    def _get_write_parameters(self, name: str, content: ContentFile | None = None) -> dict[str, Any]:
```

Sorry for being picky, we try to be as type safe as Django allows us with any new code.

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -56,6 +59,37 @@ def patch_nginx_download_redirect(self, response: HttpResponse) -> None:
...
+        # Detect content type from the original filename embedded in the
+        # versioned S3 path, so it can be properly set in the object storage
+        # metadata and later returned as a `Content-Type` header when
+        # downloading the object.
+        # Versioned paths look like:
+        #   `projects/<uuid>/files/DCIM/photo.jpg/v20260317162354-512bd29b`
+        # Note the last part of the versioned name has the version timestamp
+        # and a random UUID fragment, which prevents `mimetypes.guess_type`
+        # from guessing a proper content type, as it uses the file extension
+        # to do so.  Without this fix, the parent class's
```
suricactus:
```suggestion
        # to do so. Without this fix, the parent class's
```
nitpicking :)

suricactus:
Thanks for addressing the feedback! Enable the tests to run, so let's see how it goes. I will rework the tests a bit after merging, but so far it is good.

suricactus (approved):
Thanks!

PR #1661, opened 2026-06-15
Fix data loss scenario when applying a feature addition delta on a layer with attributes having valid default values set to apply on update

### LINKED ISSUE #1660: Applying feature addition delta will overwrite default values and lead to data loss
### What is the bug or the crash? What were your expectations and what actually happened?

When applying a feature addition delta, QFieldCloud will overwrite attribute values taken from the delta (i.e. valid values coming from QField users) when a given attribute has a non-empty default value that is set to apply on update. 

The issue arises here: https://github.com/opengisch/QFieldCloud/blob/master/docker-qgis/qfc_worker/commands/apply_deltas.py#L820-L829

We add attributes coming from the delta to the feature being created, followed by a call to QgsVectorLayerUtils.createFeature, within which default values will overwrite the attribute values we just set provided for attributes having default values re-applied on update.

The solution here is to re-apply the attributes coming from the delta _after_ calling that function.

I've seen a few bugs directly referring to this issue recently on the QField repo:

- https://github.com/opengisch/QField/issues/7555
- https://github.com/opengisch/QField/issues/7500

I'm also pretty certain I've seen bugs where I felt it also was caused by what I've described above. IMHO is a pretty critical issue as it has implications such as:

- expression-generated last updated timestamp attribute on added feature will not reflect the field value rather the time QFC added that feature
- expression-generate attributes looking at capturing GNSS device properties will also end up null
- etc.

### Steps to reproduce the issue

Here's a tiny sample project with two attributes having both default values (an epoch timestamp and a GNSS property):

[crucial.zip](https://github.com/user-attachments/files/28946580/crucial.zip)

1. Unzip that sample project in a local folder
2. In QGIS, create a cloud project, point the local folder to that project, upload to the cloud
3. In QField, download that new cloud project, making sure your GNSS device is turned on
4. Add a point, and note the quality and timestamp value
5. Push the change to the QFieldCloud server
6. Back in QGIS, download the updated crucial.gpkg dataset, and re-open the project
7. Check the attribute values of the newly-added point; you'll notice the quality value is 'error!', and the timestamp will be different from what you generated in QField

### Version

All of 'em

### Operating system name

Other (please specify in additional context)

### Operating system version

It matters not :)

### Reinstall QFieldCloud

- [ ] I have a fresh install of the latest QFieldCloud version, but the problem persists.
- [ ] I have the latest available Docker engine installed on my machine.
- [ ] Problem can be reliably reproduced, doesn't happen randomly.

### Additional context

_No response_

nirvn:
Fixes https://github.com/opengisch/QFieldCloud/issues/1660 (explanation and sample project to reproduce data loss included in that issue).

Test case incoming.


docker-qgis/qfc_worker/commands/apply_deltas.py
```diff
@@ -813,24 +813,32 @@ def create_feature(
...
+                attr_index = fields.indexFromName(attr_name)
+                feat_attrs[attr_index] = attr_value
 
     new_feat = QgsVectorLayerUtils.createFeature(layer, geometry, feat_attrs)
 
     if not new_feat.isValid():
         raise DeltaException("Unable to create a valid feature")
 
+    # Restore values that were overwritten in QgsVectorLayerUtils.createFeature for
+    # attributes having a valid default value set to apply on update
```
suricactus:
```suggestion
    # attributes having a valid default value set to apply on update
    # See https://github.com/opengisch/QFieldCloud/issues/1660
```

suricactus (commented):
Thanks for the nice issue report and the fix. Would you mind providing a test to ensure we never regress on this one?

suricactus:
@nirvn the tests are failing, but the their intention looks good!

docker-app/qfieldcloud/core/tests/test_delta.py
```diff
@@ -209,8 +264,11 @@ def test_push_apply_delta_file_empty_source_layer_id(self):
         )
 
         gdf = self.read_gpkg_layer(project, "testdata.gpkg", "points_xy")
+        f = gdf.iloc[2]
 
-        self.assertEqual(666, gdf.iloc[0]["int"])
```
suricactus, line 213: Why did we remove the check for the first feature here?

suricactus (approved):

PR #1663, opened 2026-06-16
chore: bump dependencies

gounux:
Bumping dependencies to latest versions here and there,  
See individual commits for changelogs and diffs.

One _major_ change spotted for `django-tables2`, see https://github.com/opengisch/QFieldCloud/commit/31f734330062b7d97ee49d069ebb6d1f05ae2335  
Though it seems that QFieldCloud is not impacted.

Closes #1662


suricactus (approved):
Thanks!

PR #1664, opened 2026-06-16
chore: bump `libqfieldsync` to latest

gounux:
PR only for bumping `libqfieldsync` to latest, with a few reflected API changes,

### DIFF
--- a/docker-qgis/Dockerfile
+++ b/docker-qgis/Dockerfile
@@ -82,7 +82,7 @@ ENV PYTHONPATH=/libqfieldsync
 ENV PYTHONPATH=/qfieldcloud-sdk-python:${PYTHONPATH}
 
 # QFC specific variables
-# Controls whether libqfiedlsync should log also to the QGIS message log
+# Controls whether libqfieldsync should log also to the QGIS message log
 ENV IS_WITHIN_QFC_WORKER=1
 
 # Install debugpy for local development and debuggability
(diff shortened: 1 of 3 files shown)


suricactus (approved):
@gounux wondering if we should leave this for the version after the big release that is coming now? Nothing that I think is very fragile, but just in case? I leave it up to you.

gounux:
> @gounux wondering if we should leave this for the version after the big release that is coming now? Nothing that I think is very fragile, but just in case? I leave it up to you.

It makes sense to wait for the big release indeed, though I believe that the latest `libqfieldsync` changes come with nice fixes, so I'd also be happy if this PR is embedded in the next release.

Passing the question back to you, since you're the main author of those libqfs changes :)

suricactus:
If you are around next week for some potential trivial fixes in libqfieldsync, which are very unlikely to be needed, then feel free to merge!

gounux:
Looking at [the commits and diff](https://github.com/opengisch/libqfieldsync/compare/973479e6e1ecffc566063f3b65fadec7ab150119...90eb6640869ad73a996cf4b86db10103dd60f9ec), seems that there are quite some bug fixes as well as nice features as well, e.g. location arrow, coordinate cursor styling etc.

Also pinging @SeqLaz @mohsenD98 @nirvn, as you are the contributors of those changes :)

I'd favor merging this and including this in the next release TBH,

suricactus:
Go for it!

PR #1667, opened 2026-06-16
Subscription cancellation: Use appropriate plan type

lukasgraf:
When creating the inactive dummy subscription that follows a cancelled subscription, pick the plan type of the user's most recent subscription, if available.

This ensures that if they should want to re-activate their subscription, it will be of the appropriate plan type.


duke-nyuki:
Task linked: [QF-8418 Fix subscription cancellation for new plans](https://app.clickup.com/t/2192114/QF-8418)

suricactus (approved):
Thanks!

PR #1668, opened 2026-06-17
 feat: remove project collaborations of users removed from an organization

gounux:
This PR intends to delete the project collaborations of a user that is removed from the members of an Organization.

This applies only to projects owned by the Organization, which makes sense since collaborators of such projects must be members of the Organization.

See provided test scenario for more details and exact behavior.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -823,6 +823,13 @@ class OrganizationMember(models.Model):
     def delete(self, *args, **kwargs) -> tuple[int, dict[str, int]]:
         super().delete(*args, **kwargs)
 
+        # delete the project colloborations of this deleted org member,
+        # as they are no longer part of the organization.
+        ProjectCollaborator.objects.filter(
+            project__owner=self.organization,
+            collaborator=self.member,
+        ).delete()
+
         return TeamMember.objects.filter(
             team__team_organization=self.organization,
             member=self.member,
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8457 Remove project collaborations when Person removed from Organization](https://app.clickup.com/t/2192114/QF-8457)

suricactus (approved):
I love the provided test and the fix!

I think we might have similar problem with team members too?

gounux:
> I think we might have similar problem with team members too?

Not sure I'm following you,

suricactus:
Let's say we have user1 as part of org1, that owns project1.

Let's add user1 as a member of org1. He can then be a collaborator in project1. If we remove user1 from org1 as a member, then we should remove user1 as collaborator from project1. This is what I understood this PR is doing.

Now let's add user1 as a member of org1. Let's also create team1 within org1. Let's add user1 as part of team1. Now when we remove user1 from member list for org1, user1 should also be removed from team1.

I am not sure if this is the case already, but worth verifying.

gounux:
> Let's say we have user1 as part of org1, that owns project1.
> 
> Let's add user1 as a member of org1. He can then be a collaborator in project1. If we remove user1 from org1 as a member, then we should remove user1 as collaborator from project1. This is what I understood this PR is doing.
> 
> Now let's add user1 as a member of org1. Let's also create team1 within org1. Let's add user1 as part of team1. Now when we remove user1 from member list for org1, user1 should also be removed from team1.
> 
> I am not sure if this is the case already, but worth verifying.

Oh okay I see, I believe that this TeamMember removal is already here, it's just the block below the changes of this PR :)

https://github.com/opengisch/QFieldCloud/blob/49331985cc0f48099131cbb796cbc4186884cf5e/docker-app/qfieldcloud/core/models.py#L833-L836

suricactus:
Glad to hear! Less work, more profit!

PR #1651, opened 2026-06-05
Toast error message is misleading when submitting empty collaborator input

manylon:
Message before: `User "" does not exist.`
Proposed message: `Please enter a username or email address.`

PR #1669, opened 2026-06-17
Show correct toast error when submitting empty collaborator input

manylon:
Follows up on #1651

In the referred PR, passing an empty input could silently skip the validation message. Django's AbstractUser.email is an `EmailField(blank=True)`, so accounts created without an email (e.g. via createsuperuser) store `email=""` rather than `NULL`. The DB query `Q(email=username)` with `username=""` would therefore match all such accounts, producing a non-zero user count and routing to the wrong error branch. The early return now ensures the correct validation message is always shown and avoids the unnecessary database lookup.

Before: `Adding multiple collaborators at once is not supported.`
After: `Please enter a username or email address.`

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/projects.py
+++ b/docker-app/qfieldcloud/core/utils2/projects.py
@@ -75,15 +75,17 @@ def create_collaborator_by_username_or_email(
         success, message - whether the collaborator creation was success and explanation message of the outcome
     """
     success, message = False, ""
+
+    if not username.strip():
+        return False, _("Please enter a username or email address.")
+
     users = list(
         Person.objects.filter(Q(username=username) | Q(email=username))
     ) + list(Team.objects.filter(username=username, team_organization=project.owner))
 
     if len(users) == 0:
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8392 Toast error message is misleading when submitting empty collaborator input](https://app.clickup.com/t/2192114/QF-8392)

docker-app/qfieldcloud/core/utils2/projects.py
```diff
@@ -75,15 +75,17 @@ def create_collaborator_by_username_or_email(
         success, message - whether the collaborator creation was success and explanation message of the outcome
     """
     success, message = False, ""
+
+    if not username:
```
suricactus: What will happen if I submit a bunch of empty spaces?

manylon: It will pass. Thank you for the catch I will address it!

suricactus:
Thank you very much. Would you mind updating the commit message body with the new PR description?

suricactus (approved):

PR #1670, opened 2026-06-17
Subscription cancellation: Fix plan type for org users

lukasgraf:
With the recent change (#1667) that bases the dummy follow-up subscription after a cancellation on plan type of the user's most recent subscription, a bug was introduced:

Individual, non-organization users now would be put back on an inactive "Pro" plan instead of an active "Community" plan, which wasn't intended.

This change fixes that by only considering the most recent subscription to determine the plan for organization users.

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -919,9 +919,9 @@ class AbstractSubscription(models.Model):
             .last()
         )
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8466 Fix cancellation for individual users](https://app.clickup.com/t/2192114/QF-8466)

suricactus (approved):
Good!

PR #1673, opened 2026-06-18
fix: When selecting files, ensure they are of project file type, not package files

suricactus:
This is quite error prone, we sometimes do checks on files that are no meant to be checked, e.g. package types.

For example, on deletion we check if a file with given name exists. This should not check for all files with given name, but only those that are project files.

As for the localized layers, this is less likely to be a problem, as we never package localized projects.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -1455,9 +1455,11 @@ class Project(models.Model):
             # Return all layers if the project is missing
             return self.localized_layers
 
-        available_filenames = File.objects.filter(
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8477 When selecting files, ensure they are ofo type project file, not package fiiles](https://app.clickup.com/t/2192114/QF-8477)

gounux (approved):
Thanks! Does that deserve a (little) test ?

suricactus:
Probably a good idea. But merging now to unblock the chain of PRs depending on this.

PR #1674, opened 2026-06-18
feat: remove project collabs from removed org members

gounux:
Follow-up of #1668 -> in this PR we tackle the `OrganizationMemberQuerySet`'s `delete` method, which also handles org members deletions.

### DIFF
--- a/docker-app/qfieldcloud/core/models.py
+++ b/docker-app/qfieldcloud/core/models.py
@@ -734,15 +734,24 @@ class Organization(User):
 class OrganizationMemberQueryset(models.QuerySet):
     @transaction.atomic
     def delete(self, *args, **kwargs):
+        # delete the team memberships of this deleted org member,
+        # as it is no longer part of the organization.
         team_total, team_summary = TeamMember.objects.filter(
             team__team_organization__in=self.values_list("organization"),
             member__in=self.values_list("member"),
         ).delete()
 
+        # delete the project collaborations of this deleted org member,
+        # as it is no longer part of the organization.
+        collab_total, collab_summary = ProjectCollaborator.objects.filter(
+            project__owner__in=self.values_list("organization"),
+            collaborator__in=self.values_list("member"),
+        ).delete()
+
         org_total, org_summary = super().delete(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8457 Remove project collaborations when Person removed from Organization](https://app.clickup.com/t/2192114/QF-8457)

docker-app/qfieldcloud/core/models.py
```diff
@@ -734,15 +734,24 @@ def save(self, *args, **kwargs):
...
+            project__owner__in=self.values_list("organization"),
+            collaborator__in=self.values_list("member"),
+        ).delete()
+
         org_total, org_summary = super().delete(*args, **kwargs)
 
-        total = org_total + team_total
-        summary = {**org_summary, **team_summary}
+        total = org_total + team_total + collab_total
+        summary = {**org_summary, **team_summary, **collab_summary}
```
gounux, line 754: Not sure what would be the impacts of adding those TBH,

suricactus: Sound good to me, the `*_summary` is a dict, where the key is the model's name (e.g. `core.ProjectCollaborator`) and the value is the number of deleted elements. 

See https://docs.djangoproject.com/en/6.0/ref/models/querysets/#delete

docker-app/qfieldcloud/core/models.py
```diff
@@ -823,8 +832,8 @@ def clean(self) -> None:
     def delete(self, *args, **kwargs) -> tuple[int, dict[str, int]]:
         super().delete(*args, **kwargs)
 
-        # delete the project colloborations of this deleted org member,
-        # as they are no longer part of the organization.
+        # delete the project collaborations of this deleted org member,
+        # as it is no longer part of the organization.
```
suricactus, line 836:
```suggestion
        # as the member is no longer part of the organization.
```

suricactus (approved):
LGTM!

docker-app/qfieldcloud/core/models.py
gounux: Thanks!

PR #1677, opened 2026-06-18
Ensure that produced version suffix are always matching a regex

suricactus:
Before we had a bit too much freedom with how the version suffix was generated. However since `150b65dc4fb2f6d9c6dafff3071895bc19c2f093`, the `QfcS3Boto3Storage` expects the file key to be of given format.

Centralizing the management of the matching regex and validating against it as soon as a new file is generated make me feel more confident we do not accidentally break something.

Took the opportunity to remove a few less useful tests with it.

Continuation of https://github.com/opengisch/QFieldCloud/pull/1658

### DIFF
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -1,7 +1,6 @@
 import base64
 import mimetypes
 import os
-import re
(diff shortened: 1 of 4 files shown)


lukasgraf (approved):
LGTM 👍

PR #1681, opened 2026-06-22
feat: Prepare for adding captcha on the signup page

suricactus:
Currently we were very liberal on people registering on the QFC service.

While `django-simple-captcha` is not the most advanced captcha, it is a good start to filter out the majority of the bots abusing our service.

This PR/commit also fixed a bug that the JS block on the signup page was loading twice, as the `block additional_js` was wrongly nested within the `block content`, which made it rendered twice. Once the `block additional_js` was moved outside, it will only be replaced in the root `base.html` template.

Finally, this PR does not add the captcha on the signup page, we need to setup a custom the `SignUpForm` to do so, which is out of scope.

### DIFF
--- a/docker-app/qfieldcloud/settings.py
+++ b/docker-app/qfieldcloud/settings.py
@@ -143,6 +143,7 @@ INSTALLED_APPS = [
     "bootstrap4",
     "sri",
     "corsheaders",
+    "captcha",
     # To ensure that exceptions inside other apps' signal handlers do not affect the integrity of file deletions within transactions, `django_cleanup` should be placed last in `INSTALLED_APPS`. See https://github.com/un1t/django-cleanup#configuration
     "django_cleanup.apps.CleanupConfig",
 ]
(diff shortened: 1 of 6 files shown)


docker-app/qfieldcloud/settings.py
```diff
@@ -1012,6 +1013,41 @@ def before_send(event, hint):
...
+# See https://django-simple-captcha.readthedocs.io/en/latest/advanced.html#captcha-timeout
+CAPTCHA_TIMEOUT = 5
+
+# Sets the length, in chars, of the generated captcha. (for the 'captcha.helpers.random_char_challenge' challenge)
+# See https://django-simple-captcha.readthedocs.io/en/latest/advanced.html#captcha-length
+CAPTCHA_LENGTH = 6
+
+# When set to True, the string “PASSED” (any case) will be accepted as a valid response to any CAPTCHA. Use this for testing purposes.
+# See https://django-simple-captcha.readthedocs.io/en/latest/advanced.html#captcha-test-mode
+CAPTCHA_TEST_MODE = DEBUG or ENVIRONMENT == "test"
```
gounux, line 1048:
```suggestion
CAPTCHA_TEST_MODE = (DEBUG or ENVIRONMENT == "test")
```

Find it easier to read, though it's not super important, feel free to ignore.

duke-nyuki:
Task linked: [QF-8491 Add captcha on the signup page](https://app.clickup.com/t/2192114/QF-8491)

docker-app/qfieldcloud/settings.py
suricactus: `ruff` disagrees with your suggestion, can we ignore it and get an approval to merge?

gounux (approved):

PR #1683, opened 2026-06-22
Rename the label of `Project.is_featured` from "Is sticky" to "Is featured"

suricactus:
The original name was "is_sticky" and we forgot to update the label.

The migration is directly edited, as the value does not have any effect on the database and it is safe to do.


gounux (approved):
Was wondering it this would deserve a new migration here, since a previous one is being rewritten, probably not. LGTM!

suricactus:
No migration, there is no diff between the migration tree and the state within the files, since I edited the original migration in place.

PR #1690, opened 2026-07-01
chore: bump `boto3` and related dependencies

gounux:
Since some Object Storage providers were not always compatible with recent versions of boto3 (e.g. >= `1.37`),

See individual commit messages for changelogs, diffs and version changes.

The `boto3-stubs` and `mypy-boto3-s3` are built from [this repository](https://github.com/youtype/mypy_boto3_builder), could not really find a changelog for those,

Needs to be tested on several providers before getting merged.

### DIFF
--- a/docker-app/requirements/requirements.in
+++ b/docker-app/requirements/requirements.in
@@ -1,5 +1,5 @@
-boto3-stubs==1.35.90
-boto3==1.35.90
+boto3-stubs==1.43.38
+boto3==1.43.38
 deprecated==1.3.1
 django-allauth[socialaccount]==65.18.0
 django-auditlog==3.4.1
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-5297 [sentry] boto3 >= 1.37.0 does not work with all Object Storage providers](https://app.clickup.com/t/2192114/QF-5297)

suricactus (approved):
I am not sure if we shall put all the deps in single commit in such cases, as they are quite tightly coupled. No strong opinion, arguments for both approaches.

Anyways, glad we are going forward with this dependency.

gounux:
> I am not sure if we shall put all the deps in single commit in such cases, as they are quite tightly coupled. No strong opinion, arguments for both approaches.

Good approach, indeed makes sense to group them into a single commit, so squashed with a verbose commit message.

Could test on several s3 providers, looks okey. I'll then merge when CI passes.

PR #1691, opened 2026-07-01
feat(docker): Build and push QGIS3 and QGIS4 images

toebivankenoebi:
Build and push QGIS3 and QGIS4 images to dockerhub based on the build args in `docker-compose.yml`.

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -34,7 +34,24 @@ jobs:
             enable_dockerhub: true
 
           -
+            # Keep the existing QGIS image for backwards compatibility,
+            # using the qgis3 build definition from docker-compose.yml.
             service_name: qgis
+            compose_service: qgis3
+            dockerfile: docker-qgis/Dockerfile
+            docker_context: docker-qgis
+            enable_dockerhub: true
+
+          -
+            service_name: qgis3
+            compose_service: qgis3
+            dockerfile: docker-qgis/Dockerfile
+            docker_context: docker-qgis
+            enable_dockerhub: true
+
+          -
+            service_name: qgis4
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8505 Build and push qgis3 and qgis4 docker images](https://app.clickup.com/t/2192114/QF-8505)

.github/workflows/build_and_push.yml
```diff
@@ -97,6 +114,21 @@ jobs:
...
+        id: compose_build
+        if: ${{ matrix.services.compose_service }}
+        env:
+          COMPOSE_SERVICE: ${{ matrix.services.compose_service }}
+          DEBUG: "0"
+        run: |
+          {
+            echo 'build_args<<EOF'
+            docker compose -f docker-compose.yml config --format json \
+              | jq -r --arg service "$COMPOSE_SERVICE" '.services[$service].build.args // {} | to_entries[] | "\(.key)=\(.value)"'
```
lukasgraf, line 128: Wasn't sure at first if we can rely on `jq` to be available, but it's part of the GitHub runner [Ubuntu 24.04 software list](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md). 👍

lukasgraf (approved):
LGTM! 👍

.github/workflows/build_and_push.yml
suricactus: Just pass by to say it's hard to immeditely follow what we want to do on this line.

I guess this is the easiest way to achieve what this line is doing and cannot be rewritten, so it's fine. But writing a tiny comment above what it actually does on high level would not hurt anyone IMO.

PR #1688, opened 2026-06-30
feat: add jobs logs retention days constance config

gounux:
This PR intends to add a new Constance config: `JOBS_LOGS_RETENTION_PERIOD_DAYS` with defaults to `90`.

This is the first step, with upcoming PRs coming for Jobs `output` and `feedback` automatic cleaning, after a retention period defined by this new Constance config.

Reason behind this is that Jobs "logs" can take quite some database space, and are not really relevant 90 days after a run.

PR #1692, opened 2026-07-01
feat: clear Jobs outputs after retention period

gounux:
Follow-up of #1688: add a cron job that nightly sets the Jobs outputs to null, after the `JOBS_LOGS_RETENTION_PERIOD_DAYS` retention period.

### DIFF
--- a/docker-app/qfieldcloud/core/cron.py
+++ b/docker-app/qfieldcloud/core/cron.py
@@ -107,3 +107,18 @@ class DeleteObsoleteProjectPackagesJob(CronJobBase):
         )
 
         packages.delete_obsolete_packages(projects)
+
+
+class ClearJobOutputsAfterRetentionPeriodJob(CronJobBase):
+    # runs every night at 02:00
+    schedule = Schedule(run_every_mins=1440, run_at_times=["02:00"])
+    code = "qfieldcloud.clear_jobs_outputs_after_retention_period"
+
+    def do(self):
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8432 Store the `Job.output` for specific time period and then set to null](https://app.clickup.com/t/2192114/QF-8432)

docker-app/qfieldcloud/core/cron.py
```diff
@@ -107,3 +107,18 @@ def do(self):
         )
 
         packages.delete_obsolete_packages(projects)
+
+
+class SetNullJobsOutputsAfterRetentionPeriodJob(CronJobBase):
```
lukasgraf: I would consider renaming this to something like `ClearJobOutputsAfterRetentionPeriodJob`.
(And rename the `code` accordingly).

The current name is correct, but very low level, and doesn't immediately help in unterstanding what the high level of this job is.

lukasgraf (approved):
LGTM 👍 - just a suggestion for a slightly different name 😃

docker-app/qfieldcloud/core/cron.py
gounux: Indeed :) See https://github.com/opengisch/QFieldCloud/pull/1692/changes/d9c509a03ed7acccdb5ea949d7d651ed91a5037c

suricactus (approved):
Happy to see this merged.

Issue #1694, opened 2026-07-03
S3 bucket qfieldcloud-local not creating while having multpiple apps on docker in the same VM


wojtek19022:
### What is the bug or the crash? What were your expectations and what actually happened?

During set up on my own VM with multiple docker apps i discovered that S3 bucket couldn't be correctly created. it was connected with http://172.17.0.1 addres already assigned to other app (my QFieldCloud had local addres of http://172.19.0.1. 

There is needed change in the S3 addres reference!

### Steps to reproduce the issue

```bash
sudo su
git clone --recurse-submodules -b release https://github.com/opengisch/QFieldCloud
cd QFieldCloud 
cp .env.example .env

# change HTTP port to 88
# change HTTPS port to 8443
# change NGINX_ALLOW_INTERNAL_HTTP to 1
docker compose up -d --build
```

On the container qfieldcloud-createbuckets-1 there was below error after clean run:
```bash
Creating bucket for: default
✓ Alias 'default' configured successfully.
✗ Failed to create bucket: Network error: dispatch failure
Traceback (most recent call last):
File "/createbuckets.py", line 44, in <module>
    subprocess.run(
    ~~~~~~~~~~~~~~^
        f"/usr/bin/rc mb --ignore-existing {storage_name}/{bucket_name}",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        shell=True,
        ^^^^^^^^^^^
        check=True,
        ^^^^^^^^^^^
    )
    ^
File "/usr/lib/python3.14/subprocess.py", line 578, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '/usr/bin/rc mb --ignore-existing default/qfieldcloud-local' returned non-zero exit status 3.
```

### Version

v26.18

### Operating system name

Linux

### Operating system version

Ubuntu22.04 and Ubuntu 24.01

### Reinstall QFieldCloud

- [x] I have a fresh install of the latest QFieldCloud version, but the problem persists.
- [x] I have the latest available Docker engine installed on my machine.
- [x] Problem can be reliably reproduced, doesn't happen randomly.

suricactus:
Hi @wojtek19022 , thanks for trying out QFieldCloud!

What happens when you change the `STORAGES.endpoint_url` configuration in your `.env` file at this line:

[...]

PR #1695, opened 2026-07-03
Added better config for rustfs API reference #1694

wojtek19022:


suricactus (commented):
Unfortunately this did not work in earlier versions of docker/docker compose on Windows. Are you able to verify the proposed change still works on Windows?

Also, would you mind adding a bit of description of the motivation for this PR, related issue in the backlog, why you chose the proposed approach and on what system you proved it to work?

Check our https://github.com/opengisch/QFieldCloud#collaboration guide for more info.

wojtek19022:
Actually i don't have windows machine to verify it to work on windows. It is a reference to #1694 issue. Provided solution worked on my machine with Ubuntu 22.04.5 LTS and `Docker version 29.4.2, build 055a478`

suricactus:
If changing the configuration works for you, I will close this PR, as we never intended to have QFieldCloud running along other services and if you needed the flexibility to do so, you already have it by chaning the .env file.

Since the proposed change cannot be tested on windows for now,  I would rather close this issue. If we confirm it works fine for developers on windows, I am fine to merge this PR.

PR #1701, opened 2026-07-09
Add a new enum to `Project` model for project type

manylon:
As QFieldCloud is growing, there are more and more types of projects that we will add.

Add an explicit `Project.ProjectType` IntegerChoices field (`REGULAR` / `SHARED_DATASETS`) that is derived from the project name on every `save()`, and use it everywhere the name comparison was previously used (`is_shared_datasets_project`, `get_shared_datasets_project`).

- Add `project_type` field with migration `0003_project_project_type`, backfilling existing rows named `shared_datasets`.
- Expose `project_type` as a read-only field on `ProjectSerializer` and in the admin.
- Add tests covering default type on creation, type assignment when created or renamed to/from the shared datasets name.

Admin images:
image
image


duke-nyuki:
Task linked: [QF-8181 Add a new enum about the project type](https://app.clickup.com/t/2192114/QF-8181)
Task linked: [QF-8198 Create a new "project" Django app and move the `Project` model there](https://app.clickup.com/t/2192114/QF-8198)

suricactus (approved):
Thanks!

PR #1703, opened 2026-07-10
Replace `Project.the_qgis_file_name` with `Project.the_qgis_file` that is a FK to `filestorage.File` of the `.qgs` file

manylon:
For historical reasons, **QFC** projects track their **QGIS** file by filename instead of a direct FK to the `FileVersion` instance holding the latest version of that file. That dates back to before **QFC** was aware of stored files. This PR switches all references of `Project.the_qgis_file_name` to a proper FK.

What this PR adds:
- Add `Project.the_qgis_file`, a nullable FK to `filestorage.File`, replacing the `the_qgis_file_name` `TextField`.
- Add project migration `0004`: creates the FK, backfills it from each project's matching `filestorage.File`, then drops the old field.
- Keep `the_qgis_file_name` and `has_the_qgis_file` as derived read-only properties.
- Update upload/delete flows in `filestorage/view_helpers.py` to use the new FK, including recalculating it when a file version is deleted.
- Use `select_related` for reads and `select_for_update(of=("self",))` for locking (needed to avoid Postgres's restriction on locking the nullable side of the resulting outer join).
- Update `calcprojectstorage` and `admin` to use the new FK.
- Update `get_context` for `JobRun`.
- Avoid N+1 queries when reading `the_qgis_file_name` in tests.


duke-nyuki:
Task linked: [QF-8181 Add a new enum about the project type](https://app.clickup.com/t/2192114/QF-8181)
Task linked: [QF-8547 Replace `Project.the_qgis_file_name` with `Project.the_qgis_file_version` that is a FK to `FileVersion` of the `.qgs` file](https://app.clickup.com/t/2192114/QF-8547)

docker-app/qfieldcloud/core/tests/test_qgis_file.py
```diff
@@ -112,7 +112,10 @@ def test_push_multiple_files(self):
                 Project.objects.get(pk=self.project1.pk).project_files_count, 0
             )
             self.assertEqual(
```
suricactus: `self.assertIsNone` is better here, no?

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1336,6 +1336,10 @@ def get_form(self, *args, **kwargs):
     def project_files(self, instance):
         return instance.pk
 
+    @admin.display(description=_("QGIS project file"))
+    def the_qgis_file_name(self, obj: Project) -> str | None:
+        return obj.the_qgis_file_name
```
suricactus, line 1341: Note that after this PR is merged I am ok to convert this field to a file input.

docker-app/qfieldcloud/core/tests/test_jobs.py
```diff
@@ -40,6 +40,11 @@ def setUp(self):
         # Create a project
         self.p1 = Project.objects.create(name="p1", is_public=False, owner=self.u1)
 
+    def refresh_with_qgis_file(self, project: Project) -> None:
```
suricactus:
```suggestion
    def refresh_project(self, project: Project) -> None:
```

Shorter name :)

As for the contents, I am not sure if I follow. Quoting the docs (by the end of docs block)  https://docs.djangoproject.com/en/6.0/ref/models/instances/#django.db.models.Model.refresh_from_db:

> You can cache related objects that otherwise would be cleared from the reloaded instance:
> 
>     obj.refresh_from_db(from_queryset=MyModel.objects.select_related("related_field"))

So the `from_queryset=... select_related ...` is to cache or to drop the cache? If it is for explicit caching, I don't think we need it, so we can keep simple `project.refresh_from_db()` calls.

docker-app/qfieldcloud/filestorage/tests/test_files_api.py
```diff
@@ -56,6 +56,11 @@ def setUp(self):
             file_storage="default",
         )
 
+    def refresh_p1(self) -> None:
```
suricactus:
```suggestion
    def refresh_project(self, project: Project) -> None:
```

If we need this, we need to keep the API similar across the files.

docker-app/qfieldcloud/project/migrations/0004_project_the_qgis_file_version.py
```diff
@@ -0,0 +1,59 @@
...
+
+def backfill_the_qgis_file_version(apps, schema_editor):
+    Project = apps.get_model("project", "Project")
+    File = apps.get_model("filestorage", "File")
+
+    for project in Project.objects.filter(the_qgis_file_name__isnull=False):
+        file = File.objects.filter(
+            project=project,
+            name=project.the_qgis_file_name,
+            file_type=1,  # File.FileType.PROJECT_FILE
```
suricactus:
```suggestion
            file_type=FILE_TYPE_PROJECT_FILE
```

Just add a constant within the migration/function. Avoid inline comments, add them on the line before.

docker-app/qfieldcloud/project/migrations/0004_project_the_qgis_file_version.py
```diff
@@ -0,0 +1,59 @@
...
+        ).first()
+
+        if file is None:
+            continue
+
+        project.the_qgis_file_version = file.latest_version
```
suricactus:
```suggestion
        project.the_qgis_file_version_id = file.latest_version_id
```

You can save a few seconds in the migration if we don't fetch the whole version object in `file.latest_version`.

docker-app/qfieldcloud/project/migrations/0004_project_the_qgis_file_version.py
```diff
@@ -0,0 +1,59 @@
...
+        project.save(update_fields=["the_qgis_file_version"])
+
+
+def restore_the_qgis_file_name(apps, schema_editor):
+    Project = apps.get_model("project", "Project")
+
+    for project in Project.objects.filter(the_qgis_file_version__isnull=False):
```
suricactus:
```suggestion
    project_qs = Project.objects.select_related("the_qgis_file_version__file").filter(the_qgis_file_version__isnull=False)

    for project in project_qs:
```

Don't we need to do this to save an extra query?

docker-app/qfieldcloud/project/models.py
```diff
@@ -595,7 +603,14 @@ def get_missing_localized_layers(self) -> list[dict[str, Any]]:
 
     @property
     def has_the_qgis_file(self) -> bool:
-        return bool(self.the_qgis_file_name)
+        return self.the_qgis_file_version_id is not None
+
+    @property
+    def the_qgis_file_name(self) -> str | None:
```
suricactus, line 609: If `docker-app/worker_wrapper/wrapper.py` and `docker-app/qfieldcloud/core/admin.py` are the only two places where this is used, I am ok to kill `the_qgis_file_name` as a property.

suricactus (commented):
Some small comments to be addressed before merging.

suricactus (changes requested):

docker-app/qfieldcloud/core/tests/test_qgis_file.py
manylon: aa yes, thanks!

docker-app/qfieldcloud/core/tests/test_jobs.py
manylon: `refresh_from_db()` alone would reload the fields, but drop any cached FK relations, so the next access to `the_qgis_file_version.file` would trigger extra  queries. Passing `from_queryset=...select_related("the_qgis_file_version__file")` reloads the instance and recaches that relation in the same query, so assertions on `the_qgis_file_version.file` don't hit the db again. Our tests are already taking a while, so I think it's worth keeping them optimized.

suricactus: Ok, I see. I wish this was the reason for our tests being slow :) I am -0, no strong arguments against, so let's keep what you proposed.

Then just adapt the naming to be the same across files with the cleaner `refresh_project`. I guess adding a small comment why this method exists would be of great value too, right now it's not clear if it is needed for functional or performance reasons.

---

I might become +0 (depends on the resulting complexity) if we move this method on the project instance itself, so we `project.refresh_from_db()` automatically populates `the_qgis_file_version__file` whenever the value was available.

docker-app/qfieldcloud/project/migrations/0004_project_the_qgis_file_version.py
manylon: Thanks!

docker-app/qfieldcloud/project/migrations/0004_project_the_qgis_file_version.py
manylon: yes, thanks!

docker-app/qfieldcloud/project/models.py
manylon: It is used in 10+ files, excluding migrations 🙃

docker-app/qfieldcloud/core/tests/test_jobs.py
manylon: Small steps make the difference 🙈

docker-app/qfieldcloud/project/models.py
suricactus: We wait for a better day I guess...

manylon: Resolving

docker-app/qfieldcloud/core/tests/test_jobs.py
manylon: We have a lot of other tests that use `project.refresh_from_db()` that don't actually need `the_qgis_file_version.file`. So I would not recommend overriding the  `refresh_from_db()` method.

docker-app/qfieldcloud/core/admin.py
manylon: Task created! https://app.clickup.com/t/2192114/QF-8599 and will resolve.

manylon:
Thanks for the review @suricactus. I have addressed the changes.

docker-app/qfieldcloud/project/models.py
manylon: Created a task for this - https://app.clickup.com/t/2192114/QF-8601

manylon:
Hi @suricactus , as discussed I've changed the FK from `filestorage.FileVersion` to `filestorage.File`

docker-app/qfieldcloud/project/models.py
```diff
@@ -253,7 +253,15 @@ class Meta:
     )
 
     description = models.TextField(blank=True)
-    the_qgis_file_name = models.TextField(blank=True, null=True)
+
+    the_qgis_file_id: uuid.UUID | None
```
```suggestion
    the_qgis_file_id: int | None
```



https://github.com/opengisch/QFieldCloud/blob/master/docker-app/qfieldcloud/filestorage/models.py#L76

suricactus (approved):
Last small comment.

manylon:
Thanks for the review @suricactus, will clean up the history and merge!

PR #1706, opened 2026-07-13
Disable Gunicorn control socket to fix `Permission denied` error

lukasgraf:
This fixes the following (non-fatal) error we started seeing in our `app` container logs:

```
  [ERROR] Control server error: [Errno 13] Permission denied: '/nonexistent'
```

Gunicorn recently introduced a change that [starts a control socket server](https://gunicorn.org/reference/settings/#control) by default, creating a socket at `$HOME/.gunicorn/gunicorn.ctl`. The `app` system user has `HOME=/nonexistent` (Debian's default for `adduser --system`), a mode-000 directory, causing a Permission denied error on every startup.

The `--no-control-socket` flag disables this socket, since we don't need it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
       gunicorn
         qfieldcloud.wsgi:application
         --bind 0.0.0.0:8000
+        --no-control-socket
         --timeout ${GUNICORN_TIMEOUT_S}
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8558 Disable Gunicorn control socket](https://app.clickup.com/t/2192114/QF-8558)

suricactus (approved):
Thanks!

PR #1708, opened 2026-07-14
open_qgis_file: Handle unexpected .qgs names inside .qgz

lukasgraf:
QGIS uses QFileInfo's `baseName()` to determine the stem of a `.qgs` file inside a `.qgz` file. This may lead to situations where the `.qgs` file doesn't have exactly the same stem as the `.qgz` that contains it, for example when the qgz filename contains more than one dot.

This change makes `open_qgis_file` more robust to deal with this fact, by:
  - Still trying the expected name first
  - But, when unsuccessful, falling back to the first `.qgs` file it finds and logging the expected and actual .qgs names
  
(Updated both `open_qgis_file` functions in app and worker)

### DIFF
--- a/docker-app/qfieldcloud/filestorage/utils.py
+++ b/docker-app/qfieldcloud/filestorage/utils.py
@@ -1,5 +1,6 @@
 import hashlib
 import io
+import logging
 import re
 import uuid
 import xml.etree.ElementTree as ET
@@ -18,6 +19,8 @@ from django.http import HttpRequest
 from django.utils.translation import gettext_lazy as _
 
 from qfieldcloud.core.exceptions import InvalidRangeError
+
+logger = logging.getLogger(__name__)
 
 filename_validator = RegexValidator(
     settings.STORAGES_FILENAME_VALIDATION_REGEX,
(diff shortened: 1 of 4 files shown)


duke-nyuki:
Task linked: [QF-8569 Handle unexpected qgs names inside qgz](https://app.clickup.com/t/2192114/QF-8569)

docker-app/qfieldcloud/filestorage/utils.py
```diff
@@ -268,7 +271,25 @@ def open_qgis_file(
...
+                qgs_filenames = sorted(archive_names)
+
+                if not qgs_filenames:
+                    raise KeyError(f"No .qgs file found in archive: {filename}")
+
+                actual_qgs_name = qgs_filenames[0]
+                logger.info(
+                    f"Expected .qgs file '{expected_qgs_name}' not found in qgz '{filename}', "
+                    f"found '{actual_qgs_name}' instead"
+                )
```
suricactus, line 295: I am not sure if I follow here? The archive might have many other files in there, so why we would assume that the first file in the archive is the `.qgs` file?


```suggestion
                qgs_filename = None
                for archive_file in sorted(archive_names):
                    if Path(archive_file).suffix == '.qgs':
                        qgs_filename = archive_name
                        break

                if not qgs_filename:
                    raise KeyError(f"No '.qgs' file found in the '.qgz' archive {filename}")

                logger.info(
                    f"Expected '.qgs' file '{expected_qgs_name}' not found in '.qgz' archive '{filename}', "
                    f"found and fallback to '{qgs_filename}' instead!"
                )
```

lukasgraf: Damn, good catch, thanks. I totally missed filtering for `.qgs` - that ended up on the floor somehow after I rewrote a first approach using `glob` 🤦 

(Picking the first one out of the alphabetically ordered list has a purpose - I want it do be deterministic, so that the fallback strategy is predictable).

suricactus: yup, I am totally fine with the `sorted`, therefore I kept it in the proposal above.

suricactus (approved):
Thanks a lot!

PR #1715, opened 2026-07-20
We said bye to X, let's remove it from QFieldCloud altogether

nirvn:
Aligns with https://x.com/QFieldForQGIS/status/2054926023532114234


Rakanhf (approved):
LGTM 🚀

PR #1716, opened 2026-07-21
Simplify captcha by less rotation, down from 6 to 4 chars and less noise

suricactus:
Before it was shared it is too hard to read. Let's try to simplify it.

| before | after |
|-|-|
| image | image |


lukasgraf (approved):
LGTM! 👍

PR #1722, opened 2026-07-22
typing: Refactor typing to use `Self`

suricactus:
Since we ensured newer Python version, we can now kill some TODOs to switch from returning a hard coded type to return `Self`.


Rakanhf (approved):
LGTM !

PR #1725, opened 2026-07-23
chore: bump `pillow` from `12.2.0` to `12.3.0` in `requirements.in`

gounux:
Changelog: https://github.com/python-pillow/Pillow/releases
Diff: https://github.com/python-pillow/Pillow/compare/12.2.0...12.3.0

Follow-up of #1721 which was only bumping the `pillow` version in `requirements.txt`,  
Closes #1724


suricactus (approved):

gounux:
Thanks for the approval @suricactus,  
Forgot to mention in the description that `pipcompile` was run, so we should be okay for this dependency :)

PR #1726, opened 2026-07-23
Switch from `pip` and `pip-tools` to `uv`

suricactus:
Why use two tools and pass some questionable configuration (e.g. `legacy-resolver`), when we can use one tool that is faster.

https://www.youtube.com/watch?v=gAjR4_CbPpQ

Future work may include enabling cache for local development.


scripts/object_storage_cleaner/test.py
```diff
@@ -262,16 +262,16 @@ def test_retention_period_filter(self):
...
+        # The file was deleted 2 seconds ago. 2s < 10s.
         # So the file is NEWER than the retention period. It should be PROTECTED (skipped).
         result_long_retention = self.run_script(
             [
                 "--dry-run",
                 "--prefix",
                 self.unique_prefix,
                 "--retention-period",
-                "5 seconds",
+                "10 seconds",
```
Rakanhf, line 274: I believe that these changes should be in a separate PR, but I think it's fine for now.

Rakanhf (approved):
Thank you, LGTM !

scripts/object_storage_cleaner/test.py
suricactus: They were required to make the tests pass, so they are somehow related.

PR #1732, opened 2026-07-28
Fix `'NoneType' object has no attribute 'name'` in `the_qgis_file_name`

manylon:
`the_qgis_file_id` could reference a file that no longer resolves via`the_qgis_file`, causing `has_the_qgis_file` to report `True` while `the_qgis_file` was None, so accessing `.name` raised `'NoneType' object has no attribute 'name'`. 

This PR makes the check to `the_qgis_file` itself instead of the FK `id`.

### DIFF
--- a/docker-app/qfieldcloud/project/models.py
+++ b/docker-app/qfieldcloud/project/models.py
@@ -603,11 +603,11 @@ class Project(models.Model):
 
     @property
     def has_the_qgis_file(self) -> bool:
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8620 Fix `'NoneType' object has no attribute 'name'` in `the_qgis_file_name`](https://app.clickup.com/t/2192114/QF-8620)

suricactus (approved):
Thanks!

PR #1734, opened 2026-07-29
refactor: implement single record trial subscriptions

Rakanhf:
To simplify managing trial subscriptions, we will now use a single record for trial subscriptions instead of a separate subscription record
This change will help to simplify managing trial subscriptions

- Added `cancel_current_subscriptions` method to handle cancellation of existing subscriptions when a new one is activated
- Adjusted logic in `create_subscription` to ensure proper handling of trial subscriptions
- Adjusted tests accordingly to the changes
- Added `test_trials` test case to verify the behavior of trial subscriptions
- Adjusted `active_storage_total_bytes` to exclude storage packages from the quotas granted by a trial plan

### DIFF
--- a/docker-app/qfieldcloud/subscription/models.py
+++ b/docker-app/qfieldcloud/subscription/models.py
@@ -537,10 +537,15 @@ class SubscriptionQuerySet(models.QuerySet):
                 Subscription.Status.ACTIVE_PAST_DUE,
             )
         )
+        is_trial_condition = Q(status=Subscription.Status.ACTIVE_TRIAL) & Q(
+            trial_expires_at__gt=V("now")
+        )
         return self.annotate(
             is_active=Case(
                 When(
-                    is_period_active_condition & is_status_active_condition, then=True
+                    is_period_active_condition
+                    & (is_status_active_condition | is_trial_condition),
+                    then=True,
                 ),
                 default=False,
             ),
@@ -681,6 +686,9 @@ class AbstractSubscription(models.Model):
 
     @property
     def active_storage_total_bytes(self) -> int:
+        if self.is_trialing:
+            return self.included_storage_bytes
+
         return self.included_storage_bytes + self.active_storage_package_bytes
 
     @property
@@ -928,6 +936,40 @@ class AbstractSubscription(models.Model):
         return result
 
     @classmethod
+    def cancel_current_subscriptions(
+        cls,
+        account: UserAccount,
+        active_until: datetime,
+        exclude_pk: int | None = None,
+    ) -> int:
+        """Cancels all the current subscriptions of a given account at a given moment.
+
+        An account can have at most one subscription active at a time, so whenever a
+        subscription starts being active, the one it replaces must be closed at that
+        exact boundary. Call this before a subscription starts being active, either
+        because it is born with `active_since` set, or because `active_since` is set
+        on it later on (e.g. when the remote system confirms the payment).
+
+        Args:
+            account: the account whose current subscriptions are cancelled.
+            active_until: the moment the cancelled subscriptions stop being active.
+                Usually the `active_since` of the subscription that takes over.
+            exclude_pk: primary key of a subscription that should not be cancelled.
+
+        Returns:
+            the number of cancelled subscriptions.
+        """
+        subscriptions_qs = cls.objects.current().filter(account=account)  # type: ignore
+
+        if exclude_pk is not None:
+            subscriptions_qs = subscriptions_qs.exclude(pk=exclude_pk)
+
+        return subscriptions_qs.update(
+            status=cls.Status.INACTIVE_CANCELLED,
+            active_until=active_until,
+        )
+
+    @classmethod
     def update_subscription(
         cls,
         subscription: Self,
@@ -949,18 +991,13 @@ class AbstractSubscription(models.Model):
             with advisory_lock(f"billing:account:{subscription.account_id}"):
                 cls.objects.select_for_update().get(id=subscription.id)  # type: ignore
                 update_fields = []
+                active_since = kwargs.get("active_since")
 
-                if (
-                    subscription.active_since is None
-                    and kwargs.get("active_since") is not None
-                ):
-                    cls.objects.current().filter(
-                        account=subscription.account,
-                    ).exclude(  # type: ignore
-                        pk=subscription.pk,
-                    ).update(
-                        status=Subscription.Status.INACTIVE_CANCELLED,
-                        active_until=kwargs["active_since"],
+                if subscription.active_since is None and active_since is not None:
+                    cls.cancel_current_subscriptions(
+                        subscription.account,
+                        active_until=cast(datetime, kwargs["active_since"]),
+                        exclude_pk=subscription.pk,
                     )
 
                 for attr_name, attr_value in kwargs.items():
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-8420 Add trial plan as a field of the current subscription instead of separate subscription record](https://app.clickup.com/t/2192114/QF-8420)

docker-app/qfieldcloud/subscription/migrations/0013_subscription_trial_expires_at_and_more.py
```diff
@@ -0,0 +1,114 @@
+# Generated by Django 5.2.16 on 2026-07-20 11:50
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+DROP_VIEW = "DROP VIEW IF EXISTS current_subscriptions_vw;"
+
+# The previous view definition (see migration 0009), kept so this migration is
+# reversible.
+OLD_CREATE_VIEW = """\
```
suricactus: Usually this is handled automatically by django migrations and the SQL is inlined in `migrations.RunSQL` functions (e.g. https://github.com/opengisch/QFieldCloud/blob/16cc49235a0f60237255ec0c23dcf3e6cbf89172/docker-app/qfieldcloud/core/migrations/0051_auto_20211125_0444.py#L79-L80 )? I would prefer if we keep the way we write the migrations the same - with inline SQL. Or it's the automated tooling that changed and introduced the queries on top?

Rakanhf: Not sure if i get this correctly, but I tried to maintain a similar implementation to what's already implemented in 
https://github.com/opengisch/QFieldCloud/blob/bedfa1486f6c81bda72b718dc9157175d86f8e8d/docker-app/qfieldcloud/subscription/migrations/0009_subscription_purchased_seats.py#L6-L14

suricactus: Yes, and this is wrong too and needs to be fixed.

No need for parasitic variables, we have an automated way to handle this, and it is `django makemigrations`. 

Please check my PR here https://github.com/opengisch/QFieldCloud/pull/1776 and my message here: https://chat.google.com/room/AAAABoO13iY/RxMQHgyNaio/RxMQHgyNaio?cls=10

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -644,6 +652,31 @@ class UpdateSubscriptionKwargs(TypedDict):
...
+            and self.trial_expires_at is not None
+            and self.trial_expires_at > timezone.now()
+        )
+
+    @property
+    def effective_plan(self) -> Plan:
+        if self.is_trialing and self.trial_plan_id is not None:
+            return self.trial_plan
+
+        return self.plan
```
```suggestion
    def plan(self) -> Plan:
        if self.is_trialing and self.trial_plan_id is not None:
            return self.trial_plan

        return self.regular_plan
```

Wouldn't this greatly reduce the number of changes required across the codebase if we just keep the currently active plan as `plan` instead of renaming it to `effective_plan`? 

If you insist on using `effective_plan`, which I am -0.1, I would suggest at least we rename `plan` to `regular_plan`.

Did we agree on this change somewhere, I didn't find any trace on CU, GH or chat, because the original description explicitly mentioned that `plan` is renamed to `regular_plan` and `plan` remains the currently active plan?

Rakanhf: > Wouldn't this greatly reduce the number of changes required across the codebase if we just keep the currently active plan as plan instead of renaming it to effective_plan?

No, since this is a property and the plan is already a field of the model

> If you insist on using effective_plan, which I am -0.1, I would suggest at least we rename plan to regular_plan.

When the `plan` field is changed to `regular_plan` that will still require refactoring a lot of `.plan` calls

I think we can keep `effective_plan` and `trial_plan` and change `plan` to `regular_plan` and completely ditch the concept of `plan` property, this way it will be easier to read and follow across codebase, as in effective_plan = maybe trial or regular. as of now, this is as simple as bulk replace and a migration

> Did we agree on this change somewhere, I didn't find any trace on CU, GH or chat

No, I guess I just lean toward current approach naturally, and missed this due to complexity.


Let me know what do you think about it.

suricactus: Ok, I believe you are more familiar with the details. I am surprised we need to distinguish between the old `plan` concept and new `effective_plan`. I guess if you inject a new commit as the first one in the branch that renames `plan` to `regular_plan` it will help following what is going on.

As for `effective_plan`... I find the name a bit strange, probably `active_plan`? No idea, can work with `effective_plan` for now, but if you feel insecure for the name, I am happy to find an alternative.

suricactus: Btw, I did a second check and it seems there are 11 occurances of `.plan` vs 16 of `.effective_plan`:

Image

Image

So I really think keeping the `plan` name for the what is now called `effective_plan` is less work, I would expect on `app.qfield.cloud` UI this to have even bigger effect.

Rakanhf: Well, thats a solid point, but you also need to factor in, other code not included in the search you attached, also some queries that can be easily missed during this refactor e.g `plan__`, etc. 

Now you see 11 vs 16 in terms of count, but factoring in other places, that number will be very small in comparison. 

So basically, if we changed `effective_plan` into `plan` property, and switched `plan` model field into `regular_plan`, we will need to change most of the original `.plan` to be `.regular_plan` since the business logic there requires the actual regular plan, not what is currently active.


I'm not against changing the naming of `effective_plan`

I'm more focused on trying to minimize the changes as much as possible for this refactor

suricactus: But most of the time, we use `current_subscription_vw`, which already uses `plan_id` to behave like `effective_plan` or I am missing something?

https://github.com/opengisch/QFieldCloud/blob/QF-8420_Add-trial-plan-as-a-field-of-the-current-subscription-instead-of-separate-subscription-record/docker-app/qfieldcloud/subscription/sql_config.py#L41-L45

I think already have the problem of using the same name for two different things:

If you load a subscription directly from the `Subscription` table, then the `plan` property is certainly the `plan_id` column.

If you load a subscription from `current_subscription_vw`, then the `plan` property may be either `plan_id` or `trial_plan_id`.

We definitely must change this duality by renaming the old `Subscription.plan` to `Subscription.regular_plan`, and then it will be obvious what we need to change afterwards.

Rakanhf: I might got stuck on the over simplifying the refactor and trying to keep everything similar to what it was in the past. 

I get your point towards the duality of output objects. 

So for short, now we can rename `plan` field to be `regular_plan` and refactor all queries , filters, etc, to match that

rename `effective_plan` property to be `plan`
ditch the case in the `current_subscription_vw ` entirely, since now `.plan` property will resolve it.


Just to clear out why I chose the current implementation, I mainly tried to keep everything as minimal as I can, thus didn't catch the duality issue, since for me changing quries field is something I see as high risk esspeissally when it comes to billing


Let me know what do you think about it

suricactus: > So for short, now we can rename plan field to be regular_plan and refactor all queries , filters, etc, to match that

yes, this is a must

> rename effective_plan property to be plan
> ditch the case in the current_subscription_vw  entirely, since now .plan property will resolve it.

I would keep the `current_subscription_vw` as is, and also add one more trick in the default `SubscriptionManager` - I would annotate the `Subscription` with the same `plan`. So something like this:

```
class SubscriptionManager(models.Manager):
    def get_queryset(self):
        return super().annotate(
			plan=Case(
				When(
					Q(trial_expires_at>=timezone.now()) & Q(trial__isnull=False),
					then=F("trial_plan"),
				),
				default=F("plan")
			)
		)
```

> Just to clear out why I chose the current implementation, I mainly tried to keep everything as minimal as I can, thus didn't catch the duality issue, since for me changing quries field is something I see as high risk esspeissally when it comes to billing

It's ok, you just should have raised a flag earlier that something is off with the proposal, so we can go through it. Anyways, I think what you have here looks solid at first glance, needs a few renamings and it should work as it works now.

An of course all the edge cases you collected on the way that were not working before either, but we will fix them in follow-ups.

Rakanhf: > I would keep the current_subscription_vw as is, and also add one more trick in the default SubscriptionManager - I would annotate the Subscription with the same plan. So something like this:

I'm not sure if this will be in the scope of this refactor, since if we replace the `plan` model field with `regular_plan` and add a `plan` property, I don't see any useful advantage of this  as of now for this refactor, 

Also, I don't see any reason to keep the view "case" since now we won't have a `plan_id` anymore and it will be just a property, the view should do the `CASE` in either `regular_plan` or `trial_plan`, and the property will already be enough in our use case here which will be applied to the `CurrentSubscription` model anyway.

For the annotation, This should probably on another PR, since it won't be useful for this refactor IMO. 

I will just go head and remove the `CASE` from `current_subscription_vw` and change the fields accordingly.

Let me know if there is something that I'm missing

suricactus: I am not in the details right now, probably the annotation with `plan` would be enough. What you will be able to do with the annotation/view having `plan_id` field is to filter by `plan` directly on the queryset, otherwise you need to either do it with a more complex query involving `regular_plan` and `trial_plan`, or within python. Secondly, it will save you from an extra query quering the plan, if you have not `select_related` both `regular_plan` and the `trial_plan`.

Again, I am not in the details right now, but I would expect filtering by `plan` is useful within the code, but I might be wrong and I leave it on your judgement.

duke-nyuki:
Task linked: [QF-8765 Refactor `plan` to `regular_plan` field on `subscription` model](https://app.clickup.com/t/2192114/QF-8765)

PR #1735, opened 2026-07-29
Add a new "template" project type

manylon:
Currently any QGIS project can be a template project on QFieldCloud. When using templates, there is the risk of someone using the "template" to collect actual data. This is problematic, because data will reside in the wrong project and can lead to potential data loss.
This PR introduces a new project `template` type for this reason.

- Add `Project.ProjectType.TEMPLATE`
- Keep `SHARED_DATASETS` reserved for the special `shared_datasets` project
- Block repackaging and applying deltas on template projects with a new `TemplateProjectNotAllowedError`, since templates are meant to be cloned from, not used as live field-data projects. File uploads are allowed.
- Move `project_type` test coverage into its own `test_project_type.py` file, and add tests covering types.


duke-nyuki:
Task linked: [QF-8182 Add a new "template" project type](https://app.clickup.com/t/2192114/QF-8182)

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -226,3 +226,13 @@ class NotCloneableProjectError(QFieldCloudException):
     code = "not_cloneable_project"
     message = "This project cannot be cloned."
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class TemplateProjectNotAllowedError(QFieldCloudException):
```
suricactus:
```suggestion
class OperationNotAllowedForTemplateProjectError(QFieldCloudException):
```

The name does not really match the description.

docker-app/qfieldcloud/core/exceptions.py
```diff
@@ -226,3 +226,13 @@ class NotCloneableProjectError(QFieldCloudException):
...
+    """Raised when attempting to package or apply deltas on a template project.
+
+    Template projects can only be cloned or have files uploaded to them."""
+
+    code = "template_project_not_allowed"
```
suricactus: Code needs fixing based on what we settle on as class name.

docker-app/qfieldcloud/project/models.py
```diff
@@ -592,6 +578,13 @@ def is_shared_datasets_project(self) -> bool:
         """
         return self.project_type == self.ProjectType.SHARED_DATASETS
 
+    @cached_property
+    def is_template_project(self) -> bool:
```
suricactus, line 610: Shall we add `is_regular_project` for the sake of completeness.

docker-app/qfieldcloud/project/models.py
```diff
@@ -592,6 +578,13 @@ def is_shared_datasets_project(self) -> bool:
...
+        """
+        Returns `True` if the project is a template project, otherwise `False`.
```
```suggestion
        Returns `True` if the project type is a template project, otherwise `False`.
```

docker-app/qfieldcloud/project/models.py
```diff
@@ -720,7 +713,7 @@ def has_online_vector_data(self) -> bool | None:
 
     @property
     def can_repackage(self) -> bool:
-        return True
+        return not self.is_template_project
```
suricactus: Since this is very, very sensitive part, can we be explicit here:

```
    @property
    def can_repackage(self) -> bool:
        if self.is_template_project:
            return False
        else:
            return True
```

docker-app/qfieldcloud/project/tests/test_project.py
```diff
@@ -1518,41 +1518,52 @@ def test_clone_project_invalid_requests(self):
...
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["code"], "not_cloneable_project")
+        self.assertFalse(Project.objects.filter(name="cloned_shared_datasets").exists())
 
-    def test_project_type_updates_when_renamed_away_from_shared_datasets(self):
-        """Test that renaming the shared_datasets project away reverts project_type to default value."""
-        project = Project.objects.create(
-            name=SHARED_DATASETS_PROJECT_NAME, owner=self.user1
+    def test_clone_of_template_project_is_allowed(self):
```
suricactus: In theory `test_clone_of_shared_datasets_project_is_not_allowed` and `test_clone_of_template_project_is_allowed` can be parametrized with:
```
@params-or-whatever-it-is(
  (ProjectType.REGULAR, True),
  (ProjectType.TEMPLATE, True),
  (ProjectType.SHARED_DATASETS, False),
)
def test_cloning_by_project_type(self, project_type, should_succeed):
  # ...
```

docker-app/qfieldcloud/project/tests/test_project_type.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def test_project_type_cannot_be_forced_to_shared_datasets(self):
+        """Test that a normally named project cannot be saved as of type `SHARED_DATASETS`"""
+        project = Project.objects.create(
+            name="project",
+            owner=self.user1,
+            project_type=Project.ProjectType.SHARED_DATASETS,
+        )
+
+        self.assertEqual(project.project_type, Project.ProjectType.REGULAR)
```
suricactus: This should raise rather than silently fallback to REGULAR. This may simplify the implementation that is tested in `test_api_cannot_set_project_type_to_shared_datasets`.

docker-app/qfieldcloud/project/tests/test_project_type.py
```diff
@@ -0,0 +1,230 @@
...
+            {"project_type": Project.ProjectType.SHARED_DATASETS},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["code"], "api_error")
+
+        project.refresh_from_db()
+        self.assertEqual(project.project_type, Project.ProjectType.REGULAR)
+
+    def test_api_cannot_change_project_type_of_shared_datasets_project(self):
```
suricactus: all tests above from `test_project_type_can_be_changed_from_regular_to_template` inclusive should be parametrized:
```
(
  (TEMPLATE, REGULAR, True),
  (REGULAR, SHARED_DATASETS, False),
)
def test_api_project_type_conversion(self, from_type, to_type, should_succeed)
```

docker-app/qfieldcloud/project/tests/test_project_type.py
```diff
@@ -0,0 +1,230 @@
...
+        )
+        self.assertTrue(status.is_success(response.status_code))
+
+        project.refresh_from_db()
+        self.assertTrue(project.has_the_qgis_file)
+
+        with self.assertRaises(exceptions.TemplateProjectNotAllowedError):
+            jobs.repackage(project, self.user1)
+
+    def test_file_upload_allowed_on_template_project(self):
```
suricactus: This is redundant test IMO. Template projects only have funny behaviour with QField operations.

suricactus (commented):
Round 1

docker-app/qfieldcloud/project/tests/test_project.py
manylon: I will use `self.subTest`, which is similar, just not used as a decorator, and will not introduce a new library. Also there are other tests that use it, and I will continue with the same style and approach.

nirvn:
@manylon , how are these types exposed in the json returned by /api/v1/projects/ ?

manylon:
> @manylon , how are these types exposed in the json returned by /api/v1/projects/ ?

Hi @nirvn, It is exposed as `project_type` with an integer matching the full `ProjectType` enum, any of these can appear in a JSON response. https://github.com/opengisch/QFieldCloud/blob/31fa6de019ad306f9d4b16a60211234455022e9d/docker-app/qfieldcloud/project/models.py#L192-L195 



However, only `CONFIGURABLE_PROJECT_TYPES` are accepted as input when creating/updating a project. `SHARED_DATASETS` is rejected/ignored if a client tries to set it.
https://github.com/opengisch/QFieldCloud/blob/31fa6de019ad306f9d4b16a60211234455022e9d/docker-app/qfieldcloud/project/models.py#L197-L201

docker-app/qfieldcloud/project/tests/test_project.py
manylon: Marking as resolved since I have used `self.subTest` to address the comment.

suricactus: Ah, I got used to pytest sugar :) One of the topics to be addressed regarding testing on a good day

nirvn:
For other values, such as useRole, we use a string representation. Would it make sense to keep that for project type too? I don't have a strong opinion, just curious about the thought process :)

Btw you've been rocking the pull request list, thanks ;)

manylon:
> For other values, such as useRole, we use a string representation. Would it make sense to keep that for project type too? I don't have a strong opinion, just curious about the thought process :)

I don't have a strong opinion too. Made `project_type` in string representation, that can accept case-insensitive values, and kept the integer value for storing in the database.

> 
> Btw you've been rocking the pull request list, thanks ;)

Thank you, happy to be a part of it!

suricactus:
I do have strong preference we have a string representation in the API. Please do use the lowercase representation that matches the name of the enum value. Actually we already discussed this via chat with @manylon and should be part of the PR.

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -15,8 +15,38 @@
 from qfieldcloud.subscription.exceptions import QuotaError
 
 
+def normalize_project_type_label(label: object) -> str:
+    return str(label).lower().replace(" ", "_")
```
suricactus: You don't need this. Your client is responsible to send you a sensible string. Less choice, less problems.

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -15,8 +15,38 @@
...
+        # `object` may be any `Project.ProjectType` value (e.g. `SHARED_DATASETS`),
+        # not only the configurable ones this field accepts as input.
+        return normalize_project_type_label(Project.ProjectType(object).label)
+
+    def to_internal_value(self, data):
+        for i in self._choices:
+            if normalize_project_type_label(
+                self._choices[i]
+            ) == normalize_project_type_label(data):
+                return i
```
suricactus:
```suggestion
        for choice_key in self._choices:
            ifchoice_key.lower() == data:
                return choice_key
```

Avoid using `i` or `k` or `j` variables, unless extremely clear they are a numeric index of an iterator. In general avoid single letter variables, unless extremely clear what they mean, e.g. `[str(v) for v in values]` is acceptable. 

Not sure if my proposal above is syntatically sound, please adapt it as required.

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -15,8 +15,38 @@
...
+    return str(label).lower().replace(" ", "_")
+
+
+class ProjectTypeField(serializers.ChoiceField):
+    def to_representation(self, object):
+        # `object` may be any `Project.ProjectType` value (e.g. `SHARED_DATASETS`),
+        # not only the configurable ones this field accepts as input.
+        return normalize_project_type_label(Project.ProjectType(object).label)
+
+    def to_internal_value(self, data):
```
suricactus: Typing

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -15,8 +15,38 @@
 from qfieldcloud.subscription.exceptions import QuotaError
 
 
+def normalize_project_type_label(label: object) -> str:
+    return str(label).lower().replace(" ", "_")
+
+
+class ProjectTypeField(serializers.ChoiceField):
+    def to_representation(self, object):
```
suricactus: typing

docker-app/qfieldcloud/project/serializers.py
manylon: I prefer having it, otherwise I will duplicate the same logic in 3 places in `ProjectTypeField`. I use it to return the label in a correct format to the client and to compare what the client has sent (will remove the function from the client input, this way we will expect the client to send an exact match).

docker-app/qfieldcloud/project/models.py
```diff
@@ -210,9 +223,20 @@ class PackagingOffliner(models.TextChoices):
...
+
+    # Project types configurable from clients
+    CONFIGURABLE_PROJECT_TYPES = [
+        ProjectType.REGULAR,
+        ProjectType.TEMPLATE,
+    ]
+
+    CONFIGURABLE_PROJECT_TYPE_CHOICES = get_choices_by_values(
+        ProjectType.choices, CONFIGURABLE_PROJECT_TYPES
+    )
```
suricactus: This variable is parasitic, as well as the function. Yes, it potentially saves a few CPU cycles per request, but I believe Python is fast enough so we better optimize for less dev confusion which variable to use. You can easily recreate such list every time by `[c.value, c.label for c in Project.CONFIGURABLE_PROJECT_TYPES]`. So please just inline the expression and drop `CONFIGURABLE_PROJECT_TYPE_CHOICES` and `get_choices_by_values`.

docker-app/qfieldcloud/project/models.py
```diff
@@ -1035,10 +1076,20 @@ def save(self, recompute_storage=False, *args, **kwargs):
...
 
-        self.project_type = self.ProjectType.REGULAR
-
         if self.name == SHARED_DATASETS_PROJECT_NAME:
             self.project_type = self.ProjectType.SHARED_DATASETS
+        elif self.project_type == self.ProjectType.SHARED_DATASETS:
+            raise ValidationError(
+                _(
+                    "`project_type` cannot be set to `SHARED_DATASETS` unless the project is named `{}`."
+                ).format(SHARED_DATASETS_PROJECT_NAME)
```
```suggestion
                _(
                    "`project_type` cannot be set to `{}` unless the project is named `{}`."
                ).format(ProjectType.SHARED_DATASETS, SHARED_DATASETS_PROJECT_NAME)
```
Do not hardcode enum values.

docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -395,6 +395,9 @@ def can_delete_unnecessary_file_versions(user: QfcUser, project: Project) -> boo
 
 def can_create_deltas(user: QfcUser, project: Project) -> bool:
     """Whether the user can store deltas in a project."""
+    if project.is_template_project:
```
suricactus, line 398: I think you better add `project_type` to the `Project.objects.slim()` query here, because now you add an extra query every time these permissions are checked.

https://github.com/opengisch/QFieldCloud/blob/QF-8182-template-project-type/docker-app/qfieldcloud/project/models.py#L127-L131

suricactus (changes requested):
Currently the permission checks cause N+1 queries, this is not good!

docker-app/qfieldcloud/project/models.py
manylon: I can remove `get_choices_by_values` and use list comprehension, but `CONFIGURABLE_PROJECT_TYPE_CHOICES` is used in more than one place. I can add notes in the code?

suricactus: As of this PR, they are used at exactly one place besides definition. And it is totally fine.

Curious what the "notes in the code" would say?

IMO it's obvious we will have a single source of truth - `CONFIGURABLE_PROJECT_TYPES`, if anyone needs to use it in the future, the can use that list. So any notes would be noise IMHO.

suricactus (approved):
Yay!

manylon:
Thanks for the review @suricactus, will clean history and merge.

PR #1737, opened 2026-07-30
feat: add support for mounting custom instance-specific Nginx configurations

Rakanhf:
- Added an include directive for `instance.d/*.conf` in the Nginx template this allows custom instances to cleanly inject their own Nginx server configurations or location blocks via Docker Compose volumes without shadowing or conflicting with the base `config.d` directory

### DIFF
--- a/docker-nginx/templates/default.conf.template
+++ b/docker-nginx/templates/default.conf.template
@@ -89,6 +89,9 @@ server {
   # include additional config from the current QFieldCloud instance. This is different from the default `conf.d` directory.
   include config.d/*.conf;
 
+  # Include optional instance-specific configurations
+  include instance.d/*.conf;
+
   # include proxy settings and location blocks for the QFieldCloud instance.
   include conf.d/includes/qfieldcloud.conf;
 }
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8609 Fix FirstPromoter tracking referrals server-side (fix ad-blocker tracking)](https://app.clickup.com/t/2192114/QF-8609)

docker-nginx/templates/default.conf.template
```diff
@@ -89,6 +89,9 @@ server {
   # include additional config from the current QFieldCloud instance. This is different from the default `conf.d` directory.
   include config.d/*.conf;
 
+  # Include optional instance-specific configurations
+  include instance.d/*.conf;
```
gounux, line 93: Could custom instance-specific configurations also be added in the `config.d` dir above ?  
Or should one better use this `instance.d` dedicated directory ?

Rakanhf: Since the primary goal is to inject configuration files without removing the read-only property from the `config.d` directory.

suricactus: With @Rakanhf we found a woraround to keep the `conf.d` read-only flag, but we need to create a dummy file in `conf.d/tobeoverwritten.conf` so it can be overwritten by app.qfield.cloud docker compose. We will discuss this tomorrow during the dev meeting.

suricactus (approved):
Thanks!

gounux (approved):
👌

PR #1738, opened 2026-07-30
chore: pin hadolint to v2.14.0 in pre-commit configuration to avoid breaking changes

Rakanhf:
Pinning hadolint to `v2.14.0` to avoid breaking changes, since `v2.15.0` introduces issues on github actions


suricactus (approved):
I will merge like this, but the actual problem should be addressed.

PR #1739, opened 2026-07-30
Bump hadolint from 2.14.0 to 2.15.0 and explicitly set gid & uid for the container user

suricactus:
Fixing failures like this one here: https://github.com/opengisch/QFieldCloud/actions/runs/30550620846/job/90938002257?pr=1735

Continuation of https://github.com/opengisch/QFieldCloud/pull/1738

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,11 +58,8 @@ repos:
         entry: bash -c 'mypy -p docker-app -p docker-qgis "$@"' --
 
   - repo: https://github.com/hadolint/hadolint
-    rev: v2.14.0
+    rev: v2.15.0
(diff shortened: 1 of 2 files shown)


Rakanhf (approved):
LGTM ! Thank you 

One small note before merging, since in the commit message there is a typo "3.15" instead of 2.15 

and PR title should be related to fixing the perms rather than bumping

suricactus:
@Rakanhf thanks for the typo hint, feel free to edit PR titles with typos directly.

Regardning bump vs fix, I think it's a bump as the fix was provided by you yesterday when you fixed the docker image version used, and I just moved to a newer version and applied the required changes.

PR #1740, opened 2026-07-31
Provide a nicer view name for /api/v1/deltas/{project_uuid}/ and /api/v1/members/{organization_name}/

nirvn:
Spot the difference:

| api/v1/projects/ | api/v1/files/{project_uuid} | api/v1/deltas/{project_uuid} |
| --- | --- | --- |
| image | image | image |

### DIFF
--- a/docker-app/qfieldcloud/core/views/deltas_views.py
+++ b/docker-app/qfieldcloud/core/views/deltas_views.py
@@ -196,6 +196,9 @@ class ListCreateDeltasView(generics.ListCreateAPIView):
 
         return faulty_deltafile
 
+    def get_view_name(self):
+        return _("Delta List")
+
     def get_queryset(self):
         project_id = self.request.parser_context["kwargs"]["projectid"]
         project_obj = Project.objects.get(id=project_id)
--- a/docker-app/qfieldcloud/core/views/members_views.py
+++ b/docker-app/qfieldcloud/core/views/members_views.py
@@ -1,5 +1,6 @@
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
+from django.utils.translation import gettext as _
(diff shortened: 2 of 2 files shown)


docker-app/qfieldcloud/core/views/deltas_views.py
```diff
@@ -196,6 +196,9 @@ def preserve_faulty_deltafile(
 
         return faulty_deltafile
 
+    def get_view_name(self):
+        return "Delta List"
```
suricactus:
```suggestion
        return _("Delta List")
```

Not sure if gettext imported, but this I believe should be translated.

docker-app/qfieldcloud/core/views/members_views.py
```diff
@@ -40,6 +40,9 @@ class ListCreateMembersView(generics.ListCreateAPIView):
     serializer_class = OrganizationMemberSerializer
     pagination_class = pagination.QfcLimitOffsetPagination()
 
+    def get_view_name(self):
+        return "Member List"
```
suricactus:
```suggestion
        return _("Member List")
```

Not sure if gettext imported, but this I believe should be translated.

suricactus (commented):
Tiny translation adjustment.

nirvn:
@suricactus , review addressed.

gounux (approved):
Thanks @nirvn, nice additional DRF metadata :)

Though I face challenges spotting the diff(s) in the provided PR description screenshots 🤔, I believe it's the view name next to `Api root` ?

nirvn:
@gounux , correct. Instead of using the View class name, we use a title that harmonizes with other /api/v1/ endpoints. For the /deltas/ one, it also prevents a confusion as to whether the list only includes method="create" deltas  :)

gounux:
@nirvn 👌

PR #1744, opened 2026-08-03
Change `Project.project_type` enum from `IntegerChoices` to `TextChoices`

manylon:
`project_type` should be readable and writable as a plain string ("regular", "shared_datasets") over the API, both in JSON responses and on POST/PATCH requests, instead of an internal integer code. 

We're fine storing the string values directly in the database for now rather than mapping to/from integers.


duke-nyuki:
Task linked: [QF-8629 Change `Project.project_type` enum from `IntegerChoices` to `TextChoices`](https://app.clickup.com/t/2192114/QF-8629)

suricactus (approved):
Thanks!

Just to confirm, you did a cycle migrate up -> down -> up -> down ?

manylon:
Thanks for the review @suricactus and yes, I did!

PR #1745, opened 2026-08-03
Link the project's QGIS file to its admin change page

manylon:
Replace the plain-text `the_qgis_file_name` readonly field on `ProjectAdmin` with `the_qgis_file_link`, which renders a clickable link to the `File` object's admin change page, matching how other foreign keys like `owner` are linked.

Before:
image

After:
image

### DIFF
--- a/docker-app/qfieldcloud/core/admin.py
+++ b/docker-app/qfieldcloud/core/admin.py
@@ -1405,7 +1405,7 @@ class ProjectAdmin(QFieldCloudModelAdmin):
         "owner",
         "status",
         "status_code",
-        "the_qgis_file_name",
+        "the_qgis_file",
         "overwrite_conflicts",
         "has_restricted_projectfiles",
         "file_storage_bytes",
@@ -1440,7 +1440,7 @@ class ProjectAdmin(QFieldCloudModelAdmin):
         "project_details__pre",
         "locked_at",
         "file_storage_migrated_at",
-        "the_qgis_file_name",
+        "the_qgis_file",
     )
     inlines = (
         ProjectSeedInline,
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8599 Consider converting `Project.the_qgis_file` admin field to the default form field](https://app.clickup.com/t/2192114/QF-8599)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -1405,7 +1405,7 @@ class ProjectAdmin(QFieldCloudModelAdmin):
         "owner",
         "status",
         "status_code",
-        "the_qgis_file_name",
+        "the_qgis_file_link",
```
```suggestion
        "the_qgis_file",
```

What if you drop the `_link` suffix, update the `read_only_fields` with the same name and drop the `the_qgis_file_name` method alltogether?

suricactus (changes requested):
Remove the custom method and leave django do it's magic. Also missing before/after images.

docker-app/qfieldcloud/core/admin.py
manylon: If I drop the method we get:
image

If i only drop the `_link` suffix. Django will overlook the custom display and method, because it is a FK field and will still show the same.

If this is accepted, they will be removed.

suricactus: Yes, this is fine. This will motivate us to have better representation of `File` model.

manylon: Okay, doing it!

suricactus (approved):
Çok thank you!

(update your screenshot)

manylon:
Thanks for the review @suricactus, will clean history and merge!

PR #1747, opened 2026-08-03
feat: load the debug toolbar only if we are in debug mode

suricactus:
Pretty much the title.

### DIFF
--- a/docker-app/qfieldcloud/urls.py
+++ b/docker-app/qfieldcloud/urls.py
@@ -103,7 +103,13 @@ urlpatterns = [
     path("accounts/password/change/", blocked_view),
     path("accounts/", include("allauth.urls")),
     path("invitations/", include("invitations.urls", namespace="invitations")),
-    path("__debug__/", include("debug_toolbar.urls")),
     path("captcha/", include("captcha.urls")),
     path("a/<str:username>/<str:project_name>/", redirect_to_admin_project_view),
 ]
+
+if settings.DEBUG:
+    from debug_toolbar.toolbar import debug_toolbar_urls
+
+    urlpatterns = [
+        *urlpatterns,
+    ] + debug_toolbar_urls()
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8195 Load the django-debug-toolbar conditionally only in Debug mode](https://app.clickup.com/t/2192114/QF-8195)

gounux:
CI seems to fail & throw:

```text
ERRORS:
?: (debug_toolbar.E001) The Django Debug Toolbar can't be used with tests
	HINT: Django changes the DEBUG setting to False when running tests. By default the Django Debug Toolbar is installed because DEBUG is set to True. For most cases, you need to avoid installing the toolbar when running tests. If you feel this check is in error, you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to bypass this check.
```

Rakanhf (approved):
LGTM ! 🚀 

Thank you

docker-app/qfieldcloud/urls.py
```diff
@@ -103,7 +103,13 @@
...
     path("captcha/", include("captcha.urls")),
     path("a/<str:username>/<str:project_name>/", redirect_to_admin_project_view),
 ]
+
+if "debug_toolbar" in settings.INSTALLED_APPS:
+    urlpatterns = [
+        *urlpatterns[:-1],
+        path("__debug__/", include("debug_toolbar.urls")),
+        *urlpatterns[-1:],
+    ]
```
manylon, line 115:
```suggestion
    urlpatterns.append(path("__debug__/", include("debug_toolbar.urls")))
```
Why not?

suricactus: Becauuse the very last pattern must remain the one it already is, otherwise the paths in app.qfield.cloud are broken.

PR #1752, opened 2026-08-07
Add option to exclude existing team members from the users API

manylon:
This PR adds a way for the users API to exclude members of a given team from its results.

When adding users to a team, we should exclude the users that are already team members to prevent toasts of sort `User "X" cannot be added to the team. Team member with this Team and Member already exists.`.

Changes:
- Add `exclude_members_of_team` to `querysets_utils.get_users`
- Add `exclude_members_of_team` parsing in `ListCreateUsersView.get_queryset` and pass it to `get_users`
- Add coverage in `test_queryset.py` for the plain team-exclusion case and the combined organization, invert, and team case

### DIFF
--- a/docker-app/qfieldcloud/core/querysets_utils.py
+++ b/docker-app/qfieldcloud/core/querysets_utils.py
@@ -39,6 +39,7 @@ def get_users(
     username: str,
     project: Project | None = None,
     organization: Organization | None = None,
+    exclude_members_of_team: Team | None = None,
     exclude_organizations: bool = False,
     exclude_teams: bool = False,
     invert: bool = False,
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-7277 Add more than one organization member at once](https://app.clickup.com/t/2192114/QF-7277)

suricactus:
Why is this?

Not sure if it is because of adding team members, but just to say perfectly fine to have:

- organization `org1` owner by `u1`
- team `@org1/team1` as part of `org1`
- users `u2` and `u3` as members of `org1`
- users `u2` and `u3` as members of `@org1/team1`
- project `p1` owned by `org1`
- have `@org1/team1` as collaborator on `p1` with role `READER`
- have `u2` as collaborator on `p1` with role `ADMIN`

manylon:
I need it for the team members, not for collaborators

suricactus (approved):
Now I see, you need it for the team members, not for collaborators. LGTM!

manylon:
Thanks!

PR #1755, opened 2026-08-10
Bump all indirect dependencies

suricactus:
We are bumping the direct dependencies, but the indirect ones are upgraded in a very conservative way.


manylon (approved):
LGTM

PR #1756, opened 2026-08-10
Add a new CREATOR organization role that can create organization projects

manylon:
Organizations previously could only let `ADMIN` members create new projects on the organization's behalf. This PR adds the `CREATOR` role, which grants the permission to create projects.

To let creators manage what they create, added to `Project` a new `created_by` field, and a new `post_save` signal (`project.signals.grant_creator_organization_project_admin_access`) that grants the creator explicit `ADMIN` collaborator access on every project they create on behalf of the  organization.

Also shows `created_by` in the Django admin, both on `ProjectAdmin` and the `ProjectInline` on organization pages.

Organization project inline:
image
Project page:
image


duke-nyuki:
Task linked: [QF-8572 Add a new organization role that can create organization projects](https://app.clickup.com/t/2192114/QF-8572)

docker-app/qfieldcloud/project/views.py
```diff
@@ -179,7 +179,8 @@ def get_queryset(self):
 
     @transaction.atomic
     def perform_create(self, serializer: ProjectSerializer) -> None:
-        super().perform_create(serializer)
+        # The `created_by` field is set to the user that made the request.
+        serializer.save(created_by=self.request.user)
```
suricactus, line 183: Why no more calling the super method?

docker-app/qfieldcloud/project/models.py
```diff
@@ -317,6 +317,17 @@ class Meta:
             "The project owner can be either you or any of the organization you are member of."
         ),
     )
+
+    created_by = models.ForeignKey(
+        # NOTE should be Person, but Django sometimes has troubles with Person/User (e.g. Form.full_clean()), see #514 #515
```
```suggestion
        # NOTE should be `Person`, but Django sometimes has troubles with `Person`/`User` (e.g. `Form.full_clean()`), see #514 #515
```

Curious if you experienced the same problem?

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,53 @@
...
+
+    NOTE: Any future "can create projects" role gets this automatically,
+    with no changes needed here, unless it also gains implicit access via
+    `projects_with_roles_vw`.
+    """
+    if not created or instance.created_by_id is None:
+        return
+
+    if not instance.owner.is_organization:
+        return
```
suricactus:
```suggestion
    # We want to add `ADMIN` collaborator only on creation
    if not created:
        return
    
    # We use the `created_by` to determine the `ADMIN` collaborator to add, if missed we cannot do much besides logging
    if instance.created_by is None:
        logger.warning('Project "%s" created but `created_by` field is empty!')

        return

    # For now we want to add an `ADMIN` collaborator only if the project owner is an organization
    if not instance.owner.is_organization:
        return
```

Be more explicit what we are checking.

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,53 @@
...
+
+    organization = Organization.objects.get(pk=instance.owner_id)
+
+    if instance.created_by_id == organization.organization_owner_id:
+        return
```
suricactus:
```suggestion
    # Skip adding `ADMIN` collaborator if the user is the organization owner, because ... 
    if instance.created_by_id == organization.organization_owner_id:
        return
```

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if OrganizationMember.objects.filter(
+        organization=organization,
+        member_id=instance.created_by_id,
+        role=OrganizationMember.Roles.ADMIN,
+    ).exists():
+        return
```
suricactus:
```suggestion
    # Just in case check if the user already exists as a collaborator, this should never be the case
    if OrganizationMember.objects.filter(
        organization=organization,
        member_id=instance.created_by_id,
    ).exists():
        return
```

If you are adding such check, explain why. I think if you keep `role=ADMIN` filter, you may get into a situation when you attempt to add a user that is already a collaborator, which is againt a unique key in the collaborators table. Is this something you explicitly want? 

Why don't you use create or update instead?

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,53 @@
...
+    on the new project.
+
+    Organization owners and `ADMIN` members already get implicit
+    project-admin access via `projects_with_roles_vw`. Other roles allowed
+    to create a project for the organization (today: `CREATOR`) don't have
+    that implicit access, and need this explicit grant instead.
+
+    NOTE: Any future "can create projects" role gets this automatically,
+    with no changes needed here, unless it also gains implicit access via
+    `projects_with_roles_vw`.
```
suricactus: I don't follow this comment? Or actually I follow, but not sure what you mean by "implicit access", IMO it is quite explicit. I would just drop that phrase.

docker-app/qfieldcloud/project/views.py
manylon: Because `perform_create` calls `serializer.save()` in `CreateModelMixin`

suricactus: ok

docker-app/qfieldcloud/project/models.py
manylon: I didn't hit this myself and didn't tested `created_by` as `Person`. Copied the pattern from the existing `User` FKs in the codebase, so I kept `created_by` consistent with those rather than risk hitting the same issue here.

docker-app/qfieldcloud/project/signals.py
manylon: The `role=ADMIN` check skips the grant when the creator already has admin access via `projects_with_roles_vw`, e.g., organization owners and `ADMIN` members. For the unique-key concern, I will use `get_or_create` not `create` below.

suricactus: IMO it's better if we just throw if the collaborator already exists, this is unexpected. It's always better to have explicit errors than implicit.

docker-app/qfieldcloud/project/models.py
suricactus: Let's keep it for now, the polymorphic nature of `User` is going to die soon.

docker-app/qfieldcloud/project/signals.py
manylon: Then will leave the `create` and raise.

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        ProjectCollaborator.objects.create(
+            project=instance,
+            collaborator=instance.created_by,
+            role=ProjectCollaborator.Roles.ADMIN,
+            created_by=instance.created_by,
+            updated_by=instance.created_by,
+        )
+    except IntegrityError as error:
+        raise UnexpectedProjectCollaboratorError() from error
```
suricactus:
```suggestion
    # Skip adding `ADMIN` collaborator if the user already has `ADMIN` access through the organization.
    if OrganizationMember.objects.filter(
        organization=organization,
        member=instance.created_by,
        role=OrganizationMember.Roles.ADMIN,
    ).exists():
        return

    # Since it is a completely new project, the current user should not be already a collaborator. If they are a collaborator, we should raise and cancel the operation.
    if ProjectCollaborator.objects.filter(
        project=instance,
        collaborator=instance.created_by,
    ).exists():
        raise UnexpectedProjectCollaboratorError()

    # Finally add the user as an `ADMIN` collaborator to the project.
    ProjectCollaborator.objects.create(
        project=instance,
        collaborator=instance.created_by,
        role=ProjectCollaborator.Roles.ADMIN,
        created_by=instance.created_by,
        updated_by=instance.created_by,
    )
```

Two things:

1) since we recently had a problem with using the `_id` of FK columns, let's not try to outsmart Django. We can use the regular FK fields, not their `_id` twins.
2) We should explicitly check for the scenario when the collaborator already exists. If we hope on the `IntegrityError`, it might be thrown for other reasons (e.g. another `UNIQUE` index is not satisfied in the same table.

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+            'Project "%s" created but `created_by` field is empty!', instance.name
+        )
+
+        return
+
+    # For now we want to add an `ADMIN` collaborator only if the project owner is an organization
+    if not instance.owner.is_organization:
+        return
+
+    organization = Organization.objects.get(pk=instance.owner_id)
```
suricactus:
```suggestion
    organization = Organization.objects.get(owner=instance.owner)
```

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Skip adding `ADMIN` collaborator if the user is the organization owner, because
+    # they already have `ADMIN` access through the organization.
+    if instance.created_by_id == organization.organization_owner_id:
```
suricactus:
```suggestion
    if instance.created_by == organization.organization_owner:
```

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger(__name__)
+
+
+@receiver(post_save, sender=Project)
+def grant_creator_organization_project_admin_access(
+    sender, instance, created, **kwargs
+):
+    """Whenever an organization-owned project is created by someone without
+    project-admin access, grant that person explicit `ADMIN` access
+    on the new project.
```
suricactus:
```suggestion
    """Whenever an organization-owned project is created by someone without
    project-admin access, give them explicit `ADMIN` collaborator role.
```

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Organization owners and `ADMIN` members already get project-admin access
+    via `projects_with_roles_vw`. Other roles allowed to create a project
+    for the organization, by default don't get that access,
+    and need this grant.
```
suricactus:
```suggestion
    Organization owners and `ADMIN` organization members already get project-admin access
    via `projects_with_roles_vw` and don't need to be added as collaborator.
```

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+
+    NOTE: Any future "can create projects" role gets this automatically,
+    with no changes needed here. Currently `CREATOR` is the only role
+    that can create projects for an organization, and it doesn't have
+    project-admin access by default.
```
suricactus:
```suggestion
```

We should add this as an assert for now.

docker-app/qfieldcloud/project/signals.py
```diff
@@ -0,0 +1,78 @@
...
+    if instance.created_by_id == organization.organization_owner_id:
+        return
+
+    # Skip adding `ADMIN` collaborator if the user already has `ADMIN` access through the organization.
+    if OrganizationMember.objects.filter(
+        organization=organization,
+        member_id=instance.created_by_id,
+        role=OrganizationMember.Roles.ADMIN,
+    ).exists():
+        return
```
suricactus:
```suggestion
    try:
        member_role = OrganizationMember.objects.filter(
            organization=organization,
            member_id=instance.created_by_id,
        ).value("role", flat=True).first()
    except OrganizationMember.DoesNotExist:
        member_role = None

    # When the user already has `ADMIN` organization membership, no need to add them as collaborator.
    if member_role == OrganizationMember.Roles.ADMIN:
        return

    # For now we only support creating projects by `ADMIN` and `CREATOR` organization members
    if member_role != OrganizationMember.Roles.CREATOR:
        raise NotImplementedError("Only roles `ADMIN` and `CREATOR` organization members can create projects within an organization, but {member_role} role given!")
```

suricactus (commented):
Final round I believe.

suricactus (approved):
LGTM

manylon:
Thanks, will clean the history and merge

PR #1757, opened 2026-08-10
Better object representation for File, FileVersion and Project

suricactus:
Before:
```
File(<file-id) in "<project-id>" "<filepath>"

FileVersion(<file-version-id) in "<project-id>" "<filepath>" "<version-display>"

<name> (<project-id>) owner:<owner-username>

```

After:
```
<project-id>/<filepath> [id=<file-id>]

<project-id>/<filepath>/<version-display> [id=<file-version-id>]

Project <owner-username>/<name> [id=<project-id>]

```

## Motivation

Before:
image

After:

image

### DIFF
--- a/docker-app/qfieldcloud/filestorage/models.py
+++ b/docker-app/qfieldcloud/filestorage/models.py
@@ -156,7 +156,7 @@ class File(models.Model):
         return self.versions.aggregate(size=Sum("size", default=0))["size"]
 
     def __repr__(self) -> str:
-        return f'File({self.pk}) in "{self.project_id}" "{self.name}"'  # type: ignore
+        return f"{self.project_id}/{self.name} [id={self.id}]"  # type: ignore
 
     def __str__(self) -> str:
         return self.__repr__()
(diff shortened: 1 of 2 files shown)


docker-app/qfieldcloud/filestorage/models.py
```diff
@@ -156,7 +156,7 @@ def get_total_versions_size(self) -> int:
         return self.versions.aggregate(size=Sum("size", default=0))["size"]
 
     def __repr__(self) -> str:
-        return f'File({self.pk}) in "{self.project_id}" "{self.name}"'  # type: ignore
+        return f"{self.project_id}/{self.name} [id={self.id}]"  # type: ignore
```
manylon, line 159:
```suggestion
        return f"{self.project_id} / {self.name} [id={self.id}]"  # type: ignore
```
I think it's easier to read with a little space?

manylon (approved):

docker-app/qfieldcloud/filestorage/models.py
suricactus: It's intended, it matches more or less the file structure on the object storage. And `self.name` does not have whitespaces around the slashes.

PR #1758, opened 2026-08-10
fix: fetch only the thumbnail from the database, instead of the whole `Project` row

suricactus:
This should greatly improve performance when loading the `Project` thumbnail.


manylon (approved):
LGTM

PR #1761, opened 2026-08-12
Move the `Project.project_details`  fields data into `QgisLayerField` model

manylon:
Currently the `Project.project_details`  stores QGIS information about the current project. What is worse, if we add fields  as list of field definitions on each vector layer, we increase the size of `project_details`  field 10-fold.

- Add `QgisLayerField` model for storing fields metadata from `QgisLayer` model
- Sync fields per layer via `QgisLayerFieldQuerySet.update_from_details()`, wired into layer creation/update
- Add migration and read-only Django admin for the new model
- Remove per-field data in `Project.project_details` after insertion in `QgisLayerField`
-  Keep `Project.prokect_details` in sync for now

UI:
In `QGIS projects inline`:
image
QGIS layer fields list:
image


duke-nyuki:
Task linked: [QF-8606 Move the `Project.project_details`  fields data into `QgisLayerFields` model](https://app.clickup.com/t/2192114/QF-8606)

docker-app/qfieldcloud/core/admin.py
```diff
@@ -692,6 +693,32 @@ def qgis_project_layers_list(qgis_project: "QgisProject") -> SafeText:
     )
 
 
+def qgis_layer_fields_list(layer: "QgisLayer") -> SafeText:
```
```suggestion
def render_qgis_layer_fields_table(layer: "QgisLayer") -> SafeText:
```

Same goes for `qgis_project_layers_list`.

docker-app/qfieldcloud/project/models.py
```diff
@@ -1539,3 +1557,195 @@ class Meta:
...
+    alias = models.TextField(
+        blank=True, help_text=_("Optional user-facing alias configured in QGIS.")
+    )
+
+    comment = models.TextField(
+        blank=True,
+        help_text=_("Optional field comment or description configured in QGIS."),
+    )
+
+    type = models.CharField(
```
suricactus: I think this can be enum one day too. Being a char field is ok for now.

manylon: Currently we are getting `QgsField::typeName()` from the QGIS repo. The type value comes from the provider, so whatever native type name the specific data source reports e.g. `OGR`, `Sharefile`, `GeoPackage`, `PostGIS` etc. So maybe it's better to change the name in the model to `type_name` and if we want, we can actually store the actual `type` and add a mirror enum?

suricactus: Ah, very good remark!

`QgsField::typeName()` can return very funny values, as you said they are coming from the datasource and the docs clearly states:

> The type is stored exactly as the data store reports it, with no attempt to standardize the value.
https://github.com/qgis/QGIS/blob/release-3_44/src/core/qgsfield.h#L191C52-L192C76

What we actually care about is the type in a standartized way, so we can recreate the value later.

So having the `type` and `subtype` probably is not a bad idea.

suricactus:
Rebase and ready to merge.

PR #1762, opened 2026-08-13
Revert "Merge pull request #1423 from felixxm/exclusion-constraint"

suricactus:
Unfortunately we have to revert this change because Django is trying to be a smartass and send a verification query for the `CheckConstraint` BEFORE the actual database check constraint kicks in. This is problematic because the produced query is returning that constraint is violated when it is not.

This blocks the support team to use Admin to expire existing subscriptions.

image

E.g. for the subscription overlap it sends this query:

```
SELECT
  1 AS "_check"
WHERE
  COALESCE(
    (
      EXISTS(
        SELECT
          1 AS "a"
        FROM
          "subscription_subscription" U0
        WHERE
          (
            U0."account_id" = (U0."ac count_id")
            AND TSTZRANGE(
              U0."active_since", U0."active_until",
              '[)'
            ) && (
              STZRANGE(
                '2026-08-13T11:40:36+02:00' :: timestamptz,
                '2026-08-13T11:45: 18+02:00' :: timestamptz,
                '[)'
              )
            )
            AND NOT (U0."id" = 8)
            AND U0."active_since" IS NOT NULL
          )
        LIMIT
          1
      )
    ), true
  )
```

This reverts commit f29d532c36b8823a9db69ac01bdbd6c4e8b95e6c, reversing changes made to 4f232e0ca3db20160e27fdcf24dfc3a53493acf2.

Related https://github.com/opengisch/QFieldCloud/pull/1423

### DIFF
--- a/docker-app/qfieldcloud/subscription/functions.py
+++ b/docker-app/qfieldcloud/subscription/functions.py
@@ -1,7 +1,0 @@
-from django.contrib.postgres.fields import DateTimeRangeField
-from django.db.models import Func
-
-
-class TsTzRange(Func):
-    function = "TSTZRANGE"
-    output_field = DateTimeRangeField()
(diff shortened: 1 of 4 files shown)


toebivankenoebi (approved):
Simple revert, lgtm.

PR #1766, opened 2026-08-16
Add ability to order /deltas/{project_id}/ by created_at field

nirvn:
In upcoming QField 4.3, we're (dramatically) improved the visibility of the delta/changes history dialog. 

To be on the safe side, we should upgrade that dialog to offer a paginated view, with additional items added (when available) once the user reaches the very bottom of the list view. 

Pagination for the /api/v1/deltas/ endpoint is already in, however ordering is fixed, and serves the delta changes from oldest to newest, which isn't ideal.

This PR adds the ability to order the list by created_at, which QField will then use to get pages of deltas served in reversed created_at order.

### DIFF
--- a/docker-app/qfieldcloud/core/views/deltas_views.py
+++ b/docker-app/qfieldcloud/core/views/deltas_views.py
@@ -15,6 +15,7 @@ from drf_spectacular.utils import (
     extend_schema_view,
 )
 from qfieldcloud.core import exceptions, pagination, permissions_utils, utils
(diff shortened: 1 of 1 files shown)


suricactus:
Looks good to me, @manylon please review.

manylon (approved):
Works like a charm, thanks! 

Tested with `http://localhost/api/v1/deltas/<projet_id>/?ordering=-created_at`

PR #1769, opened 2026-08-17
Refactor logging to use printf-style formatting in `qfc_worker/utils.py`

manylon:
Converts all f-string based logging calls in `qfc_worker/utils.py` to use lazy formatting instead.


suricactus (approved):

manylon:
Thanks

PR #1773, opened 2026-08-18
fix: Prevent `Delta.created_by` to cause n+1 by prefetching the `User`

suricactus:
The `DeltaSerializer` has `created_by = serializers.StringRelatedField()` field, which causes the `Delta` model to select from `core_user`.


toebivankenoebi (approved):
Discussed offline, lgtm.

PR #1776, opened 2026-08-18
Move from `django-migrate-sql-3` to `django-migrate-sql-deux`

suricactus:
The latter is OPENGIS.ch fork that was needed for Django 4.x:
https://pypi.org/project/django-migrate-sql-3/
But it does not work with Django 5.x.

The `django-migrate-sql-deux` is revived upstream version that actually
works with Django 5.x.


manylon (approved):
LGTM, thanks!

PR #1777, opened 2026-08-19
Implement multiple query optimizations on the API

suricactus:
Make some of the most commonly used endpoints faster by reducing the number of requests sent.

See individual commits for more.

### DIFF
--- a/docker-app/qfieldcloud/authentication/auth_backends.py
+++ b/docker-app/qfieldcloud/authentication/auth_backends.py
@@ -3,7 +3,6 @@ from typing import cast
 from allauth.account.auth_backends import (
     AuthenticationBackend as AllAuthAuthenticationBackend,
 )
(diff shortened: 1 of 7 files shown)


nirvn:
Sweet

t-book:
Nice ! 🍦

lukasgraf (approved):
LGTM! 👍 

I have a hard time judging the impact of the very first change though, always prefetching `account`  and `account__user` when getting the current subscription. `get_or_create_current_subscription` is invoked a lot, very liberally, via the `UserAccount.current_subscription` property.

But on the other hand, access to the account and user is then often needed just afterwards. So its hard to say whether this larger query will outweigh the cost of additional queries.

Since this will only ever be for exactly one subscription / user / account, and doesn't blow up a query for _multiple_ such entities, whatever penalty there is should be constant, and small. So I'd say let's try it.

PR #1778, opened 2026-08-19
Prevent initial packaging on template projects

manylon:
Currently, a user can make an initial package for a template project for QField, but once they make edits to that project, they cannot repackage it back.

This PR prevents packaging for template projects overall.

### DIFF
--- a/docker-app/qfieldcloud/core/permissions_utils.py
+++ b/docker-app/qfieldcloud/core/permissions_utils.py
@@ -530,6 +530,9 @@ def can_read_jobs(user: QfcUser, project: Project) -> bool:
 
 def can_create_jobs(user: QfcUser, project: Project, job_type: Job.Type) -> bool:
     """Check if the user has permission to create a job of the given type."""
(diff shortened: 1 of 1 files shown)


docker-app/qfieldcloud/core/permissions_utils.py
```diff
@@ -530,6 +530,9 @@ def can_read_jobs(user: QfcUser, project: Project) -> bool:
 
 def can_create_jobs(user: QfcUser, project: Project, job_type: Job.Type) -> bool:
     """Check if the user has permission to create a job of the given type."""
+    if job_type == Job.Type.PACKAGE and project.is_template_project:
```
suricactus:
```suggestion
    if job_type == Job.Type.PACKAGE and not project.is_regular_project:
```
I think this is more correct, so we can prevent `shared_datasets` from the same faith?

Feel free to force push in such PRs.

suricactus: Don't we need the `not` in front of the `project.is_regular_project`?

suricactus (approved):
LGTM

PR #1779, opened 2026-08-19
Move `project_type` validation into` Project.clean()`

manylon:
Currently when we change a project type from any to `shared_datasets` through Django Admin and the name is not the expected `shared_datasets` the server throws 500 error instead of giving a message with the problem to the user.

This PR fixes this by moving this validation into the `clean()` method of the `Project` model.


suricactus:
The title is missing "why" though. This is the most valuable thing in the PR/commit description usually

manylon:
FIxed, sorry!

suricactus (approved):
Thanks!

PR #1780, opened 2026-08-20
Rate limit account email actions (password reset, email confirmation)

manylon:
Currently `ACCOUNT_RATE_LIMITS` is hardcoded to `False`, so password reset and email confirmation requests are unlimited. 

This PR makes it configurable via a new `ACCOUNT_RATE_LIMITS` envvar and fixes two issues in the resend-confirmation flow: it bypassed rate limiting entirely even with `confirm_email` set in `ACCOUNT_RATE_LIMITS`, and its fallback redirect passed a raw URL name into `HttpResponseRedirect` instead of resolving it with `reverse()`, causing a 404.

Changes made:
- Configure account rate limit via a new `ACCOUNT_RATE_LIMITS` envvar
- Add a custom 429 error page, served by nginx
- Fix `resend_confirmation_email` to actually rate limit resends via `allauth.core.ratelimit.consume` (it previously called `send_confirmation()` that was bypassing rate limiting)
- Fix `redirect_to_referer_or_view` to resolve the URL name with `reverse()` before redirecting, instead of passing the raw name ("account_login") straight into `HttpResponseRedirect` — this was causing a 404 on `/api/v1/resend-confirmation/account_login` after a resend confirmation email request, because the view was redirecting to `/api/v1/resend-confirmation/account_login` instead of `/accounts/login/`

> 🚨 Note: `allauth.core.ratelimit.consume` has a "TODO: We ened to deprecate this module and keep rate limiting internal" comment. I didn't manage to get to a working alternative solution. Worth re-checking this on `allauth` upgrades.


duke-nyuki:
Task linked: [QF-8722 Add ratelimit for resending emails for verification](https://app.clickup.com/t/2192114/QF-8722)

.env.example
```diff
@@ -388,6 +388,18 @@ SOCIALACCOUNT_PROVIDERS='{}'
...
+# See https://docs.allauth.org/en/latest/account/rate_limits.html
+#
+# - reset_password: requesting a password reset email. At most 20 requests per minute per IP address, and at most 2 per minute per key (submitted email)
+# - reset_password_from_key: submitting the new-password form reached via the link from the email. At most 2 attempts per minute per IP address.
+# - confirm_email: requesting an email confirmation mail. At most 1 per 3 minutes per key (email address).
+ACCOUNT_RATE_LIMITS='{
+        "reset_password": "20/m/ip,2/m/key",
+        "reset_password_from_key": "2/m/ip",
+        "confirm_email": "1/3m/key"
+    }'
```
suricactus:
```suggestion
#
# Configuration reads as:
# "20/m/ip,2/m/key" - *20* requests per *m*inute per *IP* address AND *2* per *m*inute per *key*.
# By *key* is meant submitted email.
#
# DEFAULT='{
#     "reset_password": "20/m/ip,2/m/key",
#     "reset_password_from_key": "2/m/ip",
#     "confirm_email": "1/3m/key"
# }'
ACCOUNT_RATE_LIMITS='{
    "reset_password": "20/m/ip,2/m/key",
    "reset_password_from_key": "2/m/ip",
    "confirm_email": "1/3m/key"
}'
```

docker-app/qfieldcloud/core/views/accounts_views.py
```diff
@@ -38,29 +42,34 @@ def resend_confirmation_email(request: HttpRequest) -> HttpResponse:
...
+
         return redirect_to_referer_or_view(request, "account_login")
 
-    email_address = request.session["account_verified_email"]
+    try:
+        email_obj = EmailAddress.objects.get(email=email_address)
+    except EmailAddress.DoesNotExist:
+        messages.error(
+            request,
+            _("Email {} not found. Please register again.").format(email_address),
```
suricactus:
```suggestion
            # Do not show the email to prevent leaking of email by hijacking a session.
            _("Email not found. Please go through sign-up process again.")
```

Does not hurt to hide the email here, it's very unlikely and suspicious we end up here. We also changed the language some time ago to "sign up" instead of "register".

suricactus (approved):
Nice, thanks!

suricactus (approved):
Thanks!

PR #1781, opened 2026-08-20
Optimize GET /projects endpoint by reducing 5 queries per Project

suricactus:
Add selecting/prefetching QGIS project and layers. Since we need only the vector layers for now, prefetch only the vector layer on the `QgisProject` instance.

Sample call of GET https://localhost:8002/api/v1/projects/?format=json with 3 projects.

| before | after |
|-|-|
| image | image |

### DIFF
--- a/docker-app/qfieldcloud/project/utils/project_utils.py
+++ b/docker-app/qfieldcloud/project/utils/project_utils.py
@@ -31,9 +31,11 @@ def has_online_vector_data(project: Project) -> bool:
     if qgis_project is None:
         return False
 
-    vector_layers = qgis_project.layers.filter(layer_type=QgsLayerType.Vector)
+    for layer in qgis_project.layers.all():
+        #  skip vector layers
+        if layer.layer_type != QgsLayerType.Vector:
+            continue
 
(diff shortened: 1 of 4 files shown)


nirvn:
@suricactus , that's nice, I was starting to notice a pretty significant delay between the log in screen and the home projects' list.

docker-app/qfieldcloud/project/utils/project_utils.py
```diff
@@ -1,7 +1,7 @@
 from urllib.parse import parse_qs, urlparse
 
-from qfieldcloud.project.enums import QgsLayerType
 from qfieldcloud.project.models import Project
+from qfieldcloud.project.views import QgsLayerType
```
manylon:
```suggestion
from qfieldcloud.project.enums import QgsLayerType
```

manylon (approved):
Amazing, did some benchmarks with master, 1/3 less queries!

docker-app/qfieldcloud/project/utils/project_utils.py
suricactus: Thanks!

PR #1783, opened 2026-08-21
Add default project role for organization members

manylon:
Organization admins already get admin access to every project the organization owns, but regular members previously got nothing unless added as a collaborator. 

This PR adds a nullable `default_project_role_for_members` field on `Organization` so admins and organization owners can configure a project role that every non-admin member automatically inherits across all of the organization's projects.

Changes:
- Extract `ProjectRole` into `qfieldcloud.project.enums` so `Organization` and `ProjectCollaborator.Roles` can reference it.
- Add an `organization_member` CTE to `projects_with_roles_vw` to resolve the default project role for members.
- Add a migration to add the new field and update the view.
- Expose the field in Django admin in organization view.

UI:
image
image


duke-nyuki:
Task linked: [QF-3082 Add "default role for members" configuration in Organizations](https://app.clickup.com/t/2192114/QF-3082)

docker-app/qfieldcloud/core/models.py
```diff
@@ -1042,12 +1050,7 @@ def skip_incognito(self):
 
 
 class ProjectCollaborator(models.Model):
-    class Roles(models.TextChoices):
-        ADMIN = "admin", _("Admin")
-        MANAGER = "manager", _("Manager")
-        EDITOR = "editor", _("Editor")
-        REPORTER = "reporter", _("Reporter")
-        READER = "reader", _("Reader")
+    Roles = ProjectRole
```
suricactus: I guess you wanted to have less changes and that's why you kept this for compatibility. Please remove it and make a dedicated commit (probably the first one?) that moves the `core.models.ProjectCollaborator.Roles` to `project.enums.ProjectRole`.

docker-app/qfieldcloud/project/enums.py
```diff
@@ -2,6 +2,14 @@
 from django.utils.translation import gettext as _
 
 
+class ProjectRole(models.TextChoices):
```
suricactus: Hm, I am not sure whether it should be `ProjectRole` or `ProjectCollaboratorRole`. Because the "role" of the project can be "disaster recovery" role or "teachning" role.

docker-app/qfieldcloud/core/models.py
```diff
@@ -654,6 +654,14 @@ class Meta(User.Meta):
 
     is_initially_trial = models.BooleanField(default=False)
 
+    inherited_project_role = models.CharField(
+        max_length=10,
+        choices=ProjectCollaboratorRole.choices,
+        null=True,
+        blank=True,
+        default=None,
```
suricactus, line 662:
```suggestion
        default=None,
        help_text=_("..."),
```

Any new field we add that is not screaming it's obvious purpose should have a help text. Imagine being support and seing "Inherited project role" for the first time. What is it?

BTW while developing it, if you came with a better name, happy to discuss.

manylon: I would go with `default_project_role`, because we are setting default permissions for project access for every member in the organization.

suricactus: The main problem I have with the word "default" is that it can be interpreted "if I add add a new collaborator to a project owner by the organization, this is the role to be assigned by default when adding it" (currently that role is hardcoded as READER.

But if you don't get that confusion, I can live with it and name it "default" instead of "inherited" (or other better name).

manylon: What about `base_project_role`, because it's the lowest permission role every member gets?

manylon: Decided to go with `default_project_role_for_members`.

suricactus: Choose your fighter: default_project_role, inherited_project_role, default_project_role_for_members

suricactus (approved):
Thanks!

PR #1785, opened 2026-08-21
refactor: move `test_utils_project` from `core` to `project` app

suricactus:
It is more logical arrangement of  tests.
This will spread the load a bit more evenly between apps.


manylon (approved):
LGTM, thanks!

PR #1789, opened 2026-08-23
feat: add HTTP retries to webdav storage

gounux:
Follow-up of the closed #1553

:point_right:  Adding `urllib3`'s Retries to the HTTP requests sent by the WebDAV storage, in order to make it more resilient towards a WebDAV backend that may throw 429 / 50X errors.

Credits to @t-book, thanks mate for the initial idea :)

### DIFF
--- a/docker-app/qfieldcloud/filestorage/backend.py
+++ b/docker-app/qfieldcloud/filestorage/backend.py
@@ -10,8 +10,10 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import ContentFile
 from django.core.files.storage import Storage
 from django.http import HttpResponse
+from requests.adapters import HTTPAdapter
 from requests.auth import HTTPBasicAuth
 from storages.backends.s3 import S3Storage
+from urllib3.util.retry import Retry
 
 from qfieldcloud.filestorage.constants import VERSION_SUFFIX_REGEX
 
@@ -93,6 +95,10 @@ class QfcS3Boto3Storage(QfcBackendStorageMixin, S3Storage):
         return params
 
 
+HTTP_RETRIES_COUNT = 3
+HTTP_RETRIES_BACKOFF_FACTOR = 0.5
+
+
 class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
     """
     Storage backend using WebDAV.
@@ -134,6 +140,17 @@ class QfcWebDavStorage(QfcBackendStorageMixin, Storage):
         Authenticates using the `basic_auth` storage option.
         """
         session = requests.Session()
+
+        retries = Retry(
+            total=HTTP_RETRIES_COUNT,
+            backoff_factor=HTTP_RETRIES_BACKOFF_FACTOR,
+            status_forcelist=[429, 500, 502, 503, 504],
+            allowed_methods=["HEAD", "GET", "PUT", "DELETE", "MKCOL"],
+        )
+
+        adapter = HTTPAdapter(max_retries=retries)
+        session.mount("http://", adapter)
+        session.mount("https://", adapter)
 
         username, _, password = self.basic_auth.partition(":")
         session.auth = HTTPBasicAuth(username, password)


duke-nyuki:
Task linked: [QF-8729 Add HTTP retries to webdav storage](https://app.clickup.com/t/2192114/QF-8729)

docker-app/qfieldcloud/filestorage/backend.py
```diff
@@ -130,7 +136,20 @@ def get_requests_session(self, **kwargs) -> requests.Session:
...
         Creates a HTTP session for requesting webdav later.
         """
-        return requests.Session()
+        session = requests.Session()
+
+        retries = Retry(
+            total=HTTP_RETRIES_COUNT,
+            backoff_factor=HTTP_RETRIES_BACKOFF_FACTOR,
+            status_forcelist=[429, 500, 502, 503, 504],
+            allowed_methods=["HEAD", "GET", "PUT", "DELETE", "MKCOL"],
```
suricactus, line 148: I am always a bit sensitive on non GET/OPTIONS/HEAD requests retriability. Are we 100% sure we want/need all these other methods here?

What happens if you get errot 500 for DELETE, you cannot know if the file was deleted on the webdav storage. Then if you retry, and the file was DELETE-d, you will get an error 404 I guess. Is this properly handled in our code?

gounux: Okay, thanks for the point, not sure I'm following you :thinking:  
Suggesting retrying only on GET / OPTIONS / HEAD requests ?

t-book:
Just adding my two cents here.

I think @suricactus raises a very valid point. I had a quick look at the current delete() implementation, 

https://github.com/opengisch/QFieldCloud/blob/980258314984d264f31a9aa853c390943b43a762/docker-app/qfieldcloud/filestorage/backend.py#L249-L258

and if I understand it correctly, the specific DELETE -> 500 -> retry -> 404 case should already be handled: delete() catches and ignores HTTPError afterwards.

So I’m not sure I would necessarily limit retries to only GET / OPTIONS / HEAD. The original intention was also to make PUT/MKCOL more resilient to transient WebDAV errors such as 503, which is why I think those methods are worth considering for retries as well.

That said, I’m not familiar enough with all the WebDAV edge cases here to have a strong opinion on the final list of methods, so I’d be very happy to go with whatever the core team considers safest.

docker-app/qfieldcloud/filestorage/backend.py
suricactus: This gives some hints what I mean: https://softwareengineering.stackexchange.com/questions/382594/should-you-retry-500-api-errors

But in short - yes, I am not big fan of seeing PUT and DELETE in the list here. Not sure about `MKCOL`, it is a WebDav specific verb, so if you know the library internals and retrying is ok (e.g. two requests with the same content does not cause different responses), then I am ok with it.

suricactus: Ok, seems DELETE and PUT are idempotent: https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

Are we sure the WebDav implementations are following this? Honest question, not sure. If they are, keeping PUT and DELETE is fine.

We cannot check all WebDav implementations, but let's check at least for those we care about.

suricactus:
@gounux what is your opinion on the comments shared by me and @t-book ?

gounux:
> @gounux what is your opinion on the comments shared by me and @t-book ?

Thanks both of you for the thoughts, apologies I left some emojis but not any comment...

I'm +0 on keeping the `PUT` and `DELETE` methods in the retry, so as is: IMO this is already a win towards the current WebDAV storage behavior. That said CI is still failing on external reasons, let's try to get this merge before the next release !

suricactus (approved):
Ok, @gounux I am -0 on this, but since you are dealing with WebDAV stuff (and I definitely don't want to),  let's go with your gut feeling!

PR #1792, opened 2026-08-24
Change "filename" in job feedback from absolute to relative path

manylon:
`filename` in job feedback previously contained the full temporary path instead of just the relative path.
This PR extracts the download directory into a `PROJECT_DOWNLOAD_DIR` constant, defined where the worker already handles file downloads, and use it to strip the beginning of the path from `filename`

### DIFF
--- a/docker-qgis/qfc_worker/utils.py
+++ b/docker-qgis/qfc_worker/utils.py
@@ -51,6 +51,8 @@ from tabulate import tabulate
 # Get environment variables
 JOB_ID = os.environ.get("JOB_ID")
 
+PROJECT_DOWNLOAD_DIR = "/tmp/project"
+
 # Network timeout for QGIS. Reduced from default 60 seconds to fail fast
 # when layers are unreachable (e.g. network outage, blocked external access).
 # This prevents long blocking delays when loading project file.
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8706 Change "filename" in job feedback from absolute to relative path](https://app.clickup.com/t/2192114/QF-8706)

docker-qgis/qfc_worker/utils.py
```diff
@@ -501,8 +503,12 @@ def download_project(
     logging.info("Preparing a temporary directory for project files…")
 
     if not destination:
-        # Create a temporary directory
-        destination = Path(tempfile.mkdtemp())
+        destination = Path(PROJECT_DOWNLOAD_DIR)
+        destination.mkdir(parents=True, exist_ok=True)
```
suricactus, line 513: Probably we need these parameters if we call `download_project` multiple times. But if we don't I guess we don't need them.

You can mention this in a comment abose this line.

suricactus: """Download the project files in the project "working" directory from the Object storage.

The destination may be passed as `destination` or will be a fixed temporary directory (`PROJECT_DOWNLOAD_DIR`).
Return:
    the directory path
"""

docker-qgis/qfc_worker/utils.py
```diff
@@ -817,6 +823,9 @@ def get_layers_data(project: QgsProject) -> dict[str, dict]:
         else:
             fields = None
 
+        if filename:
+            filename = filename.replace(PROJECT_DOWNLOAD_DIR, "")
```
suricactus: What if my layer is stored at:

`./my_data/tmp/project/data.gpkg`?

This replace is a bit too naive, you should replace only the prefix of the string. Probably another `if` the filename path is absolute will not hurt.

suricactus (commented):
Nice, thank you! A few small comments.

suricactus (approved):
Thanks!

manylon:
Thanks too, will clean history and merge.

PR #1793, opened 2026-08-24
Add more ruff rules

suricactus:
Trying to get closer to the rules in the other QField python  projects.


lukasgraf:
So, from a first glance, I'm not a fan of enabling these two, which obviously also don't have a fixer available:
- [`N818`](https://docs.astral.sh/ruff/rules/error-suffix-on-exception-name/) - error-suffix-on-exception-name
- [`D417`](https://docs.astral.sh/ruff/rules/undocumented-param/) - undocumented-param

**N818** - will disallow concise and readable exception names like `SignatureMismatch`, `QuotaExceeded`, `SessionExpired`, `HandshakeFailed`, `UnsupportedCapability` (hypothetical examples). All to just make them conform to a particular pattern without any real benefit.

**D417** - will almost certainly lead to meaningless overdocumentation fluff:
```
foobar_enabled: Whether foobar is enabled or not.
```
Great, that part of the docstring is not telling me *anything* of substance, it's just additional noise. I'm generally not a fan of Google style docstrings for this reason, IMHO they're overly verbose, and they tend to get out of sync very quickly (which is much worse than no docstring at all). I much prefer PEP 257. Docstrings where they help, and actually provide additional information, but not ones that overdescribe a small function that is already self-explanatory.

suricactus:
> N818 - will disallow concise and readable exception names like SignatureMismatch, QuotaExceeded, SessionExpired, HandshakeFailed, UnsupportedCapability (hypothetical examples). All to just make them conform to a particular pattern without any real benefit.

We already use a convention that is not enforce - all custom exceptions should be suffixed with something to highlight they are exceptions - `Error` or `Exception`. The real benefit is that you immediately know you are dealing with an error, and not looking at something else. I guess that's why most Python exceptions are suffixed with "Error". E.g. `FileNotFoundError` vs `FileNotFound`.

> D417 - will almost certainly lead to meaningless overdocumentation fluff:

This is a bit more annoying, for sure if we do meaningless overdocumentation. But if I see a function with 3 params, but only 2 of them are documented, I immediately discard the docstring due to lack of trust it was updated when the third param was added. Unfortunately this `ruff` rule does not support empty docs for params:

```
def foo(a, b):
    """Do something


    Args:
        a: The first argument.
        b:
    """
    pass
```

It must have something, e.g. a dash:

```
def foo(a, b):
    """Do something


    Args:
        a: The first argument.
        b: -
    """
    pass
```

But I already caught a few situations where the documentation got silently outdated.

> I'm generally not a fan of Google style docstrings for this reason

We already have the majority of the docstrings that include params, returns etc tend to be Google docstrings.

> Docstrings where they help, and actually provide additional information, but not ones that overdescribe a small function that is already self-explanatory.

Exactly. If it is a small obvious function, you will have 1-2 params that are obvious by name, or you describe in the freetext of the function. If the function is complex with 3-5+ params, then probably is not too bad idea to describe each and every of them. Such function is either overengineered and should be split into smaller ones, so there is no need for `Args:`, or is complex enough and you want a docstring one can trust.

---

The most important thing here is that we add these rules so they can be automatically checked, modern IDEs and tools such as pre-commit do this for us.

t-book:
Maybe worth spelling out since it might address the fluff concern: D417 [only checks docstrings that already have an Args: section](https://docs.astral.sh/ruff/rules/undocumented-param/), and we're ignoring D10 anyway — so nothing forces a small self-explanatory function to grow an Args: block. It only kicks in once someone has opted into documenting params, which is exactly the "half-updated docstring" case. Or am I missing a scenario where it'd fire unprompted?

suricactus:
> It only kicks in once someone has opted into documenting params, which is exactly the "half-updated docstring" case.

Yes. This is my whole message above in one sentence.

toebivankenoebi (approved):
Looked through all the changes, only minor and looks all good to me. I like the ruleset, thanks @lukasgraf for bringing up your concerns and @t-book for chiming in. 

Don't see any problems with over-documenting with `D10x` ignored.
Don't see a problem with adding `Error` to to an exception name, doesn't change it's readability nor conciseness much.

PR #1794, opened 2026-08-24
refactor: remove unused and broken `File.remove_version` method

suricactus:
It is not only not used, it was never properly implemented.


gounux (approved):
Thanks, nice catch 👌

PR #1796, opened 2026-08-24
Add missing dependency for project migration

manylon:
Add to project migration `0003`  dependency `core/0104_alter_projectcollaborator_project_and_more`.

As I understand it, the dependency tree was missing migrations for projects that need to run after the project table rename. Because of this, now when we have changes in the SQL views, they don't execute in the correct order, and we get:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/django/db/backends/utils.py", line 103, in _execute
    return self.cursor.execute(sql)
           ~~~~~~~~~~~~~~~~~~~^^^^^
psycopg2.errors.UndefinedTable: relation "project_project" does not exist


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/src/app/manage.py", line 21, in <module>
    main()
    ~~~~^^
  File "/usr/src/app/manage.py", line 17, in main
    execute_from_command_line(sys.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
    ~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.14/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/django/core/management/base.py", line 420, in run_from_argv
    self.execute(*args, **cmd_options)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/django/core/management/base.py", line 464, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.14/site-packages/django/core/management/base.py", line 111, in wrapper
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.14/site-packages/django/core/management/commands/migrate.py", line 353, in handle
    post_migrate_state = executor.migrate(
        targets,
    ...<3 lines>...
        fake_initial=fake_initial,
    )
  File "/usr/local/lib/python3.14/site-packages/django/db/migrations/executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
[...]

PR #1797, opened 2026-08-24
Update file storage migration to ensure correct order of migrations

manylon:
> Linked with:
> - https://github.com/opengisch/QFieldCloud/pull/1796

`filestorage.0002_file_file_storage` reads the `core_project` table by its original name via raw SQL.
`core.0104_alter_projectcollaborator_project_and_more` later renames that table to `project_project`.
These two migrations had no dependency edge between them e.g `filestorage.0002` only happened to run before the rename by coincidence of Django's migration plan, not because anything enforced it. This PR fixes this by adding `run_before` in the migration.


suricactus:
Is this somehow related to https://github.com/opengisch/QFieldCloud/pull/1796 ? If so, please mention it in the initial message.

suricactus (approved):

manylon:
Thanks

PR #1798, opened 2026-08-25
Preserve source project CRS when cloning

manylon:
Currently cloning a project silently resets the source project CRS to the seed default (`EPSG:3857`) instead of keeping it. This PR fixes this by adding a `clone_from_project` guard around the CRS-setting block in `create_project.py` in `configure_qgis_project`.

Also added a unit test to ensure that the source project CRS is preserved when cloning a project in the QGIS worker `qfc_worker/tests/test_create_project.py`.

### DIFF
--- a/docker-qgis/qfc_worker/commands/create_project.py
+++ b/docker-qgis/qfc_worker/commands/create_project.py
@@ -261,11 +261,16 @@ def configure_qgis_project(project_seed: ProjectSeed, project_filename: str) -> 
 
     project.setTitle(project_seed.name)
 
-    crs = QgsCoordinateReferenceSystem(project_seed.crs)
-    if crs.isValid():
-        project.setCrs(crs)
+    if project_seed.clone_from_project:
+        logger.info("Cloning an existing project, keeping the source project's CRS.")
     else:
-        logger.warning("Project CRS parameter is invalid, skipping CRS configuration.")
+        crs = QgsCoordinateReferenceSystem(project_seed.crs)
+        if crs.isValid():
+            project.setCrs(crs)
+        else:
+            logger.warning(
+                "Project CRS parameter is invalid, skipping CRS configuration."
+            )
 
     if project_seed.settings.basemaps and not project_seed.clone_from_project:
         logger.info("Adding basemaps to the project...")
--- a/docker-app/qfieldcloud/core/tests/test_jobs.py
+++ b/docker-app/qfieldcloud/core/tests/test_jobs.py
@@ -569,6 +569,62 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
             set(self.p1.qgis_project.layers.values_list("qgis_layer_id", flat=True)),
         )
 
+    def test_clone_project_keeps_source_crs(self):
+        """Cloning a project must preserve the source project's CRS.
+
+        The source project is set up with CRS `EPSG:7801`, which is
+        deliberately different from the project seed's hardcoded fallback
+        CRS (`EPSG:3857`). If the clone incorrectly fell back to that
+        default instead of inheriting the source's CRS, this test would
+        catch it.
+        """
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.t1.key)
+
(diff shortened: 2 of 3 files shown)


duke-nyuki:
Task linked: [QF-8741 Preserve QGIS source project CRS when cloning](https://app.clickup.com/t/2192114/QF-8741)

docker-qgis/qfc_worker/tests/test_create_project.py
```diff
@@ -0,0 +1,50 @@
...
+                settings={
+                    "schemaId": "https://app.qfield.cloud/schemas/project-seed-20251201.json",
+                    "xlsform": None,
+                    "basemaps": [],
+                },
+            )
+
+            configure_qgis_project(project_seed, project_filename)
+
+            self.assertEqual(QgsProject.instance().crs().authid(), source_crs.authid())
```
suricactus: I think it will be easier if this test lives here, where most of the job related tests live:

https://github.com/opengisch/QFieldCloud/blob/7756c5ffa14a08c316eedb389ae6a99bd3d0c1f8/docker-app/qfieldcloud/core/tests/test_jobs.py#L480-L481

t-book:
Small question while reading: can a seed have both `clone_from_project` and `basemaps` set? If so, this branch
https://github.com/opengisch/QFieldCloud/blob/619b8ef7a149cda39984e2127837d6796e1ebc6d/docker-qgis/qfc_worker/commands/create_project.py#L275
 logs "No basemaps configured for this project seed" in the clone case even when the seed does have some. If that combination can't happen in practice, ignore me.

suricactus:
Basemaps should not be configured when cloning projects. We reuse the same job for two things - creating projects from scratch, and cloning projects. 

The intention here is that we eventually allow management of basemaps within QFieldCloud.

t-book:
@suricactus thanks! Makes sense!

suricactus (approved):
Thanks!

docker-qgis/qfc_worker/tests/test_create_project.py
manylon: I added a different test case in `test_jobs` with an extra fixture. Take a look at it, IMHO I would keep both tests, because they test in different levels of the code.

suricactus: Thanks. Delete the test on this level for now, the other test covers this situation fully.

We cannot afford to test the same thing multiple times for now, especially since we have great plans for our tests in the upcoming months.

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -261,11 +261,16 @@ def configure_qgis_project(project_seed: ProjectSeed, project_filename: str) ->
...
+        logger.info("Cloning an existing project, keeping the source project's CRS.")
     else:
-        logger.warning("Project CRS parameter is invalid, skipping CRS configuration.")
+        crs = QgsCoordinateReferenceSystem(project_seed.crs)
+        if crs.isValid():
+            project.setCrs(crs)
+        else:
+            logger.warning(
+                "Project CRS parameter is invalid, skipping CRS configuration."
+            )
```
suricactus, line 273: Wouldn't it be less friction if we set the `project_seed.crs` to match the one from the source project when creating the `ProjectSeed` instance instead?

docker-qgis/qfc_worker/tests/test_create_project.py
t-book: > especially since we have great plans for our tests in the upcoming months.

🚀  – Looking much forward!

docker-qgis/qfc_worker/commands/create_project.py
manylon: I think it will be harder to understand, because now you can see the intend clearly, if we are cloning, don't do anything with the crs, otherwise use the `project_seed` one

suricactus: I am 0 on this one, ok with the current approach.

suricactus (approved):

PR #1799, opened 2026-08-25
Use the correct QGIS worker image when cloning a project

manylon:
Cloning a project always ran its `CREATE_PROJECT` job under QGIS 3, regardless of the source project's actual QGIS version, because the clone's own `qgis_version` is still unset at that point, so the worker image selection fell back to the QGIS 3 default every time.

This PR fixes this by getting the QGIS version from the source project.

Changes:
- Add `resolve_project_qgis_version()`, which falls back to the clone source's `qgis_version` when the project's own is unset
- Use it in `JobRun.get_qgis_image()` instead of reading `project.qgis_version` directly
- Add tests covering the own-version, clone-fallback, and unknown-version cases

### DIFF
--- a/docker-app/worker_wrapper/wrapper.py
+++ b/docker-app/worker_wrapper/wrapper.py
@@ -715,6 +715,37 @@ class CreateProjectJobRun(JobRun):
         "%(project__id)s",
     ]
 
+    def get_qgis_image(self) -> str:
+        """Picks the QGIS image with the version of the clone project.
+
+        If the clone project doesn't have one, uses the source project's QGIS version.
+
+        A clone's own `qgis_version` is still unset at this point, so
+        without this, we'd always fall through to `JobRun.get_qgis_image()`'s
+        QGIS 3 default instead of matching the source project's version.
+        """
+        clone_from_project = None
+        if hasattr(self.job.project, "seed"):
+            clone_from_project = self.job.project.seed.clone_from_project
+
+        if (
+            self.job.project.qgis_version
+            or not clone_from_project
+            or not clone_from_project.qgis_version
+        ):
+            return super().get_qgis_image()
+
+        qgis_major_project_version = get_qgis_major_version(
+            clone_from_project.qgis_version
+        )
+
+        if qgis_major_project_version not in self.qgis_images:
+            raise JobException(
+                f"Unsupported QGIS major version {qgis_major_project_version} for project {self.job.project.id} cloned from project {clone_from_project.id} stored with {clone_from_project.qgis_version}."
+            )
+
+        return self.qgis_images[qgis_major_project_version]
+
 
 def cancel_orphaned_workers() -> None:
     client: docker.client.DockerClient = docker.from_env()
--- a/docker-app/qfieldcloud/core/tests/test_jobs.py
+++ b/docker-app/qfieldcloud/core/tests/test_jobs.py
@@ -625,6 +625,58 @@ class QfcTestCase(QfcFilesTestCaseMixin, APITransactionTestCase):
 
         self.assertEqual(cloned_project.qgis_project.crs, self.p1.qgis_project.crs)
 
+    def test_clone_project_uses_source_qgis_version_for_worker_image(self):
+        self.client.credentials(HTTP_AUTHORIZATION="Token " + self.t1.key)
+
+        ProjectSeed.objects.create(
+            project=self.p1,
+            extent=Polygon.from_bbox(projectseed_utils.DEFAULT_PROJECT_EXTENT),
+            settings={
+                "schemaId": ProjectSeed.SETTINGS_SCHEMA_ID,
+                "basemaps": [],
+                "xlsform": None,
+            },
+        )
+
+        Job.objects.create(
+            project=self.p1,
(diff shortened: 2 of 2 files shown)


duke-nyuki:
Task linked: [QF-8742 When cloning a project, use the source QGIS version to spawn a worker](https://app.clickup.com/t/2192114/QF-8742)

docker-app/worker_wrapper/wrapper.py
```diff
@@ -202,17 +205,18 @@ def get_environment(self) -> dict[str, str]:
         return environment
 
     def get_qgis_image(self) -> str:
-        if self.job.project.qgis_version:
-            qgis_major_project_version = get_qgis_major_version(
-                self.job.project.qgis_version
-            )
+        project = self.job.project
```
suricactus: Please don't rename variables, we have two uses of it.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -202,17 +205,18 @@ def get_environment(self) -> dict[str, str]:
         return environment
 
     def get_qgis_image(self) -> str:
```
suricactus, line 204: Intead of changing this method here, please overload it in `CreateProjectRun` with the inlined logic from `resolve_project_qgis_version`.  This method can and should be reverted to it's original state.

Motivation: Hiding this behind a function with such an innocent name as `resolve_project_qgis_version` might lead to disastrous misunderstanding how the current QGIS version determination works. Actually that is the sole purpose of the `get_qgis_image` method - to allow individual jobs to have individual ways to get the proper QGIS image.

docker-app/qfieldcloud/project/utils/project_utils.py
```diff
@@ -75,3 +75,19 @@ def get_qgis_major_version(version: str) -> int:
...
+
+    If `project.qgis_version` isn't set yet, falls back to the `qgis_version`
+    of its clone source, if any.
+    """
+
+    if project.qgis_version:
+        return project.qgis_version
+
+    if hasattr(project, "seed") and project.seed.clone_from_project:
+        return project.seed.clone_from_project.qgis_version
```
suricactus: This is not checking if we are trying to run a `create_job` or something else, but the purpose of this change is to fix "cloning a project", isn't it? This may lead to very unexpected results if we add other job types in the future that may run before a `create_project` job.

suricactus (changes requested):
This is the problem, this is the solution, but it must be organized in another way. See my comments.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -715,6 +715,35 @@ class CreateProjectJobRun(JobRun):
...
+        A clone's own `qgis_version` is still unset at this point, so
+        without this, we'd always fall through to `JobRun.get_qgis_image()`'s
+        QGIS 3 default instead of matching the source project's version.
+        """
+        clone_from_project = None
+        if hasattr(self.job.project, "seed"):
+            clone_from_project = self.job.project.seed.clone_from_project
+
+        if (
+            not self.job.project.qgis_version
```
suricactus: I would assert this within the `if` body, I don't think there is a situation when `self.job.project.qgis_version` is not null?

manylon: I don't completely understand, but please check the last commit that i did.

suricactus: When I read it now, I can follow, so it's good. Ignore.

suricactus (approved):
Thanks! My only comment is on the first line of the docstring, it's not immediately clear what you mean, but I can't come with a better one.

docker-app/worker_wrapper/wrapper.py
```diff
@@ -715,6 +715,36 @@ class CreateProjectJobRun(JobRun):
         "%(project__id)s",
     ]
 
+    def get_qgis_image(self) -> str:
+        """Picks the QGIS image with the version of the clone project.
+        If the clone project doesn't have one, uses the source project's QGIS version.
```
manylon, line 721: @suricactus I think it's more understandable now.

suricactus: Yes, better. Usually the first line is separated with an empty line before the next one. This allows automated tools to extract the "summary" of the docstring.

manylon: It reads well? 
image

suricactus: Read the very first sentence of https://peps.python.org/pep-0257/#multi-line-docstrings, see the example https://numpydoc.readthedocs.io/en/latest/format.html#short-summary, and the example here https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html .

So please do add a newline after your first line.

---

Tools are written to tolerate people's mistakes. In this case you used one tool to use an example that it works, have you tested all the automated tools that parse docstrings how do they behave?

manylon: Thanks for the detailed clarification on this. I will refresh my memory with the provided links. Also addressed the comment.

PR #1803, opened 2026-08-26
Use user's saved timezone instead of always falling back to server default

manylon:
`TimezoneMiddleware` checked `hasattr(request.user.useraccount, "useraccount")`,
which is always `False` since useraccount is a reverse accessor on User. As a result every request rendered timestamps in the server's `TIME_ZONE` instead of the viewer's profile timezone.

This PR fixes this and adds tests for it.

### DIFF
--- a/docker-app/qfieldcloud/core/middleware/timezone.py
+++ b/docker-app/qfieldcloud/core/middleware/timezone.py
@@ -11,6 +11,7 @@ class TimezoneMiddleware:
     def __call__(self, request):
         if request.user.is_authenticated and hasattr(request.user, "useraccount"):
             user_tz = request.user.useraccount.timezone
+            assert user_tz, "UserAccount.timezone must not be empty"
         elif settings.TIME_ZONE:
             user_tz = zoneinfo.ZoneInfo(settings.TIME_ZONE)
         else:
--- a/docker-app/qfieldcloud/core/middleware/timezone.py
+++ b/docker-app/qfieldcloud/core/middleware/timezone.py
@@ -11,7 +11,9 @@ class TimezoneMiddleware:
     def __call__(self, request):
         if request.user.is_authenticated and hasattr(request.user, "useraccount"):
             user_tz = request.user.useraccount.timezone
+
             assert user_tz, "UserAccount.timezone must not be empty"
+
         elif settings.TIME_ZONE:
             user_tz = zoneinfo.ZoneInfo(settings.TIME_ZONE)
         else:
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-3212 Incorrect timezome in the Project -> Changes page, shows UTC time instead of user's timezone](https://app.clickup.com/t/2192114/QF-3212)

docker-app/qfieldcloud/core/tests/test_middleware.py
```diff
@@ -0,0 +1,41 @@
...
+class QfcTestCase(APITransactionTestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+
+    def get_response(self, request):
+        return None
+
+    def test_activates_authenticated_users_saved_timezone(self):
+        u1 = Person.objects.create(username="u1")
+        u1.useraccount.timezone = "Europe/Sofia"
```
suricactus, line 25: :bulgaria: :exclamation:

suricactus (commented):
I think broke something else in the past. 

See https://github.com/opengisch/QFieldCloud/pull/912 and https://github.com/opengisch/QFieldCloud/pull/914 .

Also, I guess it will not hurt to check if the `timezone` field is non-empty value with an assert or something?

docker-app/qfieldcloud/core/tests/test_middleware.py
manylon: Easiest way to make sure that anything with time zones works is to verify with your own 🙈

docker-app/qfieldcloud/core/middleware/timezone.py
```diff
@@ -9,10 +9,9 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
-        if request.user.is_authenticated and hasattr(
-            request.user.useraccount, "useraccount"
-        ):
+        if request.user.is_authenticated and hasattr(request.user, "useraccount"):
             user_tz = request.user.useraccount.timezone
+            assert user_tz, "UserAccount.timezone must not be empty"
```
suricactus, line 15: Add some air around keywords on the same block level. In this case, just add a newline before.

suricactus: "Before" and "same block level" were the keywords

t-book:
Hi! Just thinking out loud again, as I’ve been doing over the last few days. 😄 I don’t want to create any extra work, so please feel free to ignore this comment if it’s not relevant.

I was just wondering about 
https://github.com/opengisch/QFieldCloud/blob/5073b0e004ac5c83bed9ac0581f5e3c86b8777c2/docker-app/qfieldcloud/core/tests/test_middleware.py#L11
 here. Since it changes the logging state globally for the whole test process, is there a specific message or side effect that it’s meant to suppress? If not, perhaps it could be omitted?

PR #1713, opened 2026-07-17
feat: add a new nullable field `trial_plan` to the `Plan` model

Rakanhf:
To simplify the logic behind the trial plans, we add a new nullable field `trial_plan` to the `Plan` model.

- Added a ForeignKey field `trial_plan` in the Plan model to link to a trial version of the plan
- Added a property `is_trialable` to determine if a plan can be trialed
- Added validation `clean_trial_plan` to ensure trial plans

PR #1805, opened 2026-08-27
feat: add `trial_expires_at` and `trial_plan` fields to the `Subscription` model

Rakanhf:
This is a follow-up to the previous [PR-1713](https://github.com/opengisch/QFieldCloud/pull/1713) that added the `trial_plan` field to `Plan` model We need to keep track of the trial plan for the current subscription instead of having a separate subscription record for the trial

- Added `trial_plan` and `trial_expires_at` fields to `AbstractSubscription` model to support trial subscriptions
- Added `is_trialing` property to determine if a subscription is currently in a trial period


duke-nyuki:
Task linked: [QF-8766 Add `trial_plan` and `trial_expires_at `to subscription model](https://app.clickup.com/t/2192114/QF-8766)

lukasgraf (approved):
LGTM! 👍 

As discussed in chat, I want to double check what the data in the re-created `current_subscriptions_vw` view is, depending on when exactly this migration will get executed. But I'll have another look at this when giving all the PRs from the stacks a test together. For now, in isolation, this change looks good to me 👍

docker-app/qfieldcloud/subscription/models.py
```diff
@@ -56,6 +56,8 @@ class SubscriptionStatus(models.TextChoices):
     ACTIVE_PAST_DUE = "active_past_due", _("Active Past Due")
     # successfully cancelled
     INACTIVE_CANCELLED = "inactive_cancelled", _("Inactive Cancelled")
+    # subscription is in trial
+    ACTIVE_TRIAL = "active_trial", _("Active Trial")
```
suricactus, line 60: Do we really need this new status? My advice is to keep only one way to determine the trialing status - checking the `trial_expires_at > now()` instead.  This way you don't can never get into a weird situation where we have `ACTIVE_TRIAL` status, but the `trial_expires_at` was 6 month ago.

Rakanhf: I think that if we keep it this way, it's true to whats reflected on stripe, which is trialing status. 

in the case you mentioned, I guess that would be another issue, since why would we even have `ACTIVE_TRIAL` status without canceling the trial on stripe ? they should both be the same now, no ?

suricactus: Well, there is no by-design reason to have the sitaution I mentioned. I am describing a potential bug, that would be never possible if the new status does not exist.

I am ok to keep the new status, just wondering what is the additional benefit besides 1:1 match with the status on Stripe. Leave it up to you.

lukasgraf: I would agree with @suricactus here.
From a data modelling perspective, it seems wrong to have the same information stored in two different ways that are redundant, and could (in theory) be contradicting each other. If this new status is introduced, it requires some additional logic to always keep the two in sync, which is easy to get wrong.

And I think the fact that our statuses don't match the Stripe statuses 1:1 is just something we have to accept, but I don't necessarily see it as a flaw.

PR #1808, opened 2026-08-28
Update active users retrieval to use organization context

manylon:
This PR fixes computing `active_users` and `active_users_count`, because before this fix, computing `active_users` for an organization crashed with:
```
AttributeError: 'User' object has no attribute 'active_users'
```


Rakanhf (approved):
LGTM ! 🚀

PR #1810, opened 2026-08-31
Enforce a hard limit on the users list endpoint

manylon:
GET `/api/v1/users/` is readable by any authenticated user and its `q` parameter matches usernames anywhere in the string. The view was only paginated when the client asked for it, so a request without pagination params returned every matching row, letting anyone enumerate the whole user base (username, type, filename, avatar_url, username_display) in a handful of queries.

- add `USER_LIST_LIMIT = 20` setting
- drop pagination on `ListCreateUsersView`
- cap `get_queryset` at `USER_LIST_LIMIT`, keeping the existing relevance ordering
- add a test asserting the endpoint never returns more than the limit

### DIFF
--- a/docker-app/qfieldcloud/core/tests/test_user.py
+++ b/docker-app/qfieldcloud/core/tests/test_user.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
 from django.db import IntegrityError
 from django.utils import timezone
 from rest_framework import status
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-8638 Add hard limit to the number of `api/v1/users/` search results](https://app.clickup.com/t/2192114/QF-8638)

Rakanhf (approved):
LGTM ! 🚀 

In the future I believe that it's would be nice to use the serializer or a filter to manage request params, so it also be better on docs

manylon:
Agree, thanks for the review!

PR #1811, opened 2026-09-01
Fix CI for building `qfieldcloud-qgis4` image with QGIS 4 installed instead of QGIS 3

manylon:
Without a `QFIELDCLOUD_WORKER_REPLICAS` and `TMP_DIRECTORY`, `docker compose config` fails on unset vars, so the resolve step yielded empty build-args and the Dockerfile fell back to its QGIS 3 defaults and the published qgis4 image shipped QGIS 3.44.

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -122,6 +122,7 @@ jobs:
           COMPOSE_SERVICE: ${{ matrix.services.compose_service }}
           DEBUG: "0"
         run: |
+          cp .env.example .env
           {
             echo 'build_args<<EOF'
             docker compose -f docker-compose.yml config --format json \
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -121,8 +121,9 @@ jobs:
         env:
           COMPOSE_SERVICE: ${{ matrix.services.compose_service }}
           DEBUG: "0"
+          QFIELDCLOUD_WORKER_REPLICAS: "1"
+          TMP_DIRECTORY: /tmp
         run: |
-          cp .env.example .env
           {
             echo 'build_args<<EOF'
             docker compose -f docker-compose.yml config --format json \
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -122,6 +122,7 @@ jobs:
           COMPOSE_SERVICE: ${{ matrix.services.compose_service }}
           DEBUG: "0"
           QFIELDCLOUD_WORKER_REPLICAS: "1"
+          TMP_DIRECTORY: /tmp
         run: |
           {
             echo 'build_args<<EOF'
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8782 Fix CI for building `qfieldcloud-qgis4` image with QGIS 4 installed instead of QGIS 3](https://app.clickup.com/t/2192114/QF-8782)

manylon:
Wondering if we should do a follow up on this with removing the default `QGIS_VERSION` in QGIS Dockerfile?

gounux:
Thanks @manylon :)

> Without a .env file, docker compose config fails on unset vars

Could you specify which env vars are missing ?  
Do I presume correctly by saying that those are `QFIELDCLOUD_QGIS3_IMAGE_NAME` and `QFIELDCLOUD_QGIS4_IMAGE_NAME` ?

Reason behind this question is that I'd favor specifying _only_ the missing env var, rather than the whole `.env[.example]` file, which may bring extra and unneeded noise IMO.

What do you think ?

manylon:
Thanks for the feedback @gounux 
Dug a little deeper and found that the problem was with `TMP_DIRECTORY` and `QFIELDCLOUD_WORKER_REPLICAS`.

`QFIELDCLOUD_QGIS3_IMAGE_NAME` and `QFIELDCLOUD_QGIS4_IMAGE_NAME` have default values in `docker-compose.yml` so they are not a problem.

.github/workflows/build_and_push.yml
```diff
@@ -121,6 +121,7 @@ jobs:
         env:
           COMPOSE_SERVICE: ${{ matrix.services.compose_service }}
           DEBUG: "0"
+          QFIELDCLOUD_WORKER_REPLICAS: "1"
```
gounux, line 124: Ok, so that'd be only this `QFIELDCLOUD_WORKER_REPLICAS` missing ? 🤔 

That surprises me a bit, since this env var is - only AFAIK - used for scaling a compose service via the `scale` key, e.g. :

https://github.com/opengisch/QFieldCloud/blob/d7d822114c7ce0c2fdb0ca17b9bdfe63bc7c5b5e/docker-compose.yml#L201

manylon: When I run
```
docker compose  -f docker-compose.yml config --format json \
              | jq -r --arg service "$COMPOSE_SERVICE" '.services[$service].build.args // {} | to_entries[] | "\(.key)=\(.value)"'
```

without `QFIELDCLOUD_WORKER_REPLICAS` I get:
```
error while interpolating services.worker_wrapper.scale: failed to cast to expected type: strconv.Atoi: parsing "": invalid syntax
```
After adding `QFIELDCLOUD_WORKER_REPLICAS ` I get:
```
invalid spec: :/tmp: empty section between colons
```

After adding `TMP_DIRECTORY`
```
DEBUG_BUILD=0
QGIS_REPOSITORY=debian
QGIS_VERSION=1:4.2.1+44resolute
UBUNTU_VERSION=resolute
```

gounux: Thanks for the details, resolving this then :ok_hand:

Rakanhf: I can indeed confirm that this was also the case with the "https://github.com/opengisch/QFieldCloud/actions/runs/33192945375/job/98922771327"`

where 
`error while interpolating services.worker_wrapper.scale: failed to cast to expected type: strconv.Atoi: parsing "": invalid syntax`

was there

Rakanhf (approved):
LGMT !

PR #1813, opened 2026-09-01
fix: use advisory lock to prevent deadlock when immediate and scheduled changes get triggered at one time

Rakanhf:
When trying to change a subscription with one immediate change and one scheduled change e.g add 1 seat and decrease 1 storage package, we get a deadlock error.

This is caused by a subscription is being modified locally while a webhook is also received, they both race to modify the local instance. So the deadlock happens, since they wait for each other to release the row lock from select_for_update in a loop.

- Added `advisory_lock` context manager wrapping `pg_advisory_xact_lock`, and use it in `update_subscription` keyed on the account, Transactions naming the same account now queue behind one another, so there is a single lock to wait on and no cycle can form see [FUNCTIONS-ADVISORY-LOCKS](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS)

### DIFF
--- a/docker-app/qfieldcloud/core/utils2/db.py
+++ b/docker-app/qfieldcloud/core/utils2/db.py
@@ -1,6 +1,9 @@
+import zlib
+from collections.abc import Iterator
+from contextlib import contextmanager
 from typing import TypeVar
 
-from django.db import models
+from django.db import models, transaction
 
 Model = TypeVar("Model", bound=models.Model)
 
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-8787 Fix deadlock when immediate and scheduled changes get triggered at one](https://app.clickup.com/t/2192114/QF-8787)

lukasgraf (approved):
LGTM, nice! 👍 🎉

PR #1817, opened 2026-09-02
Fix `QgisLayer` migration failure index row size exceeded on `qgis_layer_id ` unique constraint

manylon:
`QgisLayer.qgis_layer_id` is an unbounded `TextField`, so the `(qgis_project, qgis_layer_id)` unique constraint could build index rows too large for Postgres to store, failing inserts for layers with very long ids.

Replace `layer_qgis_project_qgis_layer_id_uniq` with an expression index on `MD5(qgis_layer_id)`, fixed at 32 bytes regardless of the source string's length. Postgres computes the hash straight from `qgis_layer_id` when it builds and checks the index.

Add tests for the new constraint: an empty `qgis_layer_id` still collides within a project, `update_from_details()` keeps rows correct through a create, rename, add and drop, and a `qgis_layer_id` long enough to have broken the old constraint no longer does.


duke-nyuki:
Task linked: [QF-8744 Fix `QgisLayer` migration failure index row size exceeded on `qgis_layer_id ` unique constraint](https://app.clickup.com/t/2192114/QF-8744)

docker-app/qfieldcloud/project/models.py
```diff
@@ -1610,13 +1619,17 @@ class QgisLayer(models.Model):
...
+
     class Meta:
         verbose_name = _("QGIS layer")
         verbose_name_plural = _("QGIS layers")
         ordering = ["qgis_project", "ordering"]
         constraints = [
             models.UniqueConstraint(
-                fields=["qgis_project", "qgis_layer_id"],
-                name="layer_qgis_project_qgis_layer_id_uniq",
+                fields=["qgis_project", "qgis_layer_id_md5"],
```
Rakanhf: I was wondering why can't we use [MD5](https://docs.djangoproject.com/en/6.1/ref/models/database-functions/#md5) Function on the index directly without having to add a new field ?

```suggestion
                F("qgis_project"),
                MD5("qgis_layer_id"),
                name="layer_qgis_project_qgis_layer_id_uniq",
```

manylon: Great idea, thanks, we will remove a lot of unnecessary complexity!

Rakanhf (approved):
LGTM ! 🚀 

Just a small comment, a commit re structure would be nice to maintain a clean commit history

manylon:
Thanks, will clean the history and merge after tests

PR #1820, opened 2026-09-03
feat: add project warning if a layer datasource contains hardcoded pg credentials

gounux:
Since the go-to way for PostGIS layers in QGIS / QFieldCloud is to use `pg_service` entries and QFC Secrets, this PR intends to add a warning to a QFieldCloud Project that contains hard-coded pg `user` or `password` values in the `datasource` definition.

Within QGIS, to create such a layer with hard-coded pg user and password in the datasource, run the following `pyqgis` code e.g. in QGIS's python console:

```python
uri = QgsDataSourceUri()
uri.setConnection("the_pg_host", "the_pg_port", "the_pg_dbname", "the_pg_username", "the_pg_password")
uri.setDataSource("the_pg_schema", "the_pg_table", "geom", "", "id")
layer = QgsVectorLayer(uri.uri(False), "the_qgis_layer_name", "postgres")
QgsProject.instance().addMapLayer(layer)
```

### DIFF
--- a/docker-app/qfieldcloud/project/models.py
+++ b/docker-app/qfieldcloud/project/models.py
@@ -979,6 +979,28 @@ class Project(models.Model):
                                 ),
                             }
                         )
+
+                # add a warning if a pg layer has a hard-coded user or pwd in the datasource
+                if layer.provider_name == "postgres" and (
+                    "user=" in layer.datasource or "password=" in layer.datasource
+                ):
+                    problems.append(
+                        {
+                            "layer": layer_name,
+                            "level": "warning",
+                            "code": "hardcoded_pg_credentials",
+                            "description": _(
+                                _(
+                                    'Layer "{}" has hard-coded postgres credentials in the datasource.'
+                                )
+                            ).format(layer_name),
+                            "solution": _(
+                                _(
+                                    "Consider using a pg_service entry and a QFieldCloud Secret instead."
+                                )
+                            ),
+                        }
+                    )
         else:
             problems.append(
                 {
--- a/docker-app/qfieldcloud/project/models.py
+++ b/docker-app/qfieldcloud/project/models.py
@@ -991,7 +991,7 @@ class Project(models.Model):
                             "code": "hardcoded_pg_credentials",
                             "description": _(
                                 _(
-                                    'Layer "{}" has hard-coded postgres credentials in the datasource.'
+                                    'Layer "{}" has a hard-coded postgres credentials in the datasource.'
                                 )
                             ).format(layer_name),
                             "solution": _(


duke-nyuki:
Task linked: [QF-8267 Add a warning when PostGIS layer with hard-coded credentials in datasource definition is detected](https://app.clickup.com/t/2192114/QF-8267)

docker-app/qfieldcloud/project/models.py
```diff
@@ -979,6 +979,28 @@ def problems(self) -> list[dict[str, Any]]:
...
+                    "user=" in layer.datasource or "password=" in layer.datasource
+                ):
+                    problems.append(
+                        {
+                            "layer": layer_name,
+                            "level": "warning",
+                            "code": "hardcoded_pg_credentials",
+                            "description": _(
+                                _(
+                                    'Layer "{}" has hard-coded postgres credentials in the datasource.'
```
manylon:
```suggestion
                                    'Layer "{}" has a hard-coded postgres credentials in the datasource.'
```

docker-app/qfieldcloud/project/models.py
```diff
@@ -979,6 +979,28 @@ def problems(self) -> list[dict[str, Any]]:
...
+                                )
+                            ).format(layer_name),
+                            "solution": _(
+                                _(
+                                    "Consider using a pg_service entry and a QFieldCloud Secret instead."
```
manylon:
```suggestion
                                    "Consider using a "pg_service" entry and a QFieldCloud Secret instead."
```

manylon (approved):
Tested it locally, works like a charm, thanks! Added some small comments, feel free to ignore 🚀

nirvn:
@gounux , shouldn't this better be fitted in libqfieldsync as part of the project checks we do?

nirvn:
@gounux , are we OK throwing a constant warning into the UI for this?

gounux:
Oh okay @nirvn maybe I pressed the _Merge_ button to quickly, apologies !

> @gounux , are we OK throwing a constant warning into the UI for this?

Would tend to answer "yes" to that question, since the pg credentials of such a layer are then displayed in the Admin's project page, under `datasource` key. What's your opinion ?

See `libqfieldsync`'s PR here: https://github.com/opengisch/libqfieldsync/pull/170

IMO it's a good thing that QFS warns that before pushing a project to QFC, which warns as well on the same concern,

nirvn:
@gounux , well, practical example: we ship a project - the QField live users map project - with plaintext credentials in the data source because the user/password is meant to be public. 

We definitively want to help users in making the right choice (hence libqfieldsync), but I don't know if we want to visually interfere when the user has made their choice already :)

gounux:
Okay I see the point, thanks @nirvn,

I assume we all agree that this warning should live in QFieldSync, question mark on its presence in QFieldCloud.

Additional opinions would definitely help here, maybe @suricactus ?

suricactus:
The whole `Project.problems` property needs rethinking one day. We do not have a sensible mechanism to export the libqfs checker info into something that qfc can use. So I would consider it a separate task.

For now this PR is doing more good, we can have the check in both qfc and libqfs. One day we will keep it on libqfc only.

nirvn:
@suricactus , just to be clear, you are OK with projects getting a permanent top-of-the-page warning when users decide to ship a project with one or more hardcoded pg credentials?

suricactus:
Absolutely, this should be discouraged as much as possible. And we should stop using it eventually for our sample projects too.

I am not convinced about the naive approach of using a substring in this PR, instead of some kind of datasource parsing, but I am not against testing it in the wild.

nirvn:
> And we should stop using it eventually for our sample projects too

That's not possible :) and that goes to my point -- but if the majority feels OK with QFC pinning warnings in this case , so be it.

suricactus:
I think the "live qfield users" project was created in another era when there was no QFC and  hardcoded credentials was because there was no alternative sync mechanism.

This is long gone and now with public project non-collaborator role feature we can rethink the whole concept. Ayways, this PR is not the place to discuss such evolution.

manylon:
> I am not convinced about the naive approach of using a substring in this PR, instead of some kind of datasource parsing, but I am not against testing it in the wild.

Agree, in a future follow up it will be better to do this check in the worker and use something like:
```
if layer.dataProvider().name() == "postgres":
  uri = QgsDataSourceUri(layer.source())
  if uri.username() or uri.passoword():
  ... store something and pass it to QFC
```

PR #1824, opened 2026-09-06
Allow adding non-organization members as project collaborators in public projects

manylon:
Organization-owned projects rejected any collaborator who wasn't an organization member, even when the project was public. Now anyone should be able to collaborate and organization membership is only checked for private projects or when a public project goes private.

UI:
Screenshot 2026-09-04 at 11 03 32


duke-nyuki:
Task linked: [QF-8419 Allow adding non-organization members as project collaborators in public projects](https://app.clickup.com/t/2192114/QF-8419)

docker-app/qfieldcloud/core/models_utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any
+
+
+def get_stored_value(instance: Any, field_name: str) -> Any:
+    """The value currently stored in the database for `field_name` on `instance`.
+
+    Returns `None` when `instance` has never been saved. Use it to compare the
+    stored value against the in-memory one, for example to act only when a field is
+    actually being changed. Be cautious when using with FK fields.
```
lukasgraf: What does `Be cautious when using with FK fields.` mean here? This comment in the docstring doesn't really tell me _why_ I should be cautious, and _how_ I would do that.

docker-app/qfieldcloud/project/models.py
```diff
@@ -1101,6 +1096,16 @@ def total_collaborators_count(self) -> int:
         """Number of people returned by `total_collaborators`."""
         return self.total_collaborators.count()
 
+    @property
+    def exceeds_private_collaborator_limit(self) -> bool:
+        """Whether the project has more direct collaborators than the owner's plan allows on a private project. Always `False` when the plan sets no limit."""
+        max_premium_collaborators_per_private_project = self.owner.useraccount.current_subscription.plan.max_premium_collaborators_per_private_project
```
lukasgraf: Can we break this up onto two lines, and also have a line break in the docstring?

I would personally write this as:
```python
    plan = self.owner.useraccount.current_subscription.plan
    max_collaborators = plan.max_premium_collaborators_per_private_project
```

`max_collaborators` makes the expression below more readable, and is _more than enough_ context, when the origin of that value is just two lines above.

docker-app/qfieldcloud/core/tests/test_queryset.py
```diff
@@ -359,9 +359,15 @@ def test_private_project_memberships(self):
...
         self.assertProjectRole(p, u1, roles.MANAGER, role_origins.COLLABORATOR, True)
 
-        # If project is made private, the collaboration is invalid
+        # The project can go private while its collaborator count is within the plan limit
         p.is_public = False
         p.save()
+        self.assertProjectRole(p, u1, roles.MANAGER, role_origins.COLLABORATOR, True)
+
+        # Dropping the plan limit below the collaborator count invalidates the role.
+        # The collaborator stays on the project but loses access.
```
lukasgraf: Are we actually sure this is what happens? The role is invalidated and the collaborator loses access? Would that affect *all* collaborators? Or which ones exactly, if only _some_ of them are over the limit?

lukasgraf (changes requested):
Looks good to me in principle, just a few nitpicks or questions for clarification. 👍

docker-app/qfieldcloud/core/tests/test_queryset.py
manylon: The project gets locked with `exceeds_private_collaborator_limit=True`, `status=failed` and `status_code=too_many_collaborators`.
`can_retrieve_project` checks with `skip_invalid=True` and when accessing `GET /api/v1/projects/{id}/` the response is `403`. This is for all collaborators of the project until the owner removes as many as needed to go under the collaborator limit

docker-app/qfieldcloud/core/models_utils.py
manylon: Thanks for addressing this, when developing I didn't tested it's behaviour with FK. Now I did and have updated the docstring.

lukasgraf (approved):

PR #1827, opened 2026-09-08
Don't cause `500` error on admin when a delta can't be applied

manylon:
Clicking "Apply delta" on a delta in the Django admin raised `NoQGISProjectError` or `NoDeltasToApplyError`, which bubbled up as a 500 error page instead of telling the user what went wrong.

Both cases now add an error message and redirect back to the change page. The "no deltas" message also lists the usual reasons: the delta isn't in `PENDING` status, it's already part of a running apply job, or the project owner can't create jobs.

UI:
image


duke-nyuki:
Task linked: [QF-3755 Re-apply changes from Django admin causes error 500](https://app.clickup.com/t/2192114/QF-3755)

lukasgraf (approved):
LGTM! 👍

PR #1828, opened 2026-09-08
fix: Worker cleanup by environment and compose project

toebivankenoebi:
Label QGIS worker containers with the compose project name and use it together with the environment when finding orphaned workers. This prevents one QFieldCloud stack from killing workers owned by another stack on the same Docker host.


lukasgraf (approved):
LGTM! 👍

suricactus (approved):
Thanks!

PR #1830, opened 2026-09-08
Return the teams that the caller is a member of within the owning organization on project retrieve

manylon:
QField will add a `@cloud_teams` expression variable. To resolve it, it needs the teams that the signed-in user belongs to in a project's owning organization.

`GET /api/v1/projects/<id>/` now returns a `teams` array. The list, create, and update responses omit it.

- `ProjectDetailSerializer` extends `ProjectSerializer` with a `teams` field and is selected only for the `retrieve` action
- `teams` holds the caller's team names with the `@organization/` prefix removed
- `teams` is empty when a person owns the project, the caller is in no team, owning or administering the organization does not by itself make the caller a team member
- the drf-spectacular `retrieve` response now points at `ProjectDetailSerializer`
- new `test_project_teams_field` covers the membership cases, cross-organization isolation, and the field's absence from the list and write responses

### DIFF
--- a/docker-app/qfieldcloud/project/serializers.py
+++ b/docker-app/qfieldcloud/project/serializers.py
@@ -6,6 +6,7 @@ from rest_framework.exceptions import ValidationError
 
 from qfieldcloud.core import exceptions
 from qfieldcloud.core.models import (
+    Team,
     User,
 )
 from qfieldcloud.project.models import (
@@ -244,6 +245,41 @@ class ProjectSerializer(serializers.ModelSerializer):
             "overwrite_conflicts",
         }
         model = Project
+
+
+class ProjectDetailSerializer(ProjectSerializer):
+    """`ProjectSerializer` with the caller's teams in the project owning organization."""
+
+    teams = serializers.SerializerMethodField()
+
+    def get_teams(self, obj: Project) -> list[str]:
+        """Returns the caller's team names in the project's owning organization, without the `@organization/` prefix.
+
+        Empty when a person owns the project or the caller is in no team.
+        """
+        # A person owns the project, so it has no teams.
+        if obj.owner.is_person:
+            return []
+
+        request = self.context["request"]
+
+        teams = Team.objects.filter(
+            team_organization=obj.owner.organization,
+            members__member=request.user,
+        ).select_related("team_organization")
+
+        team_names = []
+        for team in teams:
+            team_names.append(team.teamname)
+
+        return team_names
+
+    class Meta(ProjectSerializer.Meta):
+        fields = (*ProjectSerializer.Meta.fields, "teams")  # type: ignore[assignment]
+        read_only_fields = (
+            *ProjectSerializer.Meta.read_only_fields,
+            "teams",
+        )  # type: ignore[assignment]
 
 
 class ProjectThumbnailSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


duke-nyuki:
Task linked: [QF-6682 [QFC] Add support for @cloud_team variable](https://app.clickup.com/t/2192114/QF-6682)

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -246,6 +247,43 @@ class Meta:
         model = Project
 
 
+class ProjectDetailSerializer(ProjectSerializer):
+    """`ProjectSerializer` with the caller's teams in the project owning organization."""
+
+    teams = serializers.SerializerMethodField()
+
+    def get_teams(self, obj: Project) -> list[str]:
+        """The caller's team names in the project's owning organization, without the `@organization/` prefix.
```
gounux:
```suggestion
        """Returns the caller's team names in the project's owning organization, without the `@organization/` prefix.
```

Wondering if docstrings should start with a verb ? Also 3rd person ?  
More convention-like & linguistics comment here ;-)  
Feel free to resolve and ignore.

docker-app/qfieldcloud/project/views.py
```diff
@@ -155,6 +159,12 @@ def upload_thumbnail(self, request, projectid):
         serializer.save()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return ProjectDetailSerializer
```
gounux, line 164: Facing some challenges to understand the sense and where this `retrieve` come from,

Is that maybe worth a small comment ? Maybe not, leaving it up to you.

gounux (approved):
Apart from a few format / docstrings comment, code looks good to me, thanks :) also for the nice test cases,  
Maybe another pair of eyes from a Django expoert could help here ;-)

docker-app/qfieldcloud/project/serializers.py
```diff
@@ -246,6 +247,43 @@ class Meta:
...
+
+        Empty when there is no request, a person owns the project, or the caller is in no team.
+        """
+        request = self.context.get("request")
```
suricactus: When don't we have a request?

Make an early return if we really need to support this case and write a comment in what situation this happens. Do not mix multiple conditionals for early returns if possible, e.g. you are checking both if there is no request or there is no organization. But these two are unrelated checks. If you couple them in a single if condition, it becomes harder to follow and harder to refactoring in the future. 

Finally you can save a few CPU cycles with an early return after checking for the request.

Early returns are cheap concept to develop and maintain.

docker-app/qfieldcloud/project/views.py
manylon: From drf docs:
> A ViewSet class is simply a type of class-based View, that does not provide any method handlers such as .get() or .post(), and instead provides actions such as .list() and .create().

https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions

docker-app/qfieldcloud/project/serializers.py
manylon: Thanks for the catch, mistaken it for when drf-spectacular is generating the OpenAPI schema, we are missing the request there, but when we are using the `get_serializer_class`  it is handled.

PR #1833, opened 2026-09-11
Fix project status null by guarding `qgis_project` lookup in `Project.status`

manylon:
`Project.status` is computed each time it's read, and it decides `FAILED` by checking whether project metadata parsing succeeded. On a timed-out `process_projectfile` job,`QgisProject` row doesn't get created, and status was accessing that reverse one-to-one relation directly instead of through `getattr`. That raised `QgisProject.DoesNotExist`, which DRF's serializer quietly turns into `None`, so the API reported status as `null` instead of failed.

- Swap `self.qgis_project` for `getattr(self, "qgis_project", None)` in the `Project.status` property

### DIFF
--- a/docker-app/qfieldcloud/project/models.py
+++ b/docker-app/qfieldcloud/project/models.py
@@ -1034,7 +1034,7 @@ class Project(models.Model):
 
             # TODO use self.problems to get if there are project problems
             if (
-                not self.has_the_qgis_file or not self.qgis_project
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-8815 Fix project status `None`](https://app.clickup.com/t/2192114/QF-8815)

Rakanhf (approved):
LGTM ! 🚀

PR #1836, opened 2026-09-14
Make `django-axes` aware of the NGINX reverse proxy

manylon:
`django-axes` only reads `REMOTE_ADDR` unless told otherwise, so it kept logging the `nginx` container's own IP instead of the real client's. Also `django-axes` only reads `X-Forwarded-For` when `django-ipware` is installed, and silently falls back to `REMOTE_ADDR` otherwise.

- add `AXES_IPWARE_META_PRECEDENCE_ORDER = ("HTTP_X_FORWARDED_FOR", "REMOTE_ADDR")`
  in `settings.py`
- add `django-ipware==7.0.1` to `requirements.in`
- regenerate `requirements.txt` via the `uvcompile` service

Before:
image

After:
image


duke-nyuki:
Task linked: [QF-8565 Fix proxy configuration for django-axes](https://app.clickup.com/t/2192114/QF-8565)

lukasgraf (commented):
Have you been able to verify (locally, and possibly on staging) that this solves the issue? I.e. that the correct source IP addresses are displayed for access logs / attempts?

manylon:
> Have you been able to verify (locally, and possibly on staging) that this solves the issue? I.e. that the correct source IP addresses are displayed for access logs / attempts?

I have made a huge mistake while testing it. Thanks for the question, to actually second guess the solution. Fixed, tested locally through NGINX and also applied screenshots on before and after.

lukasgraf (approved):
Thanks! As discussed, this looks good to me! 👍 🎉

PR #1837, opened 2026-09-15
Fix admin large organization inlines

toebivankenoebi:
Fixes the organization admin page for organizations with many related objects.

Instead of rendering large projects/members/teams inline formsets, the organization admin now uses links/action panels for managing these relations. This avoids oversized admin POST requests and Django formset absolute_max issues.

 - Move organization member management to a dedicated OrganizationMemberAdmin
 - Remove the projects inline from OrganizationAdmin
 - Add project/member/team count + manage/add links to the organization admin
 - Replace member/team/secret inline contents with an action panel to manage or add itmes
 - Add manage links to secret panels for consistency
 - Preserve organization context for “save and add another” flows

Notes:
This PR was LLM assisted.
This only affects the Django admin UI. The API member endpoints are unchanged.

### Organisation -> General
Before:
image

After:
image

### Organisation -> Organisation Members
Before: 
image

After:
image
image
image

### Organisation -> Teams
Before:
image

After:
image

### Organisation -> Secrets
Before:
image

After:
image


suricactus:
Is there a reason to rush with such development when we expect to have django unfold in place in the upcoming sprint, which has a huge potential to resolve the situation with overwhelming number of poat params?

Even if it doesn't and we want to integrate these changes, I would rather focus on the new UI/UX and merge this PR after unfold.

toebivankenoebi:
Reason: We have an instance that runs into that problem on a regular basis that is currently hot patched. Need a sustainable solution. 

Regarding UX: I tried to implement this solution so that UX doesn't change and everything is still accessible through the "old" way. Current user experience with inline having to scroll down and right to the end to add an additional user is questionable.

suricactus:
Switching to unfold was thought to be the long term solution therefore planned now...

I am -1 on merging this now, but I am ok if the team vote goes in favor of this, I can live with it.

EDIT we might need this even with unfold, but I would like to avoid custom code in the admin unless really needed.

toebivankenoebi:
It solves a problem that we have now and needs to be solved in a sustainable way.

Changed to draft for now and we can use it as a basis for discussion.

PR #1832, opened 2026-09-09
Add `json2qgis_file` field to `ProjectSeed`

manylon:
A `ProjectSeed` could be built from an `XLSForm` or a cloned project. This adds a third source, a `JSON` project definition file, stored the same way as `xlsform_file`. It appears on the `ProjectSeed` admin inline, where the uploaded file is downloadable. `clean()` now rejects setting more than one of the three sources at once `xlsform_file`, `json2qgis_file` and `clone_from_project`.

- add `get_seed_json2qgis_upload_to` so it writes to `projects/<id>/seeds/json2qgis/json2qgis<ext>`,
mirroring the `xlsform` helper
- add migration file
- add `json2qgis_file` to `ProjectSeedInline.fields` in `admin.py`
- add `json2qgis_file` to the qgis worker's `ProjectSeed` dataclass

PR #1838, opened 2026-09-16
Seed a project from a JSON project definition file

manylon:
> Follow up on:
> - https://github.com/opengisch/QFieldCloud/pull/1832
> - https://github.com/opengisch/qfieldcloud-sdk-python/pull/116

Projects could previously be seeded from an XLSForm file. Now a JSON project definition works too. The worker converts it into a QGIS project in `create_project` Job.

Changes:
- add `json2qgis_file` to `ProjectSerializer`, `ProjectCreateRequestSerializer`, and the OpenAPI exclude lists on `ProjectCreateJSONRequestSerializer` and `ProjectUpdateRequestSerializer`
- add `MultipleProjectSeedSourcesError` in `core/exceptions.py` for `xlsform_file` + `json2qgis_file` conflict
- add `seed_json2qgis` action
- add `get_project_seed_json2qgis` and `_create_project_from_json2qgis`, mirroring `get_project_seed_xlsform` and `_create_project_from_xlsform`
- `prepare_project_files` takes a `json2qgis_filename` argument and builds from it
- a `Convert2QgisBaseError` from `ProjectCreator` raises `UnableToContinueException`
- add `test_create_project_with_json2qgis` in `project/tests/test_project.py`, covering the create endpoint's `seed`/`json2qgis_file`/`xlsform_file` validation rules
- add `test_seed_json2qgis_download`, covering the seed download endpoint's success and 404 paths
- add `test_create_project_from_json2qgis` in `core/tests/test_jobs.py`, running a real `CREATE_PROJECT` job against a new `testdata/json2qgis/sample.json` fixture


duke-nyuki:
Task linked: [QF-8498 Allow setting the json2qgis JSON file via the API](https://app.clickup.com/t/2192114/QF-8498)

suricactus:
Thanks!

> The worker converts it into a QGIS project at creation time.

This is a bit misleading. The worker will convert a JSON seed in create_project Job, not at "creation time". At creation time we only have the blank QFC project and the seed.

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -237,6 +295,21 @@ def prepare_project_files(
             )
         else:
             project_filename = Path(project_filename)
+    elif project_seed.json2qgis_file:
```
suricactus, line 298: With the proposed implementation, having a `xlsform` is "stronger" than having a 'json2qgis_file`. However, if we have to rank them, xlsform is one of the many frontends before getting a `json2qgis` file.

I would suggest you put a `XOR` before that and verify there is either `xlsform` or `json2qgis` file, but never both. If there are both, there is something wrong along the route.

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -419,6 +492,17 @@ def get_workflow(self, project_id: UUID) -> Workflow:  # type: ignore
...
+                    id="get_project_seed_json2qgis",
+                    name="Get Project Seed JSON2QGIS",
+                    arguments={
+                        "project_id": project_id,
+                        "project_seed": StepOutput("get_project_seed", "project_seed"),
+                        "destination_dir": WorkDirPath(),
+                    },
+                    method=get_project_seed_json2qgis,
+                    return_names=["json2qgis_filename"],
+                ),
```
suricactus, line 505: Can't we reuse the above `get_project_seed_xlsform` and rename it to `get_project_seed` instead? Then we can return both `xlsform_filename` and `json2qgis_filename`?

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -195,10 +197,66 @@ def _create_project_from_xlsform(
...
+
+    return full_filename
+
+
 def prepare_project_files(
-    project_seed: ProjectSeed, tmp_project_dir: str, xlsform_filename: str
+    project_seed: ProjectSeed,
+    tmp_project_dir: str,
+    xlsform_filename: str,
+    json2qgis_filename: str,
```
suricactus, line 257: Are we sure about the types here?

docker-qgis/qfc_worker/commands/create_project.py
```diff
@@ -195,10 +197,66 @@ def _create_project_from_xlsform(
...
+
+        return None
+
+    output_dir = Path(tmp_project_dir).joinpath("files")
+
+    try:
+        with open(json2qgis_filename) as fh:
+            project_definition = json.load(fh)
+
+        project = ProjectCreator(project_definition).build(output_dir)
```
suricactus, line 238:
```suggestion
        creator = ProjectCreator(project_definition)
        project = creator.build(output_dir)
```

Easier debugging later.

docker-app/qfieldcloud/core/tests/testdata/json2qgis/sample.json
suricactus, line 1: Instead of hard copy of the `docker-app/qfieldcloud/core/tests/testdata/json2qgis/sample.json`, can't we import it from `convert2qgis` for now? We expect a few big changes ahead and the schema is not written in stone yet. It would be nice if we can use the current sample from the convert2qgis project instead? The alternative of having `sample.json` here mean we need to copy/paste the contents of the file any time we update `json2qgis` because it might break.

suricactus (commented):
A few small comments trying to simplify the API.

PR #1839, opened 2026-09-17
feat: expose version in `/api/v1/status/` and `/api/v1/server/info/` responses

Rakanhf:
Exposing QFieldCloud `version` in `/api/v1/status/` & `/api/v1/server/info/` endpoints so it's easier to track it


duke-nyuki:
Task linked: [QF-8839 Add sever version to server info /  status endpoints](https://app.clickup.com/t/2192114/QF-8839)

gounux (approved):
Thanks 👌

suricactus:
Nice! I feel slightly off why we show the same information in two different endpoints. Can we add a small argument why did we choose these two endpoints? A bit of argumentation in the first PR message is always appreciated.

Rakanhf:
> Nice! I feel slightly off why we show the same information in two different endpoints. Can we add a small argument why did we choose these two endpoints? A bit of argumentation in the first PR message is always appreciated.

In an internal discussion, we found that it would be useful to have the version on both endpoints, since it fits for both status for monitoring, and server natively

suricactus (approved):
One day we need to rename this from `SENTRY_RELEASE` to something like `QFC_VERSION`. But out of scope for this PR.

PR #1840, opened 2026-09-17
Expose `teams` on every project endpoint without N+1 queries

manylon:
When QField logs into QFieldCloud, we do a projects refresh to insure that we get up to date details on all projects (name, description, owner, role, etc.). Project `teams` being only in the retrieval endpoint means we won't be able to rely on that project refresh to know which up-to-date teams you're assigned to for downloaded projects. So every time we open a cloud project in QField, we'll need to fetch its up-to-date detailed data.

This PR fixes this by moving `teams` from `ProjectDetailSerializer.get_teams`, which ran one query per project to onto `ProjectSerializer` and added `ProjectQueryset.with_user_teams`, prefetching each project's organization's teams for the requesting user in a single query.

Changes:
- add `ProjectQueryset.with_user_teams`, wired into `ProjectViewSet.get_queryset`, prefetching `owner.organization.user_teams`
- move `teams` and `get_teams` from the removed `ProjectDetailSerializer` onto `ProjectSerializer`; fall back to a per-object query for instances that bypass the prefetch, e.g. a freshly created or updated object
- add `PublicProjectSerializer`, dropping `teams` from `/api/v1/projects/public/`
- update `test_project_teams_field`: `teams` now appears in list/create/update responses and stays absent from the public listing

### Setup:
Total projects visible to my testing user are 211 (200 bench: 20 orgs × 10 projects and 11 personal). For every bench project, there are 3 teams that the testing user is part of out of 5.

Did a full restart of the application before running every `GET` request.
Tests are performed locally on a single machine.

### On master branch:

`GET /api/v1/projects/`
default 466.63 ms (459 queries including 455 similar and 211 duplicates )

`GET /api/v1/projects/?include-public=0&limit=20&offset=0`
default 224.31 ms (46 queries including 40 similar and 20 duplicates )

### On this branch:

`GET /api/v1/projects/`
default 488.55 ms (457 queries including 452 similar and 211 duplicates )

`GET /api/v1/projects/?include-public=0&limit=20&offset=0`
default 227.00 ms (47 queries including 40 similar and 20 duplicates )

---

Thank you @nirvn for explaining QField details 🚀.


duke-nyuki:
Task linked: [QF-8837 Optimize Project teams query and include it in the list view](https://app.clickup.com/t/2192114/QF-8837)

nirvn:
@kaustuvpokharel , FYI, this makes implementation much easier on your side

suricactus:
If I understand this correctly, this adds the teams on all projects `GET /projects`, instead on `GET /projects/uuid`? I am afra9d this is quite a heavy query to run on production repeatedly with very little benefit for end users (the users of @qfc_team are going to be very low percentage of users).

Are we sure GET /project/uuid is not sufficient?

nirvn:
@suricactus , first, there's the fact that it is the only property not exposed in the `/projects/` endpoint. Since we have other user-level properties exposed (e.g. user_role and user_role_origin), it feels off to not have [user_]teams in there.

But more importantly for QField, ATM, we are able to build _complete_ cloud project objects via the `/projects/` endpoint. 

This simplifies the handling of downloaded cloud project properties. I..e., when we download a cloud project, we don't need to trigger a `/projects/{uuid}/` call, we got all we need from the refreshing of the user's complete cloud projects list. Similarly, once we have a downloaded cloud project, when we open them, we log into QFC and refresh the user's cloud projects list. If we don't have the teams detail in there, it means we need to do an extra call to `/projects/{uuid}/` when we open the cloud project itself.

None of it is insurmountable. It'd just a/ feel more consistent, and b/ be easier to manage on the QField side of things.

suricactus:
I am very well aware how the api clients work, including QField. Thats why I thought additional call to get project details will be a good balance to keep the DB load under control and code complexity acceptable.


The reason we delayed this development for that long was exactly because on top of the heavy projects with roles query we have to add another complex query extracting the:

- owner
- organization
- organization members
- teams
- team members
- project collaborators
- users


At least 8 joins, I am not sure how this query will perform in prod.

---

On the comment that we already have the roles on the GET /projects, well that's the sole purpose of this endpoint, isn't it? Get the projects we have access to. And to determine if we have access to something,  we need to check what is our role on that project.

Teams however have a different meaning - you belong to them and we care about the highest role you have from any of the teams. That's why it's very different type of information.

---

I am not strictly against the change in this PR, but I would advice the developers behind it to try to do a napkin calculation on the performance and strain on the database, because nearly all of the calls on this API endpoint will include the teams, and we have to handle many of them. We already have heavy DB load, and my concious approach to spare some CPU on the DB was to move the teams on the GET /project/uuid that was explicitly described in the original task.

manylon:
TBH I'm -0 on this solution. Will do some calculations and give a more detailed result on the load.

manylon:
@suricactus @nirvn I did more tests with a large amount of data and updated the testing results in the PR description. I'm open to discussions on this.

